PR #3, opened 2017-05-02
Feature/refactor

garyluu:
Refactoring code for modularity.


src/app/shared/provider.service.ts
```diff
@@ -0,0 +1,61 @@
...
+    const gitUrl = tool.gitUrl;
+
+    tool.provider = this.getProvider(gitUrl);
+    tool.providerUrl = this.getProviderUrl(gitUrl, tool.provider);
+
+    return tool;
+  }
+
+  private getProvider(gitUrl: string): string {
+    if (gitUrl.includes('github.com')) {
```
denis-yuen: Add a todo here. Without an anchor, this looks fragile (for example if you had a github repo that included the string " bitbucket.org" in its name.

PR #4, opened 2017-05-09
UI2 footer DockstoreUI tag version has been updated

JWKaiqi:
Issue ga4gh/dockstore#702 has been updated.
However, the tag version number at the footer will only show up once we have created the release tag.


src/app/footer/versions.ts
```diff
@@ -0,0 +1,2 @@
+// this file is automatically generated by git.version.ts script
```
garyluu: - Don't commit generated files.
- Looks like travis needs to run "npm prebuild" for this file this file to be generated.

garyluu:
Also put the "npm run-script prebuild" instruction in the README

PR #5, opened 2017-05-10
Created tooltips for the Tools. 

JWKaiqi:
Issue ga4gh/dockstore#703 has been updated.

Please note that I replaced Kitty's app-tab with the ngx-bootstrap tab for the Info, Versions, Files tab.
Apparently the workflows page is not working properly yet.
Please let me know if I am missing something.

Thanks

### DIFF
--- a/src/app/container/container.component.ts
+++ b/src/app/container/container.component.ts
@@ -1,16 +1,17 @@
-import {Component} from '@angular/core';
-import {Router} from '@angular/router';
+import { Component } from '@angular/core';
+import { Router } from '@angular/router';
 
-import {Tool} from '../shared/tool';
+import { CommunicatorService } from '../shared/communicator.service';
+import { DateService } from '../shared/date.service';
 
-import {ToolService} from '../shared/tool.service';
-import {CommunicatorService} from '../shared/communicator.service';
-import {ProviderService} from '../shared/provider.service';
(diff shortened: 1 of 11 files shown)


src/app/container/container.component.ts
```diff
@@ -1,50 +1,47 @@
-import { Component } from '@angular/core';
-import { Router } from '@angular/router';
+import {Component} from '@angular/core';
+import {Router} from '@angular/router';
```
garyluu: Let's keep the code style same as the Angular 2+ official site.  See https://angular.io/styleguide

src/app/shared/modules/container.module.ts
```diff
@@ -28,6 +28,10 @@ import { DockerfileComponent } from '../../container/dockerfile/dockerfile.compo
 import { DescriptorsComponent } from '../../container/descriptors/descriptors.component';
 import { ParamfilesComponent } from '../../container/paramfiles/paramfiles.component';
 
+/* Bootstrap */
```
garyluu: Move these Bootstrap imports up to the top near the other third party imports.  Also arrange by alphabetical order when possible.

garyluu:
Additionally, the ngx-bootstrap tabs module also includes a 'tab' directive.  See if you can use that too instead importing another module

PR #7, opened 2017-05-15
Verified Update

JWKaiqi:
Issue ga4gh/dockstore#705 has been updated.

Note that I rearranged the formate of the new UI so that it matches the look of the old UI.
Please let me know if there is any problem.

Thanks!

Jennifer

### DIFF
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -2,7 +2,7 @@
   <div class="row">
     <h3>
       <span id="verifiedIcon" *ngIf="tool?.versionVerified">
-        <a href="https://support.twitter.com/articles/119135#">
+        <a href={{tool?.verifiedLinks}}>
           <span class="glyphicon glyphicon-ok" tooltip="Verified"></span>
         </a>
       </span>
@@ -136,11 +136,15 @@
         </tab>
         <tab heading="Versions" *ngIf="validVersions">
           <app-versions-container [versions]="tool?.tags"
-                                  [verifiedSource]="tool?.VerifiedSources"></app-versions-container>
+                                  [verifiedSource]="tool?.verifiedSources">
+
+          </app-versions-container>
         </tab>
         <tab heading="Files" *ngIf="validVersions">
           <app-files-container [id]="tool?.id" [versions]="validVersions"
-                               [default]="defaultVersion"></app-files-container>
+                               [default]="defaultVersion">
+
+          </app-files-container>
         </tab>
       </tabset>
     </div>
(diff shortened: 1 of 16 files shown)


src/app/container/container.component.html
```diff
@@ -1,138 +1,164 @@
 <div class="container">
-  <h3>{{ title }}</h3>
-  <p>
-    Last build: {{ tool?.agoMessage || 'n/a' }}
-  </p>
-</div>
+  <div class="row">
+    <h3>
+      <span id="verifiedIcon" *ngIf="tool?.versionVerified">
+        <a href="https://support.twitter.com/articles/119135#">
```
denis-yuen: Make an issue for us to write something up to describe verified tools so that we do not forget

src/app/container/versions/versions.component.html
```diff
@@ -1,76 +1,78 @@
-<div class="container">
+<div>
   <div class="row">
     <div>
       <table datatable [dtOptions]="dtOptions" class="table row-border hover">
         <thead>
-          <tr>
```
denis-yuen: Seems like an odd amount of whitespace changes

src/app/shared/dockstore.service.ts
```diff
@@ -1,17 +1,15 @@
...
-import { Dockstore } from '../shared/dockstore.model';
+import { Http } from '@angular/http';
 
 @Injectable()
 export class DockstoreService {
 
-  private static readonly months = ['Jan.', 'Feb.', 'Mar.', 'Apr.',
-                                    'May', 'Jun.', 'Jul.', 'Aug.',
-                                    'Sept.', 'Oct.', 'Nov.', 'Dec.'];
+  private static readonly months = [ 'Jan.', 'Feb.', 'Mar.', 'Apr.',
```
denis-yuen: Whitespace changes especially evident here

src/app/container/versions/versions.component.html
garyluu: Looks like <tr> is no longer nested inside <thead> due to the whitespace changes

src/app/container/container.component.html
```diff
@@ -1,138 +1,164 @@
...
+          <span class="glyphicon glyphicon-ok" tooltip="Verified"></span>
+        </a>
+      </span>
+      {{ title }}
+    </h3>
+    <p>
+      Last build: {{ tool?.agoMessage || 'n/a' }}
+    </p>
+    <div class="col-sm-9" style="background-color:white">
```
garyluu: Avoid inline styles when possible.

garyluu: In general, have col-sm, col-md, col-lg

JWKaiqi:
Please view ga4gh/dockstore#714 if you want to change the link, thanks!

PR #8, opened 2017-05-17
Updated README and package.json for setting up angular-cli

JWKaiqi:
Issue ga4gh/dockstore#715 has been updated.

So I tested this with different dockers, the global cil version is now same as the local cil.
Note that "angular/cli" in the package.json should always matches the version of the sudo npm install command:
sudo npm install -g @angular/cli@**1.0.3**

ie: make sure its 1.0.3 for both package.json and the install command.


denis-yuen:
Looks good, thanks!

PR #10, opened 2017-05-18
Document verified tools feature

denis-yuen:
Update to document verified tools for https://github.com/ga4gh/dockstore/issues/714


JWKaiqi:
Looks good to me

PR #18, opened 2017-06-06
Feature/workflows

JWKaiqi:
Build the template for Workflows, still missing Tools and Dag tab page.

Finished the Info, Labels, Version, Files tabs for Workflow, and changed table to ul for container and workflow.


src/app/container/descriptors/descriptors.service.ts
```diff
@@ -50,19 +51,42 @@ export class DescriptorsService {
   }
 
   private getCwl(id: number, versionName: string) {
-    return this.httpService.getResponse(`${ Dockstore.API_URI }/containers/${ id }/cwl?tag=${ versionName }`);
```
garyluu: most of these can be simplified to "${ this.type }s"

src/app/app.routing.ts
```diff
@@ -20,8 +19,9 @@ const APP_ROUTES: Routes = [
   { path: 'search-containers', loadChildren: 'app/containers/containers.module#ContainersModule' },
   { path: 'containers', loadChildren: 'app/containers/containers.module#ContainersModule' },
   { path: 'tools', loadChildren: 'app/containers/containers.module#ContainersModule' },
-  { path: 'search-workflows', loadChildren: 'app/workflows/workflows.module#WorkflowsModule' },
```
denis-yuen: Let's leave `search-workflows` here (in addition to `workflows`) in case anyone linked to our search for some reason

PR #21, opened 2017-06-14
Feature/dag2

garyluu:
- Populated DAG tab with a DAG that draws from the web service
- Moved away from CDNs in favor of local files for web pack optimizations
- Adding cytoscape, cytoscape-dagre, and dagre dependencies
- Fix HighlightJS by disabling it when DAG is visible - via checking DAG visibility in ngAfterViewChecked
- Fix initial load issue by initializing cytoscape once the DAG element is visible - via checking DAG visibility in ngAfterViewChecked.


src/app/workflow/dag/dag.component.html
```diff
@@ -0,0 +1,23 @@
...
+        </select>
+        <a id="exportLink" href="#" ng-show="dagJSON !== null && !notFound && !missingTool"><span class="glyphicon glyphicon-export"></span>Export to PNG</a>
+        <button type="button" id="dag_reload" ng-show="dagJSON !== null && !notFound && !missingTool" (click)="refreshDocument()" class="dag-button" title="Reset" tooltip-append-to-body="true"><span class="glyphicon glyphicon-screenshot"></span></button>
+        <button id="dag_fullscreen" ng-show="dagJSON !== null && !notFound && !missingTool" ng-click="expandDAG()" uib-tooltip="Toggle fullscreen" class="dag-button" tooltip-append-to-body="true">
+          <span class="glyphicon glyphicon-resize-full" id="resize-full-button"></span>
+          <span class="glyphicon glyphicon-resize-small no-display" id="resize-small-button"></span>
+        </button>
+      </div>
+    </div>
+<div #cy id="cy" class="row mini-dag"></div>
```
denis-yuen: seems like some formatting wonkiness, not a blocker

src/app/workflow/dag/dag.component.spec.ts
```diff
@@ -0,0 +1,28 @@
+/* tslint:disable:no-unused-variable */
```
denis-yuen: why is this warning necessary?

src/app/workflow/dag/dag.service.ts
```diff
@@ -0,0 +1,81 @@
...
+import { Injectable } from '@angular/core';
+
+
+@Injectable()
+export class DagService {
+    validVersions: Subject<any> = new Subject<any>();
+    currentWorkflowId: Subject<number> = new Subject<number>();
+    currentVersion: Subject<any> = new Subject<any>();
+    currentDagResults: Subject<any> = new BehaviorSubject<any>({
+        nodes: [
```
denis-yuen: this seems odd, is this an example or something?

src/cdn/typescript.min.js
```diff
@@ -0,0 +1 @@
...ray Int16Array Int32Array Int8Array Uint16Array Uint32Array Uint8Array Uint8ClampedArray ArrayBuffer DataView JSON Intl arguments require module console window document any number boolean string void Promise"};return{aliases:["ts"],k:r,c:[{cN:"meta",b:/^\s*['"]use strict['"]/},e.ASM,e.QSM,{cN:"string",b:"`",e:"`",c:[e.BE,{cN:"subst",b:"\\$\\{",e:"\\}"}]},e.CLCM,e.CBCM,{cN:"number",v:[{b:"\\b(0[bB][01]+)"},{b:"\\b(0[oO][0-7]+)"},{b:e.CNR}],r:0},{b:"("+e.RSR+"|\\b(case|return|throw)\\b)\\s*",k:"return throw case",c:[e.CLCM,e.CBCM,e.RM,{cN:"function",b:"(\\(.*?\\)|"+e.IR+")\\s*=>",rB:!0,e:"\\s*=>",c:[{cN:"params",v:[{b:e.IR},{b:/\(\s*\)/},{b:/\(/,e:/\)/,eB:!0,eE:!0,k:r,c:["self",e.CLCM,e.CBCM]}]}]}],r:0},{cN:"function",b:"function",e:/[\{;]/,eE:!0,k:r,c:["self",e.inherit(e.TM,{b:/[A-Za-z$_][0-9A-Za-z$_]*/}),{cN:"params",b:/\(/,e:/\)/,eB:!0,eE:!0,k:r,c:[e.CLCM,e.CBCM],i:/["'\(]/}],i:/%/,r:0},{bK:"constructor",e:/\{/,eE:!0,c:["self",{cN:"params",b:/\(/,e:/\)/,eB:!0,eE:!0,k:r,c:[e.CLCM,e.CBCM],i:/["'\(]/}]},{b:/module\./,k:{built_in:"module"},r:0},{bK:"module",e:/\{/,eE:!0},{bK:"interface",e:/\{/,eE:!0,k:"interface extends"},{b:/\$[(.]/},{b:"\\."+e.IR,r:0},{cN:"meta",b:"@[A-Za-z]+"}]}});
```
denis-yuen: What is this new cdn directory and why does it hold typescript which I would think could be handled as a dependency?

src/app/workflow/dag/dag.component.spec.ts
garyluu: Spec file is auto-generated.  I'm going to double check if it's a new 1.0.3 CLI thing or just my IDE.

src/app/workflow/dag/dag.service.ts
garyluu: Was used for testing to at least have something when the promise fails to work.

src/cdn/typescript.min.js
garyluu: All the CDNs specified in the index.html were downloaded and placed into this cdn folder instead.  This allows webpack to do further optimizations (remove all the code inside each file that will never be used).  Normally with CDNs in the index.html file, the browser will download the entire file.

denis-yuen: Does this need to be checked in then if its something webpack can generate when it runs?

garyluu: This is the source file, webpack takes it and then generates a smaller file from this.

denis-yuen: Ah, then why not point webpack at the source of this file rather than checking it into our repo?

garyluu: it is currently not possible to do that with webpack

denis-yuen: That sounds weird. 

Took a brief look, I would agree with the comment here, we shouldn't be packaging up popular libraries like typescript and just rely on the CDN. 
https://github.com/webpack/webpack/issues/150#issuecomment-287651075

Maybe look into externals as referenced here https://github.com/webpack/webpack/issues/150#issuecomment-257141360 and if not possible, just take these out and reference the CDN.

garyluu:
Putting the files back into CDN.

src/cdn/typescript.min.js
garyluu: That's not the typescript package.  Otherwise it would be more than one line of code.

PR #27, opened 2017-06-20
Feature/label editing feature

JWKaiqi:
Implemented label editing for tool and workflow.

### DIFF
--- a/src/app/shared/dockstore.service.ts
+++ b/src/app/shared/dockstore.service.ts
@@ -98,7 +98,6 @@ export class DockstoreService {
     }
   }
   setContainerLabels(containerId: number, labels) {
-    console.log('labels: ' + labels);
     const url = `${ Dockstore.API_URI }/containers/${ containerId }/labels`;
     const myParams = new URLSearchParams();
     myParams.set('labels', labels);
--- a/src/app/shared/dockstore.service.ts
+++ b/src/app/shared/dockstore.service.ts
@@ -98,7 +98,15 @@ export class DockstoreService {
     }
   }
   setContainerLabels(containerId: number, labels) {
+    console.log('labels: ' + labels);
     const url = `${ Dockstore.API_URI }/containers/${ containerId }/labels`;
+    const myParams = new URLSearchParams();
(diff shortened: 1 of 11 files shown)


src/app/workflow/workflow.component.ts
```diff
@@ -52,6 +54,40 @@ export class WorkflowComponent extends Tool implements OnDestroy {
...
   }
 
+  toggleLabelsEditMode() {
+    this.labelsEditMode = !this.labelsEditMode;
+  }
+  resetWorkflowEditData() {
+    const labelArray = this.dockstoreService.getLabelStrings(this.workflow.labels);
+    let workflowLabels = '';
+    for (let i = 0; i < labelArray.length; i++) {
+      workflowLabels += labelArray[i] + ((i !== labelArray.length - 1) ? ', ' : '');
```
garyluu: same thing here

src/app/shared/dockstore.service.ts
```diff
@@ -1,13 +1,19 @@
...
+import {Inject, Injectable} from '@angular/core';
+import { Http, RequestMethod, URLSearchParams} from '@angular/http';
+import { AuthService } from 'ng2-ui-auth';
+import { Dockstore } from '../shared/dockstore.model';
+import { HttpService } from './http.service';
+
 
 @Injectable()
 export class DockstoreService {
   private static readonly months = [ 'Jan.', 'Feb.', 'Mar.', 'Apr.',
```
denis-yuen: Kind of random, but is this used anywhere?

JWKaiqi: yes, inside the setContainerLabels and setWorkflowLabels function.

denis-yuen: The months. 
This seems private.

JWKaiqi: You are right, it is unnecessary because the month array is also defined in the date service. Only the date service is using it. Thus, I think we can remove it from the dockstore service (unless we need to use it in the future). 
What do you think?

PR #28, opened 2017-06-20
Feature/elastic experiment

denis-yuen:
See companion pull request in the Dockstore repo for readme details. 
This is going to need major refactoring, but demonstrates some of what we want to accomplish.


src/app/search/search.component.ts
```diff
@@ -0,0 +1,225 @@
...
+  styleUrls: ['./search.component.css']
+})
+export class SearchComponent implements OnInit {
+
+  private _client: Client;
+  private hits: Object[];
+  private aggs: Object;
+  private buckets: Map<string, Map<string, string>> = new Map<string, Map<string, string>>();
+  // tslint:disable-next-line
+  private initialQuery = '{"aggs":{"_type":{"terms":{"field":"_type","size":10000}},"registry":{"terms":{"field":"registry","size":10000}},"private_access":{"terms":{"field":"private_access","size":10000}},"tags_verified":{"terms":{"field":"tags.verified","size":10000}},"author":{"terms":{"field":"author","size":10000}},"namespace":{"terms":{"field":"namespace","size":10000}},"labels_value":{"terms":{"field":"labels.value","size":10000}},"tags_verifiedSource":{"terms":{"field":"tags.verifiedSource","size":10000}}},"query":{"match_all":{}}}';
```
garyluu: Might be more readable if it was multi-line, this would alleviate the need for the tslint disable

garyluu:
Something seems strange, lots of old DAG code is in there.

src/app/search/search.component.ts
denis-yuen: The whole query should be auto-generated in the end

denis-yuen:
Yeah, there bunch of strange merge conflicts when updating from develop.

PR #29, opened 2017-06-21
Feature/add tool

garyluu:
- Add tool modal
- Form validation + messages
- Error handling when adding tool
- Close modal when successfully added

Missing:
- Refresh tool and pull list of tools after adding tool
- savingActive || refreshingContainer || isInvalidCustomRegistry() || isInvalidPrivateTool() checks for submit


src/app/container/register-tool/register-tool.service.ts
```diff
@@ -0,0 +1,207 @@
...
+    };
+
+    createPath(toolObj: Tool, customDockerRegistryPath) {
+        let path = '';
+        if (customDockerRegistryPath !== null) {
+            path += customDockerRegistryPath;
+        } else {
+            path += this.getImageRegistryPath(toolObj.irProvider);
+        }
+        path += '/' + this.getImagePath(toolObj.imagePath, 'namespace') + '/' + this.getImagePath(toolObj.imagePath, 'name');
```
denis-yuen: How does this deal with alternate tool names when two tools belong to the same repo?

garyluu: The path remains the same, just as production currently does.  The toolname will be different however which would be reflected in the tool_path

denis-yuen:
Looks good, will approve after what looks like linting is fixed

PR #31, opened 2017-06-22
Feature/add tool validation

garyluu:
- Adding more validations for private registries
- Change the tool's title from path to tool_path to handle cases where multiple tools have the same path

### DIFF
--- a/src/app/container/register-tool/register-tool.service.ts
+++ b/src/app/container/register-tool/register-tool.service.ts
@@ -32,7 +32,7 @@ export class RegisterToolService {
         if (error) {
             errorObj = {
                 message: 'The webservice encountered an error trying to create this ' +
-                'container, please ensure that the container attributes are ' +
+                'tool, please ensure that the tool attributes are ' +
                 'valid and the same image has not already been registered.',
                 errorDetails: '[HTTP ' + error.status + '] ' + error.statusText + ': ' +
                 error._body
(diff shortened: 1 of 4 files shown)


src/app/container/register-tool/register-tool.service.ts
```diff
@@ -59,11 +59,23 @@ export class RegisterToolService {
...
 
-    isInvalidPrivateTool() {
-        return false;
+    isInvalidPrivateTool(toolObj: Tool) {
+        return toolObj.private_access === true && !toolObj.tool_maintainer_email;
     }
-    isInvalidCustomRegistry() {
-        return false;
+    isInvalidCustomRegistry(toolObj: Tool, customDockerRegistryPath: string) {
+            for (let i = 0; i < this.dockerRegistryMap.length; i++) {
```
denis-yuen: please add a comment for the method. 
Not sure what this is checking for

src/app/container/register-tool/register-tool.service.ts
```diff
@@ -29,15 +29,15 @@ export class RegisterToolService {
...
-            'container, please ensure that the container attributes are ' +
-            'valid and the same image has not already been registered.',
-            errorDetails: '[HTTP ' + error.status + '] ' + error.statusText + ': ' +
-            error._body
-        };
-    }
-    this.toolRegisterError.next(mapthing);
+            mapthing = {
+                message: 'The webservice encountered an error trying to create this ' +
+                'container, please ensure that the container attributes are ' +
```
denis-yuen: This should say tool now

src/app/container/register-tool/register-tool.service.ts
garyluu: If the registry is private, it checks if the customDockerRegistryPath is truthy (not null, not undefined, not "", etc)

src/app/container/register-tool/register-tool.service.ts
denis-yuen: This seems to still apply

PR #32, opened 2017-06-22
Feature/side bar

JWKaiqi:
Build the sidebar feature for both tool and workflow.

Please note that sharing count should be test it out during staging (according to the share-button wiki, twitter doesn't have this count feature), please see https://github.com/MurhafSousli/ngx-sharebuttons/wiki/Share-Counters for more information.

Also fixed some bugs for the verified tool, wasn't working properly when there is more than one branch.


JWKaiqi:
I was on the older develop branch then I merged the develop branch to my SideBar branch for updating.
That's why there are a lot of file changes...

src/styles.scss
```diff
@@ -856,6 +857,17 @@ div.tabs>div:target>div,
   background-color: #303030;
 }
 
+.form-control[disabled], .form-control[readonly], fieldset[disabled] .form-control {
+  background-color: white !important;
+  opacity: 1;
+}
+
+.btn-success {
```
garyluu: Avoid globally modifying our button styles

src/app/shared/tool.ts
```diff
@@ -34,6 +34,7 @@ export abstract class Tool implements OnInit, OnDestroy {
               private containerService: ContainerService,
               toolType: string) {
     this._toolType = toolType;
+    // this.isToolPublic = true;
```
denis-yuen: Take out commented code unless it is described well and would be useful in the future

src/app/workflow/versions/versions.component.ts
```diff
@@ -13,11 +13,12 @@ import { DockstoreService } from '../../shared/dockstore.service';
 })
 export class VersionsWorkflowComponent extends Versions {
   @Input() versions: Array<any>;
-  @Input() verifiedSource: Array<string>;
+  @Input() verifiedSource: Array<any>;
   @Input() workflowId: number;
   verifiedLink: string;
 
   setNoOrderCols(): Array<number> {
+    console.log(this.verifiedSource);
```
denis-yuen: Let's cut down on console.log calls unless the output is well described

PR #33, opened 2017-06-26
fixed the tab style

JWKaiqi:
Fixed the tab hover styling issue.

### DIFF
--- a/src/app/workflow/workflow.component.html
+++ b/src/app/workflow/workflow.component.html
@@ -90,8 +90,8 @@
               novalidate>
           <button type="submit"
                   class="btn btn-link pull-right"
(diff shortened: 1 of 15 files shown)


JWKaiqi:
Also fixed the problem with label editing, so that editing is allowed only when the tool/workflow is private.

src/app/workflow/workflow.component.html
```diff
@@ -90,11 +90,11 @@
               novalidate>
           <button type="submit"
                   class="btn btn-link pull-right"
-                  [disabled]="!labelForm.form.valid"
+                  [disabled]="!labelForm.form.valid && !isWorkflowPublic"
```
garyluu: container.component.html has different additions compared to this workflow.component.html.  The disabled expression doesn't seem quite right.

PR #34, opened 2017-06-26
Add and select tool to view

garyluu:
Contains a bunch of bug fixes and minor feature updates for tag edtior, DAG, and add tool

- Upon registering tool, add the registered response tool to the list of displayed tools without an additional API call
- Select the tool that just got registered.
- Style fix for DAG
- Bug fix for tag edtior
- Adding DAG refresh on ngAfterViewChecked
- Changes to the routing to grab published tool only when the event.url contains "containers"

### DIFF
--- a/src/app/container/view/view.component.ts
+++ b/src/app/container/view/view.component.ts
@@ -15,17 +15,15 @@ import { ViewService } from './view.service';
 @Component({
   selector: 'app-view-container',
   templateUrl: './view.component.html',
-  styleUrls: ['./view.component.css'],
-  providers: [ViewService]
+  styleUrls: ['./view.component.css']
 })
 // This is actually the tag edtior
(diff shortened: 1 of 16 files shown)


src/app/container/view/view.component.ts
```diff
@@ -200,8 +197,20 @@ export class ViewContainerComponent extends View implements OnInit, AfterViewChe
         this.mode = mode;
       }
     );
+
     this.unsavedVersion = Object.assign({}, this.version);
-    this.tool = this.communicatorService.getTool();
+    this.containerService.tool$.subscribe(tool => {
+    this.tool = tool;
```
denis-yuen: Comment on this, is this making a tool editable if it is published or just if this tag editor is in the public section of the GUI?

denis-yuen: Test

PR #39, opened 2017-06-29
Feature/deregister

garyluu:
This adds the deregister tool functionality and clears the add tool editor error info upon close and right before add.

### DIFF
--- a/src/app/container/deregister-modal/deregister-modal.component.html
+++ b/src/app/container/deregister-modal/deregister-modal.component.html
@@ -1,0 +1,21 @@
+<button id="deregisterButton" type="button" class="btn btn-danger" (click)="staticModal.show()">
+  Deregister
+</button>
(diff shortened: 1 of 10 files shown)


src/app/container/deregister-modal/deregister-modal.component.html
```diff
@@ -0,0 +1,21 @@
...
+  Deregister
+</button>
+<div class="modal fade" bsModal #staticModal="bs-modal" [config]="{backdrop: 'static'}" tabindex="-1" role="dialog" aria-labelledby="mySmallModalLabel" aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <div class="modal-title">Are you sure you wish to deregister this tool?</div>
+      </div>
+      <div class="modal-body">
+        <p>All information associated with this tool will be deleted.</p>
```
denis-yuen: Idle thought: Should this be called "Delete" rather than "Deregister"? That may require less explanation although it would parallel "register new tool" less.

garyluu: I think it's fine as it is.  The tool is deregistered from Dockstore which means the information associated with it is deleted on Dockstore

PR #42, opened 2017-07-06
Feature/add workflows

garyluu:
Add/register button and modal


src/app/shared/models/Label.ts
```diff
@@ -0,0 +1,4 @@
+export class Label {
```
denis-yuen: How are these classes created and kept up-to-date?
This may be a job for code generation https://github.com/swagger-api/swagger-codegen/tree/master/samples/client/petstore/typescript-angular2

garyluu: That was actually generated by https://www.npmjs.com/package/angular2-swagger-client-generator or one of its forks using the swagger.json  .  However, there's still quite a few bugs in the generated code, so I omitted the process from .travis.yml

PR #45, opened 2017-07-11
Feature/es2017

garyluu:
This clean-up several issues in the previous pull request such as the IP address and unused comment.
This adds es2017.object support to allow features like Object.entries
This also utilizes the bodybuilder package that was previously installed to construct the initial query.

### DIFF
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -1,7 +1,8 @@
+import * as bodybuilder from 'bodybuilder';
 import { Dockstore } from './../shared/dockstore.model';
-import {Component, OnInit, ViewChild, enableProdMode} from '@angular/core';
-import {Client} from 'elasticsearch';
-import {CommunicatorService} from '../shared/communicator.service';
+import { Component, OnInit, ViewChild, enableProdMode } from '@angular/core';
+import { Client } from 'elasticsearch';
+import { CommunicatorService } from '../shared/communicator.service';
 import { ProviderService } from '../shared/provider.service';
 import { ListContainersService } from '../containers/list/list.service';
 import { BehaviorSubject } from 'rxjs/BehaviorSubject';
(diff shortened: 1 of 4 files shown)


src/app/search/search.component.ts
```diff
@@ -1,6 +1,8 @@
-import {Component, OnInit, ViewChild, enableProdMode} from '@angular/core';
-import {Client} from 'elasticsearch';
-import {CommunicatorService} from '../shared/communicator.service';
+import * as bodybuilder from 'bodybuilder';
```
denis-yuen: Not sure if Javascript works the same way as Java, but a star import could complicate refactoring and dependency clean-up

src/app/search/search.component.ts
```diff
@@ -103,10 +98,24 @@ export class SearchComponent implements OnInit {
    */
   constructor(private providerService: ProviderService) {
     this._client = new Client({
-      host: 'http://10.11.9.73:8080/api/ga4gh/v1/extended',
+      host: Dockstore.API_URI + '/api/ga4gh/v1/extended',
       apiVersion: '2.4',
       log: 'trace'
     });
+    const shard_size = 10000;
+    const body = bodybuilder()
```
denis-yuen: looks better :+1:

src/app/search/search.component.ts
```diff
@@ -103,10 +98,24 @@ export class SearchComponent implements OnInit {
...
+      .aggregation('terms', '_type', { size: shard_size }, '_type')
+      .aggregation('terms', 'registry', { size: shard_size }, 'registry')
+      .aggregation('terms', 'private_access', { size: shard_size }, 'private_access')
+      .aggregation('terms', 'tags.verified', { size: shard_size }, 'tags_verified')
+      .aggregation('terms', 'author', { size: shard_size }, 'author')
+      .aggregation('terms', 'namespace', { size: shard_size }, 'namespace')
+      .aggregation('terms', 'labels.value', { size: shard_size }, 'labels_value')
+      .aggregation('terms', 'tags.verifiedSource', { size: shard_size }, 'tags_verifiedSource')
+      .query('match_all', {})
+      .size(500);
```
denis-yuen: For later, probably don't want to retrieve 500, probably better to grab a page worth at a time

src/app/search/search.component.ts
```diff
@@ -155,52 +164,37 @@ export class SearchComponent implements OnInit {
     });
   }
 
+  setupBuckets2(key, buckets: any) {
```
denis-yuen: Looks much better, do add comments before we forget what this does though

src/app/search/search.component.ts
```diff
@@ -160,39 +160,51 @@ export class SearchComponent implements OnInit {
...
-        if (this.buckets.get(key) == null) {
-          this.buckets.set(key, new Map<string, string>());
-          if (!this.setFilter) {
-            this.fullyExpandMap.set(key, false);
-            this.checkboxMap.set(key, new Map<string, boolean>());
-          }
-        }
-        this.buckets.get(key).set(bucket.key, bucket.doc_count);
+  /**
+   * Partially constructs the buckets, fullyExpandMap, and checkboxMap data structures
```
denis-yuen: Comments should give more insight into what we're trying to accomplish rather than mechanistically describing what is occurring. Why are these data structures being created?

Some good thoughts along these lines  https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/
https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/

PR #50, opened 2017-07-17
Feature/missing buckets

garyluu:
This fixes several things:

- Changes the registry values to friendly names
- Use must filter for certain terms like tags.verified and public/private
- Retains the bucket in the sidebar if it was checked (even if currently empty)


src/app/search/search.component.ts
```diff
@@ -378,20 +399,15 @@ export class SearchComponent implements OnInit {
...
-        let modifiedInnerFilterValue = key;
-        // private_access is the only category we do not want modify
-        if (key !== 'private_access') {
-          modifiedInnerFilterValue = key.substring(0, 1) + key.substring(1).replace(/_/g, '.');
-        }
-        if (aggKey === key) {
-          // Return some garbage output because we've decided to append a filter, there's no turning back
+        if (aggKey === key && this.searchService.exclusiveFilters.indexOf(key) === -1) {
+          // Return some garbage filter because we've decided to append a filter, there's no turning back
           // return body;  // <--- this does not work
```
garyluu: I'll leave this in because I think most people who're trying to replace the garbage filter would try to use this line first.

PR #52, opened 2017-07-17
Feature/default version

garyluu:
- Allows swagger to generate model ts files
- Added some endpoints to container and workflows (mostly generated from swagger)
- Added default versioning system to both tools and workflows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
 before_script:
   - pwd
   - wget http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/2.2.3/swagger-codegen-cli-2.2.3.jar -O swagger-codegen-cli.jar
-  - java -jar swagger-codegen-cli.jar generate -i https://dockstore.org:8443/swagger.json -l typescript-angular2 -o src/app/shared/
+  - java -jar swagger-codegen-cli.jar generate -i https://dockstore.org:8443/swagger.json -l typescript-angular2 -o src/app/shared/swagger
   - ./script
   - npm run-script prebuild
   - ng lint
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -21,6 +21,10 @@ install:
   - 'ng version'
 
 before_script:
+  - pwd
+  - wget http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/2.2.3/swagger-codegen-cli-2.2.3.jar -O swagger-codegen-cli.jar
+  - java -jar swagger-codegen-cli.jar generate -i https://dockstore.org:8443/swagger.json -l typescript-angular2 -o src/app/shared/swagger
```
denis-yuen: This could get weird if the production site gets much behind ui2.
I wonder if this could just be pointed at a tagged version of https://github.com/ga4gh/dockstore/blob/develop/dockstore-webservice/src/main/resources/swagger.yaml

.travis.yml
```diff
@@ -23,7 +23,7 @@ install:
 before_script:
   - pwd
   - wget http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/2.2.3/swagger-codegen-cli-2.2.3.jar -O swagger-codegen-cli.jar
-  - java -jar swagger-codegen-cli.jar generate -i https://dockstore.org:8443/swagger.json -l typescript-angular2 -o src/app/shared/swagger
+  - java -jar swagger-codegen-cli.jar generate -i https://raw.githubusercontent.com/ga4gh/dockstore/develop/dockstore-webservice/src/main/resources/swagger.yaml -l typescript-angular2 -o src/app/shared/swagger
```
denis-yuen: you're probably going to want a tagged version or at least a specific SHA1 commit unless you want this to break an inconvenient times

PR #53, opened 2017-07-21
Feature/workflow stub version editor

garyluu:
ga4gh/dockstore#809

This is the first pull request.  This adds:
- Workflow version editor
- Workflow stubbing
- Fix when no DAG exists
- Swagger web service API

PR #54, opened 2017-07-21
Feature/add per organization

garyluu:
ga4gh/dockstore#746
This is the 2nd pull request
- Allows registration for each organization (the '+' sign) both tool and workflows


denis-yuen:
Note to self, review after #53 so I can see the correct changed files

src/app/container/version-modal/version-modal.component.ts
```diff
@@ -0,0 +1,242 @@
...
+      case TagEditorMode.Add: {
+        this.addTag();
+        break;
+      }
+      case TagEditorMode.Edit: {
+        this.editTag();
+        break;
+      }
+      default: {
+        console.log('No idea how you submitted in neither edit or add mode');
```
denis-yuen: Let's make this a bit more formal

src/app/container/version-modal/version-modal.component.ts
```diff
@@ -0,0 +1,242 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+
+  editTag() {
+    console.log('Editing tag...');
+    const newCWL = this.unsavedCWLTestParameterFilePaths.filter(x => this.savedCWLTestParameterFilePaths.indexOf(x) === -1);
```
denis-yuen: Might be a clearer way of doing this like includes https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

src/app/container/version-modal/version-modal.component.ts
```diff
@@ -0,0 +1,242 @@
...
+        break;
+      }
+      case DescriptorType.WDL: {
+        const newTestFile = this.unsavedTestWDLFile;
+        this.unsavedWDLTestParameterFilePaths.push(newTestFile);
+        this.unsavedTestWDLFile = '';
+        break;
+      }
+      default: {
+        console.log('No idea how you submitted in neither edit or add mode');
```
denis-yuen: Same here

src/app/container/version-modal/version-modal.component.ts
```diff
@@ -0,0 +1,242 @@
...
+      case DescriptorType.CWL: {
+        this.unsavedCWLTestParameterFilePaths.splice(index, 1);
+        break;
+      }
+      case DescriptorType.WDL: {
+        this.unsavedWDLTestParameterFilePaths.splice(index, 1);
+        break;
+      }
+      default: {
+        console.log('No idea how you submitted in neither edit or add mode');
```
denis-yuen: This seems repeated multiple times as well, DRY, make this a method and change it in just one place.

PR #55, opened 2017-07-21
Feature/info tab edit

garyluu:
ga4gh/dockstore#745

This is the third pull request.
- Moved workflow and tool info tabs to its own component
- Allows workflow and tool info tabs to be edited
- Fix for old DAG displaying when new DAG is error


denis-yuen:
Hmm, I don't think GitHub is showing me a proper subset of changes as expected

garyluu:
Not sure what changed, but it looks like it's showing the correct changes now.

PR #56, opened 2017-07-24
Feature/default test parameter files

garyluu:
This is the 4th pull request.

This updates the UI to include default test parameter files in the info tab and register tool/workflow modals.  The webservice also needs to be updated.


denis-yuen:
Holy conflicts Batman.  
Are there whitespace clashes going on here or?

PR #57, opened 2017-07-24
Feature/search sortby order

JWKaiqi:
- sort category by both count amount and alphabet, with ascending and descending order.
- put selected items on the top of the list for each category.


src/app/search/search.component.ts
```diff
@@ -443,14 +396,254 @@ export class SearchComponent implements OnInit {
...
+      }
+    } else {
+      if (!this.filters.has(category)) {
+        this.filters.set(category, new Set<string>());
+      }
+      this.filters.get(category).add(categoryValue);
+    }
+  }
+  sortByAlphabet(orderedArray, orderMode): any {
+    orderedArray = orderedArray.sort(function (a, b) {
```
garyluu: Just use lambda instead of function

src/app/search/search.component.ts
```diff
@@ -443,14 +396,254 @@ export class SearchComponent implements OnInit {
...
+      } else {
+        return !orderMode ? -1 : 1;
+      }
+    });
+    return orderedArray;
+  }
+
+  sortCategoryValue (valueMap: any, sortMode: boolean, orderMode: boolean): any {
+    let orderedArray = <any>[];
+    valueMap.forEach(
```
garyluu: This looks like you're copying the array.  If it is, just use one of Object.assign or Object.create instead.

src/app/search/search.component.ts
```diff
@@ -443,14 +396,254 @@ export class SearchComponent implements OnInit {
...
+      this.activeToolBar = false;
+    } else if (this.workflowHits.length === 0 && this.toolHits.length > 0) {
+      this.activeToolBar = true;
+    } else {
+      this.activeToolBar = true;
+    }
+  }
+}
+
+export class CategorySort {
```
garyluu: Rule of One.  One class per file, probably move these 2 to the model folder or a new model folder

PR #60, opened 2017-07-31
Feature/default settings

JWKaiqi:
Issue ga4gh/dockstore#600 have been updated:
- Implemented the default page setting, which redirect the user to their latest visited page instated of the first page.
- Expand the accordion that contains the newly created tool/workflow for mytool/myworkflow page. 
- Saves user's selection filters for the search interface.
- Removed paging and searching functionalities for the home page entry table.


denis-yuen:
?

PR #61, opened 2017-07-31
Feature/default settings

JWKaiqi:
Issue ga4gh/dockstore#600 have been updated:

Implemented the default page setting, which redirect the user to their latest visited page instated of the first page.
Expand the accordion that contains the newly created tool/workflow for mytool/myworkflow page.
Saves user's selection filters for the search interface.
Removed paging and searching functionalities for the home page entry table.


src/app/shared/dockstore.model.ts
```diff
@@ -1,11 +1,11 @@
 export class Dockstore {
 
-  static readonly API_URI = 'http://localhost:8080';
+  static readonly API_URI = 'http://10.11.9.73:8080';
```
denis-yuen: I think this was checked-in accidentally

src/app/shared/dockstore.model.ts
```diff
@@ -1,11 +1,11 @@
...
   static readonly DNASTACK_IMPORT_URL= 'https://app.dnastack.com/#/app/workflow/import/dockstore';
 
-  static readonly GITHUB_CLIENT_ID = 'fill this in';
+  static readonly GITHUB_CLIENT_ID = '4e52429e7506af0a634c';
```
denis-yuen: Better invalidate this

PR #64, opened 2017-08-02
Feature/verified filter

JWKaiqi:
Issue ga4gh/dockstore#821 have been updated.


src/app/search/search.component.ts
```diff
@@ -210,17 +201,43 @@ export class SearchComponent implements OnInit {
...
+        }
       }
       if (!this.setFilter) {
         this.checkboxMap.get(key).set(bucket.key, false);
       }
     });
   }
 
+  /** this function will get the actual count number of non-verified items,
+   * Not the best way to do it, try to improve this later.
```
denis-yuen: Could you elaborate on what is occurring here actually and how it could be improved?

JWKaiqi: yes, so I added a comment above this function which explains what is doing.

PR #65, opened 2017-08-03
Fix for when no personal tools or workflows exist

garyluu:
Fix for when the user does not have any tools or workflows.  There were also some small reformatting.

### DIFF
--- a/src/app/myworkflows/myworkflows.component.html
+++ b/src/app/myworkflows/myworkflows.component.html
@@ -16,8 +16,7 @@
         </p>
         <ul>
           <li>
-            Go to 'Accounts' and link a Quay.io account that is set up to build Docker images from GitHub, Gitlab, or Bitbucket. Workflows will be synchronized
-            in the linking process.
+            Press the 'Refresh All Workflows' button to create workflow stubs for all of your soruce control repositories.  Browse to the workflow stubs and refresh/publish those that correspond to real workflows.
           </li>
           <li>
             Press the 'Register Workflow' button to manually add a workflow.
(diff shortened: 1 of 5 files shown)


src/app/myworkflows/myworkflows.component.html
```diff
@@ -1,47 +1,61 @@
...
+    <div class="col-md-12">
+      <div class="alert alert-info" role="alert">
+        <p>
+          <span class="glyphicon glyphicon-info-sign"></span>
+          There are currently no workflows registered under this account, to add your first workflow, do one of the following:
+        </p>
+        <ul>
+          <li>
+            Go to 'Accounts' and link a Quay.io account that is set up to build Docker images from GitHub, Gitlab, or Bitbucket. Workflows will be synchronized
+            in the linking process.
```
denis-yuen: Workflows don't use quay.io, replace with 

"Press "Refresh all workflows" to create workflow stubs for all of your source control repositories. Browse to the workflow stubs and publish those that correspond to real workflows."

src/app/myworkflows/myworkflows.component.html
```diff
@@ -16,8 +16,7 @@
         </p>
         <ul>
           <li>
-            Go to 'Accounts' and link a Quay.io account that is set up to build Docker images from GitHub, Gitlab, or Bitbucket. Workflows will be synchronized
-            in the linking process.
+            Press the 'Refresh All Workflows' button to create workflow stubs for all of your soruce control repositories.  Browse to the workflow stubs and refresh/publish those that correspond to real workflows.
```
denis-yuen: Typo for soruce -> source

PR #68, opened 2017-08-09
Feature/swagger fix

garyluu:
This fixes 3 things:

- Use the Dockstore.API_URI instead of localhost for all swagger API endpoitns
- The typo in the tooltip for one of the advanced search fields
- No workflow is shown in the myworkflows page due to naming mismatch. 
- Fixed upper/lower case and symbols for matching descriptions

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -1,3 +1,4 @@
+import { Dockstore } from './shared/dockstore.model';
 import { Configuration } from './shared/swagger/configuration';
 import { ApiModule } from './shared/swagger/api.module';
 import { StateService } from './shared/state.service';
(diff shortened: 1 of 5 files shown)


src/app/app.module.ts
```diff
@@ -152,7 +153,7 @@ export class AppModule {
 
 export const apiConfig = new Configuration({
   accessToken: '',
```
denis-yuen: what is the access token for in this context?

garyluu: Placeholder for now.  It apparently does not work correctly for authentication.  If it worked.  It would eliminate the need for generated files modifications.

PR #74, opened 2017-09-07
Feature/bug fixes

garyluu:
This contains a variety of small miscellaneous bug fixes.  There are still more small bug fixes to come.  This is the first pull request.  It:

- Also upgrades Angular, Angular-CLI, codelyzer, tslint...etc.


package.json
```diff
@@ -14,22 +14,20 @@
   },
   "private": true,
   "dependencies": {
-    "@angular/animations": "4.1.3",
-    "@angular/common": "4.1.3",
-    "@angular/compiler": "4.1.3",
-    "@angular/compiler-cli": "4.1.3",
-    "@angular/core": "4.1.3",
+    "@angular/animations": "^4.2.4",
```
denis-yuen: I see that this was changed from specific version numbers to version ranges. 
I believe that we changed to specific versions to avoid random failures when underlying dependencies changed. 
Is there a reason this is changing that back?

src/app/containers/list/list.component.html
```diff
@@ -19,10 +19,8 @@
             {{ tool?.starredUsers.length === 0 ? '' : tool?.starredUsers.length }}
           </td>
           <td>
-            <span id="verifiedIcon" *ngIf="getVerified(tool)">
-              <a href={{verifiedLink}}>
-                <span class="glyphicon glyphicon-ok" tooltip="Verified"></span>
-              </a>
+            <span  *ngIf="getVerified(tool)">  
+                <a style="text-decoration: none" tooltip="Verified" class="glyphicon glyphicon-ok"href={{verifiedLink}}></a>
```
denis-yuen: Cool, thanks for handling this detail

package.json
garyluu: When upgrading, by default it will change it to the ranges.  I suppose we can change it to fixed or "~" if needed.

denis-yuen: I think fixed makes more sense

PR #75, opened 2017-09-07
Tests

garyluu:
This adds karma unit tests.  This is the second pull request.  Quite a bit of refactoring was done too.  Namely using abstract classes more, removal of all custom written API calls in favor of using Swagger generated ones (so we don't have to unit test them).  At the time of this pull request, coverage reached about 75%.  This is the preliminary unit testing.  There will need to be more refinement later (higher coverage, better component tests).  A coverage badge will be displayed on develop once this pull request is merged.  

Also, this requires a small modification to the dockstore web service in order for swagger client library to work.  Swagger expects no content requests to return a no content HTTP code instead of OK.

Side note:  This is a large pull request, but all of it is just refactoring and tests.


src/app/containers/list/list.service.ts
```diff
@@ -1,14 +1,27 @@
...
     let dockerPullCmd = 'docker pull ';
+
+    // The below handles docker hub registries
     const prefix = 'registry.hub.docker.com/';
     if (path.indexOf(prefix) !== -1) {
       path = path.replace(prefix, '');
     }
+
+    // Not sure what this handles
     if (path.indexOf('_/') !== -1) {
```
denis-yuen: Looks like its killing underscores. Not sure why

denis-yuen:
Also some rebasing would be helpful

src/app/containers/list/list.service.ts
garyluu: This is in the original code, I'm not sure why it's needed.

src/app/container/paramfiles/paramfiles.service.spec.ts
```diff
@@ -0,0 +1,158 @@
+import { Tag } from './../../shared/swagger/model/tag';
```
denis-yuen: One thought, shouldn't tests belong in a specific directory rather than along side regular code like `src/app/container/paramfiles/paramfiles.component.ts`

src/app/search/search.component.spec.ts
```diff
@@ -1,25 +1,34 @@
...
+import { SearchStubService } from '../test/service-stubs';
+import { SearchService } from './search.service';
+import { ProviderService } from './../shared/provider.service';
+import { NO_ERRORS_SCHEMA } from '@angular/core';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 
 import { SearchComponent } from './search.component';
 
 describe('SearchComponent', () => {
-  let component: SearchComponent;
```
denis-yuen: This file seems totally commented out, does it do anything anymore?

src/app/container/paramfiles/paramfiles.service.spec.ts
garyluu: Angular guide suggests it this way, and Angular-CLI generates it this way.

https://angular.io/guide/styleguide#style-04-06

src/app/search/search.component.spec.ts
garyluu: The tests weren't passing and Jennifer's changes will probably modify the search component further.  Leaving this test for later.  My misc ui2 bugs issue indirectly references it.

src/app/container/error.service.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+        });
+    });
+
+    it('should be created', inject([ErrorService], (service: ErrorService) => {
+        expect(service).toBeTruthy();
+    }));
+    it('should have multiple docs', inject([ErrorService], (service: ErrorService) => {
+        service.setToolRegisterError({'status': 'status', 'statusText': 'statusText', '_body': '_body'});
+        const errorObj: any = {
+            message: 'The webservice encountered an error trying to create modify.',
```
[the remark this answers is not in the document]
garyluu: Added this to my misc bugs

PR #78, opened 2017-09-18
Feature/bugfix search

agduncan94:
This PR has working tag cloud and auto-complete for search, done by Jennifer.

I have commented out/ngif out code related to permalinks, as that will be done in another branch. We can use the existing permalink code as a base to start the implementation.

Also, made a rookie mistake and committed my dockstore.model.ts, which I have reverted and recreated a new application on github (the github client ID is visible in the history).

I have also fixed angular versions to 4.3.6. It seems 4.4.1 does not work with datatables. I will create an issue regarding this.


package.json
```diff
@@ -14,16 +14,16 @@
   },
   "private": true,
   "dependencies": {
-    "@angular/animations": "^4.2.4",
-    "@angular/common": "^4.2.4",
-    "@angular/compiler": "^4.2.4",
-    "@angular/core": "^4.2.4",
+    "@angular/animations": "4.3.6",
```
denis-yuen: Wonder if it is possible to use a variable to keep these in sync like http://ianmcnally.me/blog/2016/2/4/variables-in-npm-scripts

src/app/search/search.routing.ts
```diff
@@ -0,0 +1,13 @@
+import { RouterModule, Routes } from '@angular/router';
+
+import { SearchComponent } from './search.component';
+
+const CONTAINERS_ROUTES: Routes = [
+  {
+    path: 'admin-search', component: SearchComponent, children: [
```
denis-yuen: need to get rid of the "admin" part

package.json
agduncan94: This doesn't seem to work, when I do this it thinks the variable name is the version I am trying to install.

denis-yuen: Rats

PR #82, opened 2017-09-29
Feature/permalinks

agduncan94:
* Support for permalinks
* Fixed some package.json dependencies


.gitignore
```diff
@@ -40,6 +40,7 @@ npm-debug.log
 testem.log
 /typings
 src/app/footer/versions.ts
+cypress/videos
```
denis-yuen: Good call :+1:

src/app/shared/dockstore.model.ts
```diff
@@ -1,15 +1,19 @@
 export class Dockstore {
+  static readonly HOSTNAME = 'http://localhost';
```
denis-yuen: Maybe add a comment here indicating that a user should fill this in with something more appropriate for login/integration purposes

PR #83, opened 2017-09-29
Feature/login fix

garyluu:
- Fixes ga4gh/dockstore#911
- Gives Read:Org access for GitHub
- Handle when user was logged in but is no longer logged in error
- Remove a few console logs


src/app/shared/dockstore.model.ts
```diff
@@ -10,8 +10,9 @@ export class Dockstore {
 
   static readonly GITHUB_CLIENT_ID = 'fill this in';
   static readonly GITHUB_AUTH_URL = 'https://github.com/login/oauth/authorize';
+
   static readonly GITHUB_REDIRECT_URI = Dockstore.LOCAL_URI + '/auth/github.com';
-  static readonly GITHUB_SCOPE = 'user,user:email';
+  static readonly GITHUB_SCOPE = 'read:org,user,user:email';
```
denis-yuen: FYI, please update https://github.com/dockstore/compose_setup/blob/master/templates/dockstore.model.ts.template now with a corresponding change

garyluu: Pull request made.

PR #84, opened 2017-09-29
Feature/search refactoring

agduncan94:
* Front page search now links to search page
* Fixed search icon
* Removed search from datatables on search page
* If over 200 results, a message will be displayed (Is the message ok?)
* Moved some of the helper/initialisation functions for search component to search service

### DIFF
--- a/src/app/search/search.component.html
+++ b/src/app/search/search.component.html
@@ -168,7 +168,7 @@
           </div>
         </div>
 
-        <div class="alert alert-info" *ngIf="hits !== null && hits.length > query_size - 1">
+        <div class="alert alert-info" *ngIf="hits.length > query_size - 1">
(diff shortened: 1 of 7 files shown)


src/app/search/search.component.html
```diff
@@ -167,6 +167,12 @@
             </span>
           </div>
         </div>
+
+        <div class="alert alert-info" *ngIf="hits.length > query_size - 1">
+          <p><strong>Notice: </strong>Your search has returned greater than {{ query_size - 1 }} results, however only up to {{ query_size - 1 }} results are shown.
```
denis-yuen: If I understand correctly, this isn't actually "up to {{ query_size - 1 }} results", it should be precisely "{{ query_size - 1 }} results" 
i.e. "Your search has returned greater than {{ query_size - 1 }} results, however only {{ query_size - 1 }} results are shown. We recommend that you narrow your search to find more relevant results."

agduncan94: Yes good catch

PR #87, opened 2017-10-05
Feature/ng after view checked

garyluu:
This fixes (or works-around) the ExpressionChangedAfterItHasBeenCheckedError errors that we kept getting when enableProd() is removed ga4gh/dockstore#853:

Additionally, several other things were changed:
- all bodybuilder query building functions have been moved to its own query-builder service, no other service or component should be using bodybuilder
- search.component default test have been enabled
- default query-builder test is added
- search table (along with the tag cloud) has been moved to the search-results component

In the future, tag cloud should be its own component.  And the sidebar should also be its own component.

### DIFF
--- a/src/app/search/query-builder.service.ts
+++ b/src/app/search/query-builder.service.ts
@@ -1,7 +1,8 @@
-import { SearchService } from './search.service';
-import { AdvancedSearchObject } from './../shared/models/AdvancedSearchObject';
 import { Injectable } from '@angular/core';
 import bodybuilder from 'bodybuilder';
+
+import { AdvancedSearchObject } from './../shared/models/AdvancedSearchObject';
+import { SearchService } from './search.service';
 
 /**
  * This service constructs all the querys and should be only class that interacts with the bodybuilder library.
--- a/src/app/search/query-builder.service.ts
+++ b/src/app/search/query-builder.service.ts
@@ -12,6 +12,7 @@ import { SearchService } from './search.service';
  */
 @Injectable()
 export class QueryBuilderService {
+    // TODO: Comment on why shard_size is 10,000
     private shard_size = 10000;
     constructor(private searchService: SearchService) { }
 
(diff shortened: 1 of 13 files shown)


garyluu:
wow, that's a pretty huge coverage decrease.

src/app/search/query-builder.service.ts
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * This service constructs all the querys and should be only class that interacts with the bodybuilder library.
+ *
+ * @export
+ * @class QueryBuilderService
+ */
+@Injectable()
+export class QueryBuilderService {
+    private shard_size = 10000;
```
denis-yuen: A comment why the shard size was chosen to be 10,000 would be useful here

src/app/search/search-results/search-results.component.ts
```diff
@@ -0,0 +1,133 @@
...
+  // Tells the search service to tell the search filters to save its data
+  saveSearchFilter() {
+    this.searchService.toSaveSearch$.next(true);
+  }
+
+  tagClicked(clicked: CloudData) {
+    this.searchService.searchTerm$.next(true);
+    this.searchService.values$.next(clicked.text);
+    this.searchService.tagClicked$.next(true);
+    // this.searchTerm = true;
```
denis-yuen: Delete this or put it in a descriptive if/else

src/app/search/search.component.ts
```diff
@@ -1,23 +1,22 @@
...
+import { AdvancedSearchService } from './advancedsearch/advanced-search.service';
+import { ELASTIC_SEARCH_CLIENT } from './elastic-search-client';
+import { QueryBuilderService } from './query-builder.service';
+import { SearchService } from './search.service';
 
 /** TODO: ExpressionChangedAfterItHasBeenCheckedError is indicator that something is wrong with the bindings,
  *  so you shouldn't just dismiss it, but try to figure out why it's happening...
  **/
-enableProdMode();
+// enableProdMode();
```
denis-yuen: Same here

PR #91, opened 2017-10-05
Feature/issue 933 basic

garyluu:
Fixes ga4gh/dockstore#933 by putting a link in the footer to Dockstore UI1.  Also fixed the missing tag cloud styles introduced during the search refactoring.

### DIFF
--- a/src/app/footer/footer.component.ts
+++ b/src/app/footer/footer.component.ts
@@ -1,3 +1,4 @@
+import { Dockstore } from './../shared/dockstore.model';
 import { Metadata } from './../shared/swagger/model/metadata';
 import { Component, OnInit } from '@angular/core';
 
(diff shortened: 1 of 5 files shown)


src/app/footer/footer.component.ts
```diff
@@ -12,6 +13,8 @@ import { GA4GHService } from '../../../src/app/shared/swagger/api/gA4GH.service'
 export class FooterComponent implements OnInit {
   version: string;
   tag: string;
+  public prod = true;
```
denis-yuen: `prod=true` is probably useless this way. 
the port for the U1_WEBSITE can still come from `https://github.com/dockstore/dockstore-ui2/blob/develop/src/app/shared/dockstore.model.ts`

garyluu: - I originally had the UI1_PORT in the dockstore.model.ts, but that involved changing the dockstore.model.ts (merge conflicts with you)
- I also originally had prod set via the environment files but was told not to.  So change prod to false/true to toggle whether or not the link should be present.

PR #93, opened 2017-10-06
Feature/refresh indicator

garyluu:
This fixes the refresh workflow/tool alert message to show up when any tool or workflow is refreshed.

- The refreshing$ boolean observable was replaced by the refreshMessage$ string observable.  A non-null refreshMessage indicates that a tool/workflow is being refreshed.
- The refreshMessage$ observable is hooked up to all the refresh buttons and add workflow/tool buttons to disable them when any of them are currently running.
- The refresh alerts are seperate between the workflow and tool.  You can still refresh one workflow, then go to the mytools page and refresh another tool.

- This also fixes the grammar in the error obj message.

Additionally,

- Moves error service into the shared directory
- Renames refresh-alert to alert because it it will contain both error and refresh alerts.
- Renamed toolErrorObj to errorObj because it's shared.


garyluu:
Have to fix merge conflicts and also adding some other changes too.  Will need to re-approve.

garyluu:
And...more merge fixes

denis-yuen:
Sorry

PR #99, opened 2017-10-10
ts-node needed as a production dependency

denis-yuen:


npm-shrinkwrap.json
```diff
@@ -2,55 +2,11 @@
   "name": "dockstore-ui2",
   "version": "0.0.0",
   "dependencies": {
```
agduncan94: Seems like there is no angular-cli in the shrinkwrap, so no 'ng' installed.

agduncan94: Looks like it was removed in two places from the shrink-wrap, though that is weird since it was only removed from dev dependencies.

denis-yuen: That brings up an interesting point. For the compose_setup, I install the cli globally. https://github.com/dockstore/compose_setup/blob/master/templates/Dockerfile_ui2.template#L17 based on the readme. https://github.com/dockstore/dockstore-ui2/blob/feature/ts-node_move/README.md#bower

The .travis.yml does something different which is a mismatch

scripts/install-travis-script.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -o errexit
+set -o pipefail
+set -o nounset
+set -o xtrace
+
+if [ "$RUN_PROD" = true ]; then
+  npm install -g bower
+  npm install --production
+  bower update
```
agduncan94: Is there a reason we don't also use bower install here?

denis-yuen: Nope, just missed it. Thanks

PR #105, opened 2017-10-12
Feature/use models

garyluu:
Start taking advantage of swagger models and types.  There should be many more changes like this at some later date. 

This also uses the swagger DockstoreTool.ModeEnum.


denis-yuen:
FYI, definitely an improvement, but I think we should be switching gears to focus on testing for the next little while as opposed to this kind of improvement which can wait for 1.4

PR #107, opened 2017-10-13
Feature/version view fix

garyluu:
Fix some tooltips, prevent highlighting a non-existent element, do dag stuff when theres a dag

### DIFF
--- a/src/app/container/dockerfile/dockerfile.component.ts
+++ b/src/app/container/dockerfile/dockerfile.component.ts
@@ -56,7 +56,7 @@ export class DockerfileComponent extends VersionSelector implements AfterViewChe
     }
   }
   ngAfterViewChecked() {
-    if (this.contentHighlighted && !this.nullContent) {
+    if (this.contentHighlighted && !this.nullContent && this.elementRef.nativeElement.querySelector('.highlight')) {
       this.contentHighlighted = false;
       this.highlightJsService.highlight(this.elementRef.nativeElement.querySelector('.highlight'));
     }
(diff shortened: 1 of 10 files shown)


src/app/workflow/dag/dag.component.ts
```diff
@@ -155,6 +152,9 @@ export class DagComponent implements OnInit, AfterViewChecked, OnChanges {
         }
       }
     });
+    } else {
+      console.log('DAG is not truthy');
```
denis-yuen: Log message is not very useful, make it more descriptive

garyluu: Was for development purposes, I'll probably remove it altogether.

src/app/container/dockerfile/dockerfile.component.ts
```diff
@@ -56,7 +56,7 @@ export class DockerfileComponent extends VersionSelector implements AfterViewChe
     }
   }
   ngAfterViewChecked() {
-    if (this.contentHighlighted && !this.nullContent) {
+    if (this.contentHighlighted && !this.nullContent && this.elementRef.nativeElement.querySelector('.highlight')) {
```
denis-yuen: Not sure if this is a thing in typescript, but does `this.elementRef.nativeElement` need a nullcheck or anything similar?

garyluu: I'm currently checking if it's truthy, it previously wasn't needed but somehow it is now.

PR #109, opened 2017-10-13
Only have 'Add Tag' in manual tools, slight formatting

garyluu:
For issue ga4gh/dockstore#951

### DIFF
--- a/src/app/container/versions/versions.component.ts
+++ b/src/app/container/versions/versions.component.ts
@@ -1,5 +1,3 @@
-import { RefreshService } from '../../shared/refresh.service';
-import { NotificationsService } from 'angular2-notifications';
 /*
  *    Copyright 2017 OICR
  *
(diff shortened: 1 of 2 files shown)


src/app/container/versions/versions.component.ts
```diff
@@ -52,8 +52,16 @@ export class VersionsContainerComponent extends Versions implements OnInit {
...
+    if (this.tool && this.tool.mode === DockstoreTool.ModeEnum.MANUALIMAGEPATH) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
   setNoOrderCols(): Array<number> {
-    return [ 5, 6 ];
+    return [5, 6];
```
denis-yuen: Not a focus of this PR, but this seems likely to break at some point as new columns are added and removed

garyluu: Not actually sure what this does.  Workflow doesn't have it, just tools.

denis-yuen: Does anything break if its removed? Worth checking out in a subsequent branch

garyluu: I removed it at some point and at a glance did not notice any significant changes.  Will definitely take a closer look later.

garyluu: setNoOrderCols makes the verified and action columns unsortable and there are different columns in the workflows vs tools.  So it does look necessary.

PR #112, opened 2017-10-16
Feature/maintenance

garyluu:
Simple way to redirect to maintenance page via footer error.  

More advanced fix is included (and then reverted) because:
- It uses an http interceptor which is recommended to use with...
- the httpClient library which requires...
- a swagger codegen update which requires...
- many other fixes in enums, script.sh, api parameters, etc

Current limitations without the advanced fix:
- The page will only redirect to the maintenance page once (since footer is loaded once).  If the footer is already loaded, and then the webservice goes down, the maintenance page will not appear.

Note: This will be a merge commit so that the advanced fix can be cherry-picked easier later. https://github.com/dockstore/dockstore-ui2/pull/112/commits/700895b5b3e1b04b7bff6c21b933180803b1ef12

This also fixes the /starred page where the the tool alert looks different from the workflow alert.


src/app/maintenance/maintenance.component.ts
```diff
@@ -0,0 +1,15 @@
...
+  selector: 'app-maintenance',
+  templateUrl: './maintenance.component.html',
+  styleUrls: ['./maintenance.component.css']
+})
+export class MaintenanceComponent implements OnInit {
+
+  constructor() { }
+
+  ngOnInit() {
+  }
```
agduncan94: Just out of interest, do we actually need the empty constructor and ngOnInit?

garyluu: No, this was auto-generated, but I guess I should remove it

PR #114, opened 2017-10-16
Trims search result return by about 75%

denis-yuen:


denis-yuen:
Working on https://github.com/ga4gh/dockstore/issues/958

denis-yuen:
Working on https://github.com/ga4gh/dockstore/issues/958

src/app/search/query-builder.service.ts
```diff
@@ -44,13 +45,20 @@ export class QueryBuilderService {
...
         sidebarBody = this.appendAggregations(count, sidebarBody, bucketStubs, filters, sortModeMap);
         const builtSideBarBody = sidebarBody.build();
         const sideBarQuery = JSON.stringify(builtSideBarBody);
         return sideBarQuery;
     }
 
+    private excludeContent(body: any) {
+      // TODO: it should be possible to exclude tags and workflowVersions too
+      // however, it currently breaks the contents of the datatables for some reason
+      return body.rawOption('_source', {'excludes': ['*.content', '*.sourceFiles', 'description']});
```
agduncan94: Try [_sourceExclude](https://github.com/danpaz/bodybuilder#other-options) for bodybuilder

garyluu: Excluding workflowVersions causes datatables to break because it's using the workflowVersion to check whether the tool/workflow is verified or not

denis-yuen: Oh, seems like next version we need to equivalent of a sql `group by`

agduncan94: Could we exclude the attributes for versions except for the verified field?

garyluu:
Can straight up exclude the entire _source from everything except resultsQuery.

src/app/search/query-builder.service.ts
denis-yuen: Yup, as a temporary fix

PR #116, opened 2017-10-17
Feature/issue 723

garyluu:
For ga4gh/dockstore#723 .  The version table was protruding outside the tab contents for certain tools.  This is caused by the table contents simply being too long.  This is first bandage fixed by word-wrapping the version column when it gets too long.  It is then fixed with table-responsive so that whenever the table contents are too long, the table will be horizontally scroll-able.

This also fixes the starring component trying to use the user object when it already knows it cant' get it.

### DIFF
--- a/src/app/container/view/view.component.css
+++ b/src/app/container/view/view.component.css
@@ -13,8 +13,3 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-
-td {
-  word-wrap:break-word;
-  word-break:break-all;
-}
--- a/src/app/container/versions/versions.component.css
+++ b/src/app/container/versions/versions.component.css
@@ -14,3 +14,7 @@
  *    limitations under the License.
  */
 
+ .break {
+    word-break: break-word
+ }
+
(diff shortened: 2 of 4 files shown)


garyluu:
Oops, moving the style to css file.

src/app/container/view/view.component.css
```diff
@@ -12,9 +12,4 @@
  *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
- */
-
-td {
-  word-wrap:break-word;
-  word-break:break-all;
-}
+ */
```
denis-yuen: Is there any point to this empty file anymore?

garyluu: There is little point to this empty file except for the corresponding  component currently pointing to this.  We have alot of empty css files due to angular cli component generation.  I could remove this file, but it would only make sense if we removed the other empty css files as well for consistency.

In the event that we need to add component-specific styles, we need to re-create the file with the same name.

I personally would rather leave it as it is, takes little/no space with little performance hit.

PR #117, opened 2017-10-17
Feature/launch with fix

agduncan94:
* Only show discourse comments on public page
* Will always show launch information if tool/workflow has at least one valid version
* If default workflow is invalid, will select the first valid workflow to show launch information

### DIFF
--- a/src/app/container/container.component.ts
+++ b/src/app/container/container.component.ts
@@ -78,15 +78,22 @@ export class ContainerComponent extends Entry implements AfterViewInit {
       discourseUrl: Dockstore.DISCOURSE_URL,
       discourseEmbedUrl: decodeURIComponent(window.location.href)
     };
-  }
 
+    this.publicPageSubscription();
+
+  }
+  publicPageSubscription() {
+    this.stateService.publicPage$.subscribe(publicPage => this.publicPage = publicPage);
+  }
   // Embed Discourse comments into page
   ngAfterViewInit() {
-    (function() {
-      const d = document.createElement('script'); d.type = 'text/javascript'; d.async = true;
-      d.src = (<any>window).DiscourseEmbed.discourseUrl + 'javascripts/embed.js';
-      (document.getElementsByTagName('head')[0] || document.getElementsByTagName('body')[0]).appendChild(d);
-    })();
+    if (this.publicPage){
+      (function() {
+        const d = document.createElement('script'); d.type = 'text/javascript'; d.async = true;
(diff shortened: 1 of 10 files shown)


src/app/container/container.component.ts
```diff
@@ -78,15 +78,22 @@ export class ContainerComponent extends Entry implements AfterViewInit {
       discourseUrl: Dockstore.DISCOURSE_URL,
       discourseEmbedUrl: decodeURIComponent(window.location.href)
     };
-  }
 
+    this.publicPageSubscription();
+
+  }
+  publicPageSubscription() {
```
garyluu: This is done in entry.ts already.

src/app/shared/entry.ts
```diff
@@ -52,7 +52,7 @@ export abstract class Entry implements OnInit, OnDestroy {
   constructor(private trackLoginService: TrackLoginService,
     public providerService: ProviderService,
     public router: Router,
-    private stateService: StateService,
+    public stateService: StateService,
```
garyluu: Change shouldn't be needed if no longer subscribing from workflow/container

src/app/workflow/workflow.component.ts
```diff
@@ -66,15 +66,23 @@ export class WorkflowComponent extends Entry implements AfterViewInit {
       discourseUrl: Dockstore.DISCOURSE_URL,
       discourseEmbedUrl: decodeURIComponent(window.location.href)
     };
+
+    this.publicPageSubscription();
+
+  }
+  publicPageSubscription() {
+    this.stateService.publicPage$.subscribe(publicPage => this.publicPage = publicPage);
```
garyluu: same as container

src/app/workflow/workflow.component.ts
```diff
@@ -66,15 +66,23 @@ export class WorkflowComponent extends Entry implements AfterViewInit {
...
-      d.src = (<any>window).DiscourseEmbed.discourseUrl + 'javascripts/embed.js';
-      (document.getElementsByTagName('head')[0] || document.getElementsByTagName('body')[0]).appendChild(d);
-    })();
+    if (this.publicPage) {
+      (function() {
+        const d = document.createElement('script'); d.type = 'text/javascript'; d.async = true;
+        d.src = (<any>window).DiscourseEmbed.discourseUrl + 'javascripts/embed.js';
+        (document.getElementsByTagName('head')[0] || document.getElementsByTagName('body')[0]).appendChild(d);
+      })();
+    }
```
garyluu: This looks identical to the container.component.ts, throw it into the entry.ts

PR #126, opened 2017-10-24
Fix for bitbucket and gitlab

garyluu:
Fix for the bitbucket and gitlab.  Have default in switch statements for debugging in case similar issue arise.


src/app/loginComponents/auth/auth.component.ts
```diff
@@ -51,10 +51,14 @@ export class AuthComponent implements OnInit, OnDestroy {
             return addGitHubToken;
           case 'quay':
             return addQuayToken;
```
denis-yuen: Please add some comments here. 
This seems oddly internally inconsistent (github, not github.com vs gitlab.com) and doesn't seem to quite match the webservice either https://github.com/ga4gh/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java#L163 ((quay, vs quay.io). 

We don't have to fix it across the board for this release, but leave good notes for the next developer to understand/clean this up.

garyluu:
Using enum, should be auto-enforced and self explanatory.

src/app/shared/dockstore.model.ts
```diff
@@ -30,11 +32,11 @@ export class Dockstore {
   static readonly GITHUB_CLIENT_ID = 'fill_this_in';
   static readonly GITHUB_AUTH_URL = 'https://github.com/login/oauth/authorize';
 
-  static readonly GITHUB_REDIRECT_URI = Dockstore.LOCAL_URI + '/auth/github.com';
+  static readonly GITHUB_REDIRECT_URI = Dockstore.LOCAL_URI + '/auth/' + Provider.GITHUB;
   static readonly GITHUB_SCOPE = 'read:org,user,user:email';
 
   static readonly QUAYIO_AUTH_URL = 'https://quay.io/oauth/authorize';
-  static readonly QUAYIO_REDIRECT_URI = Dockstore.LOCAL_URI + '/auth/quay';
```
denis-yuen: Double-checking whether this been tested. This seems to change `Dockstore.LOCAL_URI + '/auth/quay';` and `Dockstore.LOCAL_URI + '/auth/gitlab';` to end with quay.io and gitlab.com respectively.

garyluu: Yes, it's been tested.

src/app/shared/enum/token-source.enum.ts
```diff
@@ -0,0 +1,6 @@
+export enum TokenSource {
```
denis-yuen: Documention: these values need to match the token format returned by Dockstore

PR #129, opened 2017-10-26
Feature/issue 994

garyluu:
Sample for ga4gh/dockstore#994  .  Should further update everything at some later date.  The code in the .ts file is mostly the same as markdown except all ` must be escaped


garyluu:
Moved everything except the textarea and the button to the .ts file as markdown.  Lower the line count, same file as all the variables (cwltool version, etc).  More importantly, it removes all the html tags for readability.

PR #136, opened 2017-10-30
Feature/update documentation

agduncan94:
* Adds documentation for the new search page


src/assets/docs/search.md
```diff
@@ -0,0 +1,25 @@
+# Search
+
+Dockstore now offers faceted search, which allows for flexible querying of tools and workflows. Tabs are used to split up the results between tools and workflows. You can search for basic terms/phrases, filter using facets, and also used advanced search queries. Note that using advanced search after setting a basic search will override your basic search, and vice versa.
```
denis-yuen: "also use advanced search queries"

src/assets/docs/search.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Filtering Your Search
+### Basic Search
+If you want to do a basic search for an exact keyword or phrase, use the search box present on the left side of the page. This will search all entries for the exact string you type.
```
denis-yuen: Hmm, I think it actually functions as a "one of" the words rather than an exact string.

src/assets/docs/search.md
```diff
@@ -0,0 +1,25 @@
...
+![Basic Search](../assets/images/docs/search-basic.png)
+
+### Advanced search
+Advanced search is useful when you want to filter based on words or phrases. It extends basic search by allowing for advanced queries. You can access advanced search using the `open advanced search` button shown in the image above. The advanced search can be applied to either files or descriptions.
+![Advanced Search](../assets/images/docs/adv-search.png)
+
+### Facets
+Facets can be used to filter which entries are shown in the search results. For example, you can filter to only shows tools from Docker Hub.
```
denis-yuen: "to only show"

src/assets/docs/search.md
agduncan94: Is it better to say something like "If you want to do a basic search for a keyword or a collection of keywords, use the search box present on the left side of the page. This will search for matches that contain at least one of the keywords. Keywords should be separated  by spaces."

denis-yuen: Sounds good

PR #137, opened 2017-10-30
Feature/info tab service

garyluu:
This adds a new extendedDockstoreTool observable that extends the original DockstoreTool.  The info-tab uses it to populate its info.  Also added test to check.  This fixes the issue with the info-tab component not showing all the extended properties.


denis-yuen:
As discussed, we'll keep this open pending an explanation for the issue that originally inspired this.

garyluu:
So it looks like the properties were being set after the tool subscription observed a change.  The additionally properties shouldn't have been able to propagate to other components, but it so happened that it did since the observed value was passed by reference.

cypress/integration/mytools.js
```diff
@@ -5,6 +5,34 @@ describe('Dockstore my tools', function() {
      cy.visit(String(global.baseUrl) + "/my-tools")
   });
 
+  describe('Should contain extended DockstoreTool properties', function() {
+    it('visit another page then come back', function() {
+      cy.get('a').contains('Tools').click()
```
denis-yuen: Let's try IDs here too, and maybe use Documentation or the Accounts page because Tools or Workflows might not be long for this world

cypress/integration/mytools.js
```diff
@@ -5,6 +5,34 @@ describe('Dockstore my tools', function() {
...
+        .get('.panel')
+        .children(':nth-child(2)')
+        .contains('a', 'b1')  
+        .click()
+        cy.contains('GitHub')
+        cy.contains('https://github.com/A2/b1')
+        cy.contains('Quay.io')
+        cy.contains('quay.io/A2/b1')
+        cy.contains('Last Build')
+        cy.contains('Mar. 15 2016 at')
```
denis-yuen: Skip the dates, they're likely to change due to #514

src/app/shared/extended-tool.service.ts
```diff
@@ -0,0 +1,37 @@
...
+    extendTool(tool: DockstoreTool): ExtendedDockstoreTool {
+        if (tool) {
+            let extendededTool: ExtendedDockstoreTool = this.providerService.setUpProvider(tool);
+              extendededTool = this.providerService.setUpProvider(tool);
+              extendededTool.buildMode = this.containerService.getBuildMode(extendededTool.mode);
+              extendededTool.buildModeTooltip = this.containerService.getBuildModeTooltip(extendededTool.mode);
+              extendededTool = this.imageProviderService.setUpImageProvider(extendededTool);
+              extendededTool.agoMessage = this.dateService.getAgoMessage(new Date(extendededTool.lastBuild).getTime());
+              extendededTool.email = this.dockstoreService.stripMailTo(extendededTool.email);
+              extendededTool.lastBuildDate = this.dateService.getDateTimeMessage(new Date(extendededTool.lastBuild).getTime());
```
denis-yuen: "extendeded"

denis-yuen: Also, this seems easy to forget to update. 
Let's add a warning in big capital letters to remember to update this if we add fields and also to look out for a better solution.

src/app/shared/extended-workflow.service.ts
```diff
@@ -0,0 +1,33 @@
...
+    extendedWorkflow$: Observable<ExtendedWorkflow>;
+    constructor(private workflowService: WorkflowService, private providerService: ProviderService,
+        private imageProviderService: ImageProviderService, private dateService: DateService,
+        private dockstoreService: DockstoreService) {
+            this.extendedWorkflow$ = this.workflowService.workflow$.map((workflow: Workflow) => this.extendWorkflow(workflow));
+        }
+
+    extendWorkflow(workflow: Workflow): ExtendedWorkflow {
+        if (workflow) {
+            const extendedWorkflow: ExtendedWorkflow = this.providerService.setUpProvider(workflow);
```
denis-yuen: same here

cypress/integration/mytools.js
garyluu: Skip the values only or also 'Last Build' and 'Last Updated'?

src/app/shared/extended-tool.service.ts
garyluu: Better solution would probably just calculate those values on the fly when its actually needed instead of always storing it.

cypress/integration/mytools.js
denis-yuen: Maybe those too

PR #138, opened 2017-10-31
Feature/incorrect redirect url handling

garyluu:
For issue ga4gh/dockstore#985.  Returning null causing an error leading to page being stuck.  Fixed by returning Observable.of(null). 

Can't test with Cypress due to `Cypress does not allow you to change superdomains within a single test.`

An option would be to:

`Alternatively you can also disable Chrome Web Security which will turn off this restriction by setting { chromeWebSecurity: false } in your 'cypress.json' file.`  which I don't really recommend.

Currently using karma to test this fix.  It correctly fails with the error prior to the fix.  It correctly passes with the fix.

### DIFF
--- a/cypress/integration/dropdown.js
+++ b/cypress/integration/dropdown.js
@@ -47,7 +47,16 @@ describe('Dropdown test', function() {
                 .get('h4').contains('GitLab').contains('Link Account')
         });
     });
-
+    var everythingOk = function() {
+        cy
+        .get('h4').contains('GitHub').contains('Unlink Account')
+        cy
+        .get('h4').contains('Quay.io').contains('Unlink Account')
+        cy
+        .get('h4').contains('Bitbucket').contains('Link Account')
+        cy
+        .get('h4').contains('GitLab').contains('Link Account')
+    }
     describe('Go to setup page', function() {
         beforeEach(function() {
             // Select dropdown setup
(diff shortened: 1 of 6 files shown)


denis-yuen:
Is it possible to unit test this by mocking bitbucket, gitlab, quay.io, etc.?
Alternatively, what are the security implications of turning off security for a browser running on Travis-CI?

garyluu:
Currently unit testing it already.  By default, bitbucket and etc will return an invalid/unspecified redirect URI so it probably doesn't need to be mocked for this specific fix.  

Turning off security lets Cypress visit any sites on Travis.  If there is a problem with the test, Cypress might visit an external site, stay there and start clicking on things, etc.  Since Cypress is meant to test our site specifically, it should remain on our site.

denis-yuen:
I think there are a few workarounds you can try here, they also seem to recommend turning off anyway if no workarounds succeed https://docs.cypress.io/guides/guides/web-security.html# so I think I'm ok with it.

garyluu:
Even with web security turned off, Cypress still refuses to go to any of the four sites:
- Refused to display 'https://github.com/login?client_id=1765bb80f70177a5e284&return_to=%2Flogin%…252Fauth%252Fgithub.com%26scope%3Dread%253Aorg%252Cuser%252Cuser%253Aemail' in a frame because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'none'".
- Refused to display 'https://quay.io/oauth/authorize?client_id=KRZC9WLMZTPRFMGHQ5ZR&redirect_uri…200/auth/quay.io&response_type=token&realm=realm&scope=repo:read,user:read' in a frame because it set 'X-Frame-Options' to 'DENY, DENY'.
- Refused to display 'https://bitbucket.org/account/signin/?next=/site/oauth2/authorize%3Fclient_id%3D5m3vqvBBZFvgLVC5Ae%26response_type%3Dcode' in a frame because it set 'X-Frame-Options' to 'SAMEORIGIN'.
- Refused to display 'https://gitlab.com/users/sign_in' in a frame because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'self'".

denis-yuen:
I don't think you actually need to go to the real site in question BTW. 

I think it would be sufficient to go to a dummy page that you control just as long as it redirects improperly and causes the failure condition. Keep your eyes on the prize, the real goal here is not to simulate the real login process. The goal is to reproduce the issue, the non-responsive UI caused by an unexpected response, fix it, and then lock it down with a test.

garyluu:
- Handle all errors, not just the null observable expected check
- Cypress test incorrect url.

cypress/integration/dropdown.js
```diff
@@ -47,7 +47,16 @@ describe('Dropdown test', function() {
                 .get('h4').contains('GitLab').contains('Link Account')
         });
     });
-
+    var everythingOk = function() {
+        cy
+        .get('h4').contains('GitHub').contains('Unlink Account')
```
denis-yuen: Is there a way to use IDs rather than just headers that contain strings? 
This seems a bit fragile

cypress/integration/dropdown.js
```diff
@@ -37,25 +37,14 @@ describe('Dropdown test', function() {
...
     var everythingOk = function() {
-        cy
-        .get('h4').contains('GitHub').contains('Unlink Account')
-        cy
-        .get('h4').contains('Quay.io').contains('Unlink Account')
-        cy
-        .get('h4').contains('Bitbucket').contains('Link Account')
-        cy
-        .get('h4').contains('GitLab').contains('Link Account')
+            cy.get('#unlink-GitHub').should('be.visible')
```
denis-yuen: Maybe just reference the values from accounts.component.ts from here rather than relying on the bare strings matching?

garyluu: The IDs are dynamically generated.  I currently can't import ts files into Cypress.io .js files

denis-yuen: If you can't reference ts from js, maybe draw both from a properties file or similar?

garyluu: Had a very simple `export enum AccountInfo` file, didn't work.

PR #141, opened 2017-11-03
Note minimum Java 8 version

denis-yuen:


denis-yuen:
Note to self, add Java 9 warning

denis-yuen:
Note incompatibility with Java 9 as well

PR #145, opened 2017-11-14
Use google url shortener on permalinks if possible

agduncan94:
Uses URL shortener for permalinks.
Note dockstore.model.ts has new field GOOGLE_SHORTENER_KEY, we will need to generate an API key with a new dockstore email address

[Issue #924](https://github.com/ga4gh/dockstore/issues/924)

### DIFF
--- a/src/app/shared/dockstore.model.ts
+++ b/src/app/shared/dockstore.model.ts
@@ -25,6 +25,9 @@ export class Dockstore {
   // Discourse URL MUST end with a slash (/)
   static readonly DISCOURSE_URL = 'http://localhost/';
 
+  // Google Shortener API key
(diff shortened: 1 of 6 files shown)


src/app/shared/dockstore.model.ts
```diff
@@ -25,6 +25,9 @@ export class Dockstore {
   // Discourse URL MUST end with a slash (/)
   static readonly DISCOURSE_URL = 'http://localhost/';
 
+  // Google Shortener API key
```
denis-yuen: Maybe a link to docs for this, I think https://developers.google.com/url-shortener/v1/getting_started#APIKey

PR #146, opened 2017-11-15
added star sorting to search page

agduncan94:
* adds sort by starring to search results
* if no starred found, still show 0

[Issue #1007](https://github.com/ga4gh/dockstore/issues/1007)

### DIFF
--- a/src/app/containers/list/list.component.html
+++ b/src/app/containers/list/list.component.html
@@ -38,7 +38,7 @@
               (tool?.toolname ? '/' + tool?.toolname : '') }}</a>
           </td>
           <td>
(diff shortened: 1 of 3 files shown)


src/app/containers/list/list.component.html
```diff
@@ -38,7 +38,7 @@
               (tool?.toolname ? '/' + tool?.toolname : '') }}</a>
           </td>
           <td>
-            {{ tool?.starredUsers.length === 0 ? '' : tool?.starredUsers.length }}
```
denis-yuen: If I'm reading this right, I think I still prefer blank over 0 stars.
Thoughts?

PR #147, opened 2017-11-16
Update Cypress

garyluu:
Suppposedly fixes the `TypeError: Cannot read property '0' of null` when running Cypress in CI


denis-yuen:
changelog in case you're curious about the big version jump https://docs.cypress.io/guides/references/changelog.html#1-0-1

PR #152, opened 2017-11-24
Feature/file tab fix

agduncan94:
* Fix weird resizing of dropdowns
* Adds download button beside copy button


agduncan94:
See ga4gh/dockstore#1038

PR #154, opened 2017-12-11
Feature/workflow source control

agduncan94:
Updates for using metadata endpoints as well as adds source control to paths.

### DIFF
--- a/travisci/db_dump.sql
+++ b/travisci/db_dump.sql
@@ -365,7 +365,8 @@ CREATE TABLE workflow (
     organization character varying(255) NOT NULL,
     path character varying(255),
     repository character varying(255) NOT NULL,
-    workflowname text
+    workflowname text,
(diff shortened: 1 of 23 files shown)


agduncan94:
Note that  https://github.com/ga4gh/dockstore/pull/1062 needs to be merged before this PR will pass travis.

travisci/db_dump.sql
```diff
@@ -365,7 +365,8 @@ CREATE TABLE workflow (
     organization character varying(255) NOT NULL,
     path character varying(255),
     repository character varying(255) NOT NULL,
-    workflowname text
+    workflowname text,
```
denis-yuen: Interesting, forgot that ui2 was using a separate sql dump

agduncan94: Me too

PR #157, opened 2018-01-02
Feature/entry version page

agduncan94:
* Reworking of the entry page (both my entries and public entries) to have versions as a top level selector
* Can link directly to specific versions of an entry
* On public pages will use a label to identify default version


agduncan94:
I can give another quick demo if needed.

src/app/container/container.component.html
```diff
@@ -116,19 +124,34 @@ <h3>
         </form>
       </tab>
       <tab heading="Versions" *ngIf="validVersions">
-        <app-versions-container [versions]="tool?.tags" id="versionsTab" [verifiedSource]="tool?.verifiedSources">
+        <app-versions-container [versions]="tool?.tags" id="versionsTab" [verifiedSource]="tool?.verifiedSources" [selectedVersion]="selectedVersion">
```
denis-yuen: This brings up an interesting question. How do we want verification to look if some versions are verified and some are not?

src/app/container/dockerfile/dockerfile.component.ts
```diff
@@ -27,10 +27,17 @@ import { ContainerService } from '../../shared/container.service';
   selector: 'app-dockerfile',
   templateUrl: './dockerfile.component.html',
 })
-export class DockerfileComponent extends VersionSelector implements AfterViewChecked {
+export class DockerfileComponent implements AfterViewChecked {
 
   @Input() id: number;
   @Input() entrypath: string;
+  _selectedVersion: any;
```
denis-yuen: Is it really not possible to narrow this down from `any` for these two lines?

src/app/container/launch/launch.component.ts
```diff
@@ -17,41 +17,87 @@
...
+              private metadataService: MetadataService) {
+    this.metadataService.getDescriptorLanguages().subscribe(map => {
+      this.descriptors = map;
+      this.validDescriptors = this.filterDescriptors(this.descriptors, this._selectedVersion);
+    });
   }
-  getDescriptors(currentVersion): any {
-    return this.toolDescriptorService.getDescriptors(this.versions, this.default);
+
+  filterDescriptors(descriptors: Array<any>, version: any): Array<any> {
```
denis-yuen: a bunch of `any` here too

agduncan94:
@garyluu Could you check this over

src/app/container/container.component.html
agduncan94: Currently we have the following:
* Checkmark beside entry name to show that at least one version is verified
* In versions tab, column to show which versions are verified
* In side bar for public viewing, we have a box with the verified versions

src/app/container/container.component.ts
```diff
@@ -68,7 +70,7 @@ export class ContainerComponent extends Entry {
     router: Router,
     private containerService: ContainerService,
     stateService: StateService,
-    errorService: ErrorService) {
+    errorService: ErrorService, route: ActivatedRoute) {
```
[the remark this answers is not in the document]
agduncan94: oops that should be deleted

src/app/container/container.component.ts
```diff
@@ -240,4 +255,8 @@ export class ContainerComponent extends Entry {
     this.containerService.setCopyBtn(copyBtn);
   }
 
+  onTagChange(tag): void {
```
garyluu: See if tag could have a certain event type.

src/app/workflow/workflow.component.ts
```diff
@@ -53,10 +53,12 @@ export class WorkflowComponent extends Entry {
   private workflowSubscription: Subscription;
   private workflowCopyBtnSubscription: Subscription;
   private workflowCopyBtn: string;
+  public selectedVersion = null;
+  public urlVersion = null;
   constructor(private dockstoreService: DockstoreService, dateService: DateService, private refreshService: RefreshService,
     private workflowsService: WorkflowsService, trackLoginService: TrackLoginService, providerService: ProviderService,
     router: Router, private workflowService: WorkflowService,
-    stateService: StateService, errorService: ErrorService) {
+    stateService: StateService, errorService: ErrorService, route: ActivatedRoute) {
```
garyluu: Same for here

PR #158, opened 2018-01-03
Feature/valid verified sorting

agduncan94:
Deals with https://github.com/ga4gh/dockstore/issues/1083


agduncan94:
@denis-yuen all passed

PR #160, opened 2018-01-04
Fix login to create tokens of the same scope as account linking

denis-yuen:
This one line/error took entirely too long to discover. 
The issue was that logging in would create a token of completely different scope from the linking account button, causing newly logged in users to be unable to refresh workflows out of the box.


denis-yuen:
FYI, I added this as a hotfix too

PR #163, opened 2018-01-04
Feature/issue 1061

garyluu:
For issue ga4gh/dockstore#1061

Changes a few things from 'cwltool' to 'cwl-runner'.  Also adds a new section to the launch section of both tools and workflows based on what's specified in the issue.  
  
Also added a test

### DIFF
--- a/src/app/container/launch/launch.component.html
+++ b/src/app/container/launch/launch.component.html
@@ -42,7 +42,7 @@
         <a href="https://github.com/common-workflow-language/cwltool">Run locally with cwl-runner (you can use '--non-strict' when the executor is cwltool)</a>
         <pre>{{ cwl }}</pre>
         </div>
-        <div class="well well-sm" tooltip="Commands for launching tools through CWLtool: the CWL reference implementation.">
+        <div class="well well-sm" tooltip="Commands for launching tools through CWLtool: the CWL reference implementation. '--non-strict' can be used.">
         Alternatively, cwltool can run a tool off of dockstore.org when all inputs and outputs are available on the local filesystem or via HTTP(S):
         <pre>{{ dockstoreSupportedCwlLaunch }}</pre>
         cwltool can also be used to create the input template:
--- a/src/app/container/launch/launch.component.html
+++ b/src/app/container/launch/launch.component.html
@@ -38,12 +38,12 @@
       <pre>{{ cli }}</pre>
     </div>
     <div *ngIf="(currentDescriptor | lowercase) === 'cwl'">
-      <div class="well well-sm" tooltip="Commands for launching tools through cwl-runner when all inputs and outputs are available on the local filesystem.">
-        <a href="https://github.com/common-workflow-language/cwltool">Run locally with cwl-runner (you can use '--non-strict' when the executor is cwltool)</a>
+      <div class="well well-sm" [tooltip]="cwlrunnerTooltip">
+        {{cwlrunnerDescription}}
         <pre>{{ cwl }}</pre>
         </div>
-        <div class="well well-sm" tooltip="Commands for launching tools through CWLtool: the CWL reference implementation. '--non-strict' can be used.">
-        Alternatively, cwltool can run a tool off of dockstore.org when all inputs and outputs are available on the local filesystem or via HTTP(S):
+        <div class="well well-sm" [tooltip]="cwltoolTooltip">
+        Alternatively, <a href="https://github.com/common-workflow-language/cwltool">cwltool</a> can run a tool off of dockstore.org when all inputs and outputs are available on the local filesystem or via HTTP(S):
         <pre>{{ dockstoreSupportedCwlLaunch }}</pre>
         cwltool can also be used to create the input template:
         <pre>{{ dockstoreSupportedCwlMakeTemplate }}</pre>
--- a/src/app/container/launch/launch.component.html
+++ b/src/app/container/launch/launch.component.html
@@ -43,9 +43,9 @@
         <a href="https://github.com/common-workflow-language/cwltool">Run locally with cwl-runner</a>
         <pre>{{ cwl }}</pre>
         Alternatively, cwltool can run a tool off of dockstore.org when all inputs and outputs are available on the local filesystem or via HTTP(S):
-        <pre>{{ altCwl1 }}</pre>
+        <pre>{{ dockstoreSupportedCwlLaunch }}</pre>
         cwltool can also be used to create the input template:
-        <pre>{{ altCwl2 }}</pre>
+        <pre>{{ dockstoreSupportedCwlMakeTemplate }}</pre>
       </div>
       <div class="well well-sm" tooltip="Commands for launching tool through Consonance: a DIY open-source framework for
       spinning up VMs and running tools on AWS EC2">
--- a/src/app/container/launch/launch.component.html
+++ b/src/app/container/launch/launch.component.html
@@ -42,6 +42,10 @@
       cwl-runner can run a tool directly when all inputs and outputs are available on the local filesystem.">
         <a href="https://github.com/common-workflow-language/cwltool">Run locally with cwl-runner</a>
         <pre>{{ cwl }}</pre>
+        Alternatively, cwltool can run a tool off of dockstore.org when all inputs and outputs are available on the local filesystem or via HTTP(S):
+        <pre>{{ altCwl1 }}</pre>
+        cwltool can also be used to create the input template:
+        <pre>{{ altCwl2 }}</pre>
       </div>
       <div class="well well-sm" tooltip="Commands for launching tool through Consonance: a DIY open-source framework for
       spinning up VMs and running tools on AWS EC2">
--- a/src/app/container/launch/launch.component.html
+++ b/src/app/container/launch/launch.component.html
@@ -43,7 +43,7 @@
         <pre>{{ cwl }}</pre>
         </div>
         <div class="well well-sm" [tooltip]="cwltoolTooltip">
-        Alternatively, <a href="https://github.com/common-workflow-language/cwltool">cwltool</a> can run a tool off of dockstore.org when all inputs and outputs are available on the local filesystem or via HTTP(S):
+        Alternatively, <a href="https://github.com/common-workflow-language/cwltool">cwltool</a> can conveniently run a tool off of dockstore.org and other <a href="https://github.com/common-workflow-language/cwltool#use-with-ga4gh-tool-registry-api">GA4GH tool registries</a> when all inputs and outputs are available on the local filesystem or via HTTP(S):
         <pre>{{ dockstoreSupportedCwlLaunch }}</pre>
         cwltool can also be used to create the input template:
         <pre>{{ dockstoreSupportedCwlMakeTemplate }}</pre>
--- a/src/app/container/launch/launch.component.html
+++ b/src/app/container/launch/launch.component.html
@@ -38,10 +38,11 @@
       <pre>{{ cli }}</pre>
     </div>
     <div *ngIf="(currentDescriptor | lowercase) === 'cwl'">
-      <div class="well well-sm" tooltip="Commands for launching tools through cwl-runner: the CWL reference implementation.
-      cwl-runner can run a tool directly when all inputs and outputs are available on the local filesystem.">
-        <a href="https://github.com/common-workflow-language/cwltool">Run locally with cwl-runner</a>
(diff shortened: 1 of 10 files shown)


src/app/workflow/version-modal/version-modal.component.ts
```diff
@@ -124,4 +124,13 @@ export class VersionModalComponent implements OnInit, AfterViewChecked {
     }
   }
   // Validation ends here
+
+  // Checks if the currently edited test parameter file already exists
+  hasDuplicateTestJson() {
```
denis-yuen: Isn't this just `return this.testParameterFilePaths.includes(this.testParameterFilePath)`?

src/app/workflow/launch/launch.component.html
```diff
@@ -28,11 +28,16 @@
...
-    <div *ngIf="currentDescriptor === 'cwl'">
-      <div class="well well-sm" tooltip="Commands for launching tools through CWLtool: the CWL reference implementation.
-      cwltool can run a tool directly when all inputs and outputs are available on the local filesystem.">
-        Alternatively, cwltool can run a workflow directly when all inputs and outputs are available on the local filesystem
+    <div *ngIf="(currentDescriptor | lowercase) === 'cwl'">
+      <div class="well well-sm" tooltip="Commands for launching workflows through cwl-runner: the CWL reference implementation.
+      cwl-runner can run a tool directly when all inputs and outputs are available on the local filesystem.">
+        Run locally with cwl-runner
         <pre>{{ cwl }}</pre>
+        <!-- The below is disabled until workflows actually work -->
```
denis-yuen: Worth pending a test for this on dockstore-webservice

src/app/shared/launch.service.ts
```diff
@@ -27,4 +27,10 @@ export abstract class LaunchService {
         return `$ consonance run --tool-dockstore-id ${path}:${versionName} ` +
             '--run-descriptor Dockstore.json --flavour \<AWS instance-type\>';
     }
+    getAltCwlString1(path: string, versionName: string) {
```
denis-yuen: Let's make this more descriptive than "1" and "2"

src/app/container/version-modal/version-modal.component.ts
```diff
@@ -265,4 +266,23 @@ export class VersionModalComponent implements OnInit, AfterViewChecked {
   getDateTimeMessage(timestamp) {
     return this.dateService.getDateTimeMessage(timestamp);
   }
+
+  // Checks if the currently edited test parameter file already exists
+  hasDuplicateTestJson(type) {
+    if (type === DescriptorType.CWL) {
+      if (this.unsavedCWLTestParameterFilePaths.includes(this.unsavedTestCWLFile)) {
```
denis-yuen: This can be simpified

src/app/container/launch/launch.component.html
```diff
@@ -38,10 +38,14 @@
...
     <div *ngIf="(currentDescriptor | lowercase) === 'cwl'">
-      <div class="well well-sm" tooltip="Commands for launching tools through CWLtool: the CWL reference implementation.
-      cwltool can run a tool directly when all inputs and outputs are available on the local filesystem.">
-        <a href="https://github.com/common-workflow-language/cwltool">Run locally with cwltool</a>
+      <div class="well well-sm" tooltip="Commands for launching tools through cwl-runner: the CWL reference implementation.
+      cwl-runner can run a tool directly when all inputs and outputs are available on the local filesystem.">
+        <a href="https://github.com/common-workflow-language/cwltool">Run locally with cwl-runner</a>
         <pre>{{ cwl }}</pre>
+        Alternatively, cwltool can run a tool off of dockstore.org when all inputs and outputs are available on the local filesystem or via HTTP(S):
+        <pre>{{ altCwl1 }}</pre>
```
denis-yuen: Let's make the naming here more descriptive too

src/app/container/launch/tool-launch.service.spec.ts
```diff
@@ -41,14 +41,18 @@ describe('LaunchService', () => {
...
     }));
     it('should getConsonanceString', inject([ToolLaunchService], (service: ToolLaunchService) => {
-        expect(service.getConsonanceString('a/b', 'latest')).toContain(
-            'consonance run --tool-dockstore-id a/b:latest --run-descriptor Dockstore.json --flavour <AWS instance-type>');
+        expect(service.getAltCwlString1('quay.io/briandoconnor/dockstore-tool-bamstats', '1.25-11')).toContain(
+            'cwltool --non-strict quay.io/briandoconnor/dockstore-tool-bamstats:1.25-11 Dockstore.json');
+    }));
+    it('should getAlternateStrings', inject([ToolLaunchService], (service: ToolLaunchService) => {
+        expect(service.getAltCwlString2('quay.io/briandoconnor/dockstore-tool-bamstats', '1.25-11')).toContain(
+            'cwltool --make-template quay.io/briandoconnor/dockstore-tool-bamstats:1.25-11 > input.yaml');
```
denis-yuen: Is cwltool correct here?

src/app/container/add-tag/add-tag.component.ts
```diff
@@ -159,4 +167,23 @@ export class AddTagComponent implements OnInit, AfterViewChecked {
     }
   }
   // Validation ends here
+  // Checks if the currently edited test parameter file already exists
+  // TODO: This code is repeated in version-modal.component.ts for tools, move it somewhere common
```
denis-yuen: Yup

src/app/container/launch/tool-launch.service.spec.ts
garyluu: The issue mentioning the new section shows cwltool as the example

src/app/workflow/version-modal/version-modal.component.ts
garyluu: Some weird merge caused this to appear.  I've rebased to remove irrelevant code

src/app/workflow/launch/launch.component.html
garyluu: The GA4GH production endpoint still works.  It might be a cwltool issue.

denis-yuen: ok, create a new ticket for us to investigate and bookmark this line then so we follow up

garyluu: Upon further investigation it looks like it does work for workflows.  It appears I was originally testing with a WDL workflow which explains why it didn't work.

src/app/container/launch/launch.component.html
```diff
@@ -38,10 +38,14 @@
       <pre>{{ cli }}</pre>
     </div>
     <div *ngIf="(currentDescriptor | lowercase) === 'cwl'">
-      <div class="well well-sm" tooltip="Commands for launching tools through CWLtool: the CWL reference implementation.
-      cwltool can run a tool directly when all inputs and outputs are available on the local filesystem.">
-        <a href="https://github.com/common-workflow-language/cwltool">Run locally with cwltool</a>
+      <div class="well well-sm" tooltip="Commands for launching tools through cwl-runner: the CWL reference implementation.
```
mr-c: `cwl-runner` is the generic way to run a CWL description using any CWL aware system. The name of the CWL reference runner is `cwltool`

src/app/container/launch/tool-launch.service.spec.ts
```diff
@@ -41,14 +41,18 @@ describe('LaunchService', () => {
 
     it('should getCWLString', inject([ToolLaunchService], (service: ToolLaunchService) => {
         expect(service.getCwlString('quay.io/a/b', 'c'))
-            .toContain('cwltool --non-strict');
+            .toContain('cwl-runner --non-strict');
```
mr-c: `--non-strict` shouldn't be here. It isn't part of the CWL standard and descriptions that need it are broken. I opened several PRs for prominent tools posted on dockstore.org to fix the "need" for `--non-strict" :-)

denis-yuen: Thanks for the heads-up!
I think I'll have to take a look for those PRs, I may have missed the notifications. In the meantime, I agree that we cannot suggest `--non-strict` for cwl-runner if it only applies for cwltool. 

@garyluu  just note that `non-strict` or its equivalent for other engines can be a troubleshooting step whether via tooltip or shell comment

src/app/container/launch/launch.component.html
mr-c: so the second half of the sentence should read "the generic interface to many CWL implementations."

src/app/shared/launch.service.ts
```diff
@@ -27,4 +27,22 @@ export abstract class LaunchService {
...
             '--run-descriptor Dockstore.json --flavour \<AWS instance-type\>';
     }
+
+    /**
+     * This creates the Dockstore-supported cwltool launch command
+     * @param path The GA4GH Tool's path
+     * @param versionName The ToolVersion's name
+     */
+    getDockstoreSupportedCwlLaunchString(path: string, versionName: string) {
+        return `$ cwltool --non-strict ${path}:${versionName} Dockstore.json`;
```
mr-c: I think this should be left out here. Users can add it back in if they wish

garyluu: the '--non-strict'?

mr-c: Correct, my apologies for the imprecision

garyluu:
Ready for re-review

src/app/container/launch/launch.component.html
```diff
@@ -38,10 +38,15 @@
       <pre>{{ cli }}</pre>
     </div>
     <div *ngIf="(currentDescriptor | lowercase) === 'cwl'">
-      <div class="well well-sm" tooltip="Commands for launching tools through CWLtool: the CWL reference implementation.
-      cwltool can run a tool directly when all inputs and outputs are available on the local filesystem.">
-        <a href="https://github.com/common-workflow-language/cwltool">Run locally with cwltool</a>
+      <div class="well well-sm" tooltip="Commands for launching tools through cwl-runner when all inputs and outputs are available on the local filesystem.">
+        <a href="https://github.com/common-workflow-language/cwltool">Run locally with cwl-runner (you can use '--non-strict' when the executor is cwltool)</a>
```
denis-yuen: Might want to explain why someone might want to do this and if it ends up being too long, maybe it can be hidden by a drop-down, info button, etc.

src/app/container/launch/launch.component.html
```diff
@@ -38,10 +38,15 @@
...
         <pre>{{ cwl }}</pre>
+        </div>
+        <div class="well well-sm" tooltip="Commands for launching tools through CWLtool: the CWL reference implementation. '--non-strict' can be used.">
+        Alternatively, cwltool can run a tool off of dockstore.org when all inputs and outputs are available on the local filesystem or via HTTP(S):
```
denis-yuen: Same here, why would someone want to use non-strict. 
Also, what distinguishes this from the above option?

src/app/workflow/launch/launch.component.html
```diff
@@ -28,12 +26,16 @@
       Run locally with the Dockstore CLI
       <pre>{{ cli }}</pre>
     </div>
-    <div *ngIf="currentDescriptor === 'cwl'">
-      <div class="well well-sm" tooltip="Commands for launching tools through CWLtool: the CWL reference implementation.
-      cwltool can run a tool directly when all inputs and outputs are available on the local filesystem.">
-        Alternatively, cwltool can run a workflow directly when all inputs and outputs are available on the local filesystem
+    <div *ngIf="(currentDescriptor | lowercase) === 'cwl'">
+      <div class="well well-sm" tooltip="Commands for launching workflows through cwl-runner when all inputs and outputs are available on the local filesystem.">
+        Run locally with cwl-runner (you can use '--non-strict' when the executor is cwltool)
```
denis-yuen: Ditto as above

denis-yuen:
Looks better, I think this just needs a bit more explanation, potentially hidden by default or linked to

garyluu:
Simplified a few things.  Moved the '--non-strict' and a brief explanation into tooltip to balance out the amount of content in tooltip vs description.  There is already a hyperlink to cwltool's GitHub.  People should honestly familiarize themselves with cwltool from there instead of Dockstore since there's quite a few flags.  Same applies for all other cwl-runners.

src/app/container/launch/launch.component.html
garyluu: Still working on this...

PR #164, opened 2018-01-04
More error messages

denis-yuen:


src/app/workflow/workflow.component.ts
```diff
@@ -188,7 +188,10 @@ export class WorkflowComponent extends Entry {
         publish: this.published
       };
       this.workflowsService.publish(this.workflow.id, request).subscribe(
-        response => this.workflow.is_published = response.is_published, err => this.published = !this.published);
+        response => this.workflow.is_published = response.is_published, err => {
+          this.published = !this.published;
+          this.refreshService.handleError('publish error', err);
```
garyluu: Slightly weird that one is 'publish' but the other is "Publish"

PR #165, opened 2018-01-05
Run chrome headless

garyluu:
Sometimes `ng test` results in `Karma - Disconnected, because no message in 10000 ms`  .  Run chrome headless appears to help mitigate the issue a bit.  Another option is to increase timeout.

### DIFF
--- a/scripts/run-travis-script.sh
+++ b/scripts/run-travis-script.sh
@@ -16,6 +16,6 @@ else
   ng lint
   ng build --progress false
   ng serve --silent &
-  ng test --watch=false --code-coverage
+  ng test --watch=false --code-coverage --browsers ChromeHeadless
   cypress run --record --config defaultCommandTimeout=10000
 fi


denis-yuen:
Can you elaborate on what we expect to see with "a bit"?

garyluu:
Until there is a better solution for the karma error, reducing execution time helps mitigate the issue that arises on Travis.  Regardless of the karma error, we should be switching to headless Chrome (see https://developers.google.com/web/updates/2017/06/headless-karma-mocha-chai for why headless Chrome is recommended on CI).  Preliminary testing on my local machine shows about a 15% or so reduction in execution time.  Times on Travis vary, but a first glance shows a 5% speed up.

Here is overview of the issue with links to some ways to mitigate it:
https://gitlab.com/gitlab-org/gitlab-ce/issues/33784

From what I gather, here's some options that might help:
- Run headless chrome
- Use async in tests
- Update Jasmine-core
- Increase timeout

denis-yuen:
ok, thanks for the explanation

PR #168, opened 2018-01-10
Feature/nextflow mods

denis-yuen:
Paired with https://github.com/ga4gh/dockstore/pull/1101 to display NextFlow workflows and their files in the UI 
Also affected by the new changes from the proposed V2 Tool Registry Service changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ jdk:
 env:
   global:
     - WEBSERVICE_VERSION="1.4.0-alpha.5"
+
   matrix:
     - RUN_PROD=false
     - RUN_PROD=true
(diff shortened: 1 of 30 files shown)


denis-yuen:
@garyluu  FYI, this works (for one of the two?) but is *terrible* 
For the change from ToolVersion to ToolVersion1 see if there's a way of defining the change in one location rather than us having to change it everywhere whenever the API version number changes. 

Might be able to use something like type aliases https://blogs.msdn.microsoft.com/typescript/2014/11/18/typescript-1-4-sneak-peek-union-types-type-guards-and-more/
https://blogs.msdn.microsoft.com/typescript/2015/09/16/announcing-typescript-1-6/  , please do some experimentation to see what's the cleanest

PR #172, opened 2018-01-12
Used main descriptor as relative path

garyluu:
For issue ga4gh/dockstore#844

Substitute the workflow cwl-runner launch string's descriptor with the relative paths endpoint instead.  Tool is also changed.


denis-yuen:
Question: is there a reason to prefer tool using the old endpoint rather than keeping them consistent?

garyluu:
The way I see it is if we switch tool to also use the relative paths endpoint, the /descriptor endpoint is then mostly unused.  The current /descriptor path works for tools as is and it's also shorter.

garyluu:
Changed tools too

PR #176, opened 2018-01-22
Feature/dockstore entry url encode

garyluu:
For issue ga4gh/dockstore#1097


denis-yuen:
Let's get some automated testing for these. 
I want to be convinced that the new URls work but also that if anyone has the old escaped URLs that they remain working.

garyluu:
My local environment changed to npm 5 so a bunch of unrelated changes appeared in this pull request.  Closing until the npm update is in develop.

PR #178, opened 2018-01-22
Feature/advanced search fix

garyluu:
Fix for issue ga4gh/dockstore#1118 mainly by performing some null checks beforehand.


denis-yuen:
@garyluu  Let's get these null checks as a hotfix too, thanks!

PR #179, opened 2018-01-23
Use stable 2.3.1 swagger codgen

garyluu:
Upgrade to use the stable 2.3.1 swagger codegen


src/app/container/container.component.ts
```diff
@@ -173,7 +173,7 @@ export class ContainerComponent extends Entry {
 
       // Only get published tool if the URI is for a specific tool (/containers/quay.io%2FA2%2Fb3)
       // as opposed to just /tools or /docs etc.
-      this.containersService.getPublishedContainerByToolPath(this.title, this._toolType)
```
denis-yuen: was an explanation discovered for this disappearing parameter?

src/app/workflow/workflow.component.ts
```diff
@@ -152,7 +152,7 @@ export class WorkflowComponent extends Entry {
 
       // Only get published workflow if the URI is for a specific workflow (/containers/quay.io%2FA2%2Fb3)
       // as opposed to just /tools or /docs etc.
-      this.workflowsService.getPublishedWorkflowByPath(this.title, this._toolType)
```
denis-yuen: same here

src/app/container/container.component.ts
garyluu: This is what I think happened.  We were originally not using the swagger generated methods and we had an extra parameter to differentiate between tool/workflow.  We swapped the methods to the swagger-generated ones but didn't remove that extra parameter during the change.  The swagger-generated methods so happen to allow an extra parameter with type 'any'.  so it still worked but didn't really do anything significant with that extra parameter.

With the new 2.3.1 swagger generated methods, it appears they removed this 'any' type parameter so it no longer works.

src/app/workflow/workflow.component.ts
garyluu: same

PR #180, opened 2018-01-24
Hotfix/advanced search fix

garyluu:
Similar to https://github.com/dockstore/dockstore-ui2/pull/178

### DIFF
--- a/src/app/search/elastic-search-client.ts
+++ b/src/app/search/elastic-search-client.ts
@@ -18,7 +18,7 @@ import { Dockstore } from './../shared/dockstore.model';
 import { Client } from 'elasticsearch';
 
 export const ELASTIC_SEARCH_CLIENT = new Client({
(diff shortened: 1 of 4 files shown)


src/app/search/elastic-search-client.ts
```diff
@@ -18,7 +18,7 @@ import { Dockstore } from './../shared/dockstore.model';
 import { Client } from 'elasticsearch';
 
 export const ELASTIC_SEARCH_CLIENT = new Client({
-    host: Dockstore.API_URI + '/api/ga4gh/v1/extended',
```
denis-yuen: is ga4gh/v2 already in hotfix?

garyluu: Not yet I believe.

garyluu: I'll revert this change then.  It looks like the hotfix branch isn't using the v2 webservice either

PR #181, opened 2018-01-25
Feature/node upgrade

garyluu:
Originally trying to fix the npm econreset and cypress install issue on Travis.  Ended up:

- Upgrading to the current node js LTS (v8.9.4) which updated npm to 5.6.0 to hopefully fix the npm econreset
- Upgrade Cypress to the latest 1.4.1 because I encounter difficulties installing otherwise
- Downgrade flexbox 2.0.0-beta.12 to 2.0.0-beta.9 because everything higher is supposed to require angular 5 (which I don't want go down that rabbit hole at this time)
- removed npm-shrinkwrap in favor of package-lock.json because it's the general recommended one.  It also auto updates upon new npm package installs. https://www.reddit.com/r/javascript/comments/6dgnnq/npm_v500_released_save_by_default_lockfile_better/di3mjuk/  The dev deps are also in the lock file which is why it's larger than the old one.  Install the different type of deps with `npm i --prod`, `npm i --dev`, and `npm i`.  The lock file's packages will be used for whichever one is about to be installed.

### Notes
- Can't solve angular-tag-cloud-module 1.4.0 requiring an old angular version because the next angular-tag-cloud-module is 2.0.0 which jumps straight to requiring angular 5.  
- https://github.com/npm/npm/issues/17979#issuecomment-332701215 shows how the new shrinkwrap works.  In short, when prod or other devs checkout the branch where package.json and package-lock.json are "compatible",  `npm i` will install the exact version of the packages listed in the lock file.  If they are not compatible, npm will install what's listed in the package.json and then update the lock file.
- I had to run `npm dedupe` to modify the lock file again otherwise a bunch of deps of deps will conflict with each other
- `npm install --prod` and `npm install` yields a slightly different package-lock.json due to angular/cli not liking being placed in deps instead of devDeps.  Using the package-lock.json from `npm install --prod` instead.  Can't move cli to devDeps because we need ng commands during production.  Can't install it outside of `npm install --prod` because it would still require an angular-cli in the package.json to link to things like the compiler-cli etc.


denis-yuen:
`flexbox 2.0.0-beta.12 to 2.0.0-beta.9 because everything higher is supposed to require angular 5` 
How is this working now since we're not actually using angular 5 but we're already using flexbox 2.0.0-beta.12?

garyluu:
I think the angular 5 requirement is something the flexbox developer manually set the requirement as (in his package.json probably).  The flex developer probably tested it with Angular 5, but still retains backwards compatibility with Angular 4.

denis-yuen:
ok, might as well keep the newer version then

garyluu:
Ok this is interesting.  apparently the shrinkwrap from before locked it down to beta 9 which completely disregarded beta 12 in the package.json which is why it works

denis-yuen:
@garyluu  could you comment on the status of this? thanks

garyluu:
Good to re-review

denis-yuen:
As discussed, let's get a check into the travis-ci production checks. 
Make sure to run 
git diff checking for an exit code to see whether package-lock and package.json are consistent

denis-yuen:
As discussed, we'll discuss this as a group at the demo and hopefully decide between yarn and npm5

denis-yuen:
As discussed, we'll move ahead with NPM5

PR #182, opened 2018-01-29
Feature/dockstore entry url encode

garyluu:
For issue ga4gh/dockstore#1097 This also contains the nodeUpgrade commits which is why it's so long.  Should be looked at after the nodeUpgrade.


cypress/integration/toolDetails.js
```diff
@@ -2,6 +2,7 @@ describe('Dockstore Tool Details', function() {
     require('./helper.js')
     beforeEach(function() {
         cy.visit(String(global.baseUrl) + "/containers/quay.io/A2/a")
+        cy.visit(String(global.baseUrl) + "/containers/quay.io%2FA2%2Fa")
```
denis-yuen: 👍

PR #184, opened 2018-01-29
Feature/entry version page

agduncan94:
* Shows current version at top of entry
* Converts version sidebar dropdown into a list of recent versions
* Can change the current version by selecting the version in the versions table
* Can link to tabs for entries
* Moved contact and starring button to give more space to the title

### DIFF
--- a/src/app/shared/entry.ts
+++ b/src/app/shared/entry.ts
@@ -19,10 +19,11 @@ import { validationPatterns } from './validationMessages.model';
 import { DockstoreTool } from './swagger/model/dockstoreTool';
 import { ErrorService } from './../shared/error.service';
 import { Workflow } from './swagger/model/workflow';
-import { Injectable, Input, OnDestroy, OnInit, AfterViewInit } from '@angular/core';
+import { Injectable, Input, OnDestroy, OnInit, AfterViewInit, ViewChild } from '@angular/core';
 import { StateService } from './state.service';
 import { Router } from '@angular/router/';
 import { Subscription } from 'rxjs/Subscription';
+import { TabsetComponent } from 'ngx-bootstrap';
 
 import { CommunicatorService } from './communicator.service';
 import { ProviderService } from './provider.service';
@@ -32,6 +33,7 @@ import { TrackLoginService } from '../shared/track-login.service';
 
 @Injectable()
 export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
+  @ViewChild('entryTabs') entryTabs: TabsetComponent;
   protected labels: string[];
   protected shareURL: string;
   public starGazersClicked = false;
@@ -195,5 +197,25 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
       selectedVersion = versions[0];
     }
     return selectedVersion;
+  }
+
+  selectTab(tab_id: number) {
+    this.entryTabs.tabs[tab_id].active = true;
+  }
+
+  entryVersionSorting(a: any, b: any): number {
+    if (a.last_modified > b.last_modified) {
+      return -1;
+    } else if (a.last_modified < b.last_modified) {
+      return 1;
+    } else {
+      if (a.verified && !b.verified) {
+        return -1;
+      } else if (!a.verified && b.verified) {
+        return 1;
+      } else {
+        return 0;
+      }
+    }
   }
 }
--- a/src/app/shared/entry.ts
+++ b/src/app/shared/entry.ts
@@ -199,10 +199,12 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
     return selectedVersion;
   }
 
-  selectTab(tab_id: number) {
-    this.entryTabs.tabs[tab_id].active = true;
+  // Selects a tab of index tabIndex
+  selectTab(tabIndex: number) {
+    this.entryTabs.tabs[tabIndex].active = true;
   }
 
+  // Sorts two entries by last modified, and then verified
   entryVersionSorting(a: any, b: any): number {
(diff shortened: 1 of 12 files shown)


agduncan94:
Ready for review

src/app/workflow/workflow.component.html
```diff
@@ -136,15 +139,23 @@ <h3>
...
       </div>
       <div class="panel-body">
-        <div class="input-group col-md-12">
-          <div class="form-group">
-            <select name="workflowVersions" class="form-control select-box" [(ngModel)]="selectedVersion">
-              <option *ngFor="let workflowVersion of workflow.workflowVersions" [ngValue]="workflowVersion">{{workflowVersion.name}}</option>
-            </select>
-          </div>
+        <div>
+          <span *ngFor="let sortedVersion of getSortedVersions(workflow.workflowVersions, defaultVersion);">
```
garyluu: Avoid using a function in places like this.

src/app/shared/entry.ts
```diff
@@ -196,4 +197,55 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
...
     return selectedVersion;
   }
+
+  // Selects a tab of index tabIndex
+  selectTab(tabIndex: number) {
+    this.entryTabs.tabs[tabIndex].active = true;
+  }
+
+  // Sorts two entries by last modified, and then verified
+  entryVersionSorting(a: any, b: any): number {
```
garyluu: Not entirely sure if this is what's wanted.  Should double-check with Denis just in case.  Need an example (might as well put it in a test too).

agduncan94: Ya I thought it was sort by last modified, and then verified to deal with tie breakers. That is what the spec said.

Test is a good idea.

src/app/shared/entry.ts
```diff
@@ -196,4 +197,55 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
...
+    let counter = 0;
+
+    // Sort versions by last_modified date and then verified
+    sortedVersions = versions.sort((a, b) => this.entryVersionSorting(a, b));
+
+    // The default version will appear first
+    finalVersions.push(defaultVersion);
+
+    // Grab the top 5 versions, ignoring the default version
+    for (const version of sortedVersions) {
```
garyluu: i feel like some sort of filter function + array.slice would be prettier

src/app/shared/entry.ts
garyluu: I kind of interpreted as sort by last modified, then take all the verified ones (while retaining order), and throw it to the top.

src/app/workflow/workflow.component.html
denis-yuen: actual question, is this a style recommendation, a performance thing?

src/app/shared/entry.ts
denis-yuen: I could be wrong, but they should be the same, assuming a stable sort

garyluu: An example would be great.  Right now, it looks like:

- Sort by date.  If same date, put verified above the other one.

I thought it'd be:

- Sort by verified (verified ones on top).  If same verified status, put newer ones above the other ones.

src/app/shared/entry.ts
```diff
@@ -196,4 +197,55 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
...
+      } else {
+        return 0;
+      }
+    }
+  }
+
+  // Sorts a list of versions by last modified date and verified, returning a subset of the versions
+  // If a default version is set, it will appear at the top
+  // If not, will set to latest
+  getSortedVersions(versions: Array<any>, defaultVersion: any): Array<any> {
```
denis-yuen: This seems to overuse `any` when these appear to always be tags.

src/app/shared/entry.ts
```diff
@@ -196,4 +197,55 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
...
+  }
+
+  // Sorts two entries by last modified, and then verified
+  entryVersionSorting(a: any, b: any): number {
+    if (a.last_modified > b.last_modified) {
+      return -1;
+    } else if (a.last_modified < b.last_modified) {
+      return 1;
+    } else {
+      if (a.verified && !b.verified) {
```
denis-yuen: Oh, I see there might be ambiguity in language here. 
I meant algorithmically, one should sort the table first by last modified and then by verified. 
I think this algorithm sorts once by modified date first and then if those are tied, sort by verified which isn't quite right. But it is worth double-checking

src/app/workflow/workflow.component.html
garyluu: The function gets ran every time the view gets checked/updated.  On my computer, it gets ran about 40+ times.

garyluu: https://stackoverflow.com/questions/42066490/call-a-component-method-from-html-in-angular2

src/app/shared/entry.ts
agduncan94: Oops ya I assumed the second case, I'll fix it.

src/app/shared/entry.ts
agduncan94: Technically it is tags or workflowversion, I wasn't sure if there was a generic version type.

garyluu: Type unions might the be solution.  `defaultVersions: Tag | WorkflowVersion`

denis-yuen: Yeah, the lack of inheritance in the generated swagger classes is annoying that way, but the type unions sound interesting

PR #187, opened 2018-01-31
Feature/915/workflow search filters

garyluu:
This is supposed to be merged in develop, but base is a different one to show differences.

This adds a few search facets to workflows (verified, verified source, organization).  It also denotes which facet is entry type specific.  It currently does not combine tool facets and workflow facets together even if they are similar.


cypress/plugins/index.js
```diff
@@ -0,0 +1,17 @@
+// ***********************************************************
```
denis-yuen: Is a cypress change intentional?

src/app/container/paramfiles/paramfiles.service.ts
```diff
@@ -32,7 +32,7 @@ export class ParamfilesService {
     if (type === 'workflows') {
       return this.workflowsService.getTestParameterFiles(id, versionName);
     } else {
-      return this.containersService.getTestParameterFiles(id, versionName, descriptor);
+      return this.containersService.getTestParameterFiles(id, descriptor, versionName);
```
denis-yuen: Why is this reversing in this PR?

src/app/search/search.service.ts
```diff
@@ -201,17 +201,20 @@ export class SearchService {
...
       ['Language', 'descriptorType'],
       ['Registry', 'registry'],
       ['Private Access', 'private_access'],
       ['Verified', 'tags.verified'],
       ['Author', 'author'],
-      ['Organization', 'namespace'],
+      ['Namespace', 'namespace'],
       ['Labels', 'labels.value.keyword'],
       ['Verified Source', 'tags.verifiedSource'],
+      ['Verified 2', 'workflowVersions.verified'],
```
denis-yuen: These names seem pretty non-descriptive (namespace, organization, verified 2, verified source 2). What is this a map used for (i.e. this map could use a code comment as well)

cypress/plugins/index.js
garyluu: New cypress allows plugins, so checked in the empty plugins file in the event that we do use plugins in the future.  Prevents unchecked file locally.

src/app/container/paramfiles/paramfiles.service.ts
garyluu: Looks like swagger codegen or the swagger.yaml reversed the parameters.

denis-yuen: It doesn't seem like either or those are changing in this PR though, so was this previous broken and undetected or was there a non-checked-in change here?

garyluu: Previous broken and undetected

denis-yuen: Ok, might be time to think of a test or extend the parameter ordering that was put in earlier.
This seems like an easy way to introduce an error without noticing

PR #188, opened 2018-01-31
Use strings only

garyluu:
For ga4gh/dockstore#1134  

Clicking on verified facet alters the filters by adding a number as the categoryValue (0 for public, 1 for private).
However, using the search path with parameters alters the filters by adding a string as the categoryValue ("0" for public, "1" for private).   This difference causes problems.  Now changing it all to strings (apparently elasticsearch doesn't care).  Also, not sure how a function with typed input parameters do not throw an error when it gets a different type.


denis-yuen:
Ok on my end, let's wait for a second opinion on hotfixes

PR #190, opened 2018-02-01
Feature/path removal

agduncan94:
Deals with the changes to the webservice removing the path column in the database.

Note this will fail until the path removal webservice PR is merged into develop.

https://github.com/ga4gh/dockstore/issues/1125
We use the field full_workflow_path to be of the form A/B/C/D, since workflow_path is already used to specify the default path of the descriptor for a tool.
So workflow.full_workflow_path is equivalent to tool.tool_path, and workflow.path is equivalent to tool.path


agduncan94:
Closing until this passes

PR #192, opened 2018-02-07
Feature/915/workflow search filters after node

garyluu:
This adds a few search facets to workflows (verified, verified source, organization). It also denotes which facet is entry type specific. It currently does not combine tool facets and workflow facets together even if they are similar.  It combines verified tools/workflows but there's still a bug with aggregation


denis-yuen:
@garyluu  do you mind quickly demoing this when you have a chance, curious how it looks

garyluu:
Closing until verified and verified sources are merged into one.

garyluu:
Re-opening because verified sources will take too long.  Splitting it into another pull request.

PR #194, opened 2018-02-09
Feature/1152/docker pull fix

garyluu:
For issue ga4gh/dockstore#1152

- Fixes the docker pull command in the version modal
- Also speed it up by removing function from template

Added one extra tool (garyluu/dockstore-cgpmap/cgpmap-cramOut) to the repository.  Will be useful in the future because it actually contains a real Dockerfile, descriptor, and test parameter file.  Currently only used to test the docker pull command.


denis-yuen:
Let's get a test for hotfixes especially
it can be based on #1152 (i.e. register a tool with a toolname, make sure the tool name doesn't show up in the docker pull command)

garyluu:
Closing until test added.

src/app/test/mocked-objects.ts
```diff
@@ -151,3 +153,118 @@ export const quayToken: Token = {
     'refreshToken': null,
     'userId': 2,
 };
+
+/**
+ * Near identical to the production 'quay.io/wtsicgp/dockstore-cgpmap/cgpmap-cramOut' tool but with many versions and dates removed
+ */
+export const realTool = {
```
denis-yuen: Sorry, let's move this into the DB dump, then we can use regular migration on upgrades.

garyluu: This is for the unit tests which don't rely on a db dump.  I could add an integration test if needed, but it seems a little overkill for a docker pull command.

denis-yuen: Yeah, integration test if necessary to avoid the mock.

travisci/db_dump.sql
```diff
@@ -584,13 +585,14 @@ INSERT INTO databasechangeloglock (id, locked, lockgranted, lockedby) VALUES (1,
 --
 
 INSERT INTO enduser (id, avatarurl, bio, company, email, isadmin, location, username) VALUES (1, NULL, NULL, NULL, NULL, false, NULL, 'user_A');
+INSERT INTO enduser (id, avatarurl, bio, company, email, isadmin, location, username) VALUES (2, 'https://avatars3.githubusercontent.com/u/24548904?v=4', NULL, 'OICR', 'garylau.work@gmail.com', false, NULL, 'garyluu');
```
denis-yuen: You might want to make this into a generic user

PR #198, opened 2018-02-15
Display view.commonwl.org visualization #1058

coverbeck:
For CWL workflows, display an SVG created by view.commonwl.org. This
is in addition to the current DAG visualization, and the change
includes radiobuttons that let the user select which visualization
to see.

This feature is off by default, as view.commonwl.org does not
support CORS. Two issues were filed and fixed in that repo, but they
have not yet been deployed. This was tested by running an instance of
the viewer locally.

To turn on the feature, initialize enableCwlViewer to true in
src/app/workflow/dag/dag.component.ts. Ideally a more fancy feature
flag feature should have been implemented that doesn't require
changing the source code to toggle the feature, but it didn't seem worth
the effort unless this is a need we anticipate coming up more often.

Todo:

The zooming in/out is pretty crude -- it doesn't let you pick which
part of the image you to see.

screen shot 2018-02-14 at 10 14 51 am

### DIFF
--- a/src/app/shared/dockstore.model.ts
+++ b/src/app/shared/dockstore.model.ts
@@ -49,4 +49,6 @@ export class Dockstore {
   static readonly GITLAB_AUTH_URL = 'https://gitlab.com/oauth/authorize';
   static readonly GITLAB_CLIENT_ID = 'fill_this_in';
   static readonly GITLAB_REDIRECT_URI = Dockstore.LOCAL_URI + '/auth/' + Provider.GITLAB;
+
+  static readonly CWL_VISUALIZER_URI = 'https://view.commonwl.org';
 }
--- a/src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
+++ b/src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
@@ -17,6 +17,7 @@
 import { Component, Input, OnInit } from '@angular/core';
 import { CwlViewerDescriptor, CwlViewerService } from './cwl-viewer.service';
 import { WorkflowVersion } from '../../../shared/swagger/model/workflowVersion';
+import { ExtendedWorkflow } from "../../../shared/models/ExtendedWorkflow";
 
 @Component({
   selector: 'app-cwl-viewer',
@@ -27,10 +28,10 @@ import { WorkflowVersion } from '../../../shared/swagger/model/workflowVersion';
 
 export class CwlViewerComponent implements OnInit {
 
-  @Input() workflow: any;
+  @Input() workflow: ExtendedWorkflow;
   @Input() set selectedVersion(value: WorkflowVersion) {
     if (value != null) {
-      this._selectedVersion = value;
+      this.version = value;
       this.onChange();
     }
   }
@@ -44,7 +45,7 @@ export class CwlViewerComponent implements OnInit {
   public cwlViewerDescriptor: CwlViewerDescriptor;
   public loading = false;
 
-  private _selectedVersion;
+  private version;
 
   constructor(private cwlViewerService: CwlViewerService) {
   }
@@ -58,11 +59,11 @@ export class CwlViewerComponent implements OnInit {
   }
 
   onChange() {
-    if (this._selectedVersion) {
+    if (this.version) {
       this.loading = true;
       this.cwlViewerDescriptor = null;
-      this.cwlViewerService.getVisualizationUrls(this.workflow.providerUrl, this._selectedVersion.reference,
-        this._selectedVersion.workflow_path)
+      this.cwlViewerService.getVisualizationUrls(this.workflow.providerUrl, this.version.reference,
+        this.version.workflow_path)
         .subscribe(
           cwlViewerDescriptor => {
             this.cwlViewerDescriptor = cwlViewerDescriptor;
--- a/src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
+++ b/src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
@@ -17,7 +17,7 @@
 import { Component, Input, OnInit } from '@angular/core';
 import { CwlViewerDescriptor, CwlViewerService } from './cwl-viewer.service';
 import { WorkflowVersion } from '../../../shared/swagger/model/workflowVersion';
-import { ExtendedWorkflow } from "../../../shared/models/ExtendedWorkflow";
+import { ExtendedWorkflow } from '../../../shared/models/ExtendedWorkflow';
 
 @Component({
   selector: 'app-cwl-viewer',
--- a/src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
+++ b/src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
@@ -1,0 +1,90 @@
+/*
+ *    Copyright 2018 OICR
+ *
+ *    Licensed under the Apache License, Version 2.0 (the 'License');
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an 'AS IS' BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+import { Component, Input, OnInit } from '@angular/core';
+import { CwlViewerDescriptor, CwlViewerService } from './cwl-viewer.service';
+import { WorkflowVersion } from '../../../shared/swagger/model/workflowVersion';
+
+@Component({
+  selector: 'app-cwl-viewer',
+  templateUrl: './cwl-viewer.html',
+  providers: [CwlViewerService],
+  styleUrls: ['./cwl-viewer.scss']
+})
+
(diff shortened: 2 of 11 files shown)


src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    this.resetZoom();
+  }
+  @Input() expanded: boolean;
+
+  public cwlViewerError = false;
+  public cwlViewerPercentageZoom = 100;
+  public cwlViewerDescriptor: CwlViewerDescriptor;
+  public loading = false;
+
+  private _selectedVersion;
```
garyluu: https://angular.io/guide/styleguide Style 03-04

Avoid prefixing private properties and methods with an underscore.

src/app/workflow/dag/cwl-viewer/cwl-viewer.service.spec.ts
```diff
@@ -0,0 +1,89 @@
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { CwlViewerService } from './cwl-viewer.service';
+import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
+import { Dockstore } from '../../../shared/dockstore.model';
+
+describe('Service: commonwl', () => {
```
garyluu: Leaving the name as the generated default for consistency.

src/app/workflow/dag/cwl-viewer/cwl-viewer.service.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Returns the CommonWL endpoint to request a visualization.
+   *
+   * @param {string} providerUrl
+   * @param {string} reference
+   * @param {string} workflow_path
+   * @returns {string}
+   */
+  cwlViewerEndpoint(providerUrl: string, reference: string, workflow_path: string) {
```
garyluu: Doesn't actually have a string return type defined `: string`

src/app/workflow/dag/cwl-viewer/cwl-viewer.service.ts
```diff
@@ -0,0 +1,119 @@
...
+    return Dockstore.CWL_VISUALIZER_URI + '/workflows?url='
+      + encodeURIComponent(providerUrl + '/blob/' + reference
+        + workflow_path);
+  }
+
+  private pollJobQueue(queueUrl: string) {
```
garyluu: missing return type

src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
```diff
@@ -0,0 +1,90 @@
...
+@Component({
+  selector: 'app-cwl-viewer',
+  templateUrl: './cwl-viewer.html',
+  providers: [CwlViewerService],
+  styleUrls: ['./cwl-viewer.scss']
+})
+
+export class CwlViewerComponent implements OnInit {
+
+  @Input() workflow: any;
```
garyluu: See if `any` can be changed

src/app/workflow/dag/dag.component.ts
```diff
@@ -190,7 +203,6 @@ export class DagComponent implements OnInit, AfterViewChecked {
   download() {
     if (this.cy) {
       const pngDAG = this.cy.png({ full: true, scale: 2 });
-      const name = this.workflow.repository + '_' + this._selectedVersion.name + '.png';
```
garyluu: Not sure what happened here.

src/app/workflow/workflow.component.html
```diff
@@ -132,7 +132,7 @@ <h3>
         <app-tool-tab [selectedVersion]="selectedVersion"></app-tool-tab>
       </tab>
       <tab id="dagTab" [disabled]="isStub()" heading="DAG" *ngIf="validVersions" (select)="setTab('DAG')">
-        <app-dag *ngIf="checkMode('DAG')" [id]="workflow?.id" [validVersions]="validVersions" [defaultVersion]="defaultVersion" [selectedVersion]="selectedVersion"></app-dag>
+        <app-dag *ngIf="checkMode('DAG')" [id]="workflow?.id" [validVersions]="validVersions" [defaultVersion]="defaultVersion" [selectedVersion]="selectedVersion" [theWorkflow]="workflow"></app-dag>
```
garyluu: If passing in the entire workflow, see if [id] can be removed

src/app/shared/dockstore.model.ts
```diff
@@ -49,4 +49,6 @@ export class Dockstore {
   static readonly GITLAB_AUTH_URL = 'https://gitlab.com/oauth/authorize';
   static readonly GITLAB_CLIENT_ID = 'fill_this_in';
   static readonly GITLAB_REDIRECT_URI = Dockstore.LOCAL_URI + '/auth/' + Provider.GITLAB;
+
+  static readonly CWL_VISUALIZER_URI = 'https://view.commonwl.org';
```
garyluu: If it's a constant that rarely changes and only used in one component and service, consider moving it out of this file.  Otherwise, we will have to modify compose_setup's dockstore.model.ts template.

denis-yuen: I think this is actually fine, I think it fits in along with the gitlab, github, etc.

src/app/workflow/dag/cwl-viewer/cwl-viewer.service.spec.ts
garyluu: Apparently the new CLI generates a different naming nowadays.  Should be "Service: CWLViewer" or similar

src/app/workflow/dag/cwl-viewer/cwl-viewer.html
```diff
@@ -0,0 +1,16 @@
+<div [ngClass]="{'fullscreen-element large-dag': expanded}" class="cwl-viewer" >
+  <div *ngIf="cwlViewerError" class="alert alert-warning">
+    <span class="glyphicon glyphicon-warning-sign"></span>&nbsp;
+    <span>The Common Workflow Language Viewer was unable to process the CWL. Note that the viewer only supports
+      branches and does not support tags.</span>
+  </div>
+  <div>
+    <span *ngIf="!cwlViewerError">See on <a [href]="cwlViewerDescriptor?.webPageUrl" target="_blank">https://view.commonwl.org <span class="glyphicon glyphicon-new-window"></span></a></span>
```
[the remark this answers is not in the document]
garyluu: To use a constant in the template, I had to import it in the component.ts file and put something strange like `CWL_VISUALIZER_URI = CWL_VISUALIZER_URI` in beginning of the class.  Not sure if there's a better way.

src/app/workflow/dag/dag.component.ts
coverbeck: Good catch, thanks! At one point I had all my changes in dag.component.ts, before I decided to pull them into their own component, and I must have left this by accident.

src/app/workflow/dag/cwl-viewer/cwl-viewer.service.spec.ts
garyluu: Not sure about everyone else, but I use IDE (which uses Angular CLI) to generate all my components/services .ts, .spec.ts, .html, .scss.  For the most part, we've been trying to keep the naming somewhat consistent.

src/app/workflow/workflow.component.html
coverbeck: Good catch, thanks.

src/app/workflow/dag/cwl-viewer/cwl-viewer.service.ts
garyluu: Was more of an observation that the jsdoc has `@returns {string}` but the method doesn't explicitly define it.  My jsdocs are auto-generated from the method signature, so I find it odd that there's a mismatch

src/app/workflow/dag/cwl-viewer/cwl-viewer.service.spec.ts
coverbeck: I've noticed it in my IDE but haven't been in the habit of using it; I'll try to start.

src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
coverbeck: Changed

src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
coverbeck: Changed

src/app/workflow/dag/dag.component.ts
```diff
@@ -34,6 +34,7 @@ export class DagComponent implements OnInit, AfterViewChecked {
   @Input() defaultVersion: any;
   @Input() id: number;
   _selectedVersion: WorkflowVersion;
+  @Input() theWorkflow: any;
```
garyluu: Check to see if theWorkflow is still used

src/app/workflow/dag/dag.component.ts
```diff
@@ -54,9 +55,21 @@ export class DagComponent implements OnInit, AfterViewChecked {
   private tooltip: string;
   public missingTool;
   private refresh = false;
+
+  public dagType: 'classic' | 'cwlviewer' = 'classic';
+  public enableCwlViewer = false;
+  public refreshCounter = 1;
+  public cwlViewerError: boolean;
```
garyluu: Check to see if cwlViewerError is still used in this component

denis-yuen:
Hurray!

PR #200, opened 2018-02-22
Feature/path removal

agduncan94:
Deals with removal of the path column.
Also fixes an issue with changing between tools when some are manual/without tags on the my tools page.


cypress/integration/mytools.js
```diff
@@ -60,6 +60,24 @@ describe('Dockstore my tools', function() {
 
   describe('manually register an Amazon ECR tool', function() {
     it("register tool", function() {
+      cy
+        .server()
+        .route({
+            method: "GET",
+            url: /refresh/,
+            response: {"id": 40000, "author":null,"description":null,"labels":[],"users":[{"id":1,"username":"user_A","isAdmin":false,"name":"user_A"}],"email":null,"defaultVersion":null,"lastUpdated":1482334377743,"gitUrl":"git@github.com:testnamespace/testname.git","mode":"MANUAL_IMAGE_PATH","name":"testname","toolname":"","namespace":"testnamespace","registry":"AMAZON_ECR","lastBuild":null,"tags":[],"is_published":false,"last_modified":null,"default_dockerfile_path":"/Dockerfile","defaultCWLTestParameterFile":"/test.cwl.json", "defaultWDLTestParameterFile":"/test.cwl.json","default_cwl_path":"/Dockstore.cwl","default_wdl_path":"/Dockstore.wdl","tool_maintainer_email":"test@email.com","private_access":true,"path":"amazon.dkr.ecr.test.amazonaws.com/testnamespace/testname","tool_path":"amazon.dkr.ecr.test.amazonaws.com/testnamespace/testname", "custom_docker_registry_path": "amazon.dkr.ecr.test.amazonaws.com"}
```
denis-yuen: Hmmm, tricky. 
I've been encouraging full returns of objects from the webservice to go into the DB so that they're automatically fixed for future changes by migration.  But this is a refresh call which is trickier

src/app/container/email.service.ts
```diff
@@ -1,6 +1,7 @@
 import { ExtendedDockstoreTool } from './../shared/models/ExtendedDockstoreTool';
 import { DockstoreService } from './../shared/dockstore.service';
 import { Injectable } from '@angular/core';
+import { DockstoreTool } from './../shared/swagger/model/dockstoreTool';
```
garyluu: Doesn't seem used

cypress/integration/mytools.js
agduncan94: Ya it would be nice if there was some way to keep these in the db somehow, but since this is testing the addition of a new tool it might be hard to do.

PR #204, opened 2018-02-27
Feature/1169/add checker workflow

garyluu:
This is the 2nd PR

For ga4gh/dockstore#1169

This adds all the user controls to the checker workflow path including:
- add/create checker workflow (endpoint should create the checker workflow and return the current workflow complete with the newly populated checker_id)
- edit checker workflow path (endpoint should receive a new string (query param) to replace the checker workflow path of the current tool/workflow (specified with id path parameter)

Additionally this does some heavy refactoring because many services are duplicated across the modules when they shouldn't be (stateService, containerService, workflowService).  More testing would be preferred.  

TODO:
- hook up to all the actual webservice endpoints and then update the refreshService, stateService, containerService/workflowService.
- integration tests


src/app/app.module.ts
```diff
@@ -1,5 +1,7 @@
-import { GA4GHService } from './shared/swagger/api/gA4GH.service';
-/*
+import { DescriptorLanguageService } from './shared/entry/descriptor-language.service';
+import { ErrorService } from './shared/error.service';
```
denis-yuen: FYI, there's some issue with your IDE placing imports before copyright notices

garyluu: Ya, I made an issue about it https://github.com/buehler/typescript-hero/issues/409.  I generally fix it when I notice it (there's been lots).  It'd probably be a 3 character fix for the plugin.  I wonder if I should make a PR to it...

garyluu:
Going to fix merge conflicts

PR #205, opened 2018-02-27
Feature/more nextflow hookup

denis-yuen:
Hook up nextflow more (test json, example path message, DAG message)

### DIFF
--- a/src/app/shared/validationMessages.model.ts
+++ b/src/app/shared/validationMessages.model.ts
@@ -33,13 +33,13 @@ export const formErrors = {
   'customDockerRegistryPath': ''
 };
 
-export const examplePatterns = {
+export const exampleDescriptorPatterns = {
   'cwl': 'e.g. /Dockstore.cwl',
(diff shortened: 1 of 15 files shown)


src/app/shared/validationMessages.model.ts
```diff
@@ -33,6 +33,12 @@ export const formErrors = {
   'customDockerRegistryPath': ''
 };
 
+export const examplePatterns = {
```
garyluu: exampleDescriptorPatterns or placeholderDescriptorPatterns might be more descriptive

garyluu: I mentioned descriptorPatterns because 
```
'cwl': 'e.g. /Dockstore.cwl',
'wdl': 'e.g. /Dockstore.wdl',
'nextflow': 'e.g. /nextflow.config'
```
seems to only contains descriptors right now.  If it will later encompass many more placeholders of different variety, then it would be fine to leave as is.  Same idea would apply to the validationPatterns naming below.

PR #208, opened 2018-02-28
Don't encode search urls either

garyluu:
For issue ga4gh/dockstore#1097

Don't encode the search entries either


denis-yuen:
FYI, link your PRs to issues for review and changelog writing purposes

PR #209, opened 2018-03-01
Show unpublished entries

garyluu:
Allow the owner's unpublished tools/workflows to be shown in the /tools and /workflows paths.  This provides the user the ability to browse to their own unpublished workflow/tool with a link (currently, they won't be able to edit it currently since it's disabled on a "public" page but there shouldn't be too much of a problem doing so in the future).

### DIFF
--- a/src/app/container/container.component.ts
+++ b/src/app/container/container.component.ts
@@ -1,6 +1,4 @@
-import { EmailService } from './email.service';
-import { ExtendedDockstoreTool } from './../shared/models/ExtendedDockstoreTool';
-/*
+/**
  *    Copyright 2017 OICR
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


denis-yuen:
What ticket is this for? I think there must have been a miscommunication here. 
The closest I can think of was that we wanted to be able to link to tools and workflows in https://dockstore.org/my-tools and https://dockstore.org/my-workflows

garyluu:
Quite a while ago, Jeltje mentioned how her own tools/workflows couldn't be modified from /tools and /workflows and was rather confusing.  This is the first step towards that.  I currently made these changes locally so I can easily test a few things before adding similar functionality to /mytools and /myworkflows

garyluu:
Opened https://github.com/ga4gh/dockstore/issues/1204  .  Closing PR since it'll take a lot more modifications for it to work.

denis-yuen:
Ok, this is straying a bit too far from our current set of tasks. 
We can discuss at stand-up, thanks for looking into this

PR #213, opened 2018-03-07
changed how we do workflow names

agduncan94:
Changes workflow names in paths to be listed as path/workflowname instead of path (workflowname). This makes it consistent with tools.

Deals with
https://github.com/ga4gh/dockstore/issues/1143

### DIFF
--- a/src/app/workflows/list/list.component.html
+++ b/src/app/workflows/list/list.component.html
@@ -33,7 +33,7 @@
                 <span class="glyphicon glyphicon-ok" tooltip="Verified"></span>
               </a>
             </span>
-            <a (click)="sendWorkflowInfo(workflow)" [routerLink]="('/workflows/' + workflow.full_workflow_path )">{{ workflow?.repository + (workflow.workflowName ? '/' + workflow.workflowName : '') }}</a>
+            <a (click)="sendWorkflowInfo(workflow)" [routerLink]="('/workflows/' + workflow?.full_workflow_path )">{{ workflow?.repository + (workflow?.workflowName ? '/' + workflow?.workflowName : '') }}</a>
           </td>
           <td>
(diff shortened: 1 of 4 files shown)


src/app/workflows/list/list.component.html
```diff
@@ -33,7 +33,7 @@
                 <span class="glyphicon glyphicon-ok" tooltip="Verified"></span>
               </a>
             </span>
-            <a (click)="sendWorkflowInfo(workflow)" [routerLink]="('/workflows/' + workflow.full_workflow_path )">{{ workflow?.repository + (workflow.workflowName ? ' (' + workflow.workflowName + ')' : '') }}</a>
+            <a (click)="sendWorkflowInfo(workflow)" [routerLink]="('/workflows/' + workflow.full_workflow_path )">{{ workflow?.repository + (workflow.workflowName ? '/' + workflow.workflowName : '') }}</a>
```
denis-yuen: Looks like a mixture of `workflow?.` and `workflow.`
Not sure if this is a problem

agduncan94: I think the ? only looks for the .field if the workflow is not null.

garyluu: I think Angular crashes when there isn't a '?' and workflow happens to be null at some point.  If there's no chance of workflow being null, it should be fine.

garyluu:
Additionally, try replacing the entire expression with `workflow?.full_workflow_path` and see what happens

src/app/workflows/list/list.component.html
agduncan94: https://stackoverflow.com/questions/45922715/what-is-the-meaning-of-question-mark-in-expressions-in-angular-2

agduncan94:
@garyluu The only problem with that is that it includes the source control, we just want the repo and workflowname.

PR #217, opened 2018-03-14
Feature/1169/add checker workflow

garyluu:
For ga4gh/dockstore#1169 and ga4gh/dockstore#1170

Additionally:
- it sorts the workflows in each organization in my-workflows by full workflow path.
- change default style extension to scss
- updates to new webservice
- removes a few duplicated HTTP calls

TODO in another PR: 
- Missing test parameter file auto-fill
- MyTools url change upon select
- Reproduce parent entry button showing when it's not supposed to
- make GA4GH endpoint path a constant


src/app/container/launch/tool-launch.service.spec.ts
```diff
@@ -57,4 +56,10 @@ describe('ToolLaunchService', () => {
         expect(service.getDockstoreSupportedCwlMakeTemplateString('quay.io/briandoconnor/dockstore-tool-bamstats', '1.25-11')).toContain(
             'cwltool --make-template quay.io/briandoconnor/dockstore-tool-bamstats:1.25-11 > input.yaml');
     }));
+    it('should get the right check tool command', inject([ToolLaunchService], (service: ToolLaunchService) => {
+        expect(service).toBeTruthy();
+        expect(service.getCheckToolString('potato', null)).toBe('$ dockstore tool check --entry potato checkparam.json');
+        expect(service.getCheckToolString('potato', 'stew')).toBe('$ dockstore tool check --entry potato:stew checkparam.json');
```
agduncan94: I think these should be dockstore checker ... instead of dockstore tool check...

src/app/myworkflows/my-workflow/my-workflow.component.html
```diff
@@ -0,0 +1,85 @@
...
+      <!--Tool Content Page-->
+      <div class="row">
+        <div class="col-md-3 containers-rsb">
+          <accordion [closeOthers]="oneAtATime">
+            <accordion-group *ngFor="let orgObj of orgWorkflows"[heading]="orgObj?.sourceControl + '/' + orgObj?.organization" panelClass="containers-accordion" [isOpen]="orgObj?.isFirstOpen">
+              <div *ngFor="let workflowObj of orgObj?.workflows">
+                <div class="panel-container-label" [ngClass]="{selected: workflow.id === workflowObj.id}">
+                  <div class="container-name-oflw pull-left" title="{{workflowObj?.repository + (workflowObj?.workflowName ? ' (' + workflowObj?.workflowName + ')' : '')}}">
+                    <a style="cursor: pointer;" (click)="goToWorkflow(workflowObj)">
+                      {{workflowObj?.repository + (workflowObj?.workflowName ? ' (' + workflowObj?.workflowName + ')' : '')}}
```
agduncan94: Make sure this merges properly to the new format with a slash instead of brackets.

src/app/container/launch/tool-launch.service.spec.ts
garyluu: I'm following the specs which say:
```
# runs the original tool or workflow
dockstore tool launch --entry sample_workflow:1.2 --json param.json
dockstore workflow launch --entry sample_tool:1.2 --json param.json
# runs the checker workflow
dockstore tool check --entry sample_workflow:1.2 --json param.json
dockstore workflow check --entry sample_tool:1.2 --json param.json
```

denis-yuen: I'm ok with either, I think we were leaning toward `dockstore checker` to group all the chcker related commands/code.  This will need to be consistent between the UI and the CLi though obviously

cypress/integration/WorkflowsSearch.js
```diff
@@ -7,23 +7,24 @@ describe('Dockstore workflow search page', function() {
    describe('Select a workflow', function() {
     it('Should display the correct url', function() {
       cy.get('tbody')
-      .children('tr')
+      .children('tr:nth-child(2)')
```
denis-yuen: the 2nd child seems fragile, is there a way to use an id or something stable?

src/app/mytools/mytools.component.ts
```diff
@@ -107,15 +108,19 @@ export class MyToolsComponent implements OnInit {
...
           matchingWorkflow = container;
         }
       });
     });
     return matchingWorkflow;
 
   }
+  goToTool(tool: ExtendedDockstoreTool) {
+    this.containerService.setTool(tool);
+    // this.router.navigateByUrl('/my-tools/' + tool.tool_path);
```
denis-yuen: New commented code, same drill

src/app/shared/checker-workflow.service.ts
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Determines if a given entry is a workflow or not (tool).
+     * Returns true if entry is a workflow, returns false if entry is not a workflow (i.e. a tool)
+     * @param entry The entry to check
+     */
+    private isEntryAWorkflow(entry: Entry): boolean {
+        if (!entry) {
+            return null;
+        }
+        if (entry.hasOwnProperty('is_checker')) {
```
denis-yuen: Minor nitpick: shouldn't these five lines just be 
`return entry.hasOwnProperty('is_checker')`?

src/app/shared/entry/launch-checker-workflow/launch-checker-workflow.component.html
```diff
@@ -0,0 +1,4 @@
+<div id="launchCheckerWorkflow" *ngIf="checkerWorkflowPath$ | async" class="well well-sm" tooltip="Commands for checking the checker workflow of this tool/workflow through Dockstore CLI">
```
denis-yuen: Hmmm, better to say that we run a checker workflow rather than check a check a checker workflow

src/app/workflow/launch/workflow-launch.service.ts
```diff
@@ -1,4 +1,7 @@
-/*
+import { Injectable } from '@angular/core';
```
denis-yuen: Fix your IDE, also update your copyright generator to 2018

src/app/workflow/launch/workflow-launch.service.ts
```diff
@@ -13,24 +16,45 @@
...
 
   getCliString(path: string, versionName: string, currentDescriptor: string) {
-    return `$ dockstore workflow launch --entry ${ path }:${ versionName } --json Dockstore.json`;
+    return `$ dockstore ${this.type} launch --entry ${path}:${versionName} --json Dockstore.json`;
   }
 
   getCwlString(path: string, versionName: string, mainDescriptor: string) {
-    return `$ cwl-runner ${ Dockstore.API_URI }/api/ga4gh/v1/tools/${ encodeURIComponent('#workflow/' + path) }` +
-      `/versions/${ encodeURIComponent(versionName) }/plain-CWL/descriptor/${mainDescriptor} Dockstore.json`;
+    return `$ cwl-runner ${Dockstore.API_URI}/api/ga4gh/v1/tools/${encodeURIComponent('#workflow/' + path)}` +
```
denis-yuen: Let's target v2

cypress/integration/WorkflowsSearch.js
garyluu: the row are generated without id.  generating a unique id per row seems hacky

src/app/mytools/mytools.component.ts
garyluu: It's to be used in my PR right after.

cypress/integration/WorkflowsSearch.js
agduncan94: Maybe a comment on what the 2nd child should be, in case more children are added/removed in the future.

garyluu: The lines below it make it obvious what it's supposed to select

src/app/mytools/mytools.component.ts
denis-yuen: Add it in that PR then

PR #220, opened 2018-03-16
Feature/extended checker workflow changes

garyluu:
Miscellaneous improvements to https://github.com/dockstore/dockstore-ui2/pull/217 

- Use a ga4gh constant to make it easier to change versions or paths later on
- Selecting my-tools' tools should change the url now
- pre-populate the test parameter file path when registering a new checker workflow

Still can't reproduce the weird issue with an entry showing it has a parent entry when it has none.


src/app/container/dockerfile/dockerfile.component.ts
```diff
@@ -73,7 +72,7 @@ export class DockerfileComponent implements AfterViewChecked {
   }
 
   getDockerfilePath(): string {
-    const basepath = Dockstore.API_URI + '/api/ga4gh/v2/tools/';
+    const basepath = Dockstore.API_URI + ga4ghPath + '/tools/';
```
denis-yuen: Good idea

src/app/mytools/my-tool/my-tool.component.scss
```diff
@@ -0,0 +1,15 @@
+/*
```
denis-yuen: Is this intentional? This looks like all copyright?

garyluu: Yes, Angular CLI auto-generates an SCSS file for each component and hooks the .ts file to it.  We've mostly kept the empty scss file in case it's needed.  My new copyright add-on automatically adds the copyright to new files.

PR #221, opened 2018-03-17
Add Feature Flags #268

coverbeck:
Add a "feature flags" property in dockstore.model.ts, so that
features can be turned on/off without doing new Git commits.
EnableCwlViewer is the only flag in there so far.

Also added a comment to dockstore.model.ts warning to keep it in sync
with the template in the dockstore/compose_setup repo, as I was
previously unaware of the need.

Not sure if this approach is overkill, but it does seem like it would be nice to be able to turn on a feature with doing a new commit, and it looks like another feature, Launch with FireCloud, may only get turned on in staging, but not prod, at first.

### DIFF
--- a/src/app/shared/dockstore.model.ts
+++ b/src/app/shared/dockstore.model.ts
@@ -16,6 +16,9 @@
 
 import { Provider } from './enum/provider.enum';
 
+// If you add new properties to this class, you need to also update
+// dockstore.model.ts.template at https://github.com/dockstore/compose_setup
+
 export class Dockstore {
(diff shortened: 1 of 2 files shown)


src/app/shared/dockstore.model.ts
```diff
@@ -16,6 +16,9 @@
 
 import { Provider } from './enum/provider.enum';
 
+// If you add new properties to this class, you need to also update
```
denis-yuen: Good idea

PR #222, opened 2018-03-20
Feature/launch with firecloud

coverbeck:
Add Launch with FireCloud button to WDL workflows that do not have import statements.

screen shot 2018-03-20 at 10 42 15 am

### DIFF
--- a/src/app/shared/firecloud.service.ts
+++ b/src/app/shared/firecloud.service.ts
@@ -1,0 +1,31 @@
+import { Injectable } from '@angular/core';
+import { Dockstore } from './dockstore.model';
+
+@Injectable()
+export class FireCloudService {
+
+  private importRegEx: RegExp = new RegExp(/^\s*import\b/, 'm');
+
+  constructor() { }
+
+  /**
+   * Simple test to see if wdl has any import statements.
+   *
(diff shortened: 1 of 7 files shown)


src/app/shared/firecloud.service.ts
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Simple test to see if wdl has any import statements.
+   *
+   * Look for any statement starting with the word `import` as the first non-blank characters at the beginning
+   * of any line.
+   *
+   * @param {string} wdl
+   * @returns {boolean}
+   */
+  wdlHasImports(wdl: string): boolean {
```
denis-yuen: This brings up a good question. 
How well is Dockstore dealing with `import` as a keyword anyway?
Do we properly create secondary files, if so this could check for >0 secondary files rather than regex although this works for now

coverbeck: http imports should work though, shouldn't they? Although FC asked us to disable for even http import for now.

denis-yuen: Yes, http imports should work no matter what. 
I was just curious what the state of relative imports was. But you're right, http imports would have to be looked for like this either way.

PR #224, opened 2018-03-23
Feature/workflow redirect

agduncan94:
Adds redirect code for workflows that do not have the expected git registry.


agduncan94:
It seems like the wrong column is being selected. Though I don't know why it would sometimes pass.

PR #225, opened 2018-03-23
Feature/logo change

agduncan94:
Updates logos to have two rows on the home page. Also adds a funding page.

Edit: https://github.com/ga4gh/dockstore/issues/1128


src/app/funding/funding.component.html
```diff
@@ -0,0 +1,71 @@
...
+          Funding
+        </h3>
+      </div>
+    </div>
+  </div>
+</div>
+
+<div class="container">
+  <div class="row-fluid" style="margin-top:20px;">
+    <p class="lead">Support for Dockstore is provided by grants from Genome Canada, the National Institutes of Health, and the Province of Ontario.</p>
```
denis-yuen: Repeated, but fix in a future PR

src/app/sponsors/sponsors.service.ts
```diff
@@ -21,10 +21,21 @@ export class SponsorsService {
     new Sponsor('collaboratory.png', new URL('https://www.cancercollaboratory.org/')),
     new Sponsor('oicr.png', new URL('https://oicr.on.ca/')),
     new Sponsor('ga4gh.png', new URL('https://genomicsandhealth.org/')),
-    new Sponsor('cwl.png', new URL('http://www.commonwl.org/'))
+    new Sponsor('ucsc.png', new URL('https://www.ucsc.edu/'))
```
denis-yuen: Hmmm, let's link to https://ucscgenomics.soe.ucsc.edu/ instead. 
(I know its technically a different logo, maybe add a comment to that effect rather than changing it now)

src/app/sponsors/sponsors.service.ts
```diff
@@ -21,10 +21,21 @@ export class SponsorsService {
...
+    // new Sponsor('broad.png', new URL('https://www.broadinstitute.org/'))
+  ];
+
+  private partners: Sponsor[] = [
+    new Sponsor('cwl.png', new URL('http://www.commonwl.org/')),
+    new Sponsor('wdl.png', new URL('https://software.broadinstitute.org/wdl/')),
```
denis-yuen: I think they prefer http://www.openwdl.org/ now

src/app/funding/funding.component.html
agduncan94: Should this just be removed?

src/app/sponsors/sponsors.component.ts
```diff
@@ -39,7 +49,22 @@ export class SponsorsComponent implements OnInit {
...
   ngOnInit() {
+    // Initialize sponsors and partners
     this.sponsors = this.sponsorsService.getSponsors();
+    this.partners = this.sponsorsService.getPartners();
+
+    // Hide second row if necessary
+    this.hideSecondRow();
+  }
+
+  hideSecondRow() {
```
garyluu: check how many times this is executed

src/app/sponsors/sponsors.component.ts
```diff
@@ -27,8 +29,16 @@ import { SponsorsService } from './sponsors.service';
 export class SponsorsComponent implements OnInit {
 
   public sponsors: Sponsor[];
+  public partners: Sponsor[];
+  public showSecondRow = false;
 
-  constructor(private sponsorsService: SponsorsService) { }
+  constructor(private sponsorsService: SponsorsService, private location: Location, private router: Router) {
+    router.events.subscribe((event) => {
```
garyluu: Any particular reason it's in the constructor instead on OnInit?

agduncan94: I was just going off of examples online.

src/app/sponsors/sponsors.component.ts
agduncan94: Ya I was calling it once during the oninit, when I only really need to call it once on route change. This isn't too big of a deal since it is only called one extra time on full refresh of a page, but I have removed it since it is redundant.

PR #226, opened 2018-03-23
Feature/1094/my entry sidebar

garyluu:
Leaving this here for now for issue ga4gh/dockstore#1094

Changes the my-tools/my-workflows' select entries component to have a toggle between published and unpublished entries.  Tests currently use `wait` in order for the rendering to finish before clicking.  Ideally, another should function should be used to determine when it's done rendering.  Additionally, change it so that the rendering can be easily detected and done once.

Try reverting the Cypress update (to autoscroll to component) and resolution change (just in case it's not visible)

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,3 +1,6 @@
 {
-  "projectId": "1ya4vj"
+  "projectId": "1ya4vj",
+  "viewportHeight": 1080,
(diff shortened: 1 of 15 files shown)


denis-yuen:
Do not delete

PR #227, opened 2018-03-23
Update webservice

garyluu:
Update webservice.  Currently builds.  hopefully doesn't require migration script for tests

Update: It did require migration :(
In general, named registry => registry_string, renamed reigstry_provider => registry.  Also ran migration script 1.4.0


garyluu:
Fixed the tests now.  When I said "build", i meant `ng build`

PR #231, opened 2018-03-28
Basically reverting a7f3ccd0def0319d077a751834a61281fbfe6a87

garyluu:
Reverting change to launch commands


agduncan94:
Just to clarify, a checker workflow should use dockstore workflow convert ... for creating a JSON.
For launching, it should use dockstore checker launch --entry, where the entry is the parent of the checker, and not the checker itself.

Is this how it is implemented?

garyluu:
The checker workflow has all the usual workflow commands.  The `dockstore checker launch --entry <parententry>` command is in the parent entry because it's confusing to have all the CLI commands reference itself, but then have another command reference some other entry.

garyluu:
Modified as per morning discussion

PR #232, opened 2018-03-28
Fix for https://github.com/ga4gh/dockstore/issues/1250

denis-yuen:
For the UI


garyluu:
`this.workflowService.upsertWorkflowToWorkflow(response);` might be safer in case the subscription doesn't update `this.workflow` faster enough

PR #234, opened 2018-03-28
Use secondaryFiles for FC Launch Eligibility

coverbeck:
We were trying to not expose Launch with Firecloud if the WDL contained
any WDL import statements. The test for this was a simple Regex that
looked for any lines that began with import. This was not a good test
because there was a workflow that had some inline Python code that
was doing an import.

On further discussion with FireCloud team, FireCloud does support
http imports. So instead rely on `workflowsService.getSecondaryWdl` to
indicate whether the WDL has an *file* imports, and disable in that
case.

This change basically gutted firecloud.service.ts, so got rid of it
as well.


src/app/workflow/workflow.component.ts
```diff
@@ -128,9 +127,9 @@ export class WorkflowComponent extends Entry {
       this.fireCloudURL = null;
       const version: WorkflowVersion = this.selectedVersion;
       if (version && this.isWdl(workflowRef)) {
-        this.workflowsService.wdl(workflowRef.id, version.name).subscribe(sourceFile => {
-          if (sourceFile.content && sourceFile.content.length && !this.firecloudService.wdlHasImports(sourceFile.content)) {
-            this.fireCloudURL = this.firecloudService.redirectUrl(workflowRef.full_workflow_path, version.name);
+        this.workflowsService.secondaryWdl(workflowRef.id, version.name).subscribe(sourceFiles => {
```
garyluu: Changing `sourceFiles =>` to `(sourceFiles: Array<SourceFile>) =>` might be helpful

coverbeck: Changed it.

PR #248, opened 2018-04-10
Add short instructions for Python 3

garyluu:
Short instructions for what to do when using Python 3

For issue ga4gh/dockstore#1242

Also added an additional check in the starring tests to make sure everything looks ok before starring.  In the future, it should be done properly instead (disable the button until it's ready to be pressed)

Moved all the pip dependencies versions to another constant in another file.

### DIFF
--- a/src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
+++ b/src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
@@ -1,9 +1,11 @@
 // tslint:disable:max-line-length
-import { Metadata } from './../../../shared/swagger/model/metadata';
-import { GA4GHService } from './../../../shared/swagger/api/gA4GH.service';
 import { Component, OnInit } from '@angular/core';
 import { AuthService } from 'ng2-ui-auth';
+
 import { Dockstore } from '../../../shared/dockstore.model';
+import { GA4GHService } from './../../../shared/swagger/api/gA4GH.service';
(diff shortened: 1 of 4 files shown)


src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -83,6 +83,8 @@ You can install the version of cwltool that we've tested for use with Dockstore
 pip install setuptools==36.5.0
 pip install cwl-runner cwltool==${this.cwltoolVersion} schema-salad==2.6.20170806163416 avro==1.8.1 ruamel.yaml==0.14.12 requests==2.18.4
 \`\`\`
+If using Python 3, install cwl-avro==1.8.3 instead of avro==1.8.1.
```
denis-yuen: This is ugly, but maybe a comment for us with the ticket to check to see when this gets fixed. 
Also 1.8.1 is repeated from above, can be a variable like cwltoolVersion

garyluu:
considering whether i should parse the requirements file in the dockstore repository now, or leave as is....

denis-yuen:
@garyluu 
I think taking the two tickets together, there's no rush so this can be done in stages 
1) Answer Brian's ticket with a fix so people can search for a solution easily
2) Add a web service endpoint ala the linked request
3) Link to it from the docs.dockstore.org FAQ and in the Angular onboarding wizard

garyluu:
Closing this until webservice part is done

PR #249, opened 2018-04-10
Now launch with is only shown if the selected version is valid

agduncan94:
The launch with section is only shown if the version selected is valid.
If it isn't then a message is shown.

Deals with https://github.com/ga4gh/dockstore/issues/1303

### DIFF
--- a/src/app/container/launch/launch.component.html
+++ b/src/app/container/launch/launch.component.html
@@ -15,45 +15,53 @@
   -->
 
 <div *ngIf="_selectedVersion">
-  <span>
-    <div class="form-inline">
-      <label tooltip="Commands to launch workflow from the command line">Launch With</label>
-      <div class="pull-right">
-        <select id="launchWithDescriptor" class="form-control input-sm" [(ngModel)]="currentDescriptor" (change)="reactToDescriptor()">
-          <option *ngFor="let descriptor of validDescriptors" [ngValue]="descriptor.value">
-            {{ descriptor.value }}
-          </option>
-        </select>
+  <div *ngIf="!_selectedVersion.valid">
+    <label tooltip="Commands to launch tool from the command line">Launch With</label>
+    <div class="alert alert-warning" role="alert">
+      This section is only available for valid versions.
+    </div>
+  </div>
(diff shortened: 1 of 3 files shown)


src/app/container/launch/launch.component.html
```diff
@@ -15,45 +15,53 @@
...
-        <select id="launchWithDescriptor" class="form-control input-sm" [(ngModel)]="currentDescriptor" (change)="reactToDescriptor()">
-          <option *ngFor="let descriptor of validDescriptors" [ngValue]="descriptor.value">
-            {{ descriptor.value }}
-          </option>
-        </select>
-      </div>
+  <div *ngIf="!_selectedVersion.valid">
+    <label tooltip="Commands to launch tool from the command line">Launch With</label>
+    <div class="alert alert-warning" role="alert">
+      This section is only available for valid versions.
```
denis-yuen: For this alert, we should probably say what the user can do to fix the situation (i.e. provide valid descriptors).
For workflows at least, still undecided about tools

src/app/workflow/launch/launch.component.html
```diff
@@ -14,37 +14,45 @@
...
     <label tooltip="Commands to launch workflow from the command line">Launch With</label>
-  </div>
-  <div class="well">
-    <div class="well well-sm" tooltip="Commands for creating a runtime JSON template">
-      Make a runtime JSON template and fill in desired inputs, outputs, and other parameters
-      <pre>{{ params }}</pre>
-      or grab one that the workflow author has provided (if applicable)
-      <pre>{{ wgetTestJsonDescription }}</pre>
+    <div class="alert alert-warning" role="alert">
+      This section is only available for valid versions.
```
denis-yuen: Same here

src/app/container/launch/launch.component.html
denis-yuen: Undecided, maybe we explain and vote in stand-up

agduncan94: Ok so this is fine to merge with just the workflow change, and we can add the tool comment at a later time?

PR #251, opened 2018-04-12
Feature/merge my entries

garyluu:
Attempt at bridging the my-tools and my-workflows components.  Can't merge the functions without casting due to types.

Also fix for when orgObjects is undefined.

Commented on why a certain test is disabled.

### DIFF
--- a/cypress/integration/mytools.js
+++ b/cypress/integration/mytools.js
@@ -144,8 +144,11 @@ describe('Dockstore my tools', function() {
                 .get('#deregisterConfirmButton')
                 .click()
 
-            
-            // This should be activated later when https://github.com/ga4gh/dockstore/issues/1318 is resolved
+            // This is deactivated because:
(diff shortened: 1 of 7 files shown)


agduncan94:
I noticed this in my local branch but we aren't properly displaying toolnames in the my tools published/unpublished tabs.

cypress/integration/mytools.js
```diff
@@ -144,8 +144,11 @@ describe('Dockstore my tools', function() {
                 .get('#deregisterConfirmButton')
                 .click()
 
-            
-            // This should be activated later when https://github.com/ga4gh/dockstore/issues/1318 is resolved
+            // This is deactivated because:
```
denis-yuen: Maybe add what can be done in the future to resolve this then (at which time this will be uncommented)
If this is likely never to be uncommented, just delete it

PR #253, opened 2018-04-16
updated error message when webservice cannot be connected to

agduncan94:
Better error message for webservice.

https://github.com/ga4gh/dockstore/issues/1321

### DIFF
--- a/src/app/shared/error.service.ts
+++ b/src/app/shared/error.service.ts
@@ -24,9 +24,18 @@ export class ErrorService {
     setErrorAlert(error: HttpErrorResponse) {
         let errorObj = null;
         if (error) {
+          if (error.status == 0) {
+            // Error code of 0 means the webservice is not responding, likely down
             errorObj = {
-                message: 'The webservice is currently down, possibly due to load. Please wait and try again later.'
+              message: 'The webservice is currently down, possibly due to load. Please wait and try again later.'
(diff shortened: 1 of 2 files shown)


src/app/shared/error.service.ts
```diff
@@ -25,9 +25,7 @@ export class ErrorService {
         let errorObj = null;
         if (error) {
             errorObj = {
-                message: 'The webservice encountered an error trying to create/modify.',
```
denis-yuen: Will this change apply for all error codes or just when the web service is down?
I'd like bad requests, internal server errors, etc. to be reported as normal but when the server is completely down (as in the linked issue), this applies

garyluu: I'm almost positive that this will change all error messages.  Also, if the webservice is known to be down and not a legit bad request, internal server error, it should just redirect to maintenance page.

agduncan94: Yes looking more into it this is the case. I'll instead show a different error message if the HTTP response is 0 (no response).

denis-yuen:
It's a good point and worth following up. 
Maybe we can tackle this initially with where the user might see a confusing "http 0" banner for now.

agduncan94:
I'll merge this and then work on a new branch to try and fix those issues.

PR #261, opened 2018-04-18
Feature/staging banner

agduncan94:
Allows you to have a banner which is only displayed on staging.

To run in staging mode, do
ng serve --env=staging

This has been added to the wiki too. Will need to update compose with this as well.

### DIFF
--- a/src/app/banner/banner.component.html
+++ b/src/app/banner/banner.component.html
@@ -1,5 +1,5 @@
 <!--
-  ~    Copyright 2017 OICR
+  ~    Copyright 2018 OICR
   ~
   ~    Licensed under the Apache License, Version 2.0 (the "License");
   ~    you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


src/app/banner/banner.component.html
```diff
@@ -0,0 +1,25 @@
+<!--
+  ~    Copyright 2017 OICR
```
garyluu: Could change the date

src/styles.scss
```diff
@@ -118,6 +118,11 @@ body {
   position: relative;
 }
 
+.staging-banner {
+  background-color: #FFEB3B;
```
garyluu: If not used in any other component, this can safely be moved into the banner component's scss.  Also, bootstrap's bg-warning color looks preferably in my opinion.

garyluu: Sadly position: relative is a bootstrap 4-only class

PR #264, opened 2018-04-19
Update cypress

garyluu:
Updating Cypress because of reasons.


garyluu:
@denis-yuen coverage for each PR will be compared against the base branch (develop).  Latest develop build on Travis seems to have error'd at the starring test.

PR #272, opened 2018-04-24
Feature/angular upgrade

garyluu:
Upgrade to Angular 5, which included updating a bunch of other packages.


src/app/test/service-stubs.ts
```diff
@@ -227,6 +227,15 @@ export class ConfigurationStub {
     };
 }
 
+export class NotificationsStubService {
+  success() {
+
```
denis-yuen: Seems odd, but possible since this is in tests.

denis-yuen:
Please follow-up with an issue/pr for compose stup

PR #274, opened 2018-04-25
Feature/mismatched webservice version

agduncan94:
On staging the banner will now specify if the webservice the UI is talking to is the expected one.

https://github.com/ga4gh/dockstore/issues/1308


src/app/banner/banner.component.ts
```diff
@@ -9,10 +13,32 @@ import { environment } from '../../environments/environment';
...
+  mismatchedVersion = false;
+  versionFromAPI: string;
+  versionBuiltWith: string;
+
+  constructor(private gA4GHService: GA4GHService) { }
 
   ngOnInit() {
     this.showBanner = environment.staging;
+
+    this.gA4GHService.metadataGet()
```
garyluu: this looks like it does a 2nd metadataGet call when the footer already did one.  double check if there's 1 or 2 metadata get calls every time the application is loaded.  If it's duplicated, somehow make the footer and banner use metadata from the same call.

agduncan94:
@garyluu Made changes you mentioned

src/app/metadata/metadata.service.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Observable } from 'rxjs/Observable';
+import { of } from 'rxjs/observable/of';
+import { Metadata } from './../shared/swagger/model/metadata';
+import { HttpErrorResponse } from '@angular/common/http';
+import { map } from 'rxjs/operators';
+import 'rxjs/add/operator/publishReplay';
+
+@Injectable()
+export class MetadataService {
+  metadata: any;
```
garyluu: Missing type here as well as methods

src/app/banner/banner.component.ts
```diff
@@ -9,10 +12,32 @@ import { environment } from '../../environments/environment';
...
+  versionFromAPI: string;
+  versionBuiltWith: string;
+
+  constructor(private metadataService: MetadataService) { }
 
   ngOnInit() {
     this.showBanner = environment.staging;
+
+    this.metadataService.getMetadata()
+      .subscribe(
```
garyluu: Missing unsubscribe, add unsubscribe to footer while ur at it.

src/app/banner/banner.component.ts
```diff
@@ -9,10 +12,32 @@ import { environment } from '../../environments/environment';
...
+        (metadata: Metadata) => {
+          if (metadata.hasOwnProperty('version')) {
+            this.versionFromAPI = metadata['version'];
+            this.versionBuiltWith = versions.version;
```
garyluu: Can probably move this out of unsubscribe

agduncan94: Do you mean out of subscribe? And move it to a function?

garyluu: out of subscribe

PR #277, opened 2018-05-01
npm updates and start of material design

garyluu:
- update almost everything for an easier transition to Angular 6 later (except ng2-auth-ui because it's more challenging)
- Also installed Angular/material as a test and changed a button to use Angular Material
- Replace notifications with Angular Material's snackbar which got rid an npm package :tada:


src/styles.scss
```diff
@@ -14,6 +14,9 @@
  *    limitations under the License.
  */
 
+// TODO: Angular Material theme to be changed to a custom one
+@import '~@angular/material/prebuilt-themes/indigo-pink.css';
```
denis-yuen: Indigo-pink? 
A snack bar? 

This might be worthy of seeing in person before merging

garyluu: A material theme refers to 5 different colours: primary, accent, warning, background, foreground.  Currently the only two components added are snackbar and a button, neither of which use the pink accent and I have no intentions of using the accent colour anytime before we get a custom theme.  The primary colour indigo currently being used looks surprisingly similar to the rest of Dockstore.

garyluu: Ok, apparently I was wrong.  The button did not use the accent colour, but the snackbar used it for the "Dismiss" word.  It just wasn't apparent to me.
selection_022

garyluu:
selection_023

PR #278, opened 2018-05-02
Feature/864/split file formats

garyluu:
Part 2 of ga4gh/dockstore#864

- Splits file format into two seperate buckets (Input and Output)
- Also fixes search that was apparently broken since Angular 5 update (elasticsearch not running in UI2 Travis, hard to tell search page is broken vs empty)
- Modifications to tag cloud query building in prep for new bodybuilder which correctly type-checks


denis-yuen:
FYI, restarted build. Looks like failure due to artifactory maintenance

PR #279, opened 2018-05-04
Feature/gitlab fix

agduncan94:
Fixes the issue where we were grouping workflows on my workflows by only the org, and not by the source control and org. 

Also started merging code for the my tools and my workflows, there is a lot of overlap now. Still lots more to be done though.


denis-yuen:
Pending builds

agduncan94:
@garyluu

PR #280, opened 2018-05-07
Feature/1372/fc launch bug

coverbeck:
Hide Launch with FireCloud if No Content

ga4gh/dockstore#1372

Created a launch-third-party component to both more easily unit
test and to add new upcoming third party integrations.

Turn on Launch with FireCloud by default
Unit test was failing with flag set to false. Since the flag has been
turned on in production, make it the default.

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -88,6 +88,7 @@ import { StargazersModule } from './stargazers/stargazers.module';
 import { StarredEntriesComponent } from './starredentries/starredentries.component';
 import { StarringModule } from './starring/starring.module';
 import { ToolDetailsComponent } from './tool-details/tool-details.component';
+import { LaunchThirdPartyComponent } from './workflow/launch-third-party/launch-third-party.component';
 
 @NgModule({
   declarations: [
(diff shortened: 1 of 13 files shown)


src/app/workflow/launch-third-party/launch-third-party.component.ts
```diff
@@ -0,0 +1,81 @@
...
+  get selectedVersion() {
+    return this._selectedVersion;
+  }
+
+  dnastackURL: string;
+  fireCloudURL: string;
+
+  constructor(private workflowsService: WorkflowsService) { }
+
+  ngOnInit() {
```
garyluu: Can possibly be removed

src/app/app.module.ts
```diff
@@ -88,6 +88,7 @@ import { StargazersModule } from './stargazers/stargazers.module';
 import { StarredEntriesComponent } from './starredentries/starredentries.component';
 import { StarringModule } from './starring/starring.module';
 import { ToolDetailsComponent } from './tool-details/tool-details.component';
+import { LaunchThirdPartyComponent } from './workflow/launch-third-party/launch-third-party.component';
```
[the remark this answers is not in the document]
coverbeck: Good catch, will remove.

src/app/workflow/launch-third-party/launch-third-party.component.ts
coverbeck: Agreed, removing.

PR #281, opened 2018-05-08
Update to alpha webservice release

denis-yuen:
Update to 1.5.0 webservice for staging release
* also syncs ui2 database structure with main dockstore repo in terms of user names, performed migrations

### DIFF
--- a/src/app/shared/entry.ts
+++ b/src/app/shared/entry.ts
@@ -198,7 +198,7 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
       }
     }
 
-    // If no url tag or default version, select first element in the dropdown
+    // If no url tag or default version, select last element in the dropdown
     if (useFirstTag && versions.length > 0) {
       selectedVersion = versions[versions.length - 1];
     }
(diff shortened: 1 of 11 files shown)


denis-yuen:
@garyluu  ran into issues updating to WS 1.5.0 
Note that the enum change was unintentional (looks like swagger doesn't understand custom serializers which broke that particular enum)

denis-yuen:
In discussion, looks like failures are all unrelated and due to some to do with the order of mocked versions. Re-running

agduncan94:
Just a note for what changes I made. If you go to a tool with no version specified then it will fallback to the last element of the versions array. There are definitely better ways to do this, I was just making two parts of the code consistent.

src/app/shared/entry.ts
```diff
@@ -200,7 +200,7 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
 
     // If no url tag or default version, select first element in the dropdown
     if (useFirstTag && versions.length > 0) {
-      selectedVersion = versions[0];
+      selectedVersion = versions[versions.length - 1];
```
garyluu: Does this now select the last element in the dropdown?

PR #287, opened 2018-05-11
Feature/hidden version fix

agduncan94:
Any version that has been hidden is not shown on public entries. They cannot be linked to.


garyluu:
If it's hidden, how come the webservice is returning it?

denis-yuen:
Maybe investigate a test for this, we're likely to switch how this is implemented to the webservice sooner rather than later, so we might as well lock down the behavior just in case we forget.

agduncan94:
@denis-yuen I added the tests for this.

PR #290, opened 2018-05-17
Fix wget test parameter file command

garyluu:
Fixes the wget test parameter file command 

TODO: return proper error message on webservice when source repo doesn't match.

### DIFF
--- a/src/app/shared/launch.service.ts
+++ b/src/app/shared/launch.service.ts
@@ -93,8 +93,12 @@ export abstract class LaunchService {
         case 'cwl':
           urlType = 'PLAIN_CWL';
           break;
-        default:
+        case 'nextflow':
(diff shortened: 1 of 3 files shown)


src/app/shared/launch.service.ts
```diff
@@ -84,11 +84,23 @@ export abstract class LaunchService {
...
+      let urlType = 'PLAIN_NFL';
+      switch (type) {
+        case 'wdl':
+          urlType = 'PLAIN_WDL';
+          break;
+        case 'cwl':
+          urlType = 'PLAIN_CWL';
+          break;
+        default:
+          urlType = 'PLAIN_NFL';
```
denis-yuen: Might be better to return a more visible error rather default to nextflow. If we go down the route of adding a fourth language, we might miss this.

PR #295, opened 2018-05-23
Feature/1422/link to test parameter file

garyluu:
For ga4gh/dockstore#1422

Changes the download test parameter file button to use the new ga4gh endpoint.  When using the relative-path endpoints, opt to not encode the last part since the webservice is able to handle it and downloading the file ends up with a better file name. 

Downloading the file, as opposed to going to the url only works with Chrome.  Firefox only allows download when it's the same origin (workaround requires using content disposition or changing content type)


denis-yuen:
1) Which version of the dockstore webservice does this use? Some of the parameters seem unfamiliar to me
2) this is required as a hotfix

garyluu:
Closing in favour of hotfix base

PR #300, opened 2018-06-01
Convert to relative paths for all

garyluu:
For ga4gh/dockstore#1422

Changing the absolute paths of the test parameter file download links to relative paths.

Conversion is mostly done using an altered nodejs path module relative function.

The relative path is based on the directory of the descriptor (not the descriptor itself).  See test for expected output.

Only works with hotfix/master webservice branch currently


denis-yuen:
Distracting environment, might be difficult for me to understand this.
But at first blush, it seems like you're calculating the relative path. 
I think you should be able to just get the relative path from the files endpoint, pending the fix https://github.com/ga4gh/dockstore/issues/1448 for test files (and maybe container files)

garyluu:
Trying to avoid chaining too many http requests in sequence.  All info needed is already available to calculate the relative path.

PR #306, opened 2018-06-14
Homepage fix tools

denis-yuen:
* Fixes layout of tables on the homepage and tools/workflows pages
* Fixes layout of entry versions table
* Comment out GitLab

### DIFF
--- a/cypress/integration/dropdown.js
+++ b/cypress/integration/dropdown.js
@@ -1,4 +1,5 @@
 describe('Dropdown test', function() {
+    // TODO: GitLab tests are commented out
     require('./helper.js')
 
     beforeEach(function() {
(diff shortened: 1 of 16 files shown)


agduncan94:
@denis-yuen I can't assign you but I made it so that GitLab does not appear in dropdowns and the token is not visible on the accounts page.

cypress/integration/dropdown.js
```diff
@@ -56,9 +57,11 @@ describe('Dropdown test', function() {
...
         it('Should show all accounts as linked (except GitLab and Bitbucket)', function() {
             everythingOk();
-            cy.visit(String(global.baseUrl) + '/auth/gitlab.com?code=somefakeid', {'failOnStatusCode': false}).then((resp) => {
-                expect(resp.status).to.eq('')
-            })
+            // cy.visit(String(global.baseUrl) + '/auth/gitlab.com?code=somefakeid', {'failOnStatusCode': false}).then((resp) => {
+            //     expect(resp.status).to.eq('')
+            // })
+            // TODO: Gitlab is being very slow, hopefully one day we can remove this
+            cy.wait(10000);
```
garyluu: After commenting it out, it looks like it's checking if everything is ok, then  waiting, then checking if everything is ok again even though nothing has changed

PR #309, opened 2018-06-18
Search for checker workflows

denis-yuen:
* needs 8fea92dba4b63338f84ca821ca31997594c317f4 from the webservice
* implementation for search page, looking for checker workflows https://github.com/ga4gh/dockstore/issues/1486
* related to TRS work but shouldn't require TRS update


denis-yuen:
Note to self: this requires a TRS update and should not be merged till then

PR #313, opened 2018-06-21
Feature/1505/remove rxjs compat

garyluu:
Update all third party packages due to rxjs 6 changes.

For issue ga4gh/dockstore#1505

Had to switch swagger-codegen to another generator because it doesn't support rxjs 6 yet, but swagger-codegen's master branch currently has it eventually we can switch back.


denis-yuen:
Hmmm, check my math but it looks like a 0.2% difference in non-prod and looks like 5% difference in production. Kinda disappointing and doesn't seem worth it. 

I'm tempted so say just to cherry-pick the easier updates like share buttons and leave the rest until swagger-codegen is updated

garyluu:
Holding off on this until swagger codegen 2.4.0 is stable

PR #314, opened 2018-06-21
Feature/register workflow disable button

agduncan94:
Fixes the issue where you cannot register a workflow because the button is disabled.
Also changes some buttons I missed to material.


garyluu:
should hook up the `refreshing` instead of just removing it.

agduncan94:
@garyluu fixed

PR #315, opened 2018-06-21
fix search label link

agduncan94:
Now clicking on a label properly goes to search page.
https://github.com/ga4gh/dockstore/issues/1197

### DIFF
--- a/src/app/workflow/workflow.component.html
+++ b/src/app/workflow/workflow.component.html
@@ -90,7 +90,7 @@
   <div *ngIf="workflow" class="col-sm-12" style="margin-bottom: 5px;">
     <span *ngIf="!labelsEditMode">
       <mat-chip-list>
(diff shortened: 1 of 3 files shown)


src/app/workflow/workflow.component.html
```diff
@@ -90,7 +90,7 @@ <h3>
   <div *ngIf="workflow" class="col-sm-12" style="margin-bottom: 5px;">
     <span *ngIf="!labelsEditMode">
       <mat-chip-list>
-        <mat-chip color="primary" *ngFor="let label of workflowEditData" (click)="goToSearch(label)">{{label}}</mat-chip>
+        <mat-chip color="primary" *ngFor="let label of workflowEditData.labels" (click)="goToSearch(label)">{{label}}</mat-chip>
```
garyluu: careful of null workflowEditData

PR #317, opened 2018-06-22
Feature/1435/commit

garyluu:
For ga4gh/dockstore#1435

Add commit id tooltip to every git reference in both tool versions and workflow versions.  

Clicking the git reference in the workflow version no longer changes the selected version.  Clicking the git reference hyperlink will take the user to the github repo page with that commit ID.  Clicking anywhere except the git reference in both tools/workflows page will change the selected version.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,9 @@
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
+
+  status:
+    project:
+      default:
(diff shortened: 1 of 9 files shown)


denis-yuen:
Noticed that codecov failed this with just a 0.02% change. 
Please copy over the settings that allow for a threshold of 1% from the main dockstore repo.

PR #318, opened 2018-06-22
Feature/1489/link to GitHub version

agduncan94:
Link to GitHub is now versioned.
Link to TRS descriptor has been added. Comments are welcome for how to best display this in the info tab (ex. label name, tooltip, etc)

https://github.com/ga4gh/dockstore/issues/1489

### DIFF
--- a/src/app/container/info-tab/info-tab.component.html
+++ b/src/app/container/info-tab/info-tab.component.html
@@ -46,12 +46,18 @@
         {{ tool?.imgProviderUrl }}
       </span>
     </li>
-    <li>
-        <strong tooltip="TRS link to the main descriptor for the selected tool version">TRS</strong>:
-        <a [href]="trsLink">
-          {{tool?.tool_path}}
-        </a>
-      </li>
+    <li *ngIf="trsLinkCWL">
+      <strong tooltip="TRS link to the main CWL descriptor for the selected tool version">TRS CWL</strong>:
+      <a [href]="trsLinkCWL">
+        {{tool?.tool_path}}
+      </a>
+    </li>
+    <li *ngIf="trsLinkWDL">
+      <strong tooltip="TRS link to the main WDL descriptor for the selected tool version">TRS WDL</strong>:
+      <a [href]="trsLinkWDL">
+        {{tool?.tool_path}}
+      </a>
+    </li>
     <li *ngIf="(tool?.default_dockerfile_path || !isPublic)">
       <form #editDockerfileForm="ngForm" class="form-inline">
         <div class="form-group">
--- a/src/app/container/info-tab/info-tab.component.html
+++ b/src/app/container/info-tab/info-tab.component.html
@@ -46,6 +46,12 @@
         {{ tool?.imgProviderUrl }}
       </span>
     </li>
+    <li>
+        <strong tooltip="TRS link to the main descriptor for the selected tool version">TRS</strong>:
+        <a [href]="trsLink">
+          {{tool?.tool_path}}
+        </a>
+      </li>
     <li *ngIf="(tool?.default_dockerfile_path || !isPublic)">
       <form #editDockerfileForm="ngForm" class="form-inline">
         <div class="form-group">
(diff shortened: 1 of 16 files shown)


src/app/container/info-tab/info-tab.component.html
```diff
@@ -46,6 +46,12 @@
         {{ tool?.imgProviderUrl }}
       </span>
     </li>
+    <li>
+        <strong tooltip="TRS link to the main descriptor for the selected tool version">TRS</strong>:
+        <a [href]="trsLink">
+          {{tool?.tool_path}}
```
denis-yuen: This seems oddly simple. Is this a placeholder?
Should this be trsLink?

src/app/shared/provider.service.ts
```diff
@@ -17,11 +17,11 @@
 export class ProviderService {
 
   /* set up project provider */
-  setUpProvider(tool) {
+  setUpProvider(tool, version) {
```
denis-yuen: from above, it seems like this method is often called as `setupProvider(tool, null)`
It would have been simpler (and more readable) to provide an overridden `setupProvider(tool) { setupProvider(tool, null)}` if thats possible in typescript. You may have IDE support for this refactoring

agduncan94: Would default parameters work too? So default version to null.

denis-yuen:
Strange coverage results might be worth looking into

agduncan94:
I think the big coverage difference is because the file was already so small, so adding like 6 lines doubled the size.

src/app/container/info-tab/info-tab.component.ts
```diff
@@ -103,4 +109,15 @@ export class InfoTabComponent implements OnInit {
...
+
+
+  /**
+   * Returns a link to the primary descriptor for the given tool version
+   * @param path tool path
+   * @param versionName name of version
+   */
+  getTRSLink(path: string, versionName: string): string {
+    return `${Dockstore.API_URI}${ga4ghPath}/tools/${encodeURIComponent(path)}` +
+      `/versions/${encodeURIComponent(versionName)}/plain-CWL/descriptor`;
```
garyluu: looks like this only handles CWL right now. make PLAIN_CWL and PLAIN_WDL something that can easily be changed once swagger has it.

src/app/workflow/info-tab/info-tab.component.ts
```diff
@@ -101,4 +108,14 @@ export class InfoTabComponent implements OnInit {
...
   }
+
+  /**
+   * Returns a link to the primary descriptor for the given workflow version
+   * @param path full workflow path
+   * @param versionName name of version
+   */
+  getTRSLink(path: string, versionName: string): string {
+    return `${Dockstore.API_URI}${ga4ghPath}/tools/${encodeURIComponent('#workflow/' + path)}` +
+      `/versions/${encodeURIComponent(versionName)}/plain-CWL/descriptor`;
```
garyluu: here too

src/app/container/info-tab/info-tab.component.ts
agduncan94: Oops

agduncan94:
How should we deal with the link for tools, where there can be both CWL and WDL. Should both be displayed?

denis-yuen:
I don't think we have a great solution for this since we have a user definable "default version" but not a "default language". Maybe we should have put it (the TRS link, not the GitHub link)  in the launch tab

agduncan94:
For now I use the languages defined in the tool.descriptorType for determining which links to show.

garyluu:
- Could use some more types in the component inputs
- ctrl -f ": any"  :cry:

src/app/container/info-tab/info-tab.component.ts
```diff
@@ -22,15 +22,30 @@ import { ContainersService } from './../../shared/swagger/api/containers.service
...
+import { Tag } from './../../shared/swagger/model/tag';
 
 @Component({
   selector: 'app-info-tab',
   templateUrl: './info-tab.component.html',
   styleUrls: ['./info-tab.component.css']
 })
 export class InfoTabComponent implements OnInit {
-  @Input() selectedVersion;
+  currentVersion;
```
garyluu: type

agduncan94: oops

PR #322, opened 2018-06-25
Feature/hosted entries fixes

agduncan94:
This deals with being able to set the default version for hosted entries from the UI (no longer calls refresh). There also needs to be changes to the webservice to allow using the update endpoint to set the default version. This will be in another PR.

https://github.com/ga4gh/dockstore/issues/1519


agduncan94:
This PR depends on the updated webservice from https://github.com/ga4gh/dockstore/pull/1534

agduncan94:
I've updated the code to use the update default version endpoints instead of a general update endpoint.

PR #327, opened 2018-06-27
Change version selection

garyluu:
For ga4gh/dockstore#1553

Changing the previous selection algorithm to be easier to read.

1. If the URL specifies a specific Tag or WorkflowVersion, choose to display that one.  Otherwise,
2. If the default version is specified, choose to display the default version.  Otherwise,
3. Choose the first version in the array (the last updated version?)

Also change the test because the checker is a stub since it can't be refreshed (no such Git repo).  Therefore, it cannot have a master branch to go to (not sure why it went there in the first place).

### DIFF
--- a/src/app/shared/entry.ts
+++ b/src/app/shared/entry.ts
@@ -175,34 +175,37 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
     });
   }
 
-  public selectVersion(versions, urlVersion, defaultVersion, selectedVersion): any {
-    let useFirstTag = true;
-    let urlTagExists = false;
-    // Determine which tag to select
-    for (const item of versions) {
-      // If a tag is specified in the URL then use it
-      if (urlVersion !== null) {
-        if (item.name === urlVersion) {
-          selectedVersion = item;
-          useFirstTag = false;
-          urlTagExists = true;
-          break;
-        }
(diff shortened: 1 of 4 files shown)


src/app/shared/entry.ts
```diff
@@ -175,34 +175,37 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
...
+   * 2. If the default version is specified, choose to display the default version.  Otherwise,
+   * 3. Choose the last version in the array (the last updated version?)
+   * @param {((Array<WorkflowVersion | Tag>))} versions
+   * @param {string} urlVersion
+   * @param {string} defaultVersion
+   * @param {((WorkflowVersion | Tag))} selectedVersion
+   * @returns {((WorkflowVersion | Tag))}
+   * @memberof Entry
+   */
+  public selectVersion(versions: (Array<WorkflowVersion | Tag>), urlVersion: string, defaultVersion: string,
```
garyluu: This is supposed to be (Array<WorkflowVersion> | Array<Tag>), no idea why errors appear when attempted to do so

denis-yuen: Probably makes more sense as a comment in code for the next developer that looks at this

denis-yuen:
FYI, pending a review request and/or build completion

PR #329, opened 2018-06-28
Show all versions

garyluu:
Showing all versions instead of just the ones on the sidebar.

Tried altering the sortVersions function to show more than 5, but the table sorting will overwrite it regardless.

### DIFF
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -143,7 +143,7 @@
       </tab>
       <tab id="versionTab" heading="Versions" (select)="setEntryTab('versions')">
         <div *ngIf="validVersions">
-          <app-versions-container [versions]="sortedVersions" (selectedVersionChange)="onSelectedVersionChange($event)" id="versionsTab" [verifiedSource]="tool?.verifiedSources" [selectedVersion]="selectedVersion">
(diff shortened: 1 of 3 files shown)


denis-yuen:
How is the order of versions in the drop-down vs the order in the versions table with this change?

agduncan94:
I think they should be the same order since they are both directly looking at workflow.workflowVersions or tool.tags

garyluu:
So this change basically reverts it back to how it is on production.  It looks the the versions table will by default sort the versions by name (overwriting any sorting that was done previously).  The versions in the dropdown however somehow manages to sort by date (which doesn't look like the order returned by webservice)

garyluu:
Just to double check, you want the dropdown to match the order of the versions table?  i.e. sorting the dropdown by name by default.  then if the versions table is changed to sort by date modified, the dropdown will also be sorted by date modified?

denis-yuen:
I don't think they need to maintain sort order if the user sorts.
However, I do expect the two to be in sync when I first land on the page.
Out of the two sorts (sort by name, sort by date modified), I think sort by date modified is probably better assuming latest first.

garyluu:
ok, should be much easier than syncing the sort then

PR #330, opened 2018-07-04
Disable instead of hide Launch with Buttons

coverbeck:
Instead of hiding Launch with FireCloud/DNAstack buttons when the WDL
is not supported, disable them so that the user knows what is going on.

For FireCloud, disable if there are file-based imports.

For DNAstack, disable if there are any imports.

Had to change to buttons from anchors, because you cannot disable
anchors (well, not with disabled attribute).

Test for http(s) imports is sort of crude; we should probably do it
server side using... wom?.

Also, there is some code duplication in checking if there are imports
for FC and DNAstack. But since this code is going into hotfix and
develop branch is on Angular 6, decided to do it that way rather than
using RxJs6 operators (which I don't know very well, plus wouldn't work
on hotfix).

Here is screenshot of buttons when disabled, with tooltip.

screen shot 2018-07-03 at 5 31 23 pm 2

### DIFF
--- a/src/app/test/mocked-objects.ts
+++ b/src/app/test/mocked-objects.ts
@@ -174,6 +174,13 @@ export const wdlSourceFile: SourceFile = {
   type: undefined
 };
 
+export const wdlSourceFileWithHttpImport: SourceFile = {
+  content: 'import "https://foo.com/some.wdl',
+  id: 1,
+  path: '',
+  type: undefined
+};
+
 export const sampleSourceFile: SourceFile = {
   content: 'potato',
   id: 1,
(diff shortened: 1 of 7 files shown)


garyluu:
The fix might possibly be outside the scope of this PR, but while testing this locally, the number of times http://localhost:8080/workflows/5906/secondaryWdl?tag=master is called has increased to 5 times from the 1 to 2 times on staging when visiting my skylab repo

coverbeck:
* Closing to reopen against a different branch.
* Good catch on extra call to get secondary files, although there should only be one more call I think, not 3. Will address and investigate when reopening new PR.
* It's slightly tricky to disable anchors in Angular: https://stackoverflow.com/questions/36985112/angular2-what-is-the-correct-way-to-disable-an-anchor-element. Combined with them looking like looking like buttons anyway, I thought it would make sense to just use buttons. But I did not notice the underline behavior on the buttons. Will investigate.

PR #332, opened 2018-07-05
Disable Launch w/FireCloud if WDL has file imports

coverbeck:
ga4gh/dockstore#330

Disable instead of hide the button so user knows what is going on, with
a tooltip explaining why.

Only did it for FireCloud, as for DNAstack the user selects the version, so it doesn't seem like we should disable the button based on the content of the current version.

Screenshot showing warning for DNAstack and disabled button with tooltip for FireCloud:

screen shot 2018-07-05 at 3 57 26 pm 2

FYI, this will be a pain to merge into develop, as the develop branch has this code in a new component.

### DIFF
--- a/src/app/workflow/workflow.component.ts
+++ b/src/app/workflow/workflow.component.ts
@@ -47,6 +47,7 @@ export class WorkflowComponent extends Entry {
   workflowEditData: any;
   dnastackURL: string;
   fireCloudURL: string;
+  fireCloudDisabledMessage: string;
   public workflow;
   public missingWarning: boolean;
   public title: string;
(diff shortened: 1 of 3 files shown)


denis-yuen:
I think it would still be worthwhile giving the user a heads-up or warning.

coverbeck:
Updating for warning with DNAstack; please hold off reviewing.

coverbeck:
This is ready for review again. Please take a look.

src/app/workflow/workflow.component.ts
```diff
@@ -110,18 +115,45 @@ export class WorkflowComponent extends Entry {
     }
   }
 
-  private isWdl(workflowRef: ExtendedWorkflow) {
-    return workflowRef.full_workflow_path && workflowRef.descriptorType === 'wdl';
+  public isWdl(workflowRef: ExtendedWorkflow) {
```
garyluu: check how many times this is called with the same workflowRef

garyluu: approved if say...3 or less

coverbeck: It was 6 times, but after a discussion went ahead with it and will address anew on develop branch.

PR #334, opened 2018-07-05
Prepare for scoped tokens on gitlab

denis-yuen:
* currently unable to determine membership or owned projects with a
token narrower than `api` which seems like a gitlab issue
https://github.com/timols/java-gitlab-api/blob/master/src/main/java/org/gitlab/api/GitlabAPI.java#L863
* tested with gitlab api 11.1.0-rc4-ee rev aedb3c9 (find at
https://gitlab.com/api/v4/version )

* note to self: follow with compose_setup PR when closing


denis-yuen:
@garyluu made https://github.com/ga4gh/dockstore/issues/1573

PR #335, opened 2018-07-05
Feature/rxjs6 lint

garyluu:
Add linter for rxjs 6 to avoid common pitfalls


garyluu:
Probably waiting until after the big PRs are merged first

garyluu:
Added even more lint rules

PR #339, opened 2018-07-09
Feature/remove angular datatables

garyluu:
Base to be changed to develop once feature/pagination is in.  This illustrates the additional changes needed to get rid of angular-datatables


src/app/container/register-tool/register-tool.service.ts
```diff
@@ -26,6 +26,8 @@ import { MetadataService } from './../../shared/swagger/api/metadata.service';
 import { DockstoreTool } from './../../shared/swagger/model/dockstoreTool';
 import { Tool } from './tool';
 import { HostedService } from './../../shared/swagger/api/hosted.service';
+// This line is super important for jQuery to work across the website for some reason
+import * as $ from 'jquery';
```
denis-yuen: Would `import * as jQuery from 'jquery';` work to avoid some of the changes below?

src/app/search/search-tool-table/search-tool-table.component.html
```diff
@@ -0,0 +1,75 @@
+<!--
+  ~    Copyright 2017 OICR
```
denis-yuen: Looks like you have a gremlin for your dates ;)

src/app/container/register-tool/register-tool.service.ts
garyluu: we had $ in some places and jQuery in some other places.  I decided it would be best to keep things consistent.  I pretty much randomly chose $ over jQuery.

PR #341, opened 2018-07-09
Fix Launch With Disabling for Checker Workflows

coverbeck:
ga4gh/dockstore#1548

Fix case when disabling of launch with FireCloud/DNAstack buttons did
not happen when it should have, when you select the checker workflow
by clicking View Checker when on the parent.

Events seem to fire in a different order when selecting a checker
workflow that way. Add a call to checkWdlForImportsAndSetFirecloudUrl in
setupPublicEntry().

To avoid too many calls to fetch secondary files, store the last time
workflow and version that was queried, and make sure we don't query
twice for the same workflow/version. Less than perfect, but will
will redo in ga4gh/dockstore#1575.

### DIFF
--- a/src/app/workflow/workflow.component.ts
+++ b/src/app/workflow/workflow.component.ts
@@ -52,6 +52,8 @@ export class WorkflowComponent extends Entry {
   public wdlHasHttpImports: boolean;
   public wdlHasFileImports: boolean;
   public enableLaunchWithFireCloud = Dockstore.FEATURES.enableLaunchWithFireCloud;
+  private lastWorkflow: ExtendedWorkflow;
+  private lastVersion: WorkflowVersion;
   public workflow;
   public missingWarning: boolean;
   public title: string;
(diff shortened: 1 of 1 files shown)


src/app/workflow/workflow.component.ts
```diff
@@ -144,18 +146,22 @@ export class WorkflowComponent extends Entry {
...
-        this.workflowsService.secondaryWdl(workflowRef.id, version.name).subscribe((sourceFiles: Array<SourceFile>) => {
-          if (!sourceFiles || sourceFiles.length === 0) {
-            this.workflowsService.wdl(workflowRef.id, version.name).subscribe((sourceFile) => {
-              this.wdlHasHttpImports = importHttpRegEx.test(sourceFile.content);
-            });
-            if (this.enableLaunchWithFireCloud) {
-              this.fireCloudURL = `${Dockstore.FIRECLOUD_IMPORT_URL}/${workflowRef.full_workflow_path}:${version.name}`;
+        if (version !== this.lastVersion || this.lastWorkflow !== workflowRef) {
+          this.lastVersion = version;
+          this.lastWorkflow = this.lastWorkflow;
```
garyluu: ?

denis-yuen: `this.lastWorkflow = workflowRef` maybe?

coverbeck: Oops, how embarrassing -- will push new commit soon.

PR #343, opened 2018-07-11
Feature/1576/register workflow dialog

garyluu:
1. Fixes the sidebar entry selection to match what's actually selected
2. Small changes in ga4gh/dockstore#1576
3. Changed how entry is selected:

- Publish, refresh, update, etc will manually upsert the workflow into workflows$ and then setWorkflow() to replace the current workflow (the URL is presumed to already be correct)
- Register will upsert the workflow and then go to the url (which then triggers it to change workflow$)
- In general:
   - workflows$ is completely seperate from url and workflow$
   - URL change => workflow$ change

Also removed replaceWorkflow() because upsert is more useful


src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
```diff
@@ -155,13 +155,13 @@ <h4 class="modal-title">Register Workflow</h4>
...
                       <button type="button" class="btn btn-default btn-sm dropdown-toggle" data-toggle="dropdown">
-                        {{hostedWorkflow?.descriptorType}}
+                        {{hostedWorkflow?.descriptorType.toUpperCase()}}
                         <span class="caret"></span>
                       </button>
                       <ul class="dropdown-menu">
                         <li *ngFor="let descriptorType of (descriptorLanguages$ | async)">
                           <a class="dropdown-item" (click)="hostedWorkflow.descriptorType = descriptorType.toLowerCase()">
-                            {{descriptorType | lowercase}}
+                            {{descriptorType | uppercase}}
```
coverbeck: I haven't actually run the code, but won't this will always uppercase `NEXTFLOW`? Presumably we'd want it to be `Nextflow`?

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
```diff
@@ -100,13 +100,13 @@ <h4 class="modal-title">Register Workflow</h4>
...
                       <button type="button" class="btn btn-default btn-sm dropdown-toggle" data-toggle="dropdown">
-                        {{workflow?.descriptorType}}
+                        {{workflow?.descriptorType.toUpperCase()}}
                         <span class="caret"></span>
                       </button>
                       <ul class="dropdown-menu">
                         <li *ngFor="let descriptorType of (descriptorLanguages$ | async)">
                           <a class="dropdown-item" (click)="changeDescriptorType(descriptorType.toLowerCase())">
-                            {{descriptorType | lowercase}}
+                            {{descriptorType | uppercase}}
```
coverbeck: Same comment as below, regarding NEXTFLOW being in caps.

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
```diff
@@ -100,13 +100,13 @@ <h4 class="modal-title">Register Workflow</h4>
                   <div class="input-group">
                     <div class="input-group-btn">
                       <button type="button" class="btn btn-default btn-sm dropdown-toggle" data-toggle="dropdown">
-                        {{workflow?.descriptorType}}
+                        {{workflow?.descriptorType.toUpperCase()}}
```
coverbeck: Same as other comments

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
garyluu: Yes.  But as it stands, there's no quick and clean way of fixing that.  I actually have a google docs sheet that will be presented to iron out these issues.

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
garyluu: Same reason

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
garyluu: same reason

garyluu:
Going to make a few changes in preparation for other future changes

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
denis-yuen: I think I'd probably prefer `Nextflow` for display purposes, but will discuss in meeting

garyluu:
Neither lowercase() or uppercase() is a good solution, using a custom pipe to directly map possible values to a user friendly one (taken from the SearchModule).  Can possibly remove it in the future once we descriptor type is unified.  Note:  Only displayed values should've been mapped, the internal ones in the .ts files should be unaffected.

garyluu:
For re-review

PR #346, opened 2018-07-13
WIP: Properly Check Permissions for Shared Workflow

coverbeck:
A fix for ga4gh/dockstore#1589

Not sure if we should take this; see my following comments.

We have to go with this approach with the current webservice API,
which makes me the question the current permissions API.

The logic is, if you can fetch permissions, you are an owner. If you
are not, then make OPTIONS calls to see what APIs you can invoke,
and the `Allow` header in the response will indicate what HTTP method
you have permissions to invoke.

My thinking was:

/path/workflow/{repository}/permissions should be an unauthorized call
if you are not an owner of {repository} -- e.g., if you only have read
permissions on the workflow, you should be able to see the workflow
but not who it's been shared with (in fact, that is how the policies
are currently defined in SAM, although we could change them).

Given that, I thought that the API could support OPTIONS calls to
indicate which endpoints are invocable with which methods. I liked
that, because if I have an endpoint that simply says for example that
you have READ permissions on a workflow, then the UI has to figure out
which API calls are allowed with that permission and which are not.
Whereas with OPTIONS you can query the API call you are going to make
and let the backend tell you whether you have permissions or not.

However:

1. I've only added a couple of OPTIONS endpoints, for the endpoints that
currently can be affected by permissions, so it's incomplete.
2. This UI code is convoluted.

So maybe we should just enable the permissions endpoint for all users,
but if you are not an owner, it only returns the current user's
permissions. That means the UI would have to figure out what API calls
it can and can't make, but it looks like it was essentially doing
that anyway.

### DIFF
--- a/src/app/workflow/workflow.component.html
+++ b/src/app/workflow/workflow.component.html
@@ -164,7 +164,7 @@
           <span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span> To see the DAG, please refresh the workflow.
         </div>
       </tab>
-      <tab *ngIf="!isPublic() && isHosted() && isUserOwner()" id="permissionsTab" heading="Permissions" (select)="setEntryTab('permissions')">
+      <tab *ngIf="!isPublic() && isHosted() && isOwner" id="permissionsTab" heading="Permissions" (select)="setEntryTab('permissions')">
         <app-permissions [workflow]="workflow"></app-permissions>
       </tab>
     </tabset>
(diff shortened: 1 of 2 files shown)


denis-yuen:
I think I agree, it seems simpler for the permissions endpoint to return the current user's permissions. If it was convoluted for the UI to work some of this out, I suspect that it will be similarly complex for the CLI as well, which we've ignored till now (reasonably) but has to be done at some point. 

We do have a hole in that we don't have a way to figure out who does in fact own a workflow (in case we wanted to ask them for more permissions), but the OPTIONS approach is tangential to that and doesn't provide that information either. 

That said, I defer to @agduncan94  and your judgment on this one.

coverbeck:
Closing this PR and will instead reimplement changes to the API discussed here.

PR #348, opened 2018-07-13
Feature/1489/fixes

garyluu:
entry.providerUrl was modified so anything that previously depended on it as parameter is broken.
Reverting that change and using a pipe to generate the version-specific providerUrl on-the-fly

Fix for when tool not loaded before selected version update for some reason (occurs on initial page load)

Fix for ga4gh/dockstore#1489 and ga4gh/dockstore#1435

Not sure how it's intended to work but now:
- Public page will always display the version-specific providerUrl
- My-entry page will always display the workflow providerUrl

### DIFF
--- a/src/app/container/container.component.ts
+++ b/src/app/container/container.component.ts
@@ -150,7 +150,7 @@ export class ContainerComponent extends Entry {
     if (tool) {
       this.tool = tool;
       if (!tool.providerUrl) {
-        this.providerService.setUpProvider(tool, this.selectedVersion);
+        this.providerService.setUpProvider(tool);
       }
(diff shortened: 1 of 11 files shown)


denis-yuen:
Could you elaborate more on "Not sure how it's intended to work but now"?

garyluu:
I'm not entirely sure how this feature was intended to work (maybe display the version-specific providerUrl in public page only, maybe show it only when the version selected is changed, etc).  

But this current PR changes the behaviour to:
- Public page will always display the version-specific providerUrl
- My-entry page will always display the workflow providerUrl

PR #352, opened 2018-07-17
Add Launch with DNAnexus

coverbeck:
ga4gh/dockstore#1537

Added launch with DNAnexus button. The order of the launch with
buttons in the UI is alphabetical.

Got rid of enableLaunchWithFireCloud feature flag. It was there
because at one point we were only able to test this feature on their
side in staging but not in prod, but now that it's been in prod
for a while, I think we can get rid of the flag and the tests for it.

Also introduced a service to generate the urls. Use HttpParams there
instead of the URLSearchParams being used previously in the
component because URLSearchParams is deprecated.

Note: I'm planning to tackle ga4gh/dockstore#1575 next in the UI, which
relates to disabling launch with buttons based on the file content.
Wanted to to do this on its own first.

Screenshot:

screen shot 2018-07-17 at 2 43 00 pm

### DIFF
--- a/src/app/workflow/launch-third-party/launch-third-party.service.ts
+++ b/src/app/workflow/launch-third-party/launch-third-party.service.ts
@@ -1,0 +1,34 @@
+import { Injectable } from '@angular/core';
+import { ga4ghPath } from '../../shared/constants';
+import { Dockstore } from '../../shared/dockstore.model';
+import { HttpParams } from '@angular/common/http';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class LaunchThirdPartyService {
+
+  constructor() { }
+
+  public dnanexusUrl(path: string, version: string): string {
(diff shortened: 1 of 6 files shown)


src/app/workflow/launch-third-party/launch-third-party.service.ts
```diff
@@ -0,0 +1,34 @@
...
+      .set('path', path)
+      .set('descriptorType', descriptorType);
+    return Dockstore.DNASTACK_IMPORT_URL + '?' + httpParams.toString();
+  }
+
+  public firecloudUrl(path: string, version: string): string {
+    return `${Dockstore.FIRECLOUD_IMPORT_URL}/${path}:${version}`;
+  }
+
+  private getTrsUrl(path: string, versionName: string): string {
```
garyluu: at some point we should move all this (and all TRS related url constructors in other files) into a separate service somewhere.

PR #353, opened 2018-07-19
Update Launch With Buttons

coverbeck:
gahgh/dockstore#1575

* Changed to Material
* Use/implement OnChanges in LaunchThirdPartyComponent; simplified
things, including getting rid of _workflow and _selectedVersion
properties.
* Changed links to buttons, as disable doesn't work on links (sort of
had it working with links, but then tests would complain about <a> not
having the disabled attribute).
* Disabling logic:
    * Disable all if there is no content for the current version. For
    DNAstack, could arguably not disable in this case, since you can
    select any version in their UI, but then it would have been a
    confusing message.
    * Disable FireCloud if there are file-path imports.
    * Display warning message if there are any imports for DNAstack, but
    don't disable button.
    * DNAnexus supports all imports, so never disable based on imports.
* Had somehow embarassingly put mock object logic into service-stubs;
moved that into mocked-objects.

Screenshots with no warnings/disablement:

screen shot 2018-07-18 at 5 39 23 pm

Screenshot with warnings/disablement:

screen shot 2018-07-18 at 5 43 32 pm 2

### DIFF
--- a/src/app/workflow/launch-third-party/launch-third-party.component.html
+++ b/src/app/workflow/launch-third-party/launch-third-party.component.html
@@ -1,4 +1,4 @@
-<div *ngIf="isWdl()" class="panel panel-default">
+<div *ngIf="isWdl" class="panel panel-default">
   <div class="panel-heading">
     <h3>Launch with</h3>
   </div>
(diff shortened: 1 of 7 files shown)


src/app/workflow/launch-third-party/launch-third-party.component.html
```diff
@@ -1,22 +1,34 @@
-<div *ngIf="dnastackURL || fireCloudURL" class="panel panel-default">
+<div *ngIf="isWdl()" class="panel panel-default">
```
agduncan94: Check how many times this is called.

coverbeck: Made this a property instead of a function call in subsequent commit.

src/app/workflow/launch-third-party/launch-third-party.component.html
```diff
@@ -1,22 +1,37 @@
...
+               [disabled]="!fireCloudURL"
+               color="primary"><mat-icon svgIcon="firecloud"></mat-icon> FireCloud &raquo;</a>
           </div>
-          <div *ngIf="dnanexusURL" class="button">
-            <p><a href="{{dnanexusURL}}"><img src="../../assets/images/thirdparty/DX_Logo_white_alpha.svg"> DNAnexus
-              &raquo;</a></p>
+          <div fxFlex="100%">
+            <a mat-raised-button
+               title="{{!wdlHasContent ? 'The WDL has no content.' : ''}}"
+               target="_blank"
```
garyluu: https://developers.google.com/web/tools/lighthouse/audits/noopener

coverbeck: Thanks, I did not know about the `rel` attribute. Pushed a new commit with it.

Regarding the workflowsService subscribe errors, right now I'm initializing everything before making the http call, so no buttons will be enabled if there is an error.

More generally, I think we need a pattern/component for displaying errors for Ajax calls. Or maybe there is, and I'm unaware of it. I added it as a topic to discuss at today's meeting.

PR #355, opened 2018-07-20
added tooltips for mode

agduncan94:
Adds tooltip for mode of workflow. We should discuss the exact wording.

https://github.com/ga4gh/dockstore/issues/1519

### DIFF
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -87,7 +87,8 @@
     <strong tooltip="Currently disabled for Hosted Workflows">Checker Workflow</strong>: n/a
   </div>
   <li *ngIf="!isPublic">
-    <strong uib-tooltip-html="modeTooltip">Mode</strong>: {{workflow?.mode}}
+    <ng-template #modeTooltip><div [innerHtml]="modeTooltipContent"></div></ng-template>
+    <strong [tooltip]="modeTooltip">Mode</strong>: {{workflow?.mode}}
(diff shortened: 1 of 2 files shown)


src/app/workflow/info-tab/info-tab.component.ts
```diff
@@ -47,6 +47,11 @@ export class InfoTabComponent extends EntryTab implements OnInit, OnChanges {
   isPublic: boolean;
   trsLink: string;
   public refreshMessage: string;
+  modeTooltipContent = `<b>STUB:</b> Basic metadata pulled from remote site.<br />
```
denis-yuen: Probably can be more specific here. 
`Stub: Basic metadata pulled from source control` 
vs
`Full: Full content synced from source control`

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -87,21 +87,22 @@
     <strong tooltip="Currently disabled for Hosted Workflows">Checker Workflow</strong>: n/a
   </div>
   <li *ngIf="!isPublic">
-    <strong uib-tooltip-html="modeTooltip">Mode</strong>: {{workflow?.mode}}
+    <ng-template #modeTooltip><div [innerHtml]="modeTooltipContent"></div></ng-template>
+    <strong [tooltip]="modeTooltip">Mode</strong>: {{workflow?.mode}}
```
garyluu: Reason why this isn't just `[tooltip]="modeTooltipContent`?

garyluu: oh, it's the bolding

PR #356, opened 2018-07-21
Fix url generated for DNAnexus

coverbeck:
ga4gh/dockstore#1537

Was passing the wrong path to generate DNAnexus link.

Tested the link by replacing pasting it into the browser, replacing
local protocol/host/port with `https://dockstore.org:8443`, since
that is the only whitelisted by DNAnexus. Thought I had tried that
before, but apparently I hadn't.


denis-yuen:
FYI, build failure seems legitimate after a retry

PR #357, opened 2018-07-23
My workflow accordion is now material

agduncan94:
The my workflows page org accordion is now material.

### DIFF
--- a/src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.html
+++ b/src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.html
@@ -1,7 +1,7 @@
 <mat-accordion>
   <mat-expansion-panel *ngFor="let orgObj of groupEntriesObject" [expanded]="orgObj?.isFirstOpen">
     <mat-expansion-panel-header>
-      <mat-panel-title>
+      <mat-panel-title class="org-accordion-header">
           {{orgObj?.sourceControl + '/' + orgObj?.organization}}
       </mat-panel-title>
     </mat-expansion-panel-header>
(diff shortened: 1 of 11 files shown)


agduncan94:
This is what it looks like now.
screenshot from 2018-07-23 11-16-32

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.spec.ts
```diff
@@ -19,7 +28,18 @@ describe('SidebarAccordionComponent', () => {
   beforeEach(async(() => {
     TestBed.configureTestingModule({
       declarations: [ SidebarAccordionComponent, RefreshWorkflowOrganizationComponent ],
-      imports: [ AccordionModule.forRoot(), ModalModule.forRoot(), TabsModule.forRoot(), RouterTestingModule],
+      imports: [
+        AccordionModule.forRoot(),
+        ModalModule.forRoot(),
+        TabsModule.forRoot(),
```
garyluu: check if these ngx-bootstrap modules are still needed in the test

src/ds-style-fix.scss
```diff
@@ -817,3 +817,25 @@ div.contact-button {
...
+  font-size: 12px !important;
+}
+
+mat-panel-title.org-accordion-header {
+  color: #487980;
+  font-size: 16px;
+}
+
+.register-workflow-modal-button {
+  margin-top: 10px !important;
```
garyluu: make it into a universal class name (mt-2)  and keep it in this file.  Otherwise move it to the component's scss file.

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.html
```diff
@@ -1,38 +1,44 @@
...
+            <p class="text-center">No unpublished workflows.</p>
+          </mat-list-item>
+        </mat-list>
+      </mat-tab>
+    </mat-tab-group>
+    <mat-action-row>
+      <span *ngIf="!orgObj?.sourceControl.startsWith('dockstore.org')">
+        <app-refresh-workflow-organization [organization]="orgObj?.organization"></app-refresh-workflow-organization>
+      </span>
+      <button mat-mini-fab color="accent" matTooltip="Add new workflow" [matTooltipShowDelay]="1000" [matTooltipPosition]="'after'" type="button" (click)="setRegisterEntryModalInfo(orgObj?.organization + '/new_workflow'); showRegisterEntryModal()">
```
coverbeck: These buttons are too close to each, IMHO. I think the edges are actually touching each other.

src/ds-style-fix.scss
agduncan94: Is there documentation on these naming conventions?

garyluu: not an official thing, but if the very common styling like padding, margin, I name it the same as the bootstrap ones: https://getbootstrap.com/docs/4.1/utilities/spacing/

PR #362, opened 2018-07-25
Experiment with json-ld, needs cleanup

denis-yuen:
* for https://github.com/ga4gh/dockstore/issues/341
* and eventually https://github.com/ga4gh/dockstore/issues/750

### DIFF
--- a/src/app/workflows/workflows.module.ts
+++ b/src/app/workflows/workflows.module.ts
@@ -29,7 +29,6 @@ import { getTooltipConfig } from './../shared/tooltip';
 import { SearchWorkflowsComponent } from './search/search.component';
 import { WorkflowsComponent } from './workflows.component';
 import { workflowsRouting } from './workflows.routing';
-import { NgxJsonLdModule } from '@ngx-lite/json-ld';
 
 @NgModule({
   declarations: [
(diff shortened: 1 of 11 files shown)


src/app/workflows/workflows.module.ts
```diff
@@ -46,7 +47,8 @@ import { workflowsRouting } from './workflows.routing';
     TabsModule.forRoot(),
     TooltipModule.forRoot(),
     WorkflowModule,
-    workflowsRouting
+    workflowsRouting,
+    NgxJsonLdModule
```
garyluu: Remove the NgxJsonLdModule from these 4 modules and throw it into EntryModule (and export inside it) and see what happens.

src/index.html
```diff
@@ -8,6 +8,32 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link rel="icon" type="image/x-icon" href="favicon.ico">
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+  <script type="application/ld+json">
+  {
+  "@context": "http://schema.org",
+  "@type": "Organization",
+  "url": "https://dockstore.org",
```
coverbeck: Just going to throw this out there, since I wasn't even aware until this PR that search engines looked for this -- should the host be dynamic, e.g., staging.dockstore.org, or in theory, if not likely, somebody tried to stand up their own instance of Dockstore?

Since you're serving up the index.html as is -- it's not a template where you can substitute values at runtime, even if my point is valid (?), it's probably more trouble than it's worth.

denis-yuen: I think, it depends. 
For staging, I don't actually really want it to rank highly. I might even put in a robots.txt to exclude it if it starts ranking highly. Similarly, for this "overall" information, I'd actually want the search engine to essentially redirect/strongly nudge someone to the main site. 

For someone standing up their own site, it's true. But I think we/they can put more resources into that when the time comes.

PR #365, opened 2018-07-26
Eliminate console error when adding user

coverbeck:
ga4gh/dockstore#1627

Toggling disabled on the input was causing the containing mat-form-field
to generate the exception.

I believe this is an Angular Material bug and the only fix I could
figure out changes the UX a bit.

Before you had to wait for the adding/removing a user operation to
complete before you could start typing in a new value. Now you don't
have to wait for the adding/removing a user to complete
before you start typing in a new one.

Although being "forced" into it, I think this behavior is OK; if you
want to add a bunch of users, there's no reason to force you to wait
for each one to complete. On top of that, in my tests, the operation
is fairly fast anyway, so depending on network connection, you
typically won't even be able to start typing in a new value before
the previous operation has already completed.

Also simplified controller by just having one add() method.

### DIFF
--- a/src/app/workflow/permissions/permissions.component.html
+++ b/src/app/workflow/permissions/permissions.component.html
@@ -18,7 +18,7 @@
     <li>
       <mat-form-field style="width: 100%;">
         <mat-chip-list #ownerlist>
-          <mat-chip *ngFor="let owner of owners" (removed)="remove(owner, 'OWNER')" [removable]="owners.length > 1">
+          <mat-chip *ngFor="let owner of owners" (removed)="remove(owner, Role.OWNER)" [removable]="owners.length > 1">
             {{owner}}
             <mat-icon matChipRemove *ngIf="owners.length > 1">cancel</mat-icon>
           </mat-chip>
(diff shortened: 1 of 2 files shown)


src/app/workflow/permissions/permissions.component.html
```diff
@@ -23,43 +23,40 @@
             <mat-icon matChipRemove *ngIf="owners.length > 1">cancel</mat-icon>
           </mat-chip>
           <input placeholder="Owners"
-                 [disabled]="updating"
                  [matChipInputFor]="ownerlist"
                  [matChipInputSeparatorKeyCodes]="separatorKeysCodes"
                  [matChipInputAddOnBlur]="addOnBlur"
-                 (matChipInputTokenEnd)="addOwner($event)"/>
+                 (matChipInputTokenEnd)="add($event, 'OWNER')"/>
```
garyluu: You can use enums in html instead of a string

coverbeck: Ah cool, did not know that. Pushed new commit.

PR #367, opened 2018-07-27
WIP: Better Handling for Permissions

coverbeck:
UI portion of ga4gh/dockstore#1589

This has a dependency on PR ga4gh/dockstore#1638, which has a new
endpoint.

Call new endpoint to get allowable actions. If SHARE action is one
of the actions, then user is an owner, and go ahead and fetch all
permissions for the workflow.

Unfortunately, generated TS code doesn't generate a type for the enum
returned by Actions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ jdk:
   - oraclejdk8
 env:
   global:
-    - WEBSERVICE_VERSION="1.5.0-alpha.7"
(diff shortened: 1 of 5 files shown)


denis-yuen:
@coverbeck  Hmmm, seems like a legitimate failure

coverbeck:
Checking...

PR #373, opened 2018-08-01
Feature/descriptor type form change

garyluu:
Makes setting the descriptor type during registration more obvious by changing from a dropdown to a vertical radio button group (many form guidelines say to avoid dropdowns when possible and also to make it more vertical).

Also added debounce time for form changes otherwise it might be overkill to check form too often.

Ends up adding more horizontal space for the workflow path which is preferable because it's sometimes the longest.

Also increase font size from 12px to 14px even though recommendations are to go 16px+

### DIFF
--- a/src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
+++ b/src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
@@ -69,7 +69,7 @@
           <form *ngIf="selectedOption?.value == 0" #registerWorkflowForm="ngForm" name="registerWorkflowForm" class="form-horizontal" (submit)="registerWorkflow()" novalidate>
             <div class="modal-body">
               <div class="form-group">
-                <label class="col-sm-3 control-label mt-0">
+                <label class="col-sm-3 control-label pt-0">
                   Source Code Repository:
                 </label>
                 <div class="col-sm-9">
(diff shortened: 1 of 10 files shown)


garyluu:
Before:

image


After:

image

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
```diff
@@ -92,42 +92,39 @@ <h4 class="modal-title">Register Workflow</h4>
                   <div *ngIf="formErrors.gitPath" class="alert alert-danger"> {{ formErrors.gitPath }} </div>
```
coverbeck: This is beyond the scope of this PR, but just mentioning it since I was trying it out -- I find it slightly disconcerting that the page comes up with errors, before the user has even had a chance to input anything. The more common UX flow that I've seen is that the required errors are indicated after the user tries to submit. And then you could have some sort of indicator to show which fields are required.

denis-yuen: I agree with this

garyluu: I think ideally the required fields are marked with a star beforehand.  The validation errors start to appear once the specific field is touched.

PR #374, opened 2018-08-01
For add checker workflow, only allow selecting of descriptor type if …

agduncan94:
…tool

See title.
https://github.com/ga4gh/dockstore/issues/1626


agduncan94:
https://github.com/ga4gh/dockstore/issues/1626

PR #375, opened 2018-08-01
Deal with Firefox copy-and-paste annoyance

denis-yuen:
Working on https://github.com/ga4gh/dockstore/issues/1622


garyluu:
Token starts disappearing at roughly col-md size

denis-yuen:
Hmmm, rats. Need a way of getting it to display properly at small screen sizes but not automatically break on Firefox vs Chrome.

denis-yuen:
(feel free to clean-up either here or separately)

garyluu:
class="small" might've worked too

PR #377, opened 2018-08-02
change launch component to be material cards and fix nextflow launch

agduncan94:
https://github.com/ga4gh/dockstore/issues/1648

No more nested wells :tada:


agduncan94:
new-launch

PR #378, opened 2018-08-02
Feature/material buttons files

agduncan94:
Changes the buttons for file copy and download to material. Ideally in the future we should have these floating above the file content.

download-button

### DIFF
--- a/src/app/container/descriptors/descriptors.component.html
+++ b/src/app/container/descriptors/descriptors.component.html
@@ -35,20 +35,21 @@
 
   <div *ngIf="content && _selectedVersion?.valid" class="row m-0">
     <div class="btn-group">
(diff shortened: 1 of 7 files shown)


garyluu:
tiny bit more spacing between the two buttons, possibly move it to the right and push everything up if possible

denis-yuen:
They also look a bit, big? if not floating

agduncan94:
They do look kind of big in that screenshot, but they are the same size as the other fab buttons. Maybe if we push them to the right it will look better since there is more empty space.

agduncan94:
How is this?
screenshot from 2018-08-02 13-01-40

garyluu:
better imo.  the buttons have to be about that size or else phone users will find it difficult to press the buttons...even though i kind of liked the previous button size + shape

garyluu:
it's probably trickier if trying to put it directly on top of the file contents because the buttons might cover up text.  whereas it can be on top of the DAG because the DAG can be panned into view if covered.

PR #379, opened 2018-08-03
Fix twitter from creating multiple scripts

garyluu:
prevent Twitter from creating a new script every time home component is visited.  Ideally, it should destroy the script when leaving but...oh well.

Partial solution for ga4gh/dockstore#1630


garyluu:
there might be a way to remove the script but it has already started executing regardless.

denis-yuen:
Yes, the implication is that a full solution would be to somehow fix it or load it in a different way so that it stops executing, is out of memory, and destroyed after browsing away.

PR #383, opened 2018-08-07
Prototype for register vs login

denis-yuen:
* needs cleanup and better appearance
* used this to test new login vs register from https://github.com/ga4gh/dockstore/issues/1691


garyluu:
probably ready for review.  UI hasn't changed too much, same page, an extra tab:

image

garyluu:
hmm...title needs fixing.  suggestions anyone?

coverbeck:
Regarding title, I would at least get rid of the word `Authentication`. I would be tempted to simply say `Login`, but then that matches/duplicates the tab text.

Hmm, actually the word `Login` is there 5 times: the tab, the title, on each button, and in the description. 7, if you count the button at the top right and the page header. :)

I wonder if you also couldn't then get rid of Login from the buttons. Use the colors from GitHub and Google... Sort of like the Stack Overflow page, although they don't have GitHub (not necessarily the size of those buttons, but the fact that it just lists the providers only and uses the colors of their logos).

screen shot 2018-08-09 at 12 50 13 pm

I'll run this locally later to see if I can think of anything else, but those are my initial thoughts from just looking at your screenshot.

garyluu:
I kept it mostly similar to what it was before, but if there's such strong objections to the previous design, I don't mind deviating from it.

denis-yuen:
For my part, I would agree with removing "Login Authentication Headers" entirely, somewhat like Gitlab's login page https://gitlab.com/users/sign_in
I think the tab headers are clear enough. A more drastic redesign can wait for a follow-up I think.

garyluu:
Using material theme...still looks strange:

image

garyluu:
the buttons are pretty much from https://lipis.github.io/bootstrap-social/

coverbeck:
1. I like the dialog header not being there.
2. I think I prefer the old buttons -- that Google red button especially is not pleasing. You could get a button for them from here: https://developers.google.com/identity/branding-guidelines

That said, I didn't mean to divert this from the original point of the PR. I think it looks much better without the title; and you could tackle the buttons, assuming you even agree in some later PR.

agduncan94:
Regarding the material theme changes, I think it makes sense to:
1) change to a grey border and make the size 1px, current one is too big.
2) remove the blue colour from the tabs, should be white for selected and grey for the other.
3) I think text should be above the buttons
4) I do agree with Charles on the buttons, though I think we should change them in this PR and not later just to get it all done with. I like the light button he linked to.

garyluu:
Added Andrew's suggestions, reverting the buttons back the old ones (even though I think https://lipis.github.io/bootstrap-social/ looks better) : 
image

Those who wish to change the buttons to something different can do so in their own PR.

src/app/app.module.ts
```diff
@@ -92,6 +93,12 @@ import { StargazersModule } from './stargazers/stargazers.module';
 import { StarredEntriesComponent } from './starredentries/starredentries.component';
 import { StarringModule } from './starring/starring.module';
 
+export const myCustomTooltipDefaults: MatTooltipDefaultOptions = {
```
agduncan94: :+1:

PR #384, opened 2018-08-09
Feature/1691/self destruct

garyluu:
Initial implementation for self-destruct.  

Partial ga4gh/dockstore#1691

Also for ga4gh/dockstore#1695

Accounts page was getting crowded, separated sections into tabs.
Also renamed the internal accounts to profiles and turned profiles into cards.
A few simplifications to the profiles page.


garyluu:
Picture of Profiles page

image

Picture of account controls currently disabled:

image

denis-yuen:
Hmmm, 
Not sure about the scale, but either the profile pictures seem kind of big or the text is rather small. 
Let's match the sizes in the stargazers page (for example https://staging.dockstore.org/workflows/github.com/h3abionet/h3agatk:1.0.1?tab=info then click on the number of starrers)

For the modal:
"All third party accounts will be unlinked ..." actually doesn't mean all that much, so I would remove it, but I would mention that profile information will be deleted.

garyluu:
I've shrunken the cards a bit.  The image size gets reduced as a result.  The image in cards are meant to be stretched to the entire width of the card (unless there's unnatural padding adding).  The width of the card can't shrink much more or else the user's email address might start wrapping too early.

src/app/loginComponents/accounts/internal/accounts.component.ts
```diff
@@ -14,41 +16,37 @@ import { UsersService } from './../../../shared/swagger/api/users.service';
...
-    this.usersService.updateLoggedInUserMetadata('github.com').subscribe((user: User) => {
+  /**
+   * Update user metadata for a service ('github.com' or 'google.com')
+   *
+   * @param {string} service  'github.com' or 'google.com'
+   * @memberof AccountsInternalComponent
+   */
+  sync(service: TokenSource) {
+    this.syncing = true;
+    this.usersService.updateLoggedInUserMetadata(service).subscribe((user: User) => {
```
coverbeck: I can't comment on lines you didn't change, but since you're no longer hard-coding the token type here, do you want to change lines 60 and 65 to use`TokenSource.GOOGLE.toString()` and `TokenSource.GITHUB.toString()`?

src/app/loginComponents/accounts/internal/accounts.component.ts
```diff
@@ -14,41 +16,37 @@ import { UsersService } from './../../../shared/swagger/api/users.service';
...
       this.hasGoogleToken$ = this.tokenService.hasGoogleToken$;
     }
 
-  syncGitHub() {
-    this.syncingWithGitHub = true;
-    this.usersService.updateLoggedInUserMetadata('github.com').subscribe((user: User) => {
+  /**
+   * Update user metadata for a service ('github.com' or 'google.com')
+   *
+   * @param {string} service  'github.com' or 'google.com'
```
coverbeck: Should that be `@param {TokenSource}`?

src/app/loginComponents/accounts/controls/delete-account-dialog/delete-account-dialog.component.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * What happens when deleteing the account has failed.
+   *
+   * @private
+   * @memberof DeleteAccountDialogComponent
+   */
+  private deleteAccountFailure(): void {
+    this.matSnackBar.open('Deleting Dockstore account failed', 'Dismiss', {
+      duration: 5000,
```
coverbeck: We should probably have this 5000 value in a shared location; I copied the 5000 for permissions component as well. Not for this PR, just mentioning it.

src/app/loginComponents/accounts/internal/accounts.component.ts
garyluu: sure, might as well :+1:

src/app/loginComponents/accounts/controls/delete-account-dialog/delete-account-dialog.component.ts
garyluu: Should be straightforward, I've done something similar for tooltips already.

garyluu:
ugh, using the new webservice causes more things to break than i thought

src/app/loginComponents/accounts/controls/controls.component.html
```diff
@@ -0,0 +1,10 @@
+<mat-card>
+  <mat-card-header>
+    <!-- <mat-card-title>Dockstore Account Controls</mat-card-title> -->
+    <mat-card-subtitle>
+      Dockstore account controls have wide overreaching consequences and should be used with extreme caution
```
denis-yuen: `overreaching` is more of a criticism, maybe "important"

denis-yuen:
Pending successful build with merge

garyluu:
hmm... i think this actually passed already

PR #386, opened 2018-08-13
Add test for self destruct

garyluu:
Adds a quick test to check that both self-destruct form buttons are disabled/enabled correctly and that it goes to login page afterwards. 

Also adds progress bar and fixes unsubscribe.


garyluu:
I ended up moving the disabled extendeduser tests into another file for simplicity

PR #395, opened 2018-08-15
Feature/register tool modal

garyluu:
Converts from ngx-bootstrap modal to material dialog.  The fields are not material though.


denis-yuen:
As we start to nail stuff down for the release, let's kick up the testing a notch. 
The codecov target of 62.5% of the diff sounds reasonable.

garyluu:
Coverage is currently not counted when tested in Cypress.

garyluu:
i think the pull base failed.

PR #400, opened 2018-08-17
show loading bar when checking for entries

agduncan94:
Shows a loading bar when fetching my tools/workflows, instead of showing blank or showing the alert that they have no tools/workflows/

### DIFF
--- a/src/app/mytools/my-tool/my-tool.component.html
+++ b/src/app/mytools/my-tool/my-tool.component.html
@@ -26,7 +26,7 @@
         <mat-sidenav-container autosize class="sidebar-set-height">
           <mat-sidenav #sidenav mode="side" [(opened)]="showSidebar" class="col-sm-3 col-xs-12">
             <span class="visible-xs" *ngIf="refreshMessage">
-              <mat-progress-bar mode="indeterminate"></mat-progress-bar><br />
+              <mat-progress-bar mode="indeterminate" class="m-2"></mat-progress-bar>
             </span>
             <div class="clearfix sidebar-header">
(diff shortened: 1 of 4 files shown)


garyluu:
This type of change can just reuse the refreshMessage variable.  Just set the refreshMessage to something truthy before the usersService call and finalize with something non-truthy

src/app/mytools/my-tool/my-tool.component.ts
```diff
@@ -68,8 +68,10 @@ export class MyToolComponent extends MyEntry implements OnInit {
     this.userService.user$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(user => {
       if (user) {
         this.user = user;
+        this.stateService.setRefreshMessage('Fetching tools');
         this.usersService.userContainers(user.id).pipe(first()).subscribe(tools => {
           this.containerService.setTools(tools);
+          this.stateService.setRefreshMessage(null);
```
garyluu: put this in finalize or error too

src/app/mytools/my-tool/my-tool.component.html
```diff
@@ -26,7 +26,7 @@
         <mat-sidenav-container autosize class="sidebar-set-height">
           <mat-sidenav #sidenav mode="side" [(opened)]="showSidebar" class="col-sm-3 col-xs-12">
             <span class="visible-xs" *ngIf="refreshMessage">
-              <mat-progress-bar mode="indeterminate"></mat-progress-bar><br />
+              <mat-progress-bar mode="indeterminate" class="m-2"></mat-progress-bar>
```
garyluu: check if this causes the bounce back

garyluu: if it does, copy the progress bar from other components

agduncan94: I meant to put this on the other one, I reverted this back.

PR #405, opened 2018-08-20
Handle errors in HTTP calls

garyluu:
For ga4gh/dockstore#1588 Split HTTP request to better handle errors.

### DIFF
--- a/src/app/myworkflows/my-workflow/my-workflow.component.ts
+++ b/src/app/myworkflows/my-workflow/my-workflow.component.ts
@@ -15,14 +15,14 @@
  */
 import { Location } from '@angular/common';
 import { Component, OnInit } from '@angular/core';
-import { MatDialog } from '@angular/material';
+import { MatDialog, MatSnackBar } from '@angular/material';
 import { NavigationEnd, Router } from '@angular/router/';
 import { AuthService } from 'ng2-ui-auth/commonjs/auth.service';
 import { combineLatest } from 'rxjs';
-import { first, takeUntil } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


src/app/myworkflows/my-workflow/my-workflow.component.ts
```diff
@@ -111,16 +111,30 @@ export class MyWorkflowComponent extends MyEntry implements OnInit {
...
         this.stateService.setRefreshMessage('Fetching workflows');
-        combineLatest(this.usersService.userWorkflows(user.id).pipe(first()), this.workflowsService.sharedWorkflows().pipe(first()))
-          .subscribe(([workflows, sharedWorkflows]) => {
-            if (workflows && sharedWorkflows) {
-              this.workflowService.setWorkflows(workflows);
-              this.workflowService.setSharedWorkflows(sharedWorkflows);
-              this.stateService.setRefreshMessage(null);
-            }
-          }, (error: any) => {
+        let count = 2;
```
denis-yuen: A bit confusing

I'd prefer two booleans (e.g. `userWorkflowsFound` and `sharedWorkflowsFound`) and test with `userWorkflowsFound && sharedWorkflowsFound` instead of `count === 0` even though its lengthier

garyluu: could change it to 2 booleans, but "found" may be misleading

denis-yuen: @garyluu fair enough, use whatever variable name is accurate yet descriptive

PR #408, opened 2018-08-20
Feature/1626/checker workflow button

garyluu:
Follow-up on ga4gh/dockstore#1626  .  Can't click view checker workflow too earlier because it's still being fetched from webservice.  Disabling button until then.

### DIFF
--- a/cypress/fixtures/refreshedChecker.json
+++ b/cypress/fixtures/refreshedChecker.json
@@ -1,0 +1,61 @@
+
+
+{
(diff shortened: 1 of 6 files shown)


denis-yuen:
Can we get a test for the issue (i.e. checker workflow button is unusable after adding checker workflow)?

denis-yuen:
Build restarted once

denis-yuen:
Build error relates to checker workflows and so seems legitimate

PR #412, opened 2018-08-21
Material colour range

agduncan94:
This changes the light blue variant of the main blue colour to be a bit lighter, so that it is visible for the progress bar.

It also takes advantage of the colour palette generated through the sketch app.

https://github.com/ga4gh/dockstore/issues/1745

screenshot from 2018-08-21 11-06-13


coverbeck:
I'm curious to see if your Travis builds pass, given the issues I'm facing on #409.

PR #414, opened 2018-08-22
Feature/1739/hosted entry copy download buttons

agduncan94:
Adds copy and download buttons to hosted entries.
https://github.com/ga4gh/dockstore/issues/1739

Currently for hosted workflows we have a toolbar with the filename and delete button (fab). For other workflows we have the filename in a dropdown and then the download and copy button above and to the right of the code editor.

Including this PR, the following two images represent the currents state of hosted and remote workflows respectively.

I feel like it is a bit weird to not be consistent. We could add a toolbar to the remote workflows along with the icons for copy/download, but then would it look weird with the dropdown above? or could we move the dropdown into the toolbar? But then would people expect that with hosted workflows?

hosted
remote-workflows

Edit:
Here is the remote workflows using a similar ui to the hosted workflows.
remote-workflows-updated

### DIFF
--- a/src/app/shared/code-editor-list/code-editor-list.component.ts
+++ b/src/app/shared/code-editor-list/code-editor-list.component.ts
@@ -22,7 +22,6 @@ export class CodeEditorListComponent {
   NEXTFLOW_CONFIG_PATH = '/nextflow.config';
   NEXTFLOW_PATH = '/main.nf';
   public DescriptorType = ToolDescriptor.TypeEnum;
-  publishDownloadLinks: Array<string>;
 
   constructor(private workflowService: WorkflowService) {
     this.published$ = this.workflowService.workflowIsPublished$;
--- a/src/app/shared/code-editor-list/code-editor-list.component.ts
+++ b/src/app/shared/code-editor-list/code-editor-list.component.ts
@@ -2,10 +2,7 @@ import { Component, Input } from '@angular/core';
 import { ToolDescriptor } from './../../shared/swagger/model/toolDescriptor';
 import { WorkflowService } from '../../shared/workflow.service';
 import { Observable } from 'rxjs';
-import { FileService } from '../file.service';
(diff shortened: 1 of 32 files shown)


denis-yuen:
I don't have a strong opinion on this one. 
Put me down for slight preference for dropdown in toolbar with the icon buttons
In terms of consistency, it doesn't bother me too much. Compare https://gist.github.com/denis-yuen/4d53effd18b6a5a90c9df3ac8c9a60c5 with the view of a single file on regular github. 
If a user is going beyond a certain file size, the dropdown makes more sense but they should probably not be on hosted workflows either.

agduncan94:
Here is what I ended up with, workflows and tools.

workflows
tools

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
```diff
@@ -18,6 +18,7 @@ export class WorkflowFileEditorComponent extends FileEditing {
   originalSourceFiles = [];
   _selectedVersion: WorkflowVersion;
   @Input() descriptorType: string;
+  @Input() entrypath: string;
```
garyluu: i prefer entryPath

src/app/shared/code-editor-list/code-editor-list.component.ts
```diff
@@ -12,10 +14,19 @@ export class CodeEditorListComponent {
...
+  @Input() entryType: string;
+  @Input() entrypath: string;
+  @Input() selectedVersion: WorkflowVersion;
+  protected published$: Observable<boolean>;
+  public downloadFilePath: string;
   NEXTFLOW_CONFIG_PATH = '/nextflow.config';
   NEXTFLOW_PATH = '/main.nf';
   public DescriptorType = ToolDescriptor.TypeEnum;
-  constructor() { }
+  publishDownloadLinks: Array<string>;
```
garyluu: unused?

garyluu:
return type for pipes

PR #426, opened 2018-08-24
Feature/1743/user fix

garyluu:
Fixing ga4gh/dockstore#1743 again

Setting userService observables to be null when exiting just in case.


denis-yuen:
What happened the first go at this?

garyluu:
Only fixed it for when the user logs in and logs out.  If the user logs into github, logout, login to google, it appears

PR #427, opened 2018-08-27
Feature/1747/linking account error

agduncan94:
* switches order of external accounts and profiles pages
* snackbar message for linking account (not success or failure, only that the process is starting) and for unlinking account (one for unlinking success and one for failure)
* moves the dockstore token to the accounts page, in a new dockstore section
* makes the copy/view token nicer (material)

Note that I wasn't able to get the error snackbar message for when linking accounts don't work, so that issue cannot be closed yet. But at least now you are redirected to the same page, so that you can see that the account was not linked.

https://github.com/ga4gh/dockstore/issues/1747

Swagger UI link in footer
https://github.com/ga4gh/dockstore/issues/1375

screenshot from 2018-08-27 10-07-27

### DIFF
--- a/cypress/integration/dropdown.js
+++ b/cypress/integration/dropdown.js
@@ -42,7 +42,6 @@ describe('Dropdown test', function () {
       cy
         .get('#dropdown-accounts')
         .click()
-      cy.contains("External Accounts").click()
     });
 
     it('Should show all accounts as linked (except GitLab and Bitbucket)', function () {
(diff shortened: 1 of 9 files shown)


denis-yuen:
@agduncan94  sorry to pile on, but if you're adding the swagger link anyway here. 
You can probably also ninja in https://github.com/ga4gh/dockstore/issues/1375

PR #430, opened 2018-08-27
Feature/1688/hide labels in starred users

garyluu:
For ga4gh/dockstore#1688  Hides labels from stargazers page

For ga4gh/dockstore#1765  Now gives the correct URL for ALL workflows (not just hosted).  Adds a test.

### DIFF
--- a/cypress/integration/starEntry.js
+++ b/cypress/integration/starEntry.js
@@ -13,6 +13,8 @@
  *     See the License for the specific language governing permissions and
  *     limitations under the License.
  */
+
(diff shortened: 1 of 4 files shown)


cypress/integration/starEntry.js
```diff
@@ -13,6 +13,8 @@
  *     See the License for the specific language governing permissions and
  *     limitations under the License.
  */
+
+ // This file can run independent of all other tests successfully and is idempotent
```
denis-yuen: :+1:

PR #433, opened 2018-08-28
Hide disabled option altogether

denis-yuen:
* would prefer tooltip, but for  https://github.com/angular/material2/issues/7953

### DIFF
--- a/src/app/search/map-friendly-values.pipe.ts
+++ b/src/app/search/map-friendly-values.pipe.ts
@@ -38,7 +38,7 @@ export class MapFriendlyValuesPipe implements PipeTransform {
     ])],
     ['descriptor_tooltip', new Map([
       ['CWL', 'Common Workflow Language'], ['WDL', 'Workflow Description Language'],
-      ['Nextflow', 'Nextflow coming soon!']
+      ['NFL', 'Nextflow coming soon!']
     ])],
   ]);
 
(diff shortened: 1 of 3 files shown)


src/app/search/map-friendly-values.pipe.ts
```diff
@@ -35,7 +35,11 @@ export class MapFriendlyValuesPipe implements PipeTransform {
     ])],
     ['source_control_provider.keyword', new Map([
       ['GITHUB', 'github.com'], ['BITBUCKET', 'bitbucket.org'], ['GITLAB', 'gitlab.com'], ['DOCKSTORE', 'dockstore.org']
-    ])]
+    ])],
+    ['descriptor_tooltip', new Map([
+      ['CWL', 'Common Workflow Language'], ['WDL', 'Workflow Description Language'],
+      ['Nextflow', 'Nextflow coming soon!']
```
garyluu: 'Nextflow' => 'NFL'

denis-yuen: These tooltips should actually provide more lengthy information.

denis-yuen: (FYI, discuss in person, I probably need to address the refresh issues)

denis-yuen:
@garyluu is there anything to follow-up on for this?
@agduncan94 probably re-review

garyluu:
don't think so

PR #440, opened 2018-09-04
Simple fix

garyluu:
Simple fix for launch command for relative file paths that have  `../` in it (like the HCA workflow).  Downside is even perfectly legit `/` are also encoded everywhere else


garyluu:
Changed to hackier solution that only encodes the necessary parts

PR #442, opened 2018-09-04
Strip versions for update calls

denis-yuen:
* not needed for updating working paths


denis-yuen:
coverage on the fritz?

PR #455, opened 2018-09-11
changed to material

agduncan94:
Changed the contact buttons (request access and contact author) to be material.


src/app/container/container.component.html
```diff
@@ -68,19 +68,15 @@ <h3 id="tool-path">
...
+          Request Access
+        </a>
       </div>
-      <div class="button contact-button" *ngIf="publicPage && tool?.email && !tool?.private_access">
-        <p>
-          <a [href]="contactAuthorHREF">
-            Contact Author
-          </a>
-        </p>
+      <div class="contact-button" *ngIf="publicPage && tool?.email && !tool?.private_access">
```
garyluu: how does it look without contact-button?

agduncan94: Hm contact button is just margin really, I'll use a b4 class.

PR #457, opened 2018-09-13
Don't set Authorization header if no token

coverbeck:
#1818

Header value was getting set to the string `Bearer null`.

We could instead delete the 'Authorization' property, but I think
this approach is better, as you can see the value is intentionally
set to null.


garyluu:
Approved if the goal of the PR/issue was to 401 without an `Authorization` header

coverbeck:
Yes, we want the 401 without auth header; the problem is that if it's set to an invalid value, then Dockstore webservice will both look it up as a Dockstore token and as a Google token, creating unnecessary load.

PR #459, opened 2018-09-13
Null check when setting authorization header

coverbeck:
#1824

### DIFF
--- a/src/app/shared/entry/GA4GHFiles.state.service.ts
+++ b/src/app/shared/entry/GA4GHFiles.state.service.ts
@@ -63,8 +63,12 @@ export class GA4GHFilesStateService {
    * @memberof GA4GHFilesStateService
    */
   public injectAuthorizationToken(ga4ghService: GA4GHService) {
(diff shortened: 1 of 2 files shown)


src/app/shared/entry/GA4GHFiles.state.service.ts
```diff
@@ -63,8 +63,12 @@ export class GA4GHFilesStateService {
    * @memberof GA4GHFilesStateService
    */
   public injectAuthorizationToken(ga4ghService: GA4GHService) {
-    ga4ghService.defaultHeaders = ga4ghService.defaultHeaders.set('Authorization',
-      ga4ghService.configuration.apiKeys['Authorization']);
+    const auth = ga4ghService.configuration.apiKeys['Authorization'];
+    if (auth) {
```
agduncan94: Could simplify to one liner like the change in my-entry

denis-yuen: As an opinion, I think this might be getting a bit long/unreadable for a ternary at, roughly 100 characters.

PR #469, opened 2018-10-04
Feature/1806/nextflow hosted workflows launch with

agduncan94:
- Enables hosted Nextflow workflows
- Updates the launch with for hosted nextflow workflows

https://github.com/ga4gh/dockstore/issues/1806

### DIFF
--- a/src/app/container/launch/launch.component.ts
+++ b/src/app/container/launch/launch.component.ts
@@ -22,6 +22,8 @@ import { Tag } from './../../shared/swagger/model/tag';
 import { ToolLaunchService } from './tool-launch.service';
 import { ContainerService } from '../../shared/container.service';
 import { Observable } from 'rxjs';
+import { Workflow } from '../../shared/swagger/model/workflow';
+import { DockstoreTool } from '../../shared/swagger/model/dockstoreTool';
 
 @Component({
   selector: 'app-launch',
@@ -29,10 +31,10 @@ import { Observable } from 'rxjs';
   styleUrls: ['./launch.component.css']
 })
 export class LaunchComponent {
-  @Input() basePath;
-  @Input() path;
-  @Input() toolname;
-  @Input() mode;
+  @Input() basePath: string;
+  @Input() path: string;
+  @Input() toolname: string;
+  @Input() mode: (DockstoreTool.ModeEnum | Workflow.ModeEnum);
 
   _selectedVersion: Tag;
   @Input() set selectedVersion(value: Tag) {
--- a/src/app/container/launch/launch.component.ts
+++ b/src/app/container/launch/launch.component.ts
@@ -32,6 +32,7 @@ export class LaunchComponent {
   @Input() basePath;
   @Input() path;
   @Input() toolname;
+  @Input() mode;
 
   _selectedVersion: Tag;
   @Input() set selectedVersion(value: Tag) {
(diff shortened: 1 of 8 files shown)


agduncan94:
screenshot from 2018-10-04 16-02-00

src/app/shared/launch.service.ts
```diff
@@ -76,7 +76,23 @@ export abstract class LaunchService {
...
     getNextflowNativeLaunchString(workflowPath: string, versionName: string) {
-      return `$ nextflow run http://${workflowPath} -r ${versionName} -with-docker`;
+      return `$ nextflow run http://${workflowPath} -r ${versionName}`;
+    }
+
+    /**
+     * Gets local launch command
+     */
+    getNextflowLocalLaunchString() {
+      return `$ nextflow run nextflow.config`;
```
coverbeck: Just curious, is there a separate ticket to run nextflow with dockstore like you can with cwl and wdl? That would allow the file provisioning plugins to be used like the are for the other languages. Or does it not make sense for Nextflow?

denis-yuen: A question well worth asking. We need to look at their engine to see if it makes sense or I think Andrew noted they have a parallel mechanism which would make this a bit redundant.

agduncan94: Nextflow offers a lot of what the Dockstore CLI does, so it would be quite redundant. It also has more features that the Dockstore CLI doesn't have. For consistency though, a user should be able to use `dockstore workflow launch` for nextflow as well, even if it just downloads the dockstore workflow and launches it with nextflow without doing any fancy file provisioning.

agduncan94: @denis-yuen Is this something you think is worth creating an issue for?

src/app/container/launch/launch.component.ts
```diff
@@ -32,6 +32,7 @@ export class LaunchComponent {
   @Input() basePath;
   @Input() path;
   @Input() toolname;
+  @Input() mode;
```
garyluu: Add type, others too while ur at it.

src/app/shared/launch.service.ts
```diff
@@ -76,7 +76,23 @@ export abstract class LaunchService {
...
+    }
+
+    /**
+     * Gets the download command
+     * @param workflowPath Path of the workflow
+     * @param versionName Name of the version
+     */
+    getNextflowDownload(workflowPath: string, versionName: string) {
```
garyluu: return type

src/app/workflow/launch/launch.component.ts
```diff
@@ -34,6 +36,7 @@ export class LaunchWorkflowComponent extends EntryTab {
   @Input() basePath;
   @Input() path;
   @Input() currentDescriptor;
+  @Input() mode;
```
garyluu: .

src/app/shared/launch.service.ts
```diff
@@ -76,7 +76,23 @@ export abstract class LaunchService {
...
      */
     getNextflowNativeLaunchString(workflowPath: string, versionName: string) {
-      return `$ nextflow run http://${workflowPath} -r ${versionName} -with-docker`;
+      return `$ nextflow run http://${workflowPath} -r ${versionName}`;
+    }
+
+    /**
+     * Gets local launch command
+     */
+    getNextflowLocalLaunchString() {
```
garyluu: .

PR #473, opened 2018-10-09
Feature/state

garyluu:
Adds state management using Akita (not a dev dependency but tree-shaking in prod removes it from use)
Moves the state.service.ts to an Akita store called "session"
Moves the old ga4gh.state.service.ts to an Akita store called "ga4gh-files"

- Fixes the odd state issue that occurs when switching between my-tools and my-workflows which causes the files to be retrieved for the wrong workflow
- Fixes the odd state issue that occurs when switching between tools or workflows which causes the wrong files to be retrieved for the new workflow


src/app/loginComponents/accounts/controls/delete-account-dialog/delete-account-dialog.component.spec.ts
```diff
@@ -16,9 +16,9 @@ describe('DeleteAccountDialogComponent', () => {
   beforeEach(async(() => {
     TestBed.configureTestingModule({
       declarations: [DeleteAccountDialogComponent],
+      schemas: [NO_ERRORS_SCHEMA],
```
agduncan94: Is this a good idea :thinking:

src/app/myworkflows/my-workflow/my-workflow.component.spec.ts
```diff
@@ -102,6 +96,7 @@ describe('MyWorkflowsComponent', () => {
     expect(registerWorkflowModalService.setWorkflowRepository).toHaveBeenCalled();
   });
   it('should refresh workflows', () => {
+    component.user = { id: 1};
```
agduncan94: spacing :open_mouth:

src/app/shared/alert/state/alert.service.ts
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * simpleSuccess() doesn't show the snackbar notification, detailedSuccess() does show it
+ * simpleError() shows the snackbar but not the alert, detailedError() shows both
+ * @export
+ * @class AlertService
+ */
+@Injectable({ providedIn: 'root' })
+export class AlertService {
+
+  constructor(private alertStore: AlertStore, private matSnackBar: MatSnackBar, private alertQuery: AlertQuery) {
```
agduncan94: Could use some function comments.

src/app/loginComponents/accounts/controls/delete-account-dialog/delete-account-dialog.component.spec.ts
garyluu: It's to ignore sub components in testing because we're not interested in those.

agduncan94: Could you add this to the style guide?

src/app/container/tool-file-editor/tool-file-editor.component.ts
```diff
@@ -77,16 +84,16 @@ export class ToolFileEditorComponent extends FileEditing {
...
         this.id,
         newSourceFiles).subscribe(result => {
           this.toggleEdit();
           this.containerService.setTool(result);
-          this.refreshService.handleSuccess(message);
+          this.alertService.detailedSuccess();
         }, error =>  {
           if (error) {
-            this.refreshService.handleError(message, error);
+            this.alertService.detailedError(error);
```
coverbeck: What if `error` is not set (can that occur)? Do you still need to call alertService?

src/app/shared/alert/state/alert.service.ts
```diff
@@ -0,0 +1,147 @@
...
+    this.setInfo(message);
+  }
+
+  /**
+   * Handles successful HTTP response and shows success to the user by matSnackBar
+   *
+   * @memberof AlertService
+   */
+  public detailedSuccess() {
+    const previousMessage = this.alertQuery.getSnapshot().message;
```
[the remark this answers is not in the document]
garyluu: this is not async

src/app/container/tool-file-editor/tool-file-editor.component.ts
garyluu: not sure how there can be no error but if there no error, call simpleError()

src/app/shared/descriptor-type-compat.service.ts
```diff
@@ -0,0 +1,74 @@
...
+   * @param {string} descriptor
+   * @memberof LaunchComponent
+   */
+  public stringToDescriptorType(descriptorType: string): ToolDescriptor.TypeEnum {
+    switch (descriptorType) {
+      case WebserviceDescriptorTypeEnum.CWL: {
+        return ToolDescriptor.TypeEnum.CWL;
+      }
+      case ToolDescriptor.TypeEnum.CWL: {
+        console.log('Unneeded conversion');
```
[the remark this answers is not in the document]
garyluu: Ya, I want to leave these in to catch future issues even though it's not entirely an error.  Even warn is too extreme imo.

src/app/shared/alert/state/alert.service.ts
```diff
@@ -0,0 +1,147 @@
...
+   * @memberof AlertService
+   */
+  public detailedError(error: HttpErrorResponse) {
+    let message: string;
+    let details: string;
+    if (error.status === 0) {
+      // Error code of 0 means the webservice is not responding, likely down
+      message = 'The webservice is currently down, possibly due to load. Please wait and try again later.';
+    } else {
+      message = 'The webservice encountered an error trying to create/modify.';
```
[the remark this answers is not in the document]
garyluu: This is copied from previous code.  And detailed messages are currently only used for create/modify actions

PR #474, opened 2018-10-25
Feature/1864/export as zip

agduncan94:
Better approach for download ZIP. Unfortunately can only use this for published entries. Benefits is that it is not blocked by popup blockers in the browser, and also the name of the ZIP file isn't random numbers.

https://github.com/ga4gh/dockstore/issues/1864

### DIFF
--- a/src/app/container/info-tab/info-tab.component.html
+++ b/src/app/container/info-tab/info-tab.component.html
@@ -166,8 +166,8 @@
     </li>
   </ul>
 
-  <span *ngIf="isValidVersion">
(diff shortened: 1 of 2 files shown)


coverbeck:
On second thought, I don't think it will work for My Workflows because of auth.

coverbeck:
Yes, I don't know if it's worth it, and if this works, then maybe best not to bother.

The only way I think you could get it work for the auth case is to prefetch the zip and use a blob, but I don't think you want to do that.

Or you could dynamically create an anchor element, like [this](http://blog.bguiz.com/2014/07/03/file-download-with-http-request-header/)

PR #477, opened 2018-10-26
Feature/hosted entry extra name

agduncan94:
Can set the toolname and workflowname for hosted tools and workflows, respectively.

addhostedtool
addhostedworkflow

Edit: I've updated the screenshots.

https://github.com/ga4gh/dockstore/issues/1738


coverbeck:
Since the repository can now be specified by the user, does this mean that I, as coverbeck, can create a hosted workflow in `dockstore.org/agduncan`? If so, are there any issues that arise from that, or is that OK?

agduncan94:
Oops actually you've pointed out an issue with the UI. How it is supposed to be is for me for example the workflow is always in "dockstore.org/agduncan94".

Before you were limited to "dockstore.org/agduncan94/wfname", but now you could have "dockstore.org/agduncan94/wfname/wf1" and "dockstore.org/agduncan94/wfname/wf2".

I'll make the fixes now.

coverbeck:
Some random thoughts:

* Should the repository be optional?
* What happens to hosted workflows that were already created before this change -- what repo will they be in? And can I create additional workflows in the same location, whatever that is?
* It would be nice if this repository input were an editable dropdown, that way I could safely pick an existing repo without worrying about making a typo in the name.

agduncan94:
@coverbeck 
Just for some background, normal paths are A/B/C(/D), where D is optional you can add. We added this in the case where a docker image is used for multiple tools. We translated this to workflows too, in the case of duplicate git repository.

For existing hosted entries, they would not have any workflow/toolname set. This is a good point, as a migration needs to be added to add the column and set it to null. These would still be valid hosted entries, as the name portion is optional.

As for your last point, similar things have been brought up in order to improve the registration process in the future, though at this point it hasn't been anything concrete decided on.

@denis-yuen 
For your point on confusion, the workflow version uses the material design guidelines, where the asterisk means required. If you click on the optional field it will also say optional. I think the confusion mostly stems from the lack of consistency between the two.

I agree that it adds complexity to the hosted workflows. I added it for completeness sake. Maybe it could be an advanced option?

I also agree that it makes more sense for hosted tools over hosted workflows.

coverbeck:
I was referring to the repository name being required, not the workflow/tool name. The repository name looks like it is required, as it has an asterisk next to it.

Today, if I create a hosted workflow it gets created in `dockstore.org/coverbeck`, e.g., `dockstore.org/coverbeck/myworkflow`.

My initial question had been, if I can specify the repo, what would stop me from creating a workflow `dockstore.org/agduncan/myworkflow`. You said that the repository comes after the username part, and you were going to update the UI to indicate that.

So if I enter a repository name `foo` and a workflow name `bar`, then that means the name will be `dockstore.org/coverbeck/foo/bar`? I think that's fine if so, but:

1. Perhaps the UI could indicate that (but not crucial).
2. What repo will pre-existing hosted workflows be under?

I hope I'm not majorly misunderstanding things here.

denis-yuen:
@agduncan94 Ah, forgot material vs older dialogues. That makes sense

For hosted workflows, I'm not sure it should even be an advanced feature if we ourselves cannot think of a use-case. It would add unnecessary confusion I think. I wouldn't remove the functionality from the webservice, but I simply wouldn't show it in the UI2 under we get a request/use-case

agduncan94:
@coverbeck Sorry for the confusion, the screenshot is out of date. So in reality the path for me is always dockstore.org/agduncan/repository/workflowname The only thing we can set are the latter two parts of the path (repo and workflowname). So you are correct with your foo/bar example.

As for your two points:
1) I agree, I think we should show the path that will be created as the used fills in the options.
2) Any existing hosted workflow will keep their naming. If you had an existing hosted workflow with the path dockstore.org/coverbeck/myworkflow, the repository is myworkflow. The workflowname is null. All this PR does is allow you to add the fourth path element, called workflowname, to new hosted workflows. I think part of the confusion stems from the UI previously asking for the workflow name, even though we store that as the repository in the db.

coverbeck:
@agduncan94 Thanks, that clarified it -- you were saying the same thing before, but it finally clicked for me with your last comment.

src/app/container/container.component.html
```diff
@@ -142,7 +142,13 @@ <h3>
         <app-info-tab [selectedVersion]="selectedVersion" [privateOnlyRegistry]="privateOnlyRegistry" [extendedDockstoreTool]="(extendedTool$ | async)" [validVersions]="validVersions"></app-info-tab>
       </tab>
       <tab id="launchTab" heading="Launch" (select)="setEntryTab('launch')">
-        <app-launch *ngIf="tool?.tags.length > 0" [basePath]="tool?.path" [selectedVersion]="selectedVersion" [path]="tool?.tool_path" [toolname]="tool?.toolname" [mode]="tool?.mode"></app-launch>
+        <app-launch *ngIf="tool?.tags.length > 0; else noTags" [basePath]="tool?.path" [selectedVersion]="selectedVersion" [path]="tool?.tool_path" [toolname]="tool?.toolname"></app-launch>
```
coverbeck: You dropped `[mode]` -- is that intentional?

agduncan94: Good catch, this was an accident during a merge with develop.

PR #479, opened 2018-10-26
removed $ from start of commands

agduncan94:
https://github.com/ga4gh/dockstore/issues/1838


src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -118,15 +118,15 @@ exec newgrp docker
 #### Part 4 - Confirm installation
 1. Run our dependencies to verify that they have been installed properly.
 \`\`\`
-$ dockstore --version
+dockstore --version
```
denis-yuen: hmm, unclear this way which lines are commands and which lines are their output

agduncan94: Good point, do you think I should readd the $ or do something else?

denis-yuen: Not sure, may need some experimentation to see what looks best or to look at other projects with documentation online. 

Brainstorming
1. Re-add $
2. Don't use a quote block and distinguish the commands from the output in some other way (italics?)

agduncan94: I'll look at some coding tutorial sites, maybe they have some good implementations.

PR #483, opened 2018-11-05
Feature/info-tab-external-links displays the important information only

rjbautis:
* Displays the important info of an external service link in the info tab of workflows and tools
* Adds new angular pipe for deconstructing a url into just the pathname
* Defines a unit test for the new pipe

ga4gh/dockstore#1047


src/app/shared/entry/url-deconstruct.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@ Pipe({
+  name: 'urlDeconstruct'
+})
+export class UrlDeconstructPipe implements PipeTransform {
+
+  transform(providerUrl: string, versionName: string): any {
+    if (providerUrl.includes('github.com') && versionName !== '') {
```
coverbeck: You can simply say `&& versionName` -- if the expression is "truthy" (not undefined, not null, not empty), it will pass. This is arguably better, as you code doesn't catch the undefined and null cases, even though they may not be possible when invoked from the HTML in Angular.

src/app/shared/entry/url-deconstruct.pipe.ts
```diff
@@ -0,0 +1,17 @@
...
+      return (new URL(providerUrl).pathname).substr(1) + '/tree/' + versionName;
+    }
+    if (providerUrl.includes('bitbucket.org') && versionName !== '') {
```
coverbeck: Same comment regarding versionName

src/app/shared/entry/url-deconstruct.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@ Pipe({
+  name: 'urlDeconstruct'
+})
+export class UrlDeconstructPipe implements PipeTransform {
+
+  transform(providerUrl: string, versionName: string): any {
+    if (providerUrl.includes('github.com') && versionName !== '') {
+      return (new URL(providerUrl).pathname).substr(1) + '/tree/' + versionName;
```
coverbeck: If not a valid URL, e.g., does not start with http(s), this code  and the other instance below will probably throw an exception.

Not sure if it's worth worrying about, but just mentioning.

coverbeck: You have `new URL(providerUrl).pathname).substr(1)` 3 times...

You could just do it once, at the top of function, and then add the bitbucket or github path in your if conditions.

src/app/shared/entry/url-deconstruct.pipe.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  it('Create an instance and return url pathname', function () {
+    const pipe = new UrlDeconstructPipe();
+
+    expect(pipe).toBeTruthy();
+    expect(pipe.transform('https://github.com/agduncan94/hello-dockstore-workflow', ''))
+      .toBe('agduncan94/hello-dockstore-workflow');
+    expect(pipe.transform('https://github.com/agduncan94/hello-dockstore-workflow', 'master'))
+      .toBe('agduncan94/hello-dockstore-workflow/tree/master');
+    expect(pipe.transform('https://bitbucket.org/rjbautis/hello-dockstore', 'develop'))
+      .toBe('rjbautis/hello-dockstore/src/develop');
```
coverbeck: Would be good to have a test if it's not GitHub nor Bitbucket, since that's another codeflow in your pipe.

src/app/shared/entry/url-deconstruct.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@ Pipe({
```
garyluu: `@ Pipe` => `@Pipe`
Add a comment on what the pipe does

src/app/shared/entry/url-deconstruct.pipe.ts
rjbautis: Good point. I'm already mutating the string URL pathname in the return statements. I will try moving away from the URL library in favor of working directly on the providerUrl string

rjbautis:
Before

screen shot 2018-11-06 at 2 57 48 pm

After
screen shot 2018-11-06 at 2 56 52 pm

src/app/shared/entry/url-deconstruct.pipe.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'urlDeconstruct'
+})
+export class UrlDeconstructPipe implements PipeTransform {
+
+  transform(providerUrl: string, versionName: string): any {
```
garyluu: Put the comment above this.  Many IDEs will automatically generate the jsdoc for it (with the `@param, @return`, etc).  Change return type to string to match your previous comment.

src/app/shared/entry/url-deconstruct.pipe.ts
```diff
@@ -0,0 +1,37 @@
...
+    const PATHNAME = 3;
+
+    if (providerUrl.includes('github.com') && versionName) {
```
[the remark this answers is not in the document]
agduncan94: If you look online there should be some regex for git urls you could use. Actually if you look in the dockstore webservice code we might already have one.

agduncan94:
I'm wondering if instead of doing gitrepo/tree/version, you could do gitrepo:version. Might look cleaner.

garyluu:
Going to be merge conflicts for every test file modified because they were all swapped to .ts files

denis-yuen:
@garyluu as discussed, please help Ryan with the merge since these merge conflicts are probably a bit too painful

coverbeck:
Ryan and I just went through resolving the merge conflicts and I think we are good to go. But we'll ping you if we get stuck. Thanks!

cypress/integration/group3/mytools.ts
```diff
@@ -60,13 +60,12 @@ describe('Dockstore my tools', () => {
...
       cy.get('a#my-tools-nav-button').click();
-      cy.contains('github.com');
       selectUnpublishedTab('quay.io/A2');
       selectTool('b1');
       cy.contains('GitHub');
-      cy.contains('https://github.com/A2/b1');
+      cy.contains('A2/b1');
       cy.contains('Quay.io');
-      cy.contains('quay.io/A2/b1');
+      cy.contains('A2/b1');
```
garyluu: These two `contains` are a little awkward with the new changes.  Add IDs to the two in the HTML and then
have two `cy.get('#theID').contains('A2/b1').should('have.attr', 'href', 'whateverthelinkissupposedtobe')`

rjbautis: Sounds good, this removes ambiguity for future references.

PR #486, opened 2018-11-09
Update README.md

garyluu:
Updating README.md to be slightly more up to date with the current process

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Table of Contents
       * [Further help](#further-help)
 
 
-# DockstoreUi2
+# Dockstore UI2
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,7 +39,7 @@ v7.10.0
 $npm -v
 4.2.0
 
-$sudo npm install -g @angular/cli@1.3.1		
+$npm install -g @angular/cli@1.3.1
```
denis-yuen: Does the non sudo version of this setup the PATH properly for the "ng" commands?

garyluu: ng commands (or any commands installed globally using npm) should work in any directory assuming that npm is set up correctly.

denis-yuen: Assuming using the specific apt-get install command above?

PR #487, opened 2018-11-12
Feature/1808/files content fix

garyluu:
For ga4gh/dockstore#1808

Bug already fixed somehow 
This changes entry components to use material tabs instead of ngx-bootstrap tabs
This also changes many ngx-bootstrap modals to material dialogs for compatibility
Files and DAG tabs are now lazy-loaded because material tabs (both eager and lazy-loaded) are not in DOM until clicked so cytoscape and ace editor which relies on DOM elements don't work when eagerly loaded (unless using hacky workarounds to determine whether it is in DOM).


garyluu:
snyk being silly

src/app/shared/code-editor/code-editor.component.ts
```diff
@@ -40,7 +43,7 @@ export class CodeEditorComponent implements AfterViewInit {
 
   constructor() {
     // The purpose of the aceId is to deal with cases where multiple editors exist on a page
-    this.aceId = 'aceEditor_' + Math.floor(Math.random() * 100000).toString();
+    this.aceId = 'aceEditor_' + ACE_EDITOR_INSTANCE++;
```
agduncan94: :+1:

PR #489, opened 2018-11-13
Add terms of service

garyluu:
Put terms of service right under privacy policy:
image


src/app/home-foot-note/home-foot-note.component.html
```diff
@@ -47,6 +47,11 @@ <h3 class="documentation">Using Dockstore</h3>
               Privacy Policy
             </a>
           </li>
+          <li>
+            <a href="../assets/docs/Dockstore_Terms_of_Service.pdf">
```
coverbeck: I would prefer if this were `target="_blank"`, since it takes you to a PDF viewer where you can only get back to dockstore.org via the back button.

But the privacy policy is already like this, so I'm just putting this comment out as food for thought for the future.

PR #493, opened 2018-11-19
Fix calls to swagger methods

denis-yuen:
* Suspected fixes needed to get hotfix to build and deploy via compose_setup


denis-yuen:
FYI, build failure with "Should display the correct url" is consistent over 4 runs total

PR #501, opened 2018-12-05
Feature/1918/verification version

garyluu:
Changes:
- Verification panel now only shows platform, platform version, and entry version it was verified, not any additional metadata (too free form to show in panel)
- Added platform version to table, renamed the verifier field back to metadata


src/app/shared/entry/verified-by/verified-by.component.ts
```diff
@@ -15,7 +15,7 @@ export class VerifiedByComponent implements OnChanges {
 
   ngOnChanges() {
     if (this.version) {
-    this.verifiedByStringArray = this.verifiedByService.getVerifiedByString(this.version.sourceFiles);
+      this.verifiedByStringArray = this.verifiedByService.getVerifiedByString(this.version.sourceFiles);
```
coverbeck: I don't think this will ever happen nor will it be an issue if it does, but just checking... can version ever get unset after it has been set? If so, then `this.verifiedByStringArray` wouldn't be cleared out.

PR #502, opened 2018-12-06
See if descriptors exist before saying they don't

coverbeck:
ga4gh/dockstore#1937

When there are no descriptors, `nullDescriptors` is set to `true`, so go
off that instead of content

### DIFF
--- a/src/app/container/dockerfile/dockerfile.component.ts
+++ b/src/app/container/dockerfile/dockerfile.component.ts
@@ -42,11 +42,11 @@ export class DockerfileComponent {
   }
   content: string;
   filePath: string;
-  nullContent: boolean;
   public published$: Observable<boolean>;
   public downloadFilePath: string;
   public customDownloadHREF: SafeUrl;
   public customDownloadPath: string;
+  public loading = false;
   constructor(public fileService: FileService, private toolQuery: ToolQuery,
               private containerService: ContainerService, private containersService: ContainersService) {
     this.filePath = '/Dockerfile';
@@ -55,14 +55,17 @@ export class DockerfileComponent {
 
   reactToVersion(): void {
     if (this._selectedVersion) {
+      this.loading = true;
       this.containersService.dockerfile(this.id, this._selectedVersion.name).pipe(first())
         .subscribe(file => {
+            this.loading = false;
             this.content = file.content;
             this.filePath = file.path;
             this.downloadFilePath = this.getContainerfilePath();
             this.customDownloadFile();
           }, error => {
             this.content = null;
+            this.loading = false;
           }
         );
     } else {
--- a/src/app/container/dockerfile/dockerfile.component.ts
+++ b/src/app/container/dockerfile/dockerfile.component.ts
@@ -16,7 +16,7 @@
 import { Component, Input } from '@angular/core';
 import { SafeUrl } from '@angular/platform-browser';
 import { Observable } from 'rxjs';
-import { first } from 'rxjs/operators';
+import {finalize, first} from 'rxjs/operators';
 
 import { ga4ghPath } from '../../shared/constants';
 import { ContainerService } from '../../shared/container.service';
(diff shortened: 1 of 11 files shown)


coverbeck:
@garyluu I know we had discussed in Slack using the alerts service, but when I looked at the code, I noticed there already was a property I could use, so I just went with a simpler solution.

If you think it's worth the effort to switch to the alerts service or that other solution you proposed, let me know and I can redo it.

garyluu:
If you slow down to the network speed to slow 3G and using nullDescriptors work, then it's fine.

garyluu:
I think it was changed from nullDescriptors previously because there was some strange bug where we have the contents of the file but for some reason nullDescriptors was true.  Perhaps state changes may have changed that.

src/app/container/dockerfile/dockerfile.component.ts
```diff
@@ -42,11 +42,11 @@ export class DockerfileComponent {
   }
   content: string;
   filePath: string;
-  nullContent: boolean;
   public published$: Observable<boolean>;
   public downloadFilePath: string;
   public customDownloadHREF: SafeUrl;
   public customDownloadPath: string;
+  public loading = false;
```
garyluu: Should initially be set to true;

Loading should be set to false once:
- this.containersService.dockerfile(...) succeeds
- this.containersService.dockerfile(...) fails
- there is no this._selectedVersion
- and possibly more

Not noticeable for this component because the selected version is passed in as Input() and takes no time to get selected version and getting the dockerfile doesn't use multiple ga4gh endpoints.

src/app/shared/selectors/entry-file-selector.ts
```diff
@@ -42,6 +42,7 @@ export abstract class EntryFileSelector implements OnDestroy {
   public downloadFilePath: string;
   public customDownloadHREF: SafeUrl;
   public customDownloadPath: string;
+  public loading = false;
```
garyluu: Should be set to true initially.  Unlike Dockerfile, this time it is noticeable with the workflow descriptor file.  Appears to have no progress bar until after the first GA4GH endpoint or so is hit.  Then the progress bar shows up.  Easily replicate by blocking the first /files endpoint on Chrome and then reloading.  However...changing this will require sticking in more `loading = false` into more places.

garyluu: Edit: Oops, the blank screen on error is easily solved by using !content on the alert

coverbeck: I discussed this in my commit message, which I realize now is not obviously visible to PR reviewers: [here](https://github.com/dockstore/dockstore-ui2/pull/502/commits/359085d13e012519ee2cee22f1c338a2d93ca68c)

garyluu: image

When the files endpoint has failed

coverbeck:
@garyluu Tweaked this some more, using `setError`. Let me know what you think.

src/app/shared/ga4gh-files/ga4gh-files.query.ts
```diff
@@ -27,14 +27,23 @@ import { GA4GHFilesState, GA4GHFilesStore } from './ga4gh-files.store';
...
+   */
   getToolFiles(descriptorType: ToolDescriptor.TypeEnum, fileTypes: Array<ToolFile.FileTypeEnum>): Observable<Array<ToolFile>> {
     if (descriptorType) {
       let toolFiles$: Observable<Array<ToolFile>>;
-      toolFiles$ = this.selectEntity(descriptorType).pipe(
-        map((gA4GHFile: GA4GHFiles) => gA4GHFile ? gA4GHFile.toolFiles : [])
-      );
+      toolFiles$ = this.selectError().pipe(switchMap(
+        error => {
+          return error != null ? of(null) : this.selectEntity(descriptorType).pipe(
```
garyluu: use `observableOf` instead of `of`, then remove `of` from imports

coverbeck: Done with a new commit

PR #503, opened 2018-12-07
Fix pipe

garyluu:
ga4gh/dockstore#1955

Fixes the pipe but doesn't solve the mystery why null or undefined is passed into it.

### DIFF
--- a/src/app/search/map-friendly-values.pipe.ts
+++ b/src/app/search/map-friendly-values.pipe.ts
@@ -57,10 +57,13 @@ export class MapFriendlyValuesPipe implements PipeTransform {
    * @memberof MapFriendlyValuesPipe
    */
   transform(key: string, subBucket: string | number): string {
-    subBucket = subBucket.toString();
-    if (!subBucket) {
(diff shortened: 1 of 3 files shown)


src/app/search/map-friendly-values.pipe.ts
```diff
@@ -57,10 +57,14 @@ export class MapFriendlyValuesPipe implements PipeTransform {
...
-      return subBucket;
+    // Handle null or undefined
+    if (subBucket === null || subBucket === undefined) {
+      console.error('null/undefined passed into the pipe along with the key: ' + key);
+      return null;
     }
+    // Handle number
+    subBucket = subBucket.toString();
+    // Handle string
     if (this.friendlyValueNames.has(key) && this.friendlyValueNames.get(key).get(subBucket.toString())) {
```
coverbeck: Do you mean to effectively call `subBucket.toString().toString()` here and on the next line? Can't you just use subBucket directly now that you already called `toString()` on it?

PR #504, opened 2018-12-18
Feature/material cleanup

agduncan94:
* Consistent spacing within tabs for tools and workflows
* Fix issue with my tools/workflows sidebar cutting out card sides
* Cleanup of spacing for public tool/workflow view
* Other small fixes

Note: Did some auto-formatting so looks like I did more changes than I actually did.


src/app/container/container.component.html
```diff
@@ -21,7 +21,8 @@
       </button>
       <span class="glyphicon glyphicon-warning-sign">
       </span>
-      {{ missingContent.length === 1 ? 'Field that is missing from file: ' : 'Fields that are missing from file: '}} {{missingContent}}
+      {{ missingContent.length === 1 ? 'Field that is missing from file: ' : 'Fields that are missing from file: '}}
```
denis-yuen: Random thought: What is the recommended best practice for localization (not that its anywhere on our roadmap, but I am curious)

PR #509, opened 2019-01-02
Feature/1958/remove glyphicons

garyluu:
Replace all glyphicons with material icons and some fontawesome

ga4gh/dockstore#1958

Part of ga4gh/dockstore#1899

Also removed tokens.component.ts and html because it does not appear to be used.

Also fixed cytoscape popper error when re-registering plugin

Fixed some state issues


garyluu:
Fixes for small bugs that were discovered while removing glyphicons.  Converted some components to material to maintain the same style as material icons

src/app/container/view/view.component.ts
```diff
@@ -82,7 +82,7 @@ export class ViewContainerComponent extends View implements OnInit {
   ngOnInit() {
     this.isPublic$ = this.sessionQuery.isPublic$;
     this.toolQuery.tool$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(tool => {
-      this.tool = tool;
+      this.tool = JSON.parse(JSON.stringify(tool));
```
agduncan94: :open_mouth:

coverbeck: I see Andrew already commented, but what is the purpose of this?

garyluu: Akita and probably most state management tools makes everything immutable.  However, we need to allow the UI to make modifications (edit).  Spread operator and Object.assign both have problems either because it doesn't copy enough properties (no deep copy) or it does deep copy and copies the read-only attribute too.

PR #514, opened 2019-01-07
Feature/organisation view

garyluu:
PR 1 of 3

Simple organizations page (without the feature organizations section) that lists organizations and case-insensitive substring search

Uses reactive form

Missing integration tests (which will be added as the last PR after the 3 PRs are done)

There's a lot of placeholders until the first 3 PRs are merged and then hooked up


src/app/navbar/navbar.component.html
```diff
@@ -26,6 +26,9 @@
         <a mat-button class="icon" (click)="resetPageNumber()" routerLink="/workflows">
             <img class="site-icons-small hidden-sm" src="../assets/images/dockstore/dockstore-workflows-purple.png"> Workflows
         </a>
+        <a mat-button class="icon" (click)="resetPageNumber()" routerLink="/organizations">
+          <mat-icon>people</mat-icon> Organizations
```
coverbeck: Are you going with the US spelling for organi*z*ation? I prefer it, being from the US, but I noticed Andrew was using the British/Canadian spelling for the API, with an s. We should probably eventually be consistent.

:)

src/app/organizations/organizations/organizations.component.html
```diff
@@ -0,0 +1,61 @@
...
+            <mat-card-title>
+              <a [routerLink]="org.name">{{org.name}}</a>
+            </mat-card-title>
+            <mat-card-subtitle>{{ org.description }}</mat-card-subtitle>
+          </mat-card-header>
+          <mat-card-content>
+            <div fxLayout="row" fxLayoutGap="10px">
+              <span *ngIf="org?.email">
+                <mat-icon>email</mat-icon>
+                <a [href]="'mailto:' + org.email" target="_top">{{ org.email }}</a>
```
coverbeck: I wonder if we should redact the visible email, although that wouldn't really stop a bot. I haven't got there yet in the review, so maybe you already handle it, but if we're displaying a user's email, we should make clear in the UI that if the user specifies an email, it will be made public.

src/app/navbar/navbar.component.html
agduncan94: :canada:

denis-yuen: Looks like the UK uses both. I guess we'll go with the zed
https://writingexplained.org/organisation-vs-organization-difference

src/app/organizations/organizations/organizations.component.html
denis-yuen: I think this is the organization's email https://github.com/broadinstitute 
Worth a warning when creating the org, but probably not an issue for bots

src/app/navbar/navbar.component.html
coverbeck: Do we want to change the API to use the zed as well? Not sure if that was implied in your comment or not, or if it's necessary.

coverbeck: Do we want to change the API to use the zed as well? Not sure if that was implied in your comment or not, or if it's necessary.

garyluu: Organisation and Organization is both accepted in most places of the world.  Only Organization is accepted in the US.  I also prefer Organization.  If there is no upside to using Organisation, I'd rather use Organization.

src/app/organizations/organizations/organizations.component.html
garyluu: Hmm...I'll make that a material form field input hint in the 2nd PR.

PR #517, opened 2019-01-10
Feature/alert component

agduncan94:
Converts the bootstrap looking alerts to material.
Old alerts looked like this:
old-alert

New alerts look like this:
new-alert

### DIFF
--- a/src/app/container/descriptors/descriptors.component.html
+++ b/src/app/container/descriptors/descriptors.component.html
@@ -14,16 +14,16 @@
   ~    limitations under the License.
   -->
 <div *ngIf="!_selectedVersion" class="p-3">
-  <alert type="warning">
-    <mat-icon>warning</mat-icon>
-    &nbsp;No versions exist for this tool.
-  </alert>
+  <mat-card class="alert alert-warning" role="alert">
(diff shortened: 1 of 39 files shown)


src/app/container/descriptors/descriptors.component.html
```diff
@@ -56,9 +56,8 @@
       </mat-toolbar>
       <app-code-editor [(content)]="content" [filepath]="filePath"></app-code-editor>
     </div>
-    <alert type="warning" *ngIf="nullDescriptors">
-      <mat-icon>warning</mat-icon>
-      &nbsp;A Descriptor File associated with this Docker container could not be found.
-    </alert>
+    <mat-card class="alert alert-warning" role="alert" *ngIf="nullDescriptors">
+      <mat-icon>warning</mat-icon> A Descriptor File associated with this Docker container could not be found.
```
coverbeck: I realize it was like this already, but I think `Descriptor File` should be `descriptor file`.

agduncan94: I think i just did it for emphasis, but there is probably a better way to do that. Maybe similar to what github does (like in your comment)

PR #518, opened 2019-01-11
Feature/1934/phase out tools and workflows page

agduncan94:
* Removes tools and workflows links from navbar and home page
* Adds redirect from tools/workflows pages to search
* Removed tests that are no longer relevant
* Removed component that is no longer relevant

https://github.com/ga4gh/dockstore/issues/1934


cypress/integration/group2/ToolsSearch.ts
```diff
@@ -1,54 +0,0 @@
...
- *    limitations under the License.
- */
-import { resetDB, setTokenUserViewPort } from '../../support/commands';
-
-describe('Dockstore tool list page', () => {
-  resetDB();
-  setTokenUserViewPort();
-  describe('Select a tool', () => {
-    it('Should be able to go to the tools search page', () => {
-      cy.visit('/search-containers');
```
denis-yuen: Sorry, on second thought. 
Let's leave dockstore.org/search-containers and dockstore.org/tools working (as with their workflow equivalents). Its good to remove them from the navigation bar and the prominent browse buttons on the homepage but they can be useful for debugging if elastic search is buggy/down. 

Maybe bury the links under the homepage in "Adding to Dockstore" or even deeper in a sitemap and leave the tests for the pages themselves. 

FYI, I think the "search tool" button on the current homepage isn't working quite right and doesn't select the entry type when clicking through, please confirm and open an issue if so.

agduncan94: So you want me to readd the pages which have the tables for tools and workflows? Do you want me to remove the redirects too?

agduncan94: Also I checked and the search tool works.

PR #520, opened 2019-01-14
hide the scrollbar on mytools and myworkflows

agduncan94:
Hides the scrollbar on the my entries page since it is ugly, can still scroll.


coverbeck:
Would it be possible to have before and after screenshots?

agduncan94:
no-scrollbar
scrollbar

coverbeck:
Thanks for the screenshots!

In your no scrollbar screenshot, are there more orgs that are not visible? If so, how do users know they can scroll to view more?

agduncan94:
@coverbeck in the no scrollbar screenshot all of the orgs are shown, none are off screen.

PR #523, opened 2019-01-15
Update files in workflow component

garyluu:
For ga4gh/dockstore#2004 

Gets the GA4GH files in the workflow component instead of files component.  Results in:

- Going to Files tab is slightly faster
- Launch tab correctly has the wget command without needing to visit the Files tab beforehand
- There is no "Expression has changed after it was checked" error when switching from Launch tab to Files tab and back 
- Going to the workflow component (and not visiting files tab) now has slightly higher data usage

This also:
- Removes a redundant select version call (it is already done so in the subscription)

Tool component has still kept the previous flow because none of its tabs use the GA4GH Files (though that can also be changed if needed).

@rjbautis to check if this actually works for his PR. maybe cherry-pick it to avoid merge conflicts

### DIFF
--- a/src/app/workflow/workflow.component.ts
+++ b/src/app/workflow/workflow.component.ts
@@ -220,9 +220,6 @@ export class WorkflowComponent extends Entry {
       this.workflowsService.getPublishedWorkflowByPath(this.title)
         .subscribe(workflow => {
           this.workflowService.setWorkflow(workflow);
-          this.selectedVersion = this.selectVersion(this.workflow.workflowVersions, this.urlVersion,
-            this.workflow.defaultVersion);
-
           this.selectTab(this.validTabs.indexOf(this.currentTab));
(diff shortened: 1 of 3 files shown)


src/app/workflow/workflow.component.ts
```diff
@@ -194,6 +196,10 @@ export class WorkflowComponent extends Entry {
           this.setPublishMessage();
           this.selectedVersion = this.selectVersion(this.workflow.workflowVersions, this.urlVersion,
             this.workflow.defaultVersion);
+          if (this.selectedVersion) {
```
rjbautis: Just for my own understanding, the workflow component updates the Ga4gh Files entity store on initialization and each time a different workflow version is selected by the user?

garyluu: yes

PR #524, opened 2019-01-15
Feature/1973/remove checker workflows in search

garyluu:
Removing is_checker from search

ga4gh/dockstore#1973
Depends on ga4gh/dockstore#2006


denis-yuen:
Is this / does this need to be paired with a webservice change to not checker content in the index in the first place. For that matter, are these search options relevant anymore?

garyluu:
This PR does two things.
- Fix the issue ga4gh/dockstore#1973
- Remove the search options that will be invalidated by the webservice change

denis-yuen:
(pending build success)

PR #526, opened 2019-01-18
Feature/1314/star unstar endpoint

NatalieEO:
#1314

Users will see an error message if starring a workflow/tool fails for some reason. E.g. a server error, or having multiple tabs open at a time and trying to star a previously starred workflow.

### DIFF
--- a/cypress/integration/group3/starErrorMessage.ts
+++ b/cypress/integration/group3/starErrorMessage.ts
@@ -15,29 +15,114 @@
  */
 import { resetDB, setTokenUserViewPort } from '../../support/commands';
 
-describe('Workflow unstarring error message', () => {
+describe('Tool and workflow starring error messages', () => {
   resetDB();
   setTokenUserViewPort();
-  beforeEach(() => {
-    cy.visit('/workflows/github.com/A/l');
+
+  function starringError(url: string, type: string, routePath: string, name: string) {
+    cy.visit(url);
+    cy.server();
+    cy.route({
+      url:routePath,
+      method:'PUT',
+      status:400,
+      response: 'You cannot star the ' + type +' '+ name +' because you have not unstarred it.'
+    })
 
     cy
-      .server();
+      .get('#starringButtonIcon')
+      .click();
 
     cy
-      .route({
-        method: 'DELETE',
-        url: /workflows\/11\/unstar/,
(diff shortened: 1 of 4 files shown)


agduncan94:
Code looks good, but looks like one of the components is failing the command `ng test`. The error is in the starring component and is `NullInjectorError: No provider for MatSnackBar!`. Usually these types of errors come from the .spec.ts files not properly importing/declaring/etc a service or component.

denis-yuen:
Link is https://github.com/ga4gh/dockstore/issues/1314 (since this is cross-repo/org)

cypress/integration/group3/starErrorMessage.ts
```diff
@@ -0,0 +1,128 @@
+/*
+ *     Copyright 2018 OICR
```
denis-yuen: Probably need to update template for 2019

src/app/starring/starring.component.ts
```diff
@@ -117,8 +119,14 @@ export class StarringComponent implements OnInit, OnDestroy, OnChanges {
       this.setStar().subscribe(
```
garyluu: `this.alertService.start('Starring/unstarring tool/workflow')` right before the subscribe.  Otherwise the snackbar errors I see right now are: 
image

and if clicked again:
image

The alert service has a short description on how to use it

NatalieEO: I put the alertService.start in the file where setStar is defined in the new commit.I can put it in the starring.component.ts instead if you think it is an issue.

NatalieEO: Nevermind, moved it back to starring.component.ts. Realized having it where setStar is defined was causing tests in travis to fail. Not sure why, but should be working now.

src/app/starring/starring.component.ts
```diff
@@ -114,11 +116,28 @@ export class StarringComponent implements OnInit, OnDestroy, OnChanges {
   setStarring() {
     this.disable = true;
     if (this.isLoggedIn) {
+      let type;
+      type = this.entryType === 'workflows' ? type = 'workflow' : type = 'tool';
```
garyluu: `const type = this.entryType === 'workflows' ? 'workflow' : 'tool';`

PR #527, opened 2019-01-18
Feature/636/link my entry to public

agduncan94:
For workflows (see topright of photo):
view-public-workflow-button

https://github.com/ga4gh/dockstore/issues/636

### DIFF
--- a/cypress/integration/group3/myworkflows.ts
+++ b/cypress/integration/group3/myworkflows.ts
@@ -155,6 +155,13 @@ describe('Dockstore my workflows', () => {
         .should('contain', 'Publish')
         .click()
         .should('contain', 'Unpublish');
+
+      cy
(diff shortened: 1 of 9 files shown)


agduncan94:
If you figure out something better than `View Public` please comment here.

cypress/integration/group3/myworkflows.ts
```diff
@@ -155,6 +155,13 @@ describe('Dockstore my workflows', () => {
         .should('contain', 'Publish')
         .click()
         .should('contain', 'Unpublish');
+
+      cy
+        .get('#viewPublicWorkflowButton')
+        .should('be.visible')
```
garyluu: Include a `.should('not.be.visible')` somewhere

PR #529, opened 2019-01-24
Feature/approve org requests

agduncan94:
Adds a request tab to the accounts page, where:
* Admins can approve or reject organisations
* Users can see their pending organisations
* Users can accept or reject organisation invites

Currently needs a new release with https://github.com/ga4gh/dockstore/pull/2044

Top of page
top-requests

Requests
accounts-request-page


coverbeck:
Screenshot feedback:

1. I suggest saying `Decline` instead of `Reject` for invitations.
1. I feel the Reject button is too jarring and prominent. I would be inclined to give it less emphasis per [Material guidelines](https://material.io/design/components/buttons.html#hierarchy-placement). On the other hand, maybe Reject should be just as prominent, as the default action won't necessarily be to approve. Not sure... It just doesn't look quite right to me.

src/app/loginComponents/requests/organization-invite-confirm-dialog.html
```diff
@@ -0,0 +1,15 @@
+<h1 mat-dialog-title>{{data.approve ? 'Accept' : 'Decline'}} Organization</h1>
```
coverbeck: If this is for accepting an invite, I think it should say you're accepting/declining the invitation, not the organization. 

Also, if you were invited to more than one org, which is probably unlikely, it would be nice to know the org name you are accepting.

src/app/loginComponents/requests/organization-request-confirm-dialog.html
```diff
@@ -0,0 +1,15 @@
+<h1 mat-dialog-title>{{data.approve ? 'Approve' : 'Decline'}} Organization</h1>
```
coverbeck: I think you should say `reject` here. :)

Sorry I wasn't clearer in my earlier suggestion -- I think you should decline invites, but I think you should reject organizations. Because rejecting an org is a bigger deal, I think it needs a stronger word.

src/app/loginComponents/requests/requests.component.html
```diff
@@ -0,0 +1,148 @@
+<ng-template #loading>
+  <mat-progress-bar mode="indeterminate"></mat-progress-bar>
+</ng-template>
+
+<div *ngIf="!(isLoading$ | async); else loading">
+
+  <!-- Curator/Admin Pending Organizations -->
+  <div *ngIf="(isAdmin$ | async) || (isCurator$ | async)" class="p-3">
+    <h3>Curator Requests</h3>
+    <p>As a curator/admin, you can approve and decline organization requests.</p>
```
coverbeck: Suggest `reject` -- see previous comment; I won't flag any more if there are any

src/app/loginComponents/requests/organization-request-confirm-dialog.html
agduncan94: Oops I'll change those back.

PR #530, opened 2019-01-25
Feature/organizations events view

agduncan94:
This displays a subset of events for organisations, based on what I thought was important. Note that the design is in no way final. Once we have the core features implemented I'll design it better.

events

https://github.com/ga4gh/dockstore/issues/2002


src/app/organizations/state/events.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  constructor(private eventsStore: EventsStore,
+              private http: HttpClient, private alertService: AlertService,
+              private organisationsService: OrganisationsService) {
+  }
+
+  /**
+   * Updates the list of events for the current organisation
+   */
+  updateOrganizationEvents(id: number): void {
+    this.alertService.start('Getting organisation events');
```
garyluu: I probably should've specified in the docs for it.  The alertService is intended to used only from calls that result from a direct action of the user (registering someone, updating something, deleting something - all via buttons clicks).  It should be started only once at any given time.  In this case, updating/getting the events is "passive" - kind of like getting members of an organization, getting collections of an organization in the organization page.  If they all call the alertService, it will cause sync issues.  

Additionally, even though default sets loading to true, this function should manually set it to true in case it's called twice without reloading the state.

src/app/organizations/state/events.service.ts
```diff
@@ -0,0 +1,34 @@
...
+@Injectable({ providedIn: 'root' })
+export class EventsService {
+
+  constructor(private eventsStore: EventsStore,
+              private http: HttpClient, private alertService: AlertService,
+              private organisationsService: OrganisationsService) {
+  }
+
+  /**
+   * Updates the list of events for the current organisation
```
garyluu: Probably could use clarification on what "update" means.

PR #532, opened 2019-01-28
Feature/curator admin ui

NatalieEO:
ga4gh/dockstore#1965
Indicates to users whether they are an admin or curator in the dropdown and in their profile.
image
image
image


src/app/navbar/navbar.component.css
```diff
@@ -59,3 +59,7 @@ mat-menu {
 .icon-sm {
   width: 25%;
 }
+
+.material-icons.md-18 {
```
garyluu: where's this used?

src/app/loginComponents/accounts/internal/accounts.component.html
```diff
@@ -25,16 +25,22 @@ <h3>Dockstore Profile</h3>
...
+            <mat-icon>account_circle</mat-icon> <strong class="align-middle">Username:</strong>
           </h1>
           <p matLine id="account-username">{{user?.username}}</p>
         </mat-list-item>
         <mat-list-item>
           <h1 matLine>
-            <strong>Administrator:</strong>
+            <mat-icon>lock</mat-icon> <strong class="align-middle">Administrator:</strong>
           </h1>
           <p matLine id="account-is-admin">{{ user?.isAdmin ? 'Yes' : 'No' }}</p>
```
denis-yuen: A thought for discussion:
a normal user probably doesn't need to know that they aren't a administrator or that they aren't a curator. 
It only really needs to show up if the logged in user is actually one of the two.
So it can be simplified to just show the icon and the title if you are (kinda like a badge), and just hide it if they aren't.

src/app/navbar/navbar.component.html
```diff
@@ -38,12 +38,23 @@
...
+          <mat-icon class="hidden-lg" *ngIf="user?.isAdmin">lock</mat-icon>
+          <mat-icon class="hidden-lg" *ngIf="user?.curator">edit</mat-icon>
+          <mat-icon class="hidden-lg" *ngIf="!user?.isAdmin && !user?.curator">account_circle</mat-icon>
+          <span class="hidden-md hidden-sm">{{user?.username}}
+          </span>
           <mat-icon>keyboard_arrow_down</mat-icon>
         </button>
         <mat-menu #menuLarge="matMenu">
-          <a mat-menu-item id="dropdown-accounts" routerLink="/accounts">
+          <a mat-menu-item id="dropdown-accounts" routerLink="/accounts" *ngIf="user?.isAdmin">
```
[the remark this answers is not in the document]
denis-yuen: Agreed, display the more consequential role (administrator) in this case if a user happens to be both.

src/app/navbar/navbar.component.css
NatalieEO: It wasn't being used anymore. Deleted now.

src/app/loginComponents/accounts/internal/accounts.component.html
```diff
@@ -25,16 +25,25 @@ <h3>Dockstore Profile</h3>
...
           </h1>
-          <p matLine id="account-is-admin">{{ user?.isAdmin ? 'Yes' : 'No' }}</p>
+          <!--<p matLine id="account-is-admin">{{ user?.isAdmin ? 'Yes' : 'No' }}</p>-->
         </mat-list-item>
+        <!--<mat-list-item>-->
+          <!--<h1 matLine>-->
+            <!--<mat-icon >edit</mat-icon><strong class="align-middle">Curator:</strong>-->
+          <!--</h1>-->
+          <!--&lt;!&ndash;<p matLine id="account-is-curator">{{ user?.curator ? 'Yes' : 'No' }}</p>&ndash;&gt;-->
+        <!--</mat-list-item>-->
```
coverbeck: I would just get rid of the commented out lines unless you expect to restore them. If we really need to go back to them, they are in version control.

src/app/loginComponents/accounts/internal/accounts.component.html
```diff
@@ -25,16 +25,25 @@ <h3>Dockstore Profile</h3>
       <mat-list>
         <mat-list-item>
           <h1 matLine>
-            <strong>Username:</strong>
+            <mat-icon>account_circle</mat-icon> <strong class="align-middle">Username:</strong>
```
coverbeck: This is not specific to this PR, but I don't think we should have a colon here, especially with the name below being inside a `<p>`, which puts it on a separate line. I think we should follow [this guideline](https://ux.stackexchange.com/questions/3611/should-label-and-field-be-separated-with-colon), where we either have a colon and don't visually distinguish the label, or distinguish the label and don't have the colon.

Since there are several panels (GitHub and Google) with this pattern, so if the OICR folks agree(?), could you log this as a separate issue?

NatalieEO:
screen shot 2019-01-29 at 4 27 55 pm

src/app/loginComponents/accounts/internal/accounts.component.html
denis-yuen: I agree, I like the second option (distinguish the label and don't have the colon)

NatalieEO: Yeah, I like the second option more as well.

NatalieEO: ga4gh/dockstore#2055

PR #533, opened 2019-01-29
Feature/pretty organizations

agduncan94:
Updates the organisations page to look nicer. Also adds very basic collection page. Note that both pages are still in their infancy and will require more modifications. It is being split up into multiple PRs to make it easier to manage and review.

Currently retrieving collection by id, but in future update will use either id or name, similar to what was done with orgs.

Organisation Page
prettyorg

Collection Page
prettycollection

### DIFF
--- a/src/app/organizations/collection/collection.component.html
+++ b/src/app/organizations/collection/collection.component.html
@@ -15,7 +15,7 @@
 -->
 <app-header>
   <mat-icon>people</mat-icon>
-  Available Collections
+  Collections
 </app-header>
 <div *ngIf="(loading$ | async); else placeholder">
   <mat-progress-bar mode="indeterminate"></mat-progress-bar>
--- a/src/app/organizations/collection/collection.component.html
+++ b/src/app/organizations/collection/collection.component.html
@@ -34,6 +34,10 @@
       </mat-card-header>
     </mat-card>
 
+    <mat-card class="alert alert-info" *ngIf="collection?.entries?.length == 0">
+      This collection has no associated entries
+    </mat-card>
+
     <div fxLayoutGap="16px grid" fxLayout="row wrap">
       <div *ngFor="let entry of collection?.entries" fxFlex="{{ 100/3 }}%">
         <mat-card class="entryCard">
--- a/src/app/organizations/collection/collection.component.html
+++ b/src/app/organizations/collection/collection.component.html
@@ -1,0 +1,75 @@
+<!--
+   Copyright 2019 OICR
+ *
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+ *
+       http://www.apache.org/licenses/LICENSE-2.0
+ *
+   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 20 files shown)


coverbeck:
Looking good!

Initial thoughts from screenshots alone:

1. I would display the organization on the individual collection page. I would maybe add the org logo, and have some thing like "<org logo> <org name> Collections".
1. I would not say "Available" in "Available Collections". Or is that a curator view where some some collections can be hidden?
1. I lean towards not displaying events unless you are a member, or maybe even a curator, although I could see it both ways.

src/app/organizations/organization/organization.component.html
```diff
@@ -23,46 +23,62 @@
...
-        <span fxLayout="row" fxLayoutGap="10px" *ngIf="org?.location">
-          <mat-icon>location_on</mat-icon>
-          <span class="ellipsis-lines">{{ org.location }}</span>
-        </span>
-        <p>Markdown description placeholder</p>
+  <div class="container" *ngIf="(organization$ | async) as org" fxLayout="column" fxLayoutGap="10px">
+    <mat-card fxFlex class="my-3">
+      <div fxLayout="row" fxLayoutGap="10px">
+        <div>
+          <img src="https://via.placeholder.com/150" alt="Logo" class="img-circle">
```
denis-yuen: Might be a good github issue to split out and get a gravatar for this.

src/app/organizations/collection/collection.component.html
```diff
@@ -0,0 +1,79 @@
...
+          <mat-card-header *ngIf="entry?.tool_path">
+            <mat-card-title>
+              <h4>
+                <a [routerLink]="['/tools/', entry.tool_path]">
+                  {{entry.tool_path}}
+                </a>
+              </h4>
+            </mat-card-title>
+            <mat-card-subtitle>
+              Last updated {{ entry.dbUpdateDate | date }}
```
garyluu: This is identical to the workflow card.  If the only change between cards is the title, just *ngIf the anchor instead of the entire header

src/app/organizations/events/events.component.html
```diff
@@ -1,53 +1,52 @@
...
-            </mat-chip-list>
-          </mat-list-item>
+  <mat-card>
+    <mat-card-header>
+      <mat-card-title>
+        Recent Events
+      </mat-card-title>
+    </mat-card-header>
+    <div fxLayout="column wrap" fxLayoutGap="20px">
+      <span *ngFor="let event of (events$ | async)" class="px-3">
```
garyluu: fxflexoffset instead of px-3

agduncan94:
@coverbeck 
1) Was going to add this in my next PR. Originally I was just gonna have a simple link, but I do like the idea of having the related org more prominent.
2) This was a copy issue from copying the organisations component, I'll change it to just collections for now.
3) Ya I agree I could be swayed either way. For now though they can just always be displayed. The events portion still needs some more designing, to include things like date and filtering what events are shown.

src/app/organizations/organization/organization.component.html
agduncan94: https://github.com/ga4gh/dockstore/issues/2052

PR #536, opened 2019-01-31
Feature/2069/no search text

garyluu:
For ga4gh/dockstore#2069

Fixes the value not being used in the search.  Cause of this.  The store has a default value of empty string which is overwriting the value in gotten from the parseParam() function.  Value is now extracted and tossed into the store.

Also cleaned up a few unused parameters, imports, variables

### DIFF
--- a/cypress/integration/group2/elasticsearch.ts
+++ b/cypress/integration/group2/elasticsearch.ts
@@ -1,0 +1,25 @@
+/*
+ *    Copyright 2019 OICR
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


cypress/integration/group2/elasticsearch.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+import { setTokenUserViewPort } from '../../support/commands';
+
+describe('Variations of URL', () => {
```
agduncan94: :open_mouth:

PR #540, opened 2019-02-06
Feature/graphic stars

NatalieEO:
ga4gh/dockstore#2060
Reordered columns in workflow/tool tables and added a star icon
image


cypress/integration/group1/app-spec.ts
```diff
@@ -69,4 +69,19 @@ describe('Dockstore Home', () => {
       cy.get('#youtubeModal').should('not.be.visible');
     });
   });
+
+  describe('Star Count', () => {
+    it.only('Tool Star Count', () => {
```
garyluu: `it.only` :cry:

NatalieEO: Oof, my bad. Gone now.

NatalieEO:
> List components shows 0, search shows empty. Is that intended?

Not sure what you mean by search shows empty? But the zero was intended if nobody has starred it yet.

NatalieEO:
> Going by the screenshot, maybe we omit the graphic and the zero if there are no stars.
> Ala current https://dockstore.org/tools or workflows

Sure I can do that instead.

garyluu:
Take a screenshot of the search page and home page entry listing component.

coverbeck:
As discussed with Cricket, suggest making the star smaller.

NatalieEO:
Still need to resize the stars on the homepage but here's what it looks like now.
image

coverbeck:
I think the size looks much better. I wonder if it would be better to have a filled in (solid) star though.

src/app/search/search-workflow-table/search-workflow-table.component.scss
```diff
@@ -0,0 +1,10 @@
+@mixin change-mat-icon-size($size: 24px) {
+  font-size: $size;
+  height: $size;
+  width: $size;
+  line-height: $size;
```
garyluu: Are changing the height, width, and line-height necessary?  As opposed to https://stackoverflow.com/a/36562672/4484759

src/app/search/search-tool-table/search-tool-table.component.scss
```diff
@@ -0,0 +1,10 @@
+@mixin change-mat-icon-size($size: 24px) {
+  font-size: $size;
+  height: $size;
+  width: $size;
+  line-height: $size;
+}
+
+.mat-icon {
```
garyluu: duplicate of the other scss. rename class, throw it in somewhere common

src/app/search/search-workflow-table/search-workflow-table.component.scss
NatalieEO: The height does seem to be necessary. Otherwise it's very off like this 
image
but you're right that the other ones don't make a difference.

agduncan94: You might be able to use flex layout for this. Have the number and star as elements in a div. The div would be fxLayout="row" with maybe  fxLayoutAlign="center center". Might need to experiment with it.

NatalieEO:
@coverbeck With it filled, it looks like this 
image

coverbeck:
I prefer the solid star as in your latest screenshot. What do others think?

src/app/search/search-tool-table/search-tool-table.component.html
```diff
@@ -57,6 +53,12 @@
         </a>
       </mat-cell>
     </ng-container>
+    <ng-container matColumnDef="starredUsers">
+      <mat-header-cell *matHeaderCellDef mat-sort-header>Stars</mat-header-cell>
+      <mat-cell class="description-cell" *matCellDef="let tool">{{ !tool?.starredUsers || tool?.starredUsers.length === 0 ? '' : tool?.starredUsers.length }}
+        <mat-icon class = "star" *ngIf="tool?.starredUsers?.length > 0 " >star_border</mat-icon>
```
coverbeck: Although your HTML is syntactically valid, I don't think there are spaces around the equals sign of attributes anywhere else, so I would remove them (`class="star"` instead of `class = "star")`

src/app/search/search-workflow-table/search-workflow-table.component.html
```diff
@@ -34,6 +30,12 @@
         </a>
       </mat-cell>
     </ng-container>
+    <ng-container matColumnDef="starredUsers">
+      <mat-header-cell *matHeaderCellDef mat-sort-header>Stars</mat-header-cell>
+      <mat-cell class="description-cell" *matCellDef="let workflow: Workflow">{{ !workflow?.starredUsers || workflow?.starredUsers.length === 0 ? '' : workflow?.starredUsers?.length }}
+        <mat-icon class = "star" *ngIf="workflow?.starredUsers?.length > 0 ">star_border</mat-icon>
```
coverbeck: Remove spaces around equals sign

NatalieEO:
@garyluu @denis-yuen @agduncan94 
I think I need to skip the test file I added that checks for the stars on the search page. Charles helped me look at what could be going wrong since it always passed for me locally. Search results do not appear when running on travis (elastic search is never run?), so my test is looking for something that won't ever be there. Here's the screenshot from the travis run.
image

agduncan94:
@NatalieEO You can probably use HTTP mock requests so that the data doesn't actually need to be in elastic search, you just mock the response. There are some examples in some of the test files. I have not tried this for ES before so I am not sure if it is so simple. But I think if you have the test but comment it out and say 'enable once we have elastic search running on travis' it should be ok.

I'm not sure if there is a reason why we don't run elastic search for the UI tests, but maybe that is something we should do in the future since search is an important component of the site that should be tested well. Though this is probably out of scope for this PR and I think mocking will be fine.

PR #543, opened 2019-02-11
Feature/wdl pipeline builder

rjbautis:
ga4gh/dockstore#1557

Epam Pipeline Builder for WDL workflow visualizations


src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+      });
+  }
+
+  ngOnDestroy() {
+    this.ngUnsubscribe.next();
+    this.ngUnsubscribe.complete();
+  }
+
+  onVersionChange(value) {
```
agduncan94: Add type

src/app/workflow/dag/dag.component.html
```diff
@@ -32,6 +32,14 @@
             <input type="radio" name="dagType" [(ngModel)]="dagType" value="cwlviewer">Common Workflow Language Viewer
           </label>
         </div>
+        <div *ngIf="enableWdlViewer && (descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL && (workflow$ | async)?.source_control_provider === 'GITHUB'" style="line-height: 20px;">
```
coverbeck: 1. Per my earlier comment, I don't think you need enableWdlViewer.
2. You have a `$isWDL` in workflow.query.ts -- is there a way you can just use that here (haven't looked at the code too closely yet)?
3. You shouldn't limit it to GitHub provider; you shouldn't care. It is like this for CWL viewer, because that external service only works with GitHub.

src/app/workflow/dag/dag.component.html
```diff
@@ -32,6 +32,14 @@
...
+          <label class="radio-inline">
+            <input type="radio" name="dagType" [(ngModel)]="dagType" value="classic">Classic
+          </label>
+          <label class="radio-inline">
+            <input type="radio" name="dagType" [(ngModel)]="dagType" value="wdlviewer">EPAM Pipeline Builder
```
coverbeck: I wonder if we should just call it `Pipeline Builder`. Even that is going to be confusing -- I'm not sure that the casual user will know what it is.

src/app/workflow/dag/dag.component.html
```diff
@@ -41,15 +49,18 @@
               <mat-icon id="resize-full-button" *ngIf="!expanded">fullscreen</mat-icon>
               <mat-icon id="resize-small-button" *ngIf="expanded">fullscreen_exit</mat-icon>
             </button>
-            <button id="dag_reload" (click)="reset()" class="dag-button btn btn-default" matTooltip="Reset">
+            <button *ngIf="dagType !== 'wdlviewer'" id="dag_reload" (click)="reset()" class="dag-button btn btn-default" matTooltip="Reset">
+              <mat-icon>control_camera</mat-icon>
+            </button>
+            <button *ngIf="dagType === 'wdlviewer' && enableWdlViewer" (click)="resetPipelineBuilder()" class="dag-button btn btn-default" matTooltip="Reset">
```
coverbeck: Last comment on `enableWdlViewer`. :) If there are other usages, your removing it from dockstore.model.ts should cause compile errors.

src/app/workflow/dag/dag.component.html
```diff
@@ -70,6 +81,9 @@
         [hidden]="dagType !== 'cwlviewer'" [workflow]="(workflow$ | async)" [selectedVersion]="selectedVersion"
         [refresh]="refreshCounter" [expanded]="expanded">
       </app-cwl-viewer>
+      <app-wdl-viewer *ngIf="enableWdlViewer && (descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL && (workflow$ | async)?.source_control_provider === 'GITHUB' && dagType === 'wdlviewer'"
```
coverbeck: Same comment about possible using `isWDL$`

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.scss
```diff
@@ -0,0 +1,22 @@
+@import "~pipeline-builder/dist/pipeline.css";
+
+.svg {
+  overflow: hidden;
+}
+
+.fullscreen-svg-element {
+  width: 100% !important;
```
coverbeck: `!important` is not ideal. I know you struggled to get it working otherwise, so maybe it's OK, but ideally these would not be necessary.

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts
```diff
@@ -0,0 +1,111 @@
...
+    this.loading = true;
+  }
+
+  ngAfterViewInit() {
+    this.visualizer = new pipeline.Visualizer(this.diagram.nativeElement, false);
+
+    // Retrieve all files for this workflow from Ga4ghFiles entity Store
+    this.wdlViewerService.getFiles(ToolDescriptor.TypeEnum.WDL).pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe(files => {
+        if (files.length > 0) {
```
coverbeck: If there is an error retrieving files, `files` will be `null` -- please test for it.

src/app/workflow/dag/dag.component.html
garyluu: Just have one button and one reset function, do the checks within a .ts file

src/app/workflow/dag/wdl-viewer/wdl-viewer.service.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Service for creating WDL workflow visualizations with EPAM Pipeline Builder library
+ */
+@Injectable()
+export class WdlViewerService {
+  private zip: JSZip = new JSZip();
+  constructor(private gA4GHFilesQuery: GA4GHFilesQuery, protected workflowsService: WorkflowsService) {
+  }
+
+  getFiles(descriptorType: ToolDescriptor.TypeEnum): Observable<Array<ToolFile>> {
+    return this.gA4GHFilesQuery.getToolFiles(descriptorType, [ToolFile.FileTypeEnum.PRIMARYDESCRIPTOR,
```
coverbeck: Are all the files guaranteed to be returned here? Not sure, just asking.

src/app/workflow/dag/wdl-viewer/wdl-viewer.service.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * Driver function for creating workflow visualizations
+   *
+   * @param files
+   * @param workflow
+   * @param version
+   */
+  create(files: Array<ToolFile>, workflow: ExtendedWorkflow, version: WorkflowVersion): Observable<any> {
```
coverbeck: Would be good not to have <any> here and the functions that are called.

src/app/workflow/workflow.component.html
```diff
@@ -265,3 +265,4 @@ <h3>DOI</h3>
       </div>
     </div> -->
   </div>
+</div>
```
coverbeck: Was this missing a closing `</div>` all along?

src/app/workflow/dag/wdl-viewer/wdl-viewer.service.ts
```diff
@@ -0,0 +1,87 @@
...
+   */
+  createMultiple(workflow: ExtendedWorkflow, version: WorkflowVersion): Observable<any> {
+    return forkJoin(this.workflowsService.wdl(workflow.id, version.name), this.workflowsService.secondaryWdl(workflow.id, version.name))
+      .pipe(
+        switchMap(res => {
+          // Store each secondary file in a zip object
+          res[1].forEach(file => this.zip.file(file.path, file.content));
+
+          return from(this.zip.generateAsync({type: 'blob'})).pipe(switchMap(zip => {
+            return from(pipeline.parse(res[0].content, {zipFile: zip}));
```
coverbeck: What happens if there is an error inside `parse()` method? Invalid WDL, cannot fetch HTTP imports because of CORS, etc.

src/app/workflow/dag/wdl-viewer/wdl-viewer.service.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Creates WDL workflow visualization for single-file workflows
+   *
+   * @param workflow
+   * @param version
+   */
+  createSingle(workflow: ExtendedWorkflow, version: WorkflowVersion): Observable<any> {
+    return this.workflowsService.wdl(workflow.id, version.name).pipe(switchMap(prim => {
+      return from(pipeline.parse(prim.content));
```
coverbeck: What happens if there is an error?

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.scss
```diff
@@ -0,0 +1,22 @@
...
+
+.mini-svg-dag {
+  width: 100% !important;
+  height: 500px !important;
+}
+
+.large-svg-dag {
+  height: 90% !important;
+  border: 1px solid #ccc !important;
+}
```
garyluu: may have overkilled it on the borders somewhere

src/app/workflow/workflow.component.html
garyluu: looks like it

src/app/workflow/dag/dag.component.html
rjbautis: I'm comfortable with renaming it, since the service being used is an open-source library. Thoughts?

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.scss
rjbautis: That's a good point, but I'm not sure how to address the issue of inline styling without `!important`. The div's dimensions are getting overridden by inline style

src/app/workflow/dag/wdl-viewer/wdl-viewer.service.ts
rjbautis: If there is an error, `parse()` throws a `Promise.rejection` with message. The message is displayed if that's the case.

coverbeck: Where is it caught?

Since there are no types for this code where this could be gleaned from, I think it would be good to have a comment regarding this behavior.

rjbautis: Sounds good. The error is caught by the Observable subscribing to this function, which in this case is inside the `wdl-viewer.component` class.

src/app/shared/dockstore.model.ts
```diff
@@ -61,6 +61,6 @@ export class Dockstore {
   static readonly CWL_VISUALIZER_URI = 'https://view.commonwl.org';
 
   static readonly FEATURES = {
-    enableCwlViewer: true
+    enableCwlViewer: true,
```
coverbeck: Trailing comma

src/app/workflow/dag/dag.component.html
```diff
@@ -32,6 +32,14 @@
             <input type="radio" name="dagType" [(ngModel)]="dagType" value="cwlviewer">Common Workflow Language Viewer
           </label>
         </div>
+        <div *ngIf="(isWDL$ | async) && (descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL" style="line-height: 20px;">
```
coverbeck: Isn't `(descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL` redundant? It looks like that is what `isWDL$` is already checking.

src/app/workflow/dag/dag.component.html
```diff
@@ -70,6 +78,9 @@
         [hidden]="dagType !== 'cwlviewer'" [workflow]="(workflow$ | async)" [selectedVersion]="selectedVersion"
         [refresh]="refreshCounter" [expanded]="expanded">
       </app-cwl-viewer>
+      <app-wdl-viewer *ngIf="(isWDL$ | async) && (descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL && (workflow$ | async)?.source_control_provider === 'GITHUB' && dagType === 'wdlviewer'"
```
coverbeck: Same as earlier comment -- isn't the second part of your and condition redundant?

src/app/workflow/dag/dag.component.html
```diff
@@ -49,7 +57,7 @@
       </div>
 
       <div *ngIf="dagType === 'classic' && !(isNFL$ | async)">
-        <mat-card class="alert alert-warning" role="alert" *ngIf="!(dagResult$ | async) && !selectedVersion?.valid">
+        <mat-card class="alert alert-warning" role="alert" *ngIf="!(dagResult$ | async) || !selectedVersion?.valid">
```
coverbeck: Why did you change this? Not saying it's wrong, but please explain.

 It was a sort of confusing logic (I think I wrote it), but I think the idea was that it would only display the invalid descriptor message if the selected version was not valid. The way you've changed it, I believe it will now say a descriptor could not be found, even if there is a descriptor -- sometimes there are errors generating the DAG that are unrelated to the version not being valid.

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      });
+  }
+
+  ngOnDestroy() {
+    this.ngUnsubscribe.next();
+    this.ngUnsubscribe.complete();
+  }
+
+  onVersionChange(value: WorkflowVersion) {
+    if (this.visualizer) {
```
coverbeck: Do you need a `this.loading = true` here? It doesn't seem like you'll get the progress bar when you change the version.

src/app/workflow/dag/wdl-viewer/wdl-viewer.service.ts
coverbeck: Can you please verify and respond to my previous question?

src/app/workflow/dag/wdl-viewer/wdl-viewer.html
```diff
@@ -0,0 +1,15 @@
+<div [ngClass]="{'fullscreen-svg-element': expanded}" class="wdl-viewer">
+  <mat-card *ngIf="wdlViewerError" class="alert alert-warning">
+    <mat-icon>warning</mat-icon>&nbsp;
+    <div>EPAM Pipeline Builder was unable to visualize this workflow. <span *ngIf="errorMessage">It returned the following error:</span></div>
```
coverbeck: You call it EPAM WDL Visualizer above (I do prefer Visualizer)

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.scss
coverbeck: OK, we can always revisit if it becomes an issue.

src/app/workflow/dag/wdl-viewer/wdl-viewer.service.ts
rjbautis: No, the function does not provide a guarantee. It returns an empty Observable array if there are no files in the store.

Currently, the tool files are placed into the store on initialization of the workflow component. I'm using this function under the assumption that the tool files have been/will be stored beforehand. I also provide a null & empty check when subscribing to this observable as an edge case

src/app/workflow/dag/dag.component.html
rjbautis: I misinterpreted this! My understanding was that an error message should display if at _least_ one check failed. I'll revert back

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts
rjbautis: Actually, I added that when the change in the `selectedVersion` variable is detected

https://github.com/dockstore/dockstore-ui2/blob/2b6135cd3a1ea2a09801e39290b34b1e94c57898/src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts#L41-L44

coverbeck: 👍

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.scss
```diff
@@ -0,0 +1,16 @@
+@import "~pipeline-builder/dist/pipeline.css";
+
+.svg {
```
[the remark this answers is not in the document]
rjbautis: There is a slight difference in the width and height when using the svg classes. But I will opt to change the classes to the classes in `bootstrap4.scss` for uniformity

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts
```diff
@@ -0,0 +1,124 @@
...
+})
+export class WdlViewerComponent implements AfterViewInit, OnInit, OnDestroy {
+
+  @Input() workflow: ExtendedWorkflow;
+  @Input() expanded: boolean;
+  @Input() set selectedVersion(value: WorkflowVersion) {
+    this.loading = true;
+    this.onVersionChange(value);
+  }
+  @Output() success: EventEmitter<boolean> = new EventEmitter<boolean>(true);
```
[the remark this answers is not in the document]
rjbautis: Comment noted in the code and issue created ga4gh/dockstore#2141

PR #544, opened 2019-02-12
Feature/title tag

NatalieEO:
ga4gh/dockstore#1990
Made title tags throughout site more specific. Ex: Dockstore | My Workflows


src/app/app.component.ts
```diff
@@ -1,11 +1,33 @@
...
   styleUrls: ['./app.component.css']
 })
-export class AppComponent {
-  constructor() {
+export class AppComponent implements OnInit {
+  constructor(
+    private router: Router,
+    private activatedRoute: ActivatedRoute,
+    private titleService: Title
+  ) {}
```
coverbeck: I think the general pattern is to have a newline after the opening brace, although it looks like it is not enforced in the linter.

src/app/app.component.ts
```diff
@@ -1,11 +1,33 @@
...
+  ngOnInit() {
```
coverbeck: There is typically a blank line between methods, although this also doesn't look like it is enforced in the linter. :)

src/app/app.component.ts
```diff
@@ -1,11 +1,33 @@
...
+    this.router.events.pipe(
+      filter(event => event instanceof NavigationEnd),
+      map(() => this.activatedRoute),
+      map((route) => {
+        while (route.firstChild) {
+          route = route.firstChild;
+        }
+
```
coverbeck: Suggest getting rid of blank line.

src/app/app.component.ts
coverbeck: Ignore this; seems like we are not consistent after all.

src/app/app.component.ts
```diff
@@ -1,11 +1,33 @@
...
+    private titleService: Title
+  ) {}
+
+  ngOnInit() {
+    this.router.events.pipe(
+      filter(event => event instanceof NavigationEnd),
+      map(() => this.activatedRoute),
+      map((route) => {
+        while (route.firstChild) {
+          route = route.firstChild;
```
garyluu: Reason for all this? `mergeMap(() => this.activatedRoute.firstChild.data)` should've worked on NavigationEnd

src/app/app.component.ts
```diff
@@ -1,11 +1,33 @@
...
+        }
+        return route;
+      }),
+      mergeMap((route) => route.data),
+    ).subscribe((event) => this.titleService.setTitle(event['title']));
```
garyluu: I know this is the appComponent but unsubscribe anyways...just in case

src/app/app.component.ts
NatalieEO: I'm unsure what the reasoning for all of it is. I looked up several examples of changing the title based on routes, and they all looked similar to this. But I think the while loop lets you get more specific title tags based on the children routes written in the other folders. So with your changes, going to /containers page will have a title of "Dockstore | Tools" but going to a specific tool will also have the same title instead of "Tools" becoming singular.

src/app/app.component.ts
NatalieEO: Added unsubscribe in latest commit.

src/app/app.component.ts
NatalieEO: The next commit does combine the two maps into one now though.

PR #547, opened 2019-02-21
Feature/version change fix

garyluu:
1. Fixes the version change not updating the files (a better fix is to put the version in the state).
2. Fixes the parameters being swapped again for add workflow test parameter files and added a test for it (previously only the tool one was fixed and tested)


denis-yuen:
@garyluu is there a ticket for this (for changelog writing purposes)

garyluu:
no ticket for this, wasn't broken in production

PR #550, opened 2019-02-25
Semaphore CI setup

garyluu:
For my personal use and those who wish to fork repo and test separate from Dockstore.  Shouldn't affect the rest of the team.

Also attempts ga4gh/dockstore#2145 (a better fix would be implement progress bar when switching between tools/workflows and have Cypress check that there's no progress bar before proceeding with other actions)

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -14,7 +14,7 @@ blocks:
       - name: Lint
         commands:
           - checkout
-          - /scripts/install-travis.sh
+          - /scripts/install-travis-script.sh
           - ng lint
           - ng test --watch=false --code-coverage --browsers ChromeHeadless
           - echo "done"
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -14,7 +14,7 @@ blocks:
           value: 1.6.0-alpha.6
       prologue:
         commands:
-          - nvm use 10.10.0
+          - sem-version node 10.13.0
           - checkout
           - npm install -g @angular/cli@6.0.2 --unsafe
           - npm ci - prefer-offline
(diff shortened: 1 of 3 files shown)


.semaphore/semaphore.yml
```diff
@@ -0,0 +1,72 @@
...
+  - name: "Build"
+    task:
+      env_vars:
+        - name: APP_ENV
+          value: prod
+        - name: WEBSERVICE_VERSION
+          value: 1.6.0-alpha.6
+      prologue:
+        commands:
+          - nvm use 10.10.0
```
coverbeck: Just curious, why is this different than .travis.yml (10.13.0)?

.semaphore/semaphore.yml
```diff
@@ -0,0 +1,72 @@
...
+  - name: "Integration Tests"
+    task:
+      env_vars:
+        - name: WEBSERVICE_VERSION
+          value: 1.6.0-alpha.6
+      secrets:
+        - name: cypress
+      prologue:
+        commands:
+          - nvm use 10.10.0
```
coverbeck: Same as previous comment

.semaphore/semaphore.yml
garyluu: Closest version that Semaphore has

.semaphore/semaphore.yml
garyluu: Same as previous comment

.semaphore/semaphore.yml
garyluu: Nevermind, was reading old docs https://semaphoreci.com/docs/languages/javascript/javascript-support-on-semaphore.html vs https://docs.semaphoreci.com/article/82-language-javascript-and-nodejs

PR #551, opened 2019-02-26
Feature/2052/organization logo

NatalieEO:
ga4gh/dockstore#2052
Users can add/update organization logo image. Any url is okay as long as it ends in .jpeg, .jpg, or png. The url is filtered through gravatar and a "gravatarized" url is stored in the database and used for display on the page. User will just see their regular url in the form.


src/app/organizations/organization/organization.component.ts
```diff
@@ -35,7 +35,8 @@ export class OrganizationComponent implements OnInit {
   organization$: Observable<Organization>;
   loading$: Observable<boolean>;
   canEdit$: Observable<boolean>;
-  constructor(private organizationQuery: OrganizationQuery, private organizationService: OrganizationService, private matDialog: MatDialog
+  constructor(private organizationQuery: OrganizationQuery, private organizationService: OrganizationService, private matDialog: MatDialog,
+              // private registerOrganizationService: RegisterOrganizationService
```
coverbeck: Remove this

src/app/organizations/organization/organization.component.ts
```diff
@@ -66,4 +67,17 @@ export class OrganizationComponent implements OnInit {
...
+  /**
+   * The database stores a gravatar link if the image was added using UI, but this is not the case if added
+   * via the API.
+   */
+  genGravatarUrl(url: string): Observable<string> {
+    if (url.match('(https://www.gravatar.com/avatar/000)(.*)')) {
+      return of(url);
+    } else {
+      return of(url);
+      // return of(this.registerOrganizationService.gravatarUrl(url));
```
coverbeck: Restore this

src/app/organizations/registerOrganization/register-organization.component.html
```diff
@@ -46,6 +46,11 @@ <h1 mat-dialog-title>{{title}}</h1>
       <mat-error *ngIf="contactEmail.hasError('email')">Please enter a valid contact email address</mat-error>
       <mat-hint>Once approved, this email address will be publicly visible</mat-hint>
     </mat-form-field>
+    <mat-form-field>
+      <input type="text" matInput formControlName="avatarUrl" placeholder="Image URL">
+      <mat-error *ngIf="avatarUrl.hasError('pattern')">Link must end in a recognizable extension (jpg, jpeg, png)</mat-error>
```
coverbeck: recognizable image extension?

src/app/organizations/registerOrganization/register-organization.component.html
```diff
@@ -46,6 +46,11 @@ <h1 mat-dialog-title>{{title}}</h1>
...
+      <mat-hint>Link to your organization's logo image. Must have a recognizable extension (jpg, jpeg, png)</mat-hint>
```
coverbeck: Ditto

src/app/organizations/state/register-organization.service.ts
```diff
@@ -198,4 +210,30 @@ export class RegisterOrganizationService {
       });
     }
   }
+
+  gravatarUrl(imgLink: string): string {
```
coverbeck: Are you saving the gravatar url in the API? I think you should save the simple url, then prepend it with the gravatar url when displaying it.

src/app/organizations/state/register-organization.service.ts
```diff
@@ -198,4 +210,30 @@ export class RegisterOrganizationService {
...
+    } else {
+      return null;
+    }
+  }
+
+  /**
+  * User can change the image url via the API (only has extension constraints), so not every url string stored in the database will be a
+  * gravatar one.
+  * */
+  extractLogoUrl(imageUrl: string): string {
```
coverbeck: See previous comment.

src/app/organizations/registerOrganization/register-organization.component.html
NatalieEO: reworded in new commit

src/app/organizations/state/register-organization.service.ts
NatalieEO: No, the url someone enters through the api is what gets saved. Not the gravatar version.

coverbeck: What does line 192 do? It looks like it's calling this method which will transform the url.

NatalieEO: yes, it makes the url into a gravatar one before saving it into the database.

coverbeck: I think you should save it without adding the gravatar portion. You already need to handle the case where there's no Gravatar url, if somebody already entered one directly via the API, so I think it makes your life simpler to just store "plain" urls.

NatalieEO: Yes, it's much simpler now doing that.

src/app/organizations/state/register-organization.service.ts
```diff
@@ -49,6 +50,12 @@ export class RegisterOrganizationService {
   // The old regex in case needed
   // readonly urlRegex = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,63}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/;
   readonly organizationNameRegex = /^[a-zA-Z][a-zA-Z\d]*$/;
+
+  readonly logoUrlRegex = new RegExp(
```
garyluu: Need a unit test for this regex, see src/app/organizations/state/register-organization.service.spec.ts for similar test.

Missing a $ at the end `.gif)$` otherwise `.png potato` works.  Also check that `.png` does not work

src/app/organizations/organization/organization.component.ts
```diff
@@ -66,4 +66,17 @@ export class OrganizationComponent implements OnInit {
       data: { description: description }, width: '600px'
     });
   }
+
+  /**
+   * The database stores a gravatar link if the image was added using UI, but this is not the case if added
+   * via the API.
+   */
+  genGravatarUrl(url: string): Observable<string> {
```
garyluu: This function is executed a few too many times (32 or so times for me) and should be in a service.  Wrapping it in an observable and async pipe isn't actually fixing it.

src/app/organizations/state/register-organization.service.ts
```diff
@@ -178,6 +189,7 @@ export class RegisterOrganizationService {
         location: organizationFormState.location,
         email: organizationFormState.contactEmail,
         status: Organization.StatusEnum.PENDING,
+        avatarUrl: this.returnImgLink(organizationFormState.avatarUrl),
```
garyluu: Try `avatarUrl: organizationFormState.avatarUrl || null` and get rid of function if it works

NatalieEO: Yes that works. Will remove.

src/app/organizations/organization/organization.component.html
```diff
@@ -28,8 +28,9 @@
     </mat-card>
     <mat-card fxFlex class="my-3">
       <div fxLayout="row" fxLayout.lt-sm="column" fxLayoutGap="10px">
-        <div>
-          <img src="https://via.placeholder.com/150" alt="Logo" class="img-circle">
+        <div >
+          <img *ngIf="org?.avatarUrl == null" src="https://via.placeholder.com/150" class="img-circle" height="150" width="150">
```
denis-yuen: Copy this into our repo, just in case.

src/app/organizations/organization/organization.component.html
```diff
@@ -28,8 +28,9 @@
...
+          <img *ngIf="org?.avatarUrl" [src]="genGravatarUrl(org.avatarUrl) | async" alt="Logo" class="img-circle" height="150" width="150">
```
garyluu: Combine the two images into something like `[src]="org.avatar.Url ? genGravatarUrl(org.avatar.url) : https://via.placeholder.com/150"`

src/app/organizations/organization/organization.component.html
NatalieEO: Added new image to repo I found doing creative commons search.

src/app/organizations/organization/organization.component.ts
NatalieEO: Redid in latest commit. Only gets called once now.

src/app/organizations/state/register-organization.service.ts
NatalieEO: Added tests, and slightly changed regex so things like `.png potato` or`.png` will fail.

src/app/organizations/organization/organization.component.ts
```diff
@@ -15,7 +15,7 @@
  */
 import { Component, OnInit } from '@angular/core';
 import { MatDialog } from '@angular/material';
-import { Observable } from 'rxjs';
+import { Observable, of } from 'rxjs';
```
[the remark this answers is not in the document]
NatalieEO: removed

src/app/organizations/organization/organization.component.ts
```diff
@@ -35,14 +36,27 @@ export class OrganizationComponent implements OnInit {
...
+
+  protected ngUnsubscribe: Subject<{}> = new Subject();
 
   ngOnInit() {
     this.loading$ = this.organizationQuery.loading$;
     this.canEdit$ = this.organizationQuery.canEdit$;
     this.organizationService.updateOrganizationFromNameORID();
     this.organization$ = this.organizationQuery.organization$;
+    this.organization$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(o => {
+      if (o) {
```
coverbeck: `if (o && o.avatarUrl)` -- if either is false you want to set gravatarUrl to null

src/app/organizations/state/register-organization.service.ts
```diff
@@ -178,6 +189,7 @@ export class RegisterOrganizationService {
         location: organizationFormState.location,
         email: organizationFormState.contactEmail,
         status: Organization.StatusEnum.PENDING,
+        avatarUrl: organizationFormState.avatarUrl || null,
```
coverbeck: The `|| null` is redundant, unless you're trying to catch an undefined case and/or you want an empty string to become null.

src/app/organizations/state/register-organization.service.ts
```diff
@@ -138,6 +148,7 @@ export class RegisterOrganizationService {
...
```
coverbeck: `|| null` is redundant

src/app/organizations/registerOrganization/register-organization.component.ts
```diff
@@ -40,7 +40,7 @@ export class RegisterOrganizationComponent implements OnInit, OnDestroy {
   public TagEditorMode = TagEditorMode;
   constructor(
     private registerOrganizationService: RegisterOrganizationService,
-    private formsManager: AkitaNgFormsManager<FormsState>, @Inject(MAT_DIALOG_DATA) public data: any
+    private formsManager: AkitaNgFormsManager<FormsState>, @Inject(MAT_DIALOG_DATA) public data: any,
```
coverbeck: I think you should revert this; not relevant to the change

src/app/organizations/registerOrganization/register-organization.component.html
```diff
@@ -46,6 +46,11 @@ <h1 mat-dialog-title>{{title}}</h1>
       <mat-error *ngIf="contactEmail.hasError('email')">Please enter a valid contact email address</mat-error>
       <mat-hint>Once approved, this email address will be publicly visible</mat-hint>
     </mat-form-field>
+    <mat-form-field>
+      <input type="text" matInput formControlName="avatarUrl" placeholder="Image URL">
+      <mat-error *ngIf="avatarUrl.hasError('pattern')">Link must end in jpg, jpeg, or png</mat-error>
+      <mat-hint>Link to your organization's logo image. Link must end in jpg jpeg or png</mat-hint>
```
coverbeck: 1. `logo image` seems redundant. I would just say `logo`.
2. Don't you now support gif (applies to previous line too)?
3. There should be commas like you have in the previous line.

src/app/organizations/state/register-organization.service.ts
NatalieEO: I do want the empty string to become null. If it's an empty string, the database will reject it because of the regex. So I can either have null here, or change the regex in the the webservice to accept an empty string.

src/app/organizations/registerOrganization/register-organization.component.ts
NatalieEO: yes, forgot to remove comma

src/app/organizations/state/register-organization.service.ts
coverbeck: :+1:

src/app/organizations/organization/organization.component.html
NatalieEO: now one

src/app/organizations/organization/organization.component.ts
NatalieEO: changed

src/app/organizations/organization/organization.component.html
```diff
@@ -28,8 +28,8 @@
     </mat-card>
     <mat-card fxFlex class="my-3">
       <div fxLayout="row" fxLayout.lt-sm="column" fxLayoutGap="10px">
-        <div>
-          <img src="https://via.placeholder.com/150" alt="Logo" class="img-circle">
+        <div >
+          <img [src]="org.avatarUrl != null ? gravatarUrl : '../../../assets/images/dockstore/PlaceholderLC.png'" class="img-circle" height="150" width="150">
```
coverbeck: I would probably have `gravatarUrl || '../.....png`. This code assumes that if avatarUrl is not null, then gravatarUrl is not as well. They probably always are in sync, but just test against gravatarUrl itself. Assuming gravatarUrl is not set if avatarUrl isn't set.

src/app/organizations/organization/organization.component.html
```diff
@@ -28,8 +28,8 @@
     </mat-card>
     <mat-card fxFlex class="my-3">
       <div fxLayout="row" fxLayout.lt-sm="column" fxLayoutGap="10px">
-        <div>
-          <img src="https://via.placeholder.com/150" alt="Logo" class="img-circle">
+        <div >
```
coverbeck: Generally there don't seem to be spaces in `<div>`

src/app/organizations/state/organization.service.ts
```diff
@@ -96,4 +96,13 @@ export class OrganizationService {
...
       });
   }
+
+  genGravatarUrl(url: string): string {
+    if (url) {
+      const gravatarUrl = 'https://www.gravatar.com/avatar/' + '000' + '?d=' + url;
+      return gravatarUrl;
+    } else {
+      return null;
+    }
```
agduncan94: return url ? 'https://www.gravatar.com/avatar/' + '000' + '?d=' + url : null'

src/app/organizations/organization/organization.component.html
```diff
@@ -28,8 +28,8 @@
     </mat-card>
     <mat-card fxFlex class="my-3">
       <div fxLayout="row" fxLayout.lt-sm="column" fxLayoutGap="10px">
-        <div >
-          <img [src]="org.avatarUrl != null ? gravatarUrl : '../../../assets/images/dockstore/PlaceholderLC.png'" class="img-circle" height="150" width="150">
+        <div>
+          <img [src]="(gravatarUrl$ | async ) || '../../../assets/images/dockstore/PlaceholderLC.png'" class="img-circle" height="150" width="150">
```
coverbeck: I'm not sure if this is legit, although it worked when @NatalieEO tried it. @garyluu , is this OK?

garyluu: I think it should.  From the pipe map, `(gravatarUrl$ | async)` looks like it will always return null or the actual gravatarUrl.  Beyond that, it's just business as usual

src/app/organizations/organization/organization.component.ts
```diff
@@ -35,14 +35,18 @@ export class OrganizationComponent implements OnInit {
   organization$: Observable<Organization>;
   loading$: Observable<boolean>;
   canEdit$: Observable<boolean>;
-  constructor(private organizationQuery: OrganizationQuery, private organizationService: OrganizationService, private matDialog: MatDialog
-  ) { }
+  gravatarUrl$: Observable<string>;
+  constructor(private organizationQuery: OrganizationQuery, private organizationService: OrganizationService,
+              private matDialog: MatDialog) { }
+
+  protected ngUnsubscribe: Subject<{}> = new Subject();
```
garyluu: This is unused as well as the import

PR #555, opened 2019-03-01
Feature/entry collection organizations

garyluu:
Changed the collectionOrganization object

Moved the current-collection-an-entry-is-part-of component to its own component.
When trying to add to your own collection, filters out the ones that already have it.


src/app/entry/state/current-collections.query.ts
```diff
@@ -0,0 +1,31 @@
...
+    collectionOrganizations.sort((a, b) => this.sortCollectionOrganizations(a, b));
+    return collectionOrganizations;
+  }));
+  currentCollectionIds$: Observable<Array<number>> = this.currentCollectionOrganizations$.pipe(
+    map(collectionOrganizations => collectionOrganizations.map(collectionOrganization => collectionOrganization.collectionId)));
+  constructor(protected store: CurrentCollectionsStore) {
+    super(store);
+  }
+
+  sortCollectionOrganizations(a: CollectionOrganization, b: CollectionOrganization): number {
```
coverbeck: Is organizationName all lower or upper case? If no, then this should be case-insensitive sort.

PR #557, opened 2019-03-04
Move sitemap link

denis-yuen:
Small edit to https://github.com/ga4gh/dockstore/issues/1934

### DIFF
--- a/src/app/home-foot-note/home-foot-note.component.html
+++ b/src/app/home-foot-note/home-foot-note.component.html
@@ -26,7 +26,7 @@
             </a>
           </li>
           <li>
-            <a target="_blank" href="/sitemap">
+            <a target="_blank" routerLink="/sitemap">
               Sitemap
             </a>
           </li>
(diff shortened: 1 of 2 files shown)


src/app/home-foot-note/home-foot-note.component.html
```diff
@@ -26,8 +26,8 @@ <h3 class="documentation">Using Dockstore</h3>
             </a>
           </li>
           <li>
-            <a target="_blank" href="https://docs.dockstore.org/docs/publisher-tutorials/advanced-features/">
-              Advanced Features
+            <a target="_blank" href="/sitemap">
```
garyluu: Anything that is on the site should be a routerLink

coverbeck: I thought so to at first, but because it's `target="_blank"`, I thought the whole SPA would get loaded on the new tab anyway, so it doesn't matter.

garyluu: ah, i missed that

garyluu: ...unless it prevents it from opening in new tab

denis-yuen: It seems to work locally

PR #564, opened 2019-03-06
Point to latest webservice build.

coverbeck:


garyluu:
Change my semaphore file while you're at it

.semaphore/semaphore.yml
```diff
@@ -1,71 +1,71 @@
...
+          - ng lint
+          - npm run build.prod
+          - npm install codecov
+          - ng test --watch=false --code-coverage --browsers ChromeHeadless
+          - ./node_modules/codecov/bin/codecov
+  - name: "Integration Tests"
+    task:
+      env_vars:
+        - name: WEBSERVICE_VERSION
+          value: 1.6.0-alpha.7
```
garyluu: Here too, thanks :smile:

denis-yuen:
@coverbeck updating semaphore is really not necessary until we use it more broadly across the team

garyluu:
Takes very little time to update the semaphore.yml so that I don't get any unexpected surprises and have to update it myself after it fails

denis-yuen:
nevertheless, I'd like to consider updating these as a favour that cannot block PRs 
sorry @garyluu

coverbeck:
@denis-yuen Saw your comment after I'd started the change to semaphore.

garyluu:
Ok, I will make my own commits alongside these when I see it since it actually is necessary for my work.

PR #567, opened 2019-03-07
Feature/1945/google refresh token error message

garyluu:
UI change for ga4gh/dockstore#1945.
There is also a webservice change.

Handles the error in a slightly better way.  Previously, the error alert card would show for a split second before disappearing because the other GET request succeeded and overwrites the error.

Also changes the Lint, Unit, and Prod Test to be its own stage so that integration tests don't trigger if something simple like 'ng build' fails

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,29 +25,26 @@ install:
 
 jobs:
   include:
-    - stage: test
+    - stage: unit test
+      name: "Lint, Unit Test, and Prod Test"
       script:
         - ng lint
         - ng test --watch=false --code-coverage --browsers ChromeHeadless
         - ng version
         - npm run build.prod
+    - stage: integration test
+      name: "Cypress Group 1"
       env:
-        - NAME='Lint, Unit Test, and Prod Test'
-    - stage: test
-      env:
-        - NAME='Integration test 1'
         - TEST='cypress/integration/group1/**/*'
       script: ./scripts/run-travis-script.sh
-
-    - stage: test
+    - stage: integration test
+      name: "Cypress Group 2"
       env:
-        - NAME='Integration test 2'
         - TEST='cypress/integration/group2/**/*'
       script: ./scripts/run-travis-script.sh
-
-    - stage: test
+    - stage: integration test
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -25,7 +25,7 @@ install:
 
 jobs:
   include:
-    - stage: test
```
denis-yuen: Use a simpler one word for the stage (less typo prone), Travis already lower cases too 

You can use the longer description as a name
https://docs.travis-ci.com/user/build-stages/#naming-your-build-stages

src/app/shared/alert/state/alert.service.ts
```diff
@@ -81,6 +81,20 @@ export class AlertService {
...
+  /**
+   * Handles HTTP error response and show a detailed message only in the matSnackBar.
+   * Use when there's is no alert component which normally displays the detailed message.
+   *
+   * @param {HttpErrorResponse} error  The HttpErrorResponse return by the failed Http call
+   * @memberof AlertService
+   */
+  public detailedSnackBarError(error: HttpErrorResponse) {
+    this.clearEverything();
+    const detailedError = '[HTTP ' + error.status + '] ' + error.statusText + ': ' +
```
coverbeck: I think I've commented on this elsewhere, and there may be a ticket for it already, but I don't think we should be surfacing HTTP status codes to the user, at least not by default -- maybe there is a detailed view where it gets shown. I would think the typical Dockstore user would know/understand http status codes...

But since we're doing it elsewhere, this is fine.

denis-yuen: I'm in the camp that I'd like to have them shown, if only to make it easier to debug bug reports. 
But that they can't be the only thing that is shown (i.e. a more descriptive message accompanying is  helpful)

coverbeck: OK... The "debug bug reports" reminds me, one thing we did in a previous job was send all 4xx and 5xx errors in the browser to Google Analytics, then periodically try to cull the most common ones.

PR #568, opened 2019-03-08
Feature/2129/cwl viewer broken

NatalieEO:
ga4gh/dockstore#2129
CWL viewer no longer broken. Provider url no longer null.

### DIFF
--- a/src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
+++ b/src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
@@ -14,7 +14,7 @@
  *    limitations under the License.
  */
 
-import { Component, Input, OnDestroy, OnInit } from '@angular/core';
+import {Component, Input, OnChanges, OnDestroy, OnInit} from '@angular/core';
 import { CwlViewerDescriptor, CwlViewerService } from './cwl-viewer.service';
 import { WorkflowVersion } from '../../../shared/swagger/model/workflowVersion';
 import { ExtendedWorkflow } from '../../../shared/models/ExtendedWorkflow';
@@ -31,13 +31,12 @@ import {takeUntil} from 'rxjs/operators';
   styleUrls: ['./cwl-viewer.scss']
 })
 
-export class CwlViewerComponent implements OnInit, OnDestroy {
+export class CwlViewerComponent implements OnInit, OnChanges, OnDestroy {
 
   @Input() workflow: ExtendedWorkflow;
   @Input() set selectedVersion(value: WorkflowVersion) {
     if (value != null) {
       this.version = value;
-      this.onChange();
     }
   }
   @Input() set refresh(value: any) {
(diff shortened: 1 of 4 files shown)


src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
```diff
@@ -27,13 +31,12 @@ import { Subject } from 'rxjs';
   styleUrls: ['./cwl-viewer.scss']
 })
 
-export class CwlViewerComponent implements OnInit, OnDestroy {
+export class CwlViewerComponent implements OnInit, OnChanges, OnDestroy {
 
   @Input() workflow: ExtendedWorkflow;
```
garyluu: check whether this is still used (since you have the actual extendedWorkflow now).  if it's not used, remove it.  if it is, change the type to Workflow.

NatalieEO: Removed the variable and looks like all my imports are being used.

src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
```diff
@@ -27,13 +31,11 @@ import { Subject } from 'rxjs';
   styleUrls: ['./cwl-viewer.scss']
 })
 
-export class CwlViewerComponent implements OnInit, OnDestroy {
+export class CwlViewerComponent implements OnInit, OnChanges, OnDestroy {
 
-  @Input() workflow: ExtendedWorkflow;
   @Input() set selectedVersion(value: WorkflowVersion) {
```
coverbeck: The only reason to have this setter was to invoke `this.onChange()`. Now that you've made the code better by just using `ngOnChanges`, I think you should just be able to change this to `@Input() version: WorkflowVersion`.

src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
garyluu: Check extendedWorkflowService and extendedWorkflowStore

src/app/workflow/dag/cwl-viewer/cwl-viewer.component.ts
garyluu: reminder, version is only set there it seems.  if you remove the setter, you might as well remove the version variable altogether and use the selectedVersion instead.

coverbeck: Sorry, yes I meant `@Input selectedVersion: WorkflowVersion`, which then means as Gary says, that your code should reference this.selectedVersion.

NatalieEO: Hm, actually I realize now that the image won't appear if the setter is removed.

NatalieEO: I think the version stays null and so the ngOnChanges part doesn't never goes because of the if check

PR #570, opened 2019-03-08
Feature/2160/alias redirect

agduncan94:
Removes previous alias code, now instead does the following:
* Added an aliases component at /aliases/type/alias, where `type` is in the set {organisations, collections} and `alias` is the alias
* If type is valid, will attempt to find the associated object for that alias using the appropriate alias endpoint
* If found object, redirect to page
* If not found, message saying no object with given alias found of some type

The benefit of this approach is that the organisations/collections code doesn't need to worry about aliases, they are all done in a separate component. Any new entity types (ex. entry) can be added to this component pretty easily.

The alias page shows the following, though this will only be visible for a split second, if that, unless on a slow connection.
redirect-alias

This PR also removes any links to organisations or collections using DB ids. **Now ONLY names will be used for these links.**

### DIFF
--- a/cypress/integration/group3/organizations.ts
+++ b/cypress/integration/group3/organizations.ts
@@ -132,25 +132,11 @@ describe('Dockstore Organizations', () => {
       cy.contains('veryFakeCollectionName');
       cy.contains('very fake collection topic');
     });
-
-    it('be able to update a collection description', () => {
-      cy.visit('/organizations/1/collections/1');
-      cy.get('#editCollectionDescription').click();
-      cy.get('#updateOrganizationDescriptionButton').should('be.visible').should('not.be.disabled');
-      typeInTextArea('Description', '* fake collection description');
-      cy.contains('Preview Mode').click();
-      cy.contains('fake collection description');
-      cy.contains('* fake collection description').should('not.exist');
-      cy.get('#updateOrganizationDescriptionButton').should('be.visible').should('not.be.disabled').click();
-      cy.get('#updateOrganizationDescriptionButton').should('not.be.visible');
-      cy.contains('fake collection description');
-      cy.contains('* fake collection description').should('not.exist');
-    });
   });
 
   describe('Should be able to update description', () => {
     it('be able to update an organization description with markdown', () => {
-      cy.visit('/organizations/1');
+      cy.visit('/organizations/Potatoe');
       cy.get('#editOrgDescription').click();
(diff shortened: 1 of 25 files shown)


src/app/aliases/aliases.component.html
```diff
@@ -0,0 +1,26 @@
+<app-header>
+  Aliases
+</app-header>
+<div class="container">
+  <mat-card class="alert alert-warning" role="alert" *ngIf="!types.includes(type)">
```
garyluu: have a boolean variable for this ngIf

cypress/integration/group3/organizations.ts
```diff
@@ -260,13 +260,72 @@ describe('Dockstore Organizations', () => {
...
+      cy.visit('/organizations/Potatoe/collections/veryFakeCollectionName');
       cy.title().should('eq', 'Dockstore | Collection');
     });
   });
+
+  describe('Find organization and collection by alias', () => {
+    it('organization alias', () => {
+      cy.server();
+      cy.route({
+        url: '/organizations/fakeAlias/aliases',
```
denis-yuen: So it I understand a visit to `/aliases/organizations/fakeAlias` should redirect to `Cypress.config().baseUrl + '/organizations/Potatoe`. But what role is `/organizations/fakeAlias/aliases` playing here?

src/app/aliases/aliases.component.html
```diff
@@ -0,0 +1,26 @@
+<app-header>
+  Aliases
```
denis-yuen: Could probably just call it "redirecting..." or something similar.

src/app/aliases/aliases.component.ts
```diff
@@ -0,0 +1,59 @@
...
+    super();
+  }
+
+  ngOnInit() {
+    this.type = this.route.snapshot.paramMap.get('type');
+    this.alias = this.route.snapshot.paramMap.get('alias');
+
+    this.loading$ = this.aliasesQuery.loading$;
+
+    if (this.types.includes(this.type)) {
```
coverbeck: I don't think this if is necessary, given the explicit tests you have for `organizations` and `collections` inside this if, but I could be missing something.

agduncan94: I think that is a relic of something I was trying out before, I'll remove it.

cypress/integration/group3/organizations.ts
agduncan94: This is the API call being used to get an organizations alias called fakeAlias, this is not what is seen in the browser URL.

PR #576, opened 2019-03-13
Feature/1947/search table bug

NatalieEO:
ga4gh/dockstore#1947
Fixed by setting workflow tab to be active when workflow facet is selected and search request has no hits.


src/app/search/state/search.query.ts
```diff
@@ -47,10 +47,14 @@ export class SearchQuery extends Query<SearchState> {
     if (!tools || !workflows) {
       return true;
     }
+    const param = this.route.snapshot.queryParams['_type'];
+
     if (tools.length === 0 && workflows.length > 0) {
       return false;
     } else if (workflows.length === 0 && tools.length > 0) {
       return true;
+    } else if (workflows.length === 0 && param === 'workflow') {
```
coverbeck: A comment here might be nice, but not crucial. We can always figure it out by looking at the Git blame.

PR #583, opened 2019-03-14
Fix error message not going away

NatalieEO:
ga4gh/dockstore#2124
After looking more into it, seems like the error message wouldn't disappear only if you navigated to the search page. Fixed by calling alertService.clearEverything() in search component.


NatalieEO:
@garyluu from my understanding, the other pages happen to have alertService.start() called in some component which calls setInfo(). setInfo() sets the type in alert store back to 'info' instead of 'error'. But the search component never calls alertService.start() so the old error state doesn't get overwritten.

PR #589, opened 2019-03-18
Feature/2072/highlight current version

agduncan94:
Highlights the current version if available in the public side box thing.
https://github.com/ga4gh/dockstore/issues/2072


denis-yuen:
odd build issue?

agduncan94:
@denis-yuen Build issue is with the organisation tests occasionally failing. It is out of scope for this issue, though I think needs to be fixed. I think we have created an issue to do so.

PR #590, opened 2019-03-18
Feature/fix tool table

agduncan94:
* Order of workflow project links and stars matches that of tools
* Remove docker pull from tools table


denis-yuen:
snyk looks like a false positive

PR #592, opened 2019-03-19
Feature/2244/rejected orgs should be hidden

agduncan94:
Depends on https://github.com/ga4gh/dockstore/pull/2245

When an organisation is rejected:
* It appears under my requests on the accounts page (request tab)
* Users can request a re-review from the accounts page
* There is now an alert on organisations and collections that are rejected

Also https://github.com/ga4gh/dockstore/issues/2244

### DIFF
--- a/src/app/loginComponents/requests/requests.component.html
+++ b/src/app/loginComponents/requests/requests.component.html
@@ -46,9 +46,9 @@
 
   <!-- My Pending Organization Requests -->
   <div class="p-3">
-    <h3>My Pending Requests</h3>
+    <h3>My Requests</h3>
     <p>The following organizations that you have created are pending approval.</p>
-    <mat-card class="alert alert-info" role="alert" *ngIf="(myPendingOrganizationRequests$ | async)?.length == 0">
+    <mat-card class="alert alert-info" role="alert" *ngIf="(myPendingOrganizationRequests$ | async)?.length == 0 && (myRejectedOrganizationRequests$ | async)?.length == 0">
       <mat-icon>info</mat-icon> You have no organizations pending approval.
     </mat-card>
 
@@ -56,6 +56,21 @@
       <mat-card *ngFor="let membership of (myPendingOrganizationRequests$ | async); let i = index" [id]="'my-pending-org-card-' + i">
         <mat-card-header>
           <mat-card-title>
+            <a [routerLink]="('/organizations/' + membership.organization.name )" [matTooltip]="membership?.organization?.name">{{membership.organization.displayName}}</a>
+          </mat-card-title>
+          <mat-card-subtitle>
+            Requested on {{ membership.dbCreateDate | date}}
+          </mat-card-subtitle>
+        </mat-card-header>
+        <mat-card-content class="px-4">
+          {{ membership.organization.topic }}
+        </mat-card-content>
+      </mat-card>
+
+      <mat-card *ngFor="let membership of (myRejectedOrganizationRequests$ | async); let i = index" [id]="'my-pending-org-card-' + i">
+        <mat-card-header>
+          <mat-card-title>
+            <mat-icon color="warn" matTooltip="This organization has been rejected">warning</mat-icon>
             <a [routerLink]="('/organizations/' + membership.organization.name )" [matTooltip]="membership?.organization?.name">{{membership.organization.displayName}}</a>
           </mat-card-title>
           <mat-card-subtitle>
--- a/src/app/loginComponents/requests/requests.component.html
+++ b/src/app/loginComponents/requests/requests.component.html
@@ -31,12 +31,11 @@
           {{ org.topic }}
         </mat-card-content>
 
-        <mat-card-actions>
-          <div fxFlex></div>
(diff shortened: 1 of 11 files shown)


src/app/loginComponents/state/requests.service.ts
```diff
@@ -135,7 +139,23 @@ export class RequestsService {
...
       }, () => {
-        this.updateMyMembershipState(null, null, null);
+        this.updateMyMembershipState(null, null, null, null);
+        this.requestsStore.setError(true);
+        this.alertService.simpleError();
+      });
+  }
+
+  requestRereview(id: number): void {
+    this.alertService.start('Rerquesting review for organization ' + id);
```
garyluu: typo

src/app/loginComponents/requests/requests.component.html
```diff
@@ -67,6 +66,32 @@ <h3>My Pending Requests</h3>
...
+    <p class="pt-2">The following organizations that you have created are rejected.</p>
+    <mat-card class="alert alert-info" role="alert" *ngIf="(myRejectedOrganizationRequests$ | async)?.length == 0">
+      <mat-icon>info</mat-icon> You have no rejected organizations.
+    </mat-card>
+    <div fxLayout="column" fxLayoutGap="32px">
+      <mat-card *ngFor="let membership of (myRejectedOrganizationRequests$ | async); let i = index" [id]="'my-rejected-org-card-' + i">
+        <mat-card-header>
+          <mat-card-title>
+            <mat-icon color="primary" matTooltip="This organization has been rejected">warning</mat-icon>
+            <a [routerLink]="('/organizations/' + membership.organization.name )" [matTooltip]="membership?.organization?.name">{{membership.organization.displayName}}</a>
```
coverbeck: Inconsistent on the use of `?`. I'm assuming the one after membership is unnecessary; if the one after organization is necessary, then it should be added  for the other usages.

src/app/loginComponents/requests/requests.component.html
```diff
@@ -31,12 +31,11 @@ <h3>Curator Requests</h3>
...
 
-        <mat-card-actions>
-          <div fxFlex></div>
-          <button mat-flat-button [id]="'reject-pending-org-' + i" class="mr-1" (click)="openDialog(org.displayName, org.id, false)">
+        <mat-card-actions fxLayout="row" fxLayoutAlign="end center" fxLayoutGap="10px">
+          <button mat-flat-button [id]="'reject-pending-org-' + i" class="mr-1" (click)="openDialog(org.displayName, org.id, false)" matTooltip="Reject an organization request">
             <mat-icon>clear</mat-icon>Reject
           </button>
-          <button mat-flat-button color="primary" [id]="'approve-pending-org-' + i" (click)="openDialog(org.displayName, org.id, true)">
+          <button mat-flat-button color="primary" [id]="'approve-pending-org-' + i" (click)="openDialog(org.displayName, org.id, true)" matTooltip="Approve an organization request">
```
coverbeck: Just curious, why are you generating ids for here and elsewhere in the HTML? Is this because Angular can get confused if there is nothing unique distinguishing (I've seen that in AngularJS)?

src/app/organizations/collection/collection.component.html
```diff
@@ -30,6 +30,10 @@
           You can view your pending organization requests on the requests tab of the <a [routerLink]="('/accounts')" style="text-decoration: underline">accounts page</a>.
         </ng-template>
       </mat-card>
+
+      <mat-card fxFlex class="my-3 alert alert-info" *ngIf="organization?.status === 'REJECTED'">
```
coverbeck: I don't get the whole context, but assuming this is a public page (if it's not, ignore the rest of this):

Is this for the case where an org was approved then rejected?

If not, I think rejected orgs should essentially be treated as non-existent to non-members.

If yes, it's trickier because of the whole deletion/archiving thing we've been discussing.

src/app/organizations/organization/organization.component.html
```diff
@@ -31,6 +31,11 @@
         You can view your pending organization requests on the requests tab of the <a [routerLink]="('/accounts')" style="text-decoration: underline">accounts page</a>.
       </ng-template>
     </mat-card>
+
+    <mat-card fxFlex class="my-3 alert alert-info" *ngIf="org?.status === 'REJECTED'">
```
coverbeck: Same question/comment as previous question/comment.

src/app/organizations/collection/collection.component.html
agduncan94: So this should only show if the user is a curator/admin or has a membership in the organisation. It should not appear to anyone else (public).

src/app/loginComponents/requests/requests.component.html
agduncan94: I am doing it for testing specifically.

garyluu: I generally follow https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements  , but I don't mind others using ID

PR #594, opened 2019-03-19
Feature/2227/org logo should not be circle

NatalieEO:
ga4gh/dockstore#2227
Org logo is now a square or rectangle depending on the original image's dimensions. Uses `max-width` and `max-height` so image doesn't get distorted. Max-width is bigger because it seems more likely that logos would have larger widths than heights.


denis-yuen:
(pending build success, restarted builds)

NatalieEO:
Push still shows pending but clicking details shows it's actually done and passing. Will merge.

PR #595, opened 2019-03-20
Don't display error message if user cannot perform account changes

rshreya:
ga4gh/dockstore#2084

### DIFF
--- a/src/app/loginComponents/accounts/controls/controls.component.html
+++ b/src/app/loginComponents/accounts/controls/controls.component.html
@@ -1,10 +1,8 @@
 
 <div class="p-3">
-  <div *ngIf="(canChangeUsername$ | async)">
-    <mat-card class="alert alert-danger" role="alert">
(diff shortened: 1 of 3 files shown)


src/app/loginComponents/accounts/controls/controls.component.html
```diff
@@ -1,8 +1,10 @@
 
 <div class="p-3">
-  <mat-card class="alert alert-danger" role="alert">
-    <mat-icon>error</mat-icon> Dockstore account controls have a large impact and should be used with extreme caution.
-  </mat-card>
+  <div *ngIf="(canChangeUsername$ | async)">
```
coverbeck: You can just put the `*ngIf` directly on the `mat-card` and get rid of this `<div>`.

agduncan94:
For the failing test I think you have to mock the /extended API call to properly mock this (in the file disabledAccountControls.ts). See the beforeEach on line 22 of dropdown.ts for an example of mocking this API call.

I'm not sure if it is in the primer, but you can view your cypress tests at https://dashboard.cypress.io.

PR #596, opened 2019-03-20
If webservice on port 443, omit port in url

coverbeck:
ga4gh/dockstore#1594

This doesn't drop the port 80 if on default port of http, but
doesn't seem worth it.

There are a lot of places that use API_URL; instead of massaging
them one by one to remove the 443 port, it seems like we
should just do it here. A lot of those places have user-facing
URLs, e.g., the TRS url, the launch with instructions...

Will create a corresponding PR in compose_setup if this gets
approved.

### DIFF
--- a/src/app/shared/dockstore.model.ts
+++ b/src/app/shared/dockstore.model.ts
@@ -29,7 +29,8 @@ export class Dockstore {
   static readonly DISCOURSE_URL = 'http://localhost/';
 
   static readonly LOCAL_URI = Dockstore.HOSTNAME + ':' + Dockstore.UI_PORT;
-  static readonly API_URI = Dockstore.HOSTNAME + ':' + Dockstore.API_PORT;
+  // @ts-ignore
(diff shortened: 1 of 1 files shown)


garyluu:
What's the ts-lint warning?

coverbeck:
> What's the ts-lint warning?

`ERROR in src/app/shared/dockstore.model.ts(32,51): error TS2367: This condition will always return 'true' since the types '"8080"' and '"443"' have no overlap.`

coverbeck:
I did the change this way  (`Dockstore.HOSTNAME + (Dockstore.API_PORT !== '443' ? (':' + Dockstore.API_PORT) : '')`) because I incorrectly assumed that install_bootstrap prompted for the port and I'd have to change it to accept an empty value. But now I see the port is just set directly in the template.

Now I think I should instead just see if port is set. But I'd also like to get this in the upcoming release to staging, so I'll just go with this for now.

PR #598, opened 2019-03-20
Handle 204

garyluu:
For issue ga4gh/dockstore#2164 

It was expected that the endpoint would always return a workflow object on success.  But since there's a possible 204 now, the UI needs to know how to handle it.

Edit: This also has the same problems with tools

### DIFF
--- a/src/app/shared/file-editing.ts
+++ b/src/app/shared/file-editing.ts
@@ -14,7 +14,7 @@
  *    limitations under the License.
  */
 
-import { Input } from '@angular/core';
+import { AlertService } from './alert/state/alert.service';
 import { Files } from './files';
 import { SourceFile } from './swagger/model/sourceFile';
 
@@ -25,6 +25,14 @@ export class FileEditing extends Files {
    */
   toggleEdit() {
     this.editing = !this.editing;
+  }
+
+  constructor(protected alertService: AlertService) {
+    super();
+  }
+
+  handleNoContentResponse() {
+    this.alertService.detailedSuccess('Version did not change (no changes detected in file contents)');
   }
 
   /**
--- a/src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
+++ b/src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
@@ -1,6 +1,5 @@
 import { Component, Input } from '@angular/core';
 import { Observable } from 'rxjs';
-
 import { AlertService } from '../../shared/alert/state/alert.service';
 import { FileEditing } from '../../shared/file-editing';
 import { WorkflowQuery } from '../../shared/state/workflow.query';
@@ -10,6 +9,7 @@ import { RefreshService } from './../../shared/refresh.service';
 import { HostedService } from './../../shared/swagger/api/hosted.service';
 import { WorkflowsService } from './../../shared/swagger/api/workflows.service';
 import { WorkflowVersion } from './../../shared/swagger/model/workflowVersion';
+
 
 @Component({
   selector: 'app-workflow-file-editor',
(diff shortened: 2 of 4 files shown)


garyluu:
I cancelled the push because I don't care about it.

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
```diff
@@ -77,23 +77,30 @@ export class WorkflowFileEditorComponent extends FileEditing {
...
-          });
-        }, error =>  {
-          if (error) {
+          }, error => {
             this.alertService.detailedError(error);
-          }
+          });
+        } else {
+          // Probably encountered a 204
+          this.alertService.detailedSuccess();
```
denis-yuen: What does the end user end up seeing here?

garyluu: There will just be a snackBar saying updating version succeeded.

denis-yuen: I think this might be linked to the wrong issue up top. 
Possible to say something like "no detected changes, update ignored" as a warning?

garyluu: Oops, definitely linked to wrong issue

garyluu: Ok, I added a way to override the success message.

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
```diff
@@ -77,23 +77,29 @@ export class WorkflowFileEditorComponent extends FileEditing {
...
             this.workflowService.setWorkflow(workflow2);
-          });
-        }, error =>  {
-          if (error) {
+          }, error => {
             this.alertService.detailedError(error);
-          }
+          });
+        } else {
+          // Probably encountered a 204
```
coverbeck: The downside of using a generated client library. :) Good thing this method never returns 201; how would you ever distinguish a 201 and 204?

garyluu: there's an observe parameter in the function that you can specify `response` instead of the default which is `body`.   Once you have the response, you can figure out what the status code was.  I personally thought i was overkill for this, so I just left it as is.

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
```diff
@@ -77,23 +77,29 @@ export class WorkflowFileEditorComponent extends FileEditing {
...
+          this.alertService.detailedSuccess('Version did not change');
```
denis-yuen: Please say why it didn't change (e.g. no file changes provided, new files are not different, etc.)

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
coverbeck: 👍 Did not know that. I was frustrated by the inability to write the API I wanted for POSTing zips because of client code generation and vented incorrectly here.

coverbeck: 👍 Did not know that. I was frustrated by the inability to write the API I wanted for POSTing zips because of client code generation and vented incorrectly here.

src/app/shared/file-editing.ts
```diff
@@ -27,6 +27,14 @@ export class FileEditing extends Files {
     this.editing = !this.editing;
   }
 
+  constructor(protected alertService: AlertService) {
+    super();
+  }
+
+  handleNoContentResponse() {
```
denis-yuen: do inline to match the workflow version

PR #599, opened 2019-03-20
Clear alert state on route change

NatalieEO:
ga4gh/dockstore#2124
ga4gh/dockstore#2230
Alert state gets cleared when route changes.

### DIFF
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -136,7 +136,7 @@ export class SearchComponent implements OnInit, OnDestroy {
 
   ngOnInit() {
     // Set alert store back to its initial state. Otherwise an error created on another page might appear when search page is loaded.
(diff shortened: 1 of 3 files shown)


src/app/search/search.component.ts
```diff
@@ -136,7 +136,7 @@ export class SearchComponent implements OnInit, OnDestroy {
 
   ngOnInit() {
     // Set alert store back to its initial state. Otherwise an error created on another page might appear when search page is loaded.
-    this.alertService.clearEverything();
+    // this.alertService.clearEverything();
```
garyluu: If this is not used, just remove it

NatalieEO: Ah forgot to delete that. Canceled other build and pushed new commit.

PR #600, opened 2019-03-21
Feature/2264/incompatibleVersions

garyluu:
Use incompatible versions for ga4gh/dockstore#2264

Revert later!  Or maybe don't merge at all!

Previous tests using compatible versions take less than 11 mins on Travis.  This PR changed it to over 28 mins, but it doesn't affect me since I'm using Semaphore CI!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 dist: trusty
 language: node_js
 node_js:
-  - '10.13.0'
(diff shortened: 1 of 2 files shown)


denis-yuen:
@garyluu looks like this needs a travis_wait https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

PR #604, opened 2019-03-22
2167/fix my unpublished workflows bug

NatalieEO:
ga4gh/dockstore#2167
Honestly not sure why this wasn't working before after going through the code.
Interestingly, this bug didn't exist before on Safari (and still doesn't).

### DIFF
--- a/src/app/myworkflows/my-workflow/my-workflow.component.ts
+++ b/src/app/myworkflows/my-workflow/my-workflow.component.ts
@@ -145,6 +145,10 @@ export class MyWorkflowComponent extends MyEntry implements OnInit {
           const sortedSharedWorkflows = this.myworkflowService.sortGroupEntries(sharedWorkflows, this.user.username, 'workflow');
           this.setGroupSharedEntriesObject(sortedSharedWorkflows);
 
+          // If a user navigates directly to an unpublished workflow on their my-workflows page (via bookmark, refresh), the url needs to be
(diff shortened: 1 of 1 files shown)


src/app/myworkflows/my-workflow/my-workflow.component.ts
```diff
@@ -145,6 +145,10 @@ export class MyWorkflowComponent extends MyEntry implements OnInit {
           const sortedSharedWorkflows = this.myworkflowService.sortGroupEntries(sharedWorkflows, this.user.username, 'workflow');
           this.setGroupSharedEntriesObject(sortedSharedWorkflows);
 
+          // If a user navigates directly to an unpublished workflow on their my-workflows page (via bookmark, refresh), the url needs to be
```
coverbeck: How does this work if you don't directly navigate to the my-workflows page? Has something else already invoked `workflowService.setWorkflow`?

NatalieEO: Do you mean if you don't directly navigate to a specific workflow on the my-workflows page (i.e. just go to my-workflows)? If so, the next if statement, `if (!this.workflow)`, will get executed which will eventually set the workflow to your first published entry, or if none are published, just the first workflow.

coverbeck: Yes, I meant that. Thanks for the clarification.

PR #607, opened 2019-03-26
Same hacky fix like organization

garyluu:
For ga4gh/dockstore#2294

We have a PUT and don't have the entire object, so it was setting it to null every time. Now giving it more of the entire object.

### DIFF
--- a/src/app/organizations/collections/state/create-collection.service.ts
+++ b/src/app/organizations/collections/state/create-collection.service.ts
@@ -4,13 +4,13 @@ import { FormBuilder, FormGroup, Validators } from '@angular/forms';
 import { MatDialog, MatSnackBar } from '@angular/material';
 import { AkitaNgFormsManager } from '@datorama/akita-ng-forms-manager';
 import { finalize } from 'rxjs/operators';
-
 import { AlertService } from '../../../shared/alert/state/alert.service';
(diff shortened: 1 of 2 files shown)


denis-yuen:
> For [ga4gh/dockstore#2294](https://github.com/ga4gh/dockstore/issues/2294)
> 
> We have a PUT not don't have the entire object, so it was setting it to null every time. Now giving it more of the entire object.

Think there's a typo here.
Maybe, "We have a PUT that was sending an incomplete object" ...

src/app/organizations/collections/state/create-collection.service.ts
```diff
@@ -141,14 +143,16 @@ export class CreateCollectionService {
...
    * @memberof CreateCollectionService
    */
-  updateCollection(collectionFormState: FormsState['createOrUpdateCollection'], collectionID: number) {
+  updateCollection(collectionFormState: FormsState['createOrUpdateCollection'], collectionID: number, collectionDescripton: string) {
     let collection: Collection;
     collection = {
       name: collectionFormState.name,
       topic: collectionFormState.topic,
-      displayName: collectionFormState.displayName
+      displayName: collectionFormState.displayName,
```
denis-yuen: Code comment:
Why are the other three coming from state but description is passed in?

PR #610, opened 2019-03-29
Use transaction postgres + DAG fix

garyluu:
- Fixes DAG labels overlapping ga4gh/dockstore#2281
- Revert the cy.exec timeout back to the default because it appears if it does timeout with the default, it will timeout with a longer one too
- Use transaction for postgres reset script because it's better practice that way


garyluu:
The postgres changes should be made regardless because timeout was way too long.

As for the transaction, I think it's a better practice to have it like that.

PR #615, opened 2019-04-02
fix discourse location

agduncan94:
https://github.com/ga4gh/dockstore/issues/1305


garyluu:
It's the infamous postgres timeout.  I have a fix attempt for it but not entirely confident it actually works.

PR #616, opened 2019-04-02
Hosted entry path

agduncan94:
Hosted entry paths default to be absolute.


src/app/shared/code-editor-list/code-editor-list.component.ts
```diff
@@ -129,7 +129,7 @@ export class CodeEditorListComponent {
     } else if (this.fileType === 'testParam') {
       return '.json';
     } else if (this.fileType === 'dockerfile') {
-      return '/Dockerfile';
+      return 'Dockerfile';
```
denis-yuen: Is Dockerfile losing the leading slash on purpose?

agduncan94: Yes, since it is added above in a more general case.

PR #621, opened 2019-04-10
Feature/2142/cache-epam-visualization

rjbautis:
For ga4gh/dockstore#2142
Adds a new Entity Store for WDL workflow visualizations produced by EPAM pipeline builder library

### DIFF
--- a/src/app/workflow/dag/wdl-viewer/state/wdl-viewer.model.ts
+++ b/src/app/workflow/dag/wdl-viewer/state/wdl-viewer.model.ts
@@ -1,10 +1,12 @@
 /**
  * Defines types for the response of Pipeline Builder library
  */
-export interface WdlViewerPipeline {
+export interface WdlViewerPipelineResponse {
   status: boolean;
   message: string;
   model: Array<any>;
   actions: Array<any>;
-  id?: number;
+
+  // Not originally part of the response object, but all responses should be identified by the workflow version used to generated it
+  workflowVersion?: number;
 }
--- a/src/app/workflow/dag/wdl-viewer/state/wdl-viewer.model.ts
+++ b/src/app/workflow/dag/wdl-viewer/state/wdl-viewer.model.ts
@@ -6,7 +6,4 @@ export interface WdlViewerPipelineResponse {
   message: string;
   model: Array<any>;
   actions: Array<any>;
-
-  // Not originally part of the response object, but all responses should be identified by the workflow version used to generated it
-  workflowVersion?: number;
 }
--- a/src/app/workflow/dag/wdl-viewer/state/wdl-viewer.model.ts
+++ b/src/app/workflow/dag/wdl-viewer/state/wdl-viewer.model.ts
@@ -1,0 +1,10 @@
+/**
+ * Defines types for the response of Pipeline Builder library
+ */
+export interface WdlViewerPipeline {
+  status: boolean;
+  message: string;
+  model: Array<any>;
+  actions: Array<any>;
+  id?: number;
+}
--- a/src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts
+++ b/src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts
@@ -17,12 +17,12 @@
 import { AfterViewInit, Component, ElementRef, Input, OnDestroy, Renderer2, ViewChild, ViewEncapsulation } from '@angular/core';
 import * as pipeline from 'pipeline-builder';
 import { Subject } from 'rxjs';
-import { finalize, take, takeUntil, delay } from 'rxjs/operators';
+import { take, takeUntil, delay } from 'rxjs/operators';
 import { FileService } from '../../../shared/file.service';
 import { GA4GHFilesService } from '../../../shared/ga4gh-files/ga4gh-files.service';
 import { WorkflowQuery } from '../../../shared/state/workflow.query';
 import { ToolDescriptor, Workflow, WorkflowsService, WorkflowVersion } from '../../../shared/swagger';
-import { WdlViewerPipeline } from './state/wdl-viewer.model';
+import { WdlViewerPipelineResponse } from './state/wdl-viewer.model';
 import { WdlViewerQuery } from './state/wdl-viewer.query';
 import { WdlViewerService } from './state/wdl-viewer.service';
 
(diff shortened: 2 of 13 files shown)


src/app/workflow/dag/wdl-viewer/state/wdl-viewer.model.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Defines types for the response of Pipeline Builder library
+ */
+export interface WdlViewerPipeline {
+  status: boolean;
+  message: string;
+  model: Array<any>;
+  actions: Array<any>;
+  id?: number;
```
coverbeck: This is getting nit-picky, but it looks like the id isn't actually part of the response, but something you add on -- the id of the workflow version. I was going to comment in the `wdl-viewer.service.update()` method that your code might fail because if the id is not set (because it's declared optional here), but it looks like you always ensure the id is set -- but you're not letting the compiler do the checking for you.

You don't need to over-architect this, so I suggest at a bare minimum commenting and/or renaming the id field to clearly indicate that that it is not part of the response.

More elaborately, you could try using [intersection types](https://www.typescriptlang.org/docs/handbook/advanced-types.html), where WdlViewerPipeline is the response, and you have a new type that is an intersection of WdlViewerPipeline and id, where is not optional.

src/app/workflow/dag/wdl-viewer/state/wdl-viewer.model.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Defines types for the response of Pipeline Builder library
+ */
+export interface WdlViewerPipeline {
```
coverbeck: I know it was like this already, but I would call this something like WdlViewerPipelineResponse or WdlViewerPipelineModel to clearly indicate this is not a pipeline.

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts
```diff
@@ -64,29 +67,42 @@ export class WdlViewerComponent implements AfterViewInit, OnDestroy {
           return;
         }
 
-        if (files && files.length > 0) {
+        // Add delay so parent component can finish rendering it's view before the status is set
```
coverbeck: Typo, should be `its`

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts
```diff
@@ -64,29 +67,42 @@ export class WdlViewerComponent implements AfterViewInit, OnDestroy {
...
+            } else if (files && files.length > 0) {
+
+              // Create the Epam WDL visualization and attach the result to the DOM element. Stop subscribing after first completion
+              this.wdlViewerService.create(files, this.workflow, this.version).pipe(take(1), finalize(() => {
+                this.clearProgressBar();
+              }))
+                .subscribe((res: WdlViewerPipeline) => {
+                    res.id = this.version.id;
+                    this.visualizer.attachTo(res.model[0]);
+                    this.wdlViewerService.update(this.workflow.id, res);
```
coverbeck: I haven't tried running this, but won't this cause the selectEntity subscription on line 71 to get an observable and you'll end up attaching to visualizer twice?

rjbautis: This doesn't attach the object to the visualizer twice because the selectEntity subscription on line 71 pipes the `take` operator. The operator limits the subscription to emit only 1 value (or entity in this case) and then stops subscribing. On the other hand, if the subscription piped `takeUntil(this.ngUnsubscribe)` then the visualizer would attach it twice

I also verified that this code doesn't attach twice with breakpoints in Chrome debugger.

coverbeck: Thanks for verifying!

I'm still wondering if it wouldn't be simpler though to just drive the visualization through the store. Right now you have attachTo in two places. If you just always visualized from the store, instead of sometimes visualizing from the store and sometimes from the result of an Ajax call.

src/app/workflow/dag/wdl-viewer/state/wdl-viewer.model.ts
rjbautis: Regarding the intersection type, does this mean that the store and query should extend an intersection of `WdlViewerPipeline` and `id`?

coverbeck: Looks like it would have to, and may not be worth the effort.

To be clear, my concern was that you have a property, id, that is declared optional, yet your code relies on it always being present, and you don't do null checks.

OTOH, the code is not that complex, and it's unlikely there will be any additional users of this particular code, so it's probably overkill to spend too much time on this.

src/app/workflow/dag/wdl-viewer/wdl-viewer.component.ts
rjbautis: That would be possible if I utilize `takeUntil(this.ngUnsubscribe)` instead of `take(1)` on line 71, and remove `attachTo` from line 90. This code snippet produces the situation described in the original comment where the selectEntity subscription emits at two times: first time, the cached object is null; second time, the cached object exists.

Should this be the way to go?

coverbeck: I think that's the preferred way, yes. Look at [the diagram](https://netbasal.gitbook.io/akita/); in a perfect world the component is getting data via the query.

garyluu:
Can you update and test pipeline-builder@0.3.10-dev.308 to get rid of a few vulnerabilities?

garyluu:
Possibly have to rebase for PR Travis build to pass.

src/app/workflow/dag/wdl-viewer/state/wdl-viewer.model.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Defines types for the response of Pipeline Builder library
+ */
+export interface WdlViewerPipelineResponse {
+  status: boolean;
+  message: string;
+  model: Array<any>;
+  actions: Array<any>;
+
+  // Not originally part of the response object, but all responses should be identified by the workflow version used to generated it
+  workflowVersion?: number;
```
coverbeck: Looking at the code as it is now, this feels weird to me. The only place you use this is in `wdl-viewer-service.update()`. I would suggest getting rid of this from here, and the interface will cleanly model the response you get from the library, and then just adding the workflow version id as an additional parameter to the update method.

denis-yuen:
(pending passing builds)

garyluu:
I restarted the Travis CI PR build and it looks like it passes.  Didn't bother with the push though.

PR #628, opened 2019-04-22
Feature/2386/config

coverbeck:
Endpoint for Ajax requests is now always the same domain. The calls are against `window.location.protocol + '//' + window.location.host + '/api'` -- see app.module.ts.

Load `/config.json` when initializing app.

Use Angular proxy for dev mode.

Several tests that are mocking responses needed to be updated to handle the new endpoint.

For CircleCI, run a barebones nginx, because CircleCI is running UI in prod mode, and the Angular proxy can only be used for dev mode. Remove angular-http-proxy as part of this.

For CircleCI, the nginx.conf needs absolute paths, so use `sed` to help configure.


garyluu:
What with the package-lock.json changes?

coverbeck:
Whoops, shouldn't have committed that. That said, why is that happening, or maybe that's what you're asking? Looks like I am running into [this](https://npm.community/t/package-lock-json-keeps-changing-between-platforms-and-runs/1129)

coverbeck:
Reverted the change

src/app/app.module.ts
```diff
@@ -208,7 +219,7 @@ export class AppModule {
 
 export const apiConfig = new Configuration({
   apiKeys: {},
-  basePath: Dockstore.API_URI
+  basePath: window.location.protocol + '//' + window.location.host + '/api'
```
garyluu: Dockstore.API_URI is changed to this, what is Dockstore.API_URI nowadays?

cypress/integration/group1/checkerWorkflowFromWorkflow.ts
```diff
@@ -55,14 +55,13 @@ describe('Checker workflow test from my-workflows', () => {
       cy.fixture('refreshedChecker').then((json) => {
         cy.route({
           method: 'GET',
-          url: '/workflows/*/refresh',
+          url: '/api/workflows/*/refresh',
```
garyluu: `/api` is supposed to be part of the API URL.  Test changes should not be necessary

garyluu:
Approving because the config.json for all the properties is a definite improvement.  Not a fan of the API_URL though but I don't have a better solution at the moment.  Also assuming s3 + cloudfront doesn't have the same problems as GitHub pages.

cypress/integration/group1/checkerWorkflowFromWorkflow.ts
coverbeck: I like it because it more closely matches the UI invocations in staging/production. But it was unfortunate I had to change all the test files.

garyluu: so for production, if we end up changing the web service to dockstore-webservice.org with no /api, we'll have to get rid of all these again :cry:

coverbeck: Good point.

PR #629, opened 2019-04-23
feature/2291/Add alert service to login

NatalieEO:
ga4gh/dockstore#2291
image

### DIFF
--- a/src/app/login/login.component.html
+++ b/src/app/login/login.component.html
@@ -24,6 +24,9 @@
     </div>
   </div>
 </div>
+<div class="text-center">
+  <app-alert></app-alert>
+</div>
 <div class="container separation">
   <div class="row">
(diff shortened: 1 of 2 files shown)


src/app/login/login.component.html
```diff
@@ -24,6 +24,9 @@ <h3 class="available-containers">
     </div>
   </div>
 </div>
+<div class="text-center">
```
garyluu: Ideally the page would be converted to use the header component for consistency, but this is fine too.

coverbeck:
I can sort of visualize what this would look like, but could you please attach a screenshot to the PR?

NatalieEO:
> I can sort of visualize what this would look like, but could you please attach a screenshot to the PR?

Added pic

PR #631, opened 2019-04-23
Allow spaces before and after form inputs

wshands:
An example of how I approached allowing the user to enter whitespace before and after an input where the UI will remove the whitespace before completing the form request. 

If I do it this way over 28 files will have to be altered in this manner. Is there a more efficient way to do this?

### DIFF
--- a/src/app/container/register-tool/register-tool.service.ts
+++ b/src/app/container/register-tool/register-tool.service.ts
@@ -167,7 +167,7 @@ export class RegisterToolService {
     getImagePath(imagePath, part) {
         /** Defines the regex that an image path (namespace/name) must match.
          Group 1 = namespace, Group 2 = name*/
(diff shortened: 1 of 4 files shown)


garyluu:
Try https://github.com/khashayar/ng-trim-value-accessor

denis-yuen:
Sorry, think this was a bit more of a pain than we expected while discussing. 
I'm not sure this is worth it unless a cleaner/easier way is found/suggested

coverbeck:
What if we just had the validator display a message when there are leading and/or trailing spaces?

I think I've seen such UI on other sites, although I can't remember one right now.

It's less friendly than automatically fixing it for the user, but on the other hand it makes it very clear what is going on, and is easier to implement.

coverbeck:
@wshands What is the status of this? It shows up in my PRs for review, but it is in draft status.

Would it make sense to either close it or move it out of draft status, whichever one applies?

Thanks.

PR #633, opened 2019-04-24
feature/1600/improve-pwa-score

rjbautis:
Improving Dockstore's lighthouse audit score for Progress Web Application, i.e. one aspect of ga4gh/dockstore#1600

* Adds a service worker to Dockstore to control current page and start url
* Caches key assets in the service worker such as material icons, fonts, and js/css

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1597,6 +1597,14 @@
         "tslib": "1.9.0"
       }
     },
+    "@angular/service-worker": {
+      "version": "7.2.13",
+      "resolved": "https://registry.npmjs.org/@angular/service-worker/-/service-worker-7.2.13.tgz",
+      "integrity": "sha512-gMDUYtaxQfJtLLLGSuGEjde7z13eFmWPY/xweNa0+HVyGWOOw3GiP+q/1LUhW2f9gNxBR/2QvcWSEcWYLH9Sgg==",
+      "requires": {
+        "tslib": "^1.9.0"
+      }
+    },
     "@babel/code-frame": {
       "version": "7.0.0",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
@@ -9187,7 +9195,7 @@
       "dependencies": {
         "chalk": {
           "version": "1.1.3",
-          "resolved": "https://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
+          "resolved": "http://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
           "integrity": "sha1-qBFcVeSnAv5NFQq9OHKCKn4J/Jg=",
           "dev": true,
           "optional": true,
@@ -9974,7 +9982,7 @@
       "dependencies": {
         "async": {
           "version": "1.5.2",
-          "resolved": "https://registry.npmjs.org/async/-/async-1.5.2.tgz",
+          "resolved": "http://registry.npmjs.org/async/-/async-1.5.2.tgz",
           "integrity": "sha1-7GphrlZIDAw8skHJVhjiCJL5Zyo=",
           "dev": true
         }
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
           "integrity": "sha512-JTWmoY9tWCs7zvIk/CvRjhjGaOd+OVBM987mxFo+OW66cGpdKjZcpmc74ES1sB//7Kl/PAe8+wEakuhG4pcgOw==",
           "dev": true,
           "requires": {
-            "tslib": "1.9.0"
+            "tslib": "^1.9.0"
           }
         }
       }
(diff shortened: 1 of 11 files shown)


src/app/app.module.ts
```diff
@@ -168,7 +169,8 @@ export const myCustomSnackbarDefaults: MatSnackBarConfig = {
     ApiModule.forRoot(getApiConfig),
     CustomMaterialModule,
     RefreshAlertModule,
-    RequestsModule
+    RequestsModule,
+    ServiceWorkerModule.register('ngsw-worker.js', { enabled: environment.production })
```
coverbeck: How will we test this locally?

agduncan94:
@rjbautis how much does this increase the score by?

src/app/app.module.ts
coverbeck: Also, where is this file? Is this a generic file that is installed via a dependency that goes off the manifest, or does it need to be committed?

src/app/navbar/navbar.component.css
```diff
@@ -45,7 +45,7 @@
 
 .toolbar-holder-sm {
   height: auto !important;
-  margin: 5px;
+  padding: 5px;
```
coverbeck: Why did this change? Can you please add a comment, in the commit message or in the PR explaining why.

src/manifest.json
```diff
@@ -1,20 +1,22 @@
 {
   "name": "Dockstore",
   "short_name": "Dockstore",
-  "description": "Create, Share, Use",  
-  "icons": [{
-    "src": "assets/images/dockstore/dockstore.png",
+  "theme_color": "#21335B",
```
coverbeck: Could you also explain these differences?

Also this is getting really nit-picky, but it would be nice if the diff showed just what you changed without having rearranged everything. If something goes wrong, it would be nice to tell what changed. I think it's that you added scope and start_url?

rjbautis:
> @rjbautis how much does this increase the score by?

The PWA score is actually a [baseline checklist](https://developers.google.com/web/progressive-web-apps/checklist) of 12 items verified by Lighthouse. Currently dockstore.org passes 4/12 items. The changes increase the number (on a local build) to 9/12. 

Factoring in the http/https checklist items that dockstore.org already passes (but fails on local build), the number of passing items should be 11/12

src/app/app.module.ts
rjbautis: `ngsw-worker.js` is generated when the production build is created with `npm build.prod`; I'm not sure how we can test this locally but I can look into it. 

The committed file `ngsw-config.json` is a service worker config file that specifies the caching behaviors for `ngsw-worker.js` to include.

rjbautis: From what I've read, the service worker only works in production apps

src/app/navbar/navbar.component.css
rjbautis: One PWA is verifying that the content is sized correctly for the viewport. Specifically for mobile viewports.

Margin affects the outer space of an element, which increases the `innerWidth` of the window's content area but not the `outerWidth` of the browser's window. As is, the `innerWidth !== outerWidth` because margin added `5px` to the content area

Padding affects the inner space of an element but does not increase the `innerWidth`, so the viewport content is correctly sized.

coverbeck: Thanks for the explanation! I'm aware of the difference between padding and margin in general, but didn't know why this was being done in particular. I think you should add this to the commit message when you merge.

I'm also a little wary that this might introduce some unintended styling changes on a page with the navbar. Do pages visually look the same with and without this change?

src/manifest.json
rjbautis: Sorry about that, the committed `manifest.json` was auto-generated by Angular CLI when the service worker was installed. Since this worked with the prod build, I deleted the old manifest file and manually moved the important info to the new file

coverbeck: Ah, if it's a generated file, then it may not be worth the effort

src/app/navbar/navbar.component.css
rjbautis: The difference is noticeable but also pretty negligible. Left is local, right is dockstore.org

Screen Shot 2019-04-24 at 2 37 37 PM

package-lock.json
```diff
@@ -3701,7 +3709,7 @@
         },
         "util": {
           "version": "0.10.3",
-          "resolved": "https://registry.npmjs.org/util/-/util-0.10.3.tgz",
+          "resolved": "http://registry.npmjs.org/util/-/util-0.10.3.tgz",
```
garyluu: check why your npm is downgrading to http from https

rjbautis:
There's an ongoing open issue on the Cypress repo (as of 05/29/2019) https://github.com/cypress-io/cypress/issues/702, which talks about service workers causing confusing behavior.

Some of our integration tests that involve network request stubbing are not passing consistently due to the service worker. I exhausted some possible solutions like unregistering the sw and clearing the cache before each test, but the tests are still flaky

denis-yuen:
Likely obselete

PR #635, opened 2019-04-25
Feature/2003/starring orgs

NatalieEO:
ga4gh/dockstore#2003
I know this needs to be abstracted out...


src/app/organizations/organization/organization-starring/organization-starring.component.ts
```diff
@@ -0,0 +1,157 @@
...
+  // Determine if user has starred organization.
+  calculateRate(starredUsers: User[]): boolean {
+    if (!this.user || !starredUsers) {
+      return false;
+    } else {
+      let matchingUser: User;
+      if (!starredUsers) {
+        return false;
+      }
+      matchingUser = starredUsers.find(user => user.id === this.user.id);
```
coverbeck: You can simplify this whole block to `return starredUsers.some(user => user.id === this.user.id);`

src/app/organizations/organization/organization-starring/organization-starring.component.ts
```diff
@@ -0,0 +1,157 @@
...
+   */
+  setStarring() {
+    this.disable = true;
+    if (this.isLoggedIn) {
+
+      if (this.rate) {
+        const message = 'Unstarring ' + this.organization.name;
+        this.alertService.start(message);
+      } else {
+        const message = 'Starring ';
```
coverbeck: Did you mean to say `Starring <organization name>` similar to how you did on line 107?

Also, FYI on that line and here if you add it, TypeScript has string interpolation, so you can optionally do `` `Unstarring ${this.organization.name}`` `.

coverbeck: Also, I'm not wild about having the same variable declared twice so close to each other; even though they are in their own blocks and this is legal. I would probably inline and not have the variable. And/or only invoke `this.alertService.start` only once.

Just my personal opinion, so feel free to ignore.

src/app/organizations/organization/organization-starring/organization-starring.component.ts
```diff
@@ -0,0 +1,157 @@
...
+
+        },
+        (error) => {
+          this.alertService.detailedError(error);
+          this.disable = false;
+        });
+    }
+  }
+
+  setStar(): any {
```
coverbeck: Last comment on `any`, as in, I'm not going to flag other instances, if any :) Please try to avoid using it whenever possible.

src/app/organizations/organization/organization-starring/organization-starring.component.ts
```diff
@@ -0,0 +1,144 @@
...
+export class OrganizationStarringComponent implements OnInit, OnDestroy, OnChanges {
+
+  @Input() organization: Organization;
+  @Output() change: EventEmitter<boolean> = new EventEmitter<boolean>();
+  private user: User;
+  public isLoggedIn: boolean;
+  public rate = false;
+  public total_stars = 0;
+  public disable = false;
+  private ngUnsubscribe: Subject<{}> = new Subject();
```
coverbeck: Gary already commented on this below,  but you can just have your class `extend Base`, and then you don't need this either.

src/app/shared/starring.ts
```diff
@@ -0,0 +1,25 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+import { User } from './swagger';
+
+export function calculateRate(starredUsers: User[], currentUser: User): boolean {
```
coverbeck: The name of this function is sort of confusing given the return type -- when I first saw it I was expecting a rate, e.g., a number. 

I see it was already like this and you moved it here for reuse... I still find it confusing. :)

src/app/starredentries/starredentries.component.html
```diff
@@ -51,21 +62,52 @@ <h3 class="starred-header">Starred Workflows</h3>
...
+          <mat-card class="alert alert-info starred-header" role="alert" *ngIf="starredOrganizations.length==0">
+            <mat-icon>warning</mat-icon>
+            You have no starred organizations.
+          </mat-card>
+        </div>
+
+        <mat-list>
+          <mat-list-item *ngFor="let organization of starredOrganizations; let last = last;">
+            <h3 class="spacing" mat-line>
+              <a href="/organizations/{{organization.name}}" [matTooltip]="organization?.name">{{organization.displayName}}</a>
```
coverbeck: These should use routerLink instead of href so that navigation stays inside the SPA.

src/app/starredentries/starredentries.component.html
```diff
@@ -51,21 +62,52 @@ <h3 class="starred-header">Starred Workflows</h3>
...
+            </h3>
+            <span mat-line>
+            <small class="spacing">Email: <a [href]="'mailto:' + organization.email">{{ organization.email }}</a></small>
+            <small class="spacing">Website: <a [href]="organization?.link">{{ organization?.link }}</a> </small>
```
coverbeck: Should this have `_target="blank"`? We're not consistent on it, but I tend to think we should.

src/app/organizations/organization/organization-starring/organization-starring.component.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.ngUnsubscribe.next();
+    this.ngUnsubscribe.complete();
+  }
+
+  setUpOrganization(organization: Organization) {
+    this.organization = organization;
+    this.getStarredUsers();
+  }
+
+  // Determine if user has starred organization.
```
coverbeck: Unless you expect to restore this block, you should just delete it.

src/app/organizations/organization/organization.component.html
```diff
@@ -60,13 +65,14 @@ <h1 [matTooltip]="org?.name">{{org.displayName}}</h1>
           </span>
           <span fxLayout="row" fxLayoutGap="10px" *ngIf="org?.link">
             <mat-icon>link</mat-icon>
-            <span><a [href]="org.link" target="_blank">{{ org.link }}</a></span>
+            <span><a [href]="org.link" target="_blank">{{ org?.link }}</a></span>
```
coverbeck: You should add `rel="noopener"`. See [this](https://developers.google.com/web/tools/lighthouse/audits/noopener)

src/app/organizations/organization/organization-starring/organization-starring.component.ts
```diff
@@ -0,0 +1,144 @@
...
+})
+export class OrganizationStarringComponent implements OnInit, OnDestroy, OnChanges {
+
+  @Input() organization: Organization;
+  @Output() change: EventEmitter<boolean> = new EventEmitter<boolean>();
+  private user: User;
+  public isLoggedIn: boolean;
+  public rate = false;
+  public total_stars = 0;
+  public disable = false;
```
coverbeck: I don't think this property is used at all. I was going to comment that it's poorly named, that it have a more specific name, e.g., disableRateButton, so I was looking to see where it's used in order to suggest a better one, and I don't think it's used in HTML anywhere, and all you do in this file is set it, but never read it.

src/app/organizations/organization/organization-starring/organization-starring.component.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    TestBed.configureTestingModule({
+      imports: [MatIconModule, MatSnackBarModule],
+      declarations: [ OrganizationStarringComponent ],
+      providers: [
+        {provide: TrackLoginService, useClass: TrackLoginStubService},
+        {provide: OrganizationStarringService, useClass: OrganizationStarringStubService},
+        {provide: StarentryService, useClass: StarEntryStubService},
+        {provide: ContainerService, useClass: ContainerStubService},
+        {provide: StarringService, useClass: StarringStubService},
+        { provide: StarOrganizationService, useClass: StarOrganizationStubService }
```
coverbeck: Formatting inconsistent with previous lines.

src/app/organizations/organization/organization-starring/organization-starring.component.ts
NatalieEO: Realized this is used in starring.component.html. Put it back and renamed to what you suggested

coverbeck: Sorry about that; my WebStorm Find Usages let me down. :( Or I read the output incorrectly.

src/app/starredentries/starredentries.component.html
```diff
@@ -1,6 +1,6 @@
 <app-header>
   <mat-icon>star</mat-icon>
-  Starred Tools and Workflows
+  Starred Tools, Workflows, and Organizations
```
agduncan94: I think `Starred` or `My Starred` works better, as the tools, workflows, and orgs are mentioned in the tabs already.

PR #638, opened 2019-04-25
2354/file not showing

agduncan94:
If not ace editor with given ID found, don't bother doing stuff to it (stops error message).

Lazy load descriptor tab for workflows and dockerfile tab for tools for https://github.com/dockstore/dockstore/issues/2354


agduncan94:
@coverbeck I did not get the chance.

PR #643, opened 2019-04-29
Feature/2348/aws logs

garyluu:
This is one of the 3 PRs for dockstore/dockstore#2348

It adds a new dialog for detailed verification information + logs for files that have been tested by Dockstore's ToolTester.  Removes the previous verification information into this modal.

Only Phase 1.  Also missing integration tests

Also:
- Fixes the count error for ngx-sharebuttons by outright disabling it (we never had count working anyways)
- Adds openapi 3 support.  There will be two directories src/app/shared/swagger and src/app/shared/openapi .  When importing services, make sure your auto-import is importing the correct one because there will be duplicate services where it only works with a specific one.


garyluu:
Codacy, you so silly

src/app/shared/state/checker-workflow.query.ts
```diff
@@ -44,11 +45,29 @@ export class CheckerWorkflowQuery extends Query<CheckerWorkflowState> {
    * Returns true if entry is a workflow, returns false if entry is not a workflow (i.e. a tool)
    * @param entry The entry to check
    */
-  public isEntryAWorkflow(entry: Entry): boolean {
+  public isEntryAWorkflow(entry: Entry | null): boolean {
```
coverbeck: If you're declaring the parameter can be null, then shouldn't you add to the return type as well? Not sure if this is common or not.

Also, although it was like this already, I find it slightly weird that a function named like a boolean function and returning a boolean can return a `null`. It's really a function that returns 3 states.

PR #647, opened 2019-05-01
Feature/2442/circle ci

garyluu:
For dockstore/dockstore#2442

My commit messages are messed up :cry: 

- Removing Travis CI and Semaphore CI (Circle CI is fast enough to not need Semaphore CI)
- Only use Circle CI
- 22m14s with CircleCI vs 78m11s with Travis CI but CircleCI has 4 jobs whereas Travis CI has 5
- Removed Cypress from package.json and only installing the latest one in tests (can revert this change)
- Attempt number 4 at trying to fix the postgres issue
- Moved running the webservice outside the run-webservice script because of CircleCI being weird with background tasks
- Still missing npm caching (would save 40 seconds in total)
- **Running Cypress tests with a production build (reaching the memory limit of CircleCI otherwise)**
- Had to downgrade the pipeline-builder because it doesn't work in production


coverbeck:
> Removed Cypress from package.json and only installing the latest one in tests (can revert this change)

The README.md should be updated to reflect how to run Cypress locally now that it won't be installed automatically

.circleci/config.yml
```diff
@@ -1,52 +1,125 @@
...
+      - install_container_dependencies
+      - run:
+          name: Install dependencies
+          command: bash -i -c 'npm ci'
+      - run:
+          name: Check dependencies
+          command: bash -i -c 'npm ls'
+      - run:
+          name: Lint
+          command: bash -i -c 'npx ng lint'
```
coverbeck: Or `npm run lint`

PR #651, opened 2019-05-03
Feature/logged off page

agduncan94:
New homepage geared towards logged off/new users. Please check out the branch and try locally.

Open to feedback on the layout, what is missing, changes to text and whatnot. Note that any images are purely placeholders.

Below are some screenshots.

Screen Shot 2019-05-03 at 2 30 57 PM

Screen Shot 2019-05-03 at 2 31 45 PM

Screen Shot 2019-05-03 at 2 31 48 PM

Screen Shot 2019-05-03 at 2 31 52 PM

Screen Shot 2019-05-03 at 2 31 58 PM


coverbeck:
Going off screenshots only, initial thoughts:

1. Overall, looks really great
1. The `Browse tools and workflows` doesn't look clickable.
1. On the `Developers` and `Scientists` buttons, some contradictory thoughts:
    1. I wonder if it should be singular
    1. Or, maybe `I am a developer` and `I am scientist`
    1. I wonder if `developer` is still too overloaded of a word. Arguably what scientists is also development (?). What about if there was a brief description, like `I want to create and share workflows` and `I want to find a workflow for use in my research`? Terrible wording in general, and the longer verbiage takes is not good. On the other hand, while we know what we mean by developers and scientists, I'm wondering if a user coming in from the cold will know what he/she is. Not sure; I could be overthinking it.

agduncan94:
@coverbeck 
2. I'll add an underline to browse tools and workflows
3. I like the `I am an x` idea. I agree that developer vs scientist has some overlap and may be confusing. I like what Cricket had said in the meeting about differentiating between producers and consumers. I am not sure what the corresponding words would be, or if such words exist. Maybe a short phrase is better, as you stated.

Maybe something like `Create workflows` and `Use/Launch/Run Workflows`.

src/app/home/home.component.ts
```diff
@@ -56,6 +59,7 @@ export class HomeComponent implements OnInit, AfterViewInit {
 
   ngOnInit() {
     this.user$ = this.userQuery.user$;
+    this.userQuery.user$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(user => this.user = user);
```
garyluu: No need to subscribe just for the username.
Make username another observable or just (user$ | async).username
Could've extend `Base` to avoid things like missing onDestroy

src/app/home/home.component.scss
```diff
@@ -76,3 +76,87 @@ span {
...
+     display: none !important;
+  }
+
+  .flex-column-on-mobile {
+    flex-direction: column !important;
+  }
+}
+
+.youtube {
+  z-index: 1000;
```
garyluu: Use https://getbootstrap.com/docs/4.3/layout/overview/#z-index values.  Put the new class above
```
.bootstrap-tooltip-z-index {
  z-index: 1070;

}
```

src/app/home/home.component.html
```diff
@@ -16,103 +16,204 @@
...
+        <div fxLayout="row wrap" fxLayoutAlign="center center" fxLayoutGap="20px" class="pb-5">
+          <button mat-raised-button>
+            <a href="#developerIntroduction" class="button-link">Developers</a>
+          </button>
+          <button mat-raised-button>
+            <a href="#scientistIntroduction" class="button-link">Scientists</a>
+          </button>
+        </div>
+
+        <!-- <div fxLayout="row wrap" fxLayoutAlign="center" fxLayoutGap="20px">
```
garyluu: Comment on this

src/app/home/home.component.scss
```diff
@@ -76,3 +76,87 @@ span {
   margin-bottom: 2px;
   font-weight: 700;
 }
+
+.button-link {
+  text-decoration: none !important;
```
garyluu: Comment on what the `!important` is overriding

src/app/home/home.component.html
```diff
@@ -16,103 +16,204 @@
...
+
+<div id="developerIntroduction" class="container">
+  <h2 class="text-center">Developers</h2>
+  <p class="lead text-center">
+    Make your tools more accessible to the scientific community.
+  </p>
+  <div fxLayout="row wrap" fxLayoutGap="30px" fxLayoutAlign="center center" class="developer-holder flex-column-on-mobile">
+    <div fxFlex>
+      <div fxLayout="column">
+        <mat-icon fxFlexAlign="center" style="font-size: 50px; height: 50px">account_balance</mat-icon>
```
garyluu: style repeated enough to deserve a class (which applies the style to all mat-icons within the div)

src/app/home/home.component.html
```diff
@@ -16,103 +16,204 @@
...
+    </div>
+
+    <div fxFlex fxLayout="column" fxLayoutGap="30px" class="m-3 hidden-on-mobile" fxLayoutAlign="center center">
+      <img src="../assets/images/dockstore/graph-shadow.png">
     </div>
   </div>
+
 </div>
 
-<app-home-foot-note></app-home-foot-note>
```
garyluu: Delete component if no longer used

src/app/home/home.component.html
```diff
@@ -16,103 +16,204 @@
...
-                <a mat-raised-button color="accent" href="https://discuss.dockstore.org/">Discuss <mat-icon>chevron_right</mat-icon></a>
-              </div>
-            </div>
-        </div>
+    </div>
+    <div fxFlex>
+      <div fxLayout="column">
+        <mat-icon fxFlexAlign="center" style="font-size: 50px; height: 50px">description</mat-icon>
+        <h3 class="text-center">Describe tools and workflows</h3>
+        <p>Programs are described with CWL, WDL or Nextflow and stored on GitHub, BitBucket, or GitLab.</p>
```
garyluu: Find and replace all "BitBucket" with "Bitbucket"

src/app/home/home.component.html
```diff
@@ -16,103 +16,213 @@
 
 <div class="header">
   <div class="container">
-    <div class="row header1">
-      <div class="col-md-6 splash">
```
garyluu: Remove slash class if no longer used

src/app/home/home.component.html
```diff
@@ -16,103 +16,209 @@
...
-            </span>
-          </tab>
-        </tabset>
+
+<div class="container stripe-holder">
+  <div fxFlex fxLayout="column" fxLayoutGap="20px">
+    <h3 class="text-center">Share your tools now</h3>
+
+    <div fxLayout="row wrap" fxLayoutAlign="center center">
+      <a mat-stroked-button href="https://docs.dockstore.org/docs/prereqs" target="_blank" class="mr-2">
```
garyluu: fxLayoutGap instead of mr-2 on two of them

src/app/home/home.component.html
```diff
@@ -44,10 +43,11 @@ <h2 class="title caption">Docker Tools and Workflows for the Sciences</h2>
       <div fxLayout="column wrap" fxLayoutAlign="center center" fxLayoutGap="20px">
         <h3 class="search-title">New to Dockstore? Start here</h3>
         <div fxLayout="row wrap" fxLayoutAlign="center center" fxLayoutGap="20px" class="pb-5">
-          <a mat-raised-button href="#developerIntroduction" class="button-link">Developers</a>
-          <a mat-raised-button href="#scientistIntroduction" class="button-link">Scientists</a>
+          <a mat-raised-button href="#developerIntroduction" class="button-link">Create and share</a>
```
coverbeck: You'll get different opinions on this, but I don't like "Create and share" and "Discover and launch". If I'm a brand new user, that's still too vague to me, when I'm trying to figure out what Dockstore is. Maybe it's just harder for me, because I don't fall into either category.

I prefer what you had before, but again, I'm probably not the target audience.

src/app/home/home.component.html
```diff
@@ -88,7 +88,7 @@ <h3 class="text-center">Reproducible environment</h3>
       <div fxLayout="column">
         <mat-icon fxFlexAlign="center" style="font-size: 50px; height: 50px">description</mat-icon>
         <h3 class="text-center">Describe tools and workflows</h3>
-        <p>Programs are described with <a href="https://www.commonwl.org/" target="_blank">CWL</a>, <a href="http://openwdl.org/" target="_blank">WDL</a> or <a href="https://www.nextflow.io/" target="_blank">Nextflow</a> and stored on <a href="https://github.com/" target="_blank">GitHub</a>, <a href="https://bitbucket.org/" target="_blank">BitBucket</a>, or <a href="https://www.gitlab.com" target="_blank">GitLab</a>.</p>
+        <p>Programs are described with <a href="https://www.commonwl.org/" target="_blank">CWL</a>, <a href="http://openwdl.org/" target="_blank">WDL</a> or <a href="https://www.nextflow.io/" target="_blank">Nextflow</a> and stored on <a href="https://github.com/" target="_blank">GitHub</a>, <a href="https://bitbucket.org/" target="_blank">Bitbucket</a>, or <a href="https://www.gitlab.com" target="_blank">GitLab</a>.</p>
```
coverbeck: It was like this already, but since you're in there... You should add `rel="noopener noreferrer"` when using `target=_blank`.

src/app/home/home.component.html
agduncan94: I'll revert to what I had before, and we can further discuss naming and stuff in the google docs I am making.

src/app/home/home.component.html
```diff
@@ -16,103 +16,209 @@
...
+
+    <div fxLayout="row wrap" fxLayoutAlign="center center" fxLayoutGap="20px">
+      <a mat-stroked-button href="https://docs.dockstore.org/docs/prereqs" target="_blank" rel="noopener noreferrer">
+        Getting Started Guide
+      </a>
+      <button mat-stroked-button routerLink="/organizations">
+        Discover Organizations
+      </button>
+      <button mat-stroked-button routerLink="/search">
+        Discover Tools and Workflows
```
NatalieEO: When you click both discover buttons (here and in the scientist section), the new page opens up scrolled to the bottom.

src/app/home/home.component.html
```diff
@@ -16,103 +16,209 @@
...
+      <img src="../assets/images/dockstore/scientist-flow.png">
     </div>
   </div>
+
+  <div class="text-center">
+    <p>
+      Or, use the Dockstore CLI to run tools and workflows locally.
+    </p>
+    <button mat-stroked-button color="accent" routerLink="/quick-start">
+      Setup CLI
```
NatalieEO: This button also opens the page scrolled to the bottom.

src/app/home/home.component.html
```diff
@@ -16,103 +16,209 @@
...
+        </button>
+
+        <ng-template #loggedIn>
+          <button mat-flat-button color="accent" routerLink="/accounts">
+            Welcome {{ (user$ | async).username }}
+          </button>
+        </ng-template>
+
+        <button mat-button routerLink="/search" class="underlined-link">
+          Browse tools and workflows
```
NatalieEO: Total opinion from me so up to you but it's kind of weird to me to see a button and link ) right next to each other (because they're different formats). Since the search bar is below and "browse tools and workflows" also takes you to the same page, I think I would like it more either

1. Browse tools and workflows wasn't a link anymore, but on a new line and the search bar would be implied that that's how you browse them.
2. Get rid of the browse tools and workflows line and maybe change the text inside the search bar from "Search..." to "Browse tools and workflows"

NatalieEO:
Also, this looks so much better than before!

src/app/home/home.component.html
agduncan94: With the `browse tools and workflows` button I was trying to make it less prominent, as the major thing we would want is to get the user to sign up, however I still wanted to make it obvious to the user how to get to the search page. Perhaps a more neutral colour for the `Browse tools and workflows button` and make it a button and not just a link.

The issue with removing the button and only having the search bar is that it requires the user to know what they are searching for. They may not know exactly what they need and are just trying to see what tools/workflows we offer.

agduncan94: browse-button

NatalieEO: Ah yeah that makes sense. It definitely looks better to me as a button in that screenshot.

david4096:
I really like this!

Suggests to me a dynamic counter of how many workflows there are.

"Browse from over 400 workflows" or, a counter that says, "7 workflows added this week". "featured workflow" or some kind of spinner with a list of workflows or orgs.

Instead of "Create an Account" we might have the signup box right on that page (log in with GH/Google logo center of screen).

One consideration is that we currently DO support a lot of features for CLI only consumers. You can search and run without having created an account and I wonder if we are overemphasizing web based clients. Would a box with code like `wget dockstore.jar && dockstore workflow run hello-world` prove interesting to first time visitors?

agduncan94:
@david4096 I like the idea of a counter, would be cool to see the total count.

For the signup box, I took it out because I thought there might be too much going on with the homepage (at least the top visible section). Are you thinking a signup box in the top section of the homepage? Would changing `create an account` to something like `register with github/google` work?

I quite like the CLI focus idea, doing something similar to what nextflow does https://www.nextflow.io/. Get them to install it very quick, and then can direct them to tutorials or tools/workflows.

david4096:
@agduncan94 regarding the login page, the idea is to reduce the number of clicks before someone wants to start, putting the GH and Google login buttons themselves lowers click inertia since the client is accustomed to that login logo. You'll get some of the people drop-off if you have a to click "make an account" then "login with GH". I think that two "register with github" "register with google" links are what I'm imagining.

src/app/home/home.component.html
Ldcabansay: I like the button. Will this new user/not logged in view still have the 'My Tools' and 'My Workflows' on the top right?

agduncan94: Only logged in users will see my tools/workflows, so a not logged in view will not have the my tools/workflows.

cricketsloan:
I agree that we would like to show that accounts are created with GH or Google.  That will lower the inertia.


I also miss the graphic but I think you have that on your todo list.


Do we want a 3rd category that is Organization?  And show that you can Collect and curate your workflows?

agduncan94:
Round 2 of screenshots. Main difference is instead of having developers and scientists, we use the `create, share, use` to explain Dockstore.

Screenshot from 2019-05-08 14-56-53
Screenshot from 2019-05-08 14-57-07
Screenshot from 2019-05-08 14-57-21
Screenshot from 2019-05-08 14-57-36

NatalieEO:
> Round 2 of screenshots. Main difference is instead of having developers and scientists, we use the `create, share, use` to explain Dockstore.
I like both, but I think I like the flow of this one more.

david4096:
@agduncan94 Yep, that's the idea!

src/app/home/home.component.html
```diff
@@ -16,103 +16,219 @@
...
+  <h2 class="text-center">Create</h2>
+  <p class="lead text-center">
+    Create reproducible workflows using cutting-edge technology and standards.
+  </p>
+  <div fxLayout="row wrap" fxLayout.lt-md="column" fxLayoutGap="30px" fxLayoutAlign="center center"  class="developer-holder">
+    <div fxFlex>
+      <mat-card fxLayout="column" class="small-card">
+        <h3 class="text-center">Reproducible environment</h3>
+        <p>
+          The execution environment is described by <a href="https://www.docker.com/" target="_blank" rel="noopener noreferrer">Docker</a> and stored and built on <a href="https://quay.io/" target="_blank" rel="noopener noreferrer">Quay.io</a> or <a href="https://hub.docker.com/" target="_blank" rel="noopener noreferrer">DockerHub</a>.
```
garyluu: Docker Hub

agduncan94:
@david4096 Gary brought up a good point that there are four separate login buttons.
1. Sign up with Google
2. Sign up with GitHub
3. Login with Google
4. Login with GitHub

Due to this, we can't really only have two buttons on the homepage as they wouldn't always go to the direct place. We would have to have four buttons in total. I don't think this is a good idea.

I see there being two options:
1. Change from `Create an account` button to `Login with Google/GitHub`, which will take you to the login page. At least this way users still instantly know that you can sign in with Google/GitHub.
2. Embed the `/login` screen somewhere on the homepage. I think this would add too much content to the homepage and the user would be overwhelmed.

I think for now I am leaning towards option 1.

agduncan94:
@denis-yuen Would a button saying view on twitter work too? or do you want to keep the widget on the homepage. I am just struggling to figure out where to put it as it looks bad taking up a lot of width.

agduncan94:
Here is the final version for this PR. I have removed images and added twitter. I also removed the highlighted features section and moved some of that content to other sections.

Note that updates to the text will come in a following PR, so it shouldn't necessarily be a blocker (unless there are big issues).

Will need to find a better way to display twitter/discourse but for now they will go at the bottom.

Screenshot from 2019-05-15 14-34-25
Screenshot from 2019-05-15 14-34-40
Screenshot from 2019-05-15 14-35-04
Screenshot from 2019-05-15 14-35-16

cricketsloan:
Side note:  DataStage really wants their logo on this page somehow.  @denis-yuen likely has opinions.  I would not have brought it up but I am looking at the page and how the logos don't have space to grow

cricketsloan:
Should the video be more prominent?  It seems like it gets hidden

I thought that we agreed to make the "Docker Tools and Workflows for the Sciences" to be more emphasized

Maybe these comments can go into the content tickets

agduncan94:
@cricketsloan I wanted to get this merged in so that we could see it in our day to day development and make changes as we see. I did increase the size of the "Docker tools and workflows for the sciences" though probably not enough it seems.

We could change the button so that the video stands out more. When you click next steps it scrolls down with the video button featured at the top.

I think we should put all these ideas into a ticket, so I've made this https://github.com/dockstore/dockstore/issues/2487

PR #652, opened 2019-05-06
DRAFT: An example check for null assignments

wshands:
A draft pull request that shows an example of an approach to fixing the null checks errors. Is this a good way to fix them? I would follow this approach in other files; and there might be around 30 files to change.


agduncan94:
It seems like a lot of work over 30 files, but I cannot think of a better way to do it.

src/app/aliases/aliases.component.ts
```diff
@@ -18,8 +18,8 @@ export class AliasesComponent extends Base implements OnInit {
   organization$: Observable<Organization>;
   collection$: Observable<Collection>;
 
-  public type: string;
-  public alias: string;
+  public type: string | null;
```
coverbeck: I would prefer `public type?: string`. Can't say exactly why, but it feels better to me.

@garyluu In the GitHub issue, you suggested this approach. Why do you like it?

garyluu: I think I only prefer it in interfaces (like ExtendedWorkflow.ts) where the property is optional because the compiler would complain when you try to create the object without that property.

wshands:
Closing to submit as non draft PR

src/app/aliases/aliases.component.ts
```diff
@@ -32,20 +32,26 @@ export class AliasesComponent extends Base implements OnInit {
...
+    /*
+    let myType: string | null;
+    myType = this.route.snapshot.paramMap.get('type') ;
+    // Avoid assigning null to this.type
+    this.type = (myType) ? myType : '';
+     */
+    this.type = this.route.snapshot.paramMap.get('type') ;
     this.alias = this.route.snapshot.paramMap.get('alias');
-    this.validType = this.types.includes(this.type);
+    this.validType = (this.type) ? this.types.includes(this.type) : false;
```
coverbeck: Isn't this redundant? Won't `this.types.includes(null)` already return `false`?

src/app/aliases/aliases.component.ts
```diff
@@ -32,20 +32,26 @@ export class AliasesComponent extends Base implements OnInit {
...
 
   ngOnInit() {
-    this.type = this.route.snapshot.paramMap.get('type');
+    /*
+    let myType: string | null;
+    myType = this.route.snapshot.paramMap.get('type') ;
+    // Avoid assigning null to this.type
+    this.type = (myType) ? myType : '';
+     */
+    this.type = this.route.snapshot.paramMap.get('type') ;
```
coverbeck: It seems like this was an experiment and all of this can be removed?

PR #653, opened 2019-05-06
Feature/discourse topics

agduncan94:
No longer generates a discourse topic if one does not exist on first visit. Instead shows discourse topic only if one is set in the database. If not, will show an error.

Edit:
https://github.com/dockstore/dockstore/issues/1816


src/app/container/container.component.html
```diff
@@ -185,7 +185,15 @@ <h3>
         </mat-tab>
       </mat-tab-group>
     </div>
-    <div *ngIf="publicPage" class="mt-2" id='discourse-comments'></div>
+
+    <div *ngIf="publicPage" class="mt-2 mr-3">
+      <div *ngIf="tool?.topicId !== null else noTopicId" id='discourse-comments'></div>
+      <ng-template #noTopicId>
+        <mat-card class="alert alert-warning" role="alert">
+          <mat-icon>info</mat-icon> No Discourse topic exists for this tool. Contact a Dockstore curator to create a topic.
```
coverbeck: How do you contact a Dockstore curator? I don't even know myself. :) Could this be a link?

src/app/shared/entry.ts
```diff
@@ -353,10 +343,10 @@ export abstract class Entry implements OnInit, OnDestroy, AfterViewInit {
   }
 
   /**
-   * Deals with redirecting to canonical URL and running discourse call
+   * Deals with redirecting to canonical URL
    * @return {void}
    */
-  redirectAndCallDiscourse(myPage: string): void {
+  redirectToCanonicalURL(myPage: string): void {
```
coverbeck: I don't think this method is doing anything now, is it? It calculates some things, but it doesn't do anything with any of the variables. Or am I overlooking something?

src/app/workflow/workflow.component.html
```diff
@@ -190,7 +190,15 @@ <h3>
       </mat-tab>
     </mat-tab-group>
   </div>
-  <div *ngIf="publicPage" class="mt-2" id='discourse-comments'></div>
+
+  <div *ngIf="publicPage" class="mt-2 mr-3">
+    <div *ngIf="workflow?.topicId !== null else noTopicId" id='discourse-comments'></div>
+    <ng-template #noTopicId>
+      <mat-card class="alert alert-warning" role="alert">
+        <mat-icon>info</mat-icon> No Discourse topic exists for this workflow. Contact a Dockstore curator to create a topic.
```
coverbeck: Same comment about contacting a Dockstore curator

src/app/container/container.component.ts
```diff
@@ -98,7 +99,30 @@ export class ContainerComponent extends Entry {
...
-    this.redirectAndCallDiscourse('/my-tools');
+    this.redirectToCanonicalURL('/my-tools');
+  }
+
+  ngAfterViewInit() {
+    if (this.publicPage) {
+      this.toolQuery.tool$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(
+        tool => {
+
+        if (tool && tool.topicId) {
```
coverbeck: This is probably not a scenario, but can either these be set and then not set as the events fire? If so, the DiscourceEmbed would end up pointing to an older topic.

src/app/workflow/workflow.component.ts
```diff
@@ -92,13 +93,36 @@ export class WorkflowComponent extends Entry {
...
     this.isRefreshing$ = this.alertQuery.showInfo$;
     this.descriptorType$ = this.workflowQuery.descriptorType$;
   }
 
+  ngAfterViewInit() {
+    if (this.publicPage) {
+      this.workflowQuery.workflow$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(
+        workflow => {
+
+        if (workflow && workflow.topicId) {
```
coverbeck: Same as comment for tools

src/app/shared/entry.ts
agduncan94: It updates the path to conform to certain characteristics, though I do think there is code that can be deleted.

src/app/container/container.component.html
agduncan94: That is a good question :laughing:

src/app/shared/entry.ts
coverbeck: Oh, I see, it calls `this.decodeURL()`, which changes a property.

Never mind...

agduncan94: I do agree we probably need to do some reorganising of this code, but for now I'll leave it as is.

src/app/container/container.component.ts
agduncan94: I do not think this would happen, but I don't want to say with absolute certainty that it can't. Unfortunately I can't really do much testing until this is one staging.

src/app/workflow/workflow.component.html
```diff
@@ -190,7 +190,15 @@ <h3>
       </mat-tab>
     </mat-tab-group>
   </div>
-  <div *ngIf="publicPage" class="mt-2" id='discourse-comments'></div>
+
+  <div *ngIf="publicPage" class="mt-2 mr-3">
+    <div *ngIf="workflow?.topicId !== null else noTopicId" id='discourse-comments'></div>
```
coverbeck: Can the topicId be an empty string if it's not set or will it be null? If there's any doubt, you should change this to `*ngIf="workflow?topicId else noTopicId`.

src/app/container/container.component.html
```diff
@@ -185,7 +185,15 @@ <h3>
         </mat-tab>
       </mat-tab-group>
     </div>
-    <div *ngIf="publicPage" class="mt-2" id='discourse-comments'></div>
+
+    <div *ngIf="publicPage" class="mt-2 mr-3">
+      <div *ngIf="tool?.topicId !== null else noTopicId" id='discourse-comments'></div>
```
coverbeck: Same comment as in workflow.component.html.

src/app/workflow/workflow.component.html
agduncan94: It should either be null or set, but if that handles both cases I'll change it.

PR #654, opened 2019-05-08
Feature/2290/google tag manager

garyluu:
For dockstore/dockstore#2290

   This injects the google tag manager script into the index.html HEAD.
   Injecting it is slower than having it in the HTML, but because the Google Tag Manager ID changes,
   we would've needed multiple index.html and multiple builds.
   The initial page load time will always be inaccurate but we have lighthouse to check that.
   TODO: Pull the Google Tag Manager Id from the config.json

Also:
- Increased timeout for Cypress checks so that it doesn't time out for star entry (because webservice is too slow)
- Added some checks to better help figure out why postgres fails to reset DB


garyluu:
> This looks good, but I wonder if we shouldn't have everything in place before merging it:
> 
> 1. The `/metadata/config.json` endpoint serving up the GTM id. Both the webservice serving it, and compose_setup up populating the dockstore.yml.
> 2. The code that reads the config.json, my #628, whatever we agree on.
> 3. The general approach on how we will be handling boostrapping the UI.
> 
> Otherwise the branch will be in a state where the GTM stuff is not working.
> 
> Your call...

Ya, that's true.  I'll sit on this for a while the other stuff is done.  But in any case, it'll either:
- Get the ID from the environment files
- Get it from the config.json locally
- Get it from the webservice

PR #658, opened 2019-05-13
feature/2392/org UI fixes

NatalieEO:
dockstore/dockstore#2392
Stargazers visually consistent for tools/workflows/organizations
Made stargazer images smaller
remove redundant titles
don't show info on my starred organizations if info null

Screen Shot 2019-05-14 at 9 36 13 AM
Screen Shot 2019-05-14 at 9 35 19 AM

### DIFF
--- a/src/app/organizations/organization/organization.component.ts
+++ b/src/app/organizations/organization/organization.component.ts
@@ -27,6 +27,7 @@ import {
   UpdateOrganizationOrCollectionDescriptionComponent
 } from './update-organization-description/update-organization-description.component';
 import { UserQuery } from '../../shared/user/user.query';
+import { DateService } from '../../shared/date.service';
 
 @Component({
   selector: 'organization',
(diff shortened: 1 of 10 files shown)


coverbeck:
Could you please attach a screenshot of how it looks now?

NatalieEO:
@coverbeck attached images

coverbeck:
I think the Organization Stargazers page should show the organization in question, like it does for a workflow/tool. There are other elements on the page too (last modified, the starring component, that should probably match as well). I think the most important one is the organization name.

In a perfect world, they would be the same component and you would get this for free.

If you want to track this separately in a separate ticket, that's fine.

BTW, I'm not wild about the Back to details button, but we have that pattern already for tools and workflows.

coverbeck:
Also on line 21 of stargarzers.component.html, it will say there are no gazers for this entry instead of organization.

NatalieEO:
Screen Shot 2019-05-16 at 11 23 33 AM

src/app/organizations/organization/organization.component.ts
```diff
@@ -84,4 +85,8 @@ export class OrganizationComponent implements OnInit {
   organizationStarGazersChange(): void {
     this.organizationStarGazersClicked = !this.organizationStarGazersClicked;
   }
+
+  getTimeAgoMessage(date: Date) {
```
coverbeck: Gary comments whenever I do this -- how often is this invoked?  Probably a lot. It's preferable not to invoke methods for display values from your html.

If it's invoked a lot, you may want to either make this a pipe, or perhaps have the OrganizationQuery return it, although I'm not sure if that's a good practice.

PR #661, opened 2019-05-17
Feature/workflow and tool alias redirect

agduncan94:
https://github.com/dockstore/dockstore/issues/2495

Depends on https://github.com/dockstore/dockstore/pull/2496

Support for redirecting using tool/workflow aliases.


src/app/aliases/aliases.component.ts
```diff
@@ -17,12 +17,14 @@ export class AliasesComponent extends Base implements OnInit {
...
   collection$: Observable<Collection>;
+  workflow$: Observable<Workflow>;
+  tool$: Observable<DockstoreTool>;
 
   public type: string;
   public alias: string;
   public validType: boolean;
   // Types contains resource types that support aliases
-  public types = [ 'organizations', 'collections' ];
+  public types = [ 'organizations', 'collections', 'workflows', 'tools', 'containers' ];
```
coverbeck: Just curious, how can the type of a tool be either tools or containers? Is this some legacy thing?

agduncan94: Yes, we called tools containers a long time ago, so we don't want to break links.

PR #663, opened 2019-05-23
Generalizable work for new workflow languages

denis-yuen:
* adapt to simplification of descriptor endpoints
* document places to add a new language

Accompanies https://github.com/dockstore/dockstore/pull/2428


denis-yuen:
Can't review changes (as author) but thanks @coverbeck for finishing this off after I forgot about it

PR #664, opened 2019-05-23
Upload build to s3

coverbeck:
dockstore/dockstore#2446

* Only run for tags (for now)
* Requires the following environment variables to be set in CircleCI
   1. AWS_ACCESS_KEY_ID
   2. AWS_SECRET_ACCESS_KEY
   3. AWS_BUCKET
   4. AWS_REGION (needed even though no region is needed for s3 API)
* If the AWS_BUCKET is not set, the upload is skipped.

For now, upload the builds to <TAG NAME>-<sha> folder.

This ticket only uploads to s3; subsequent tickets will handle
the actual serving of content from s3. Some tweaks may need to be
made then, but this gets the core uploading in place.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,7 +91,7 @@ workflows:
   everything:
     jobs:
       # Add the tags filter to all jobs so they will run before upload_to_s3
-      - lint_unit_test_coverage
+      - lint_unit_test_coverage:
           filters:
             tags:
               only: /.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,6 @@
 version: 2.1
+orbs:
+  aws-s3: circleci/aws-s3@1.0.0
 executors:
   integration_test_exec: # declares a reusable executor
     docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -65,20 +67,63 @@ jobs:
...
-      - lint_unit_test_coverage
+      # Add the tags filter to all jobs so they will run before upload_to_s3
+      - lint_unit_test_coverage:
+          filters:
+            tags:
+              only: /.*/
       - integration_test_1:
+          filters:
+            tags:
+              only: /.*/
```
denis-yuen: ~Does this limit integration tests to tags only as well?~

Actually, I see tests above as part of jobs standalone and here. Will that lead to the tests running twice? 
What's the interaction between the tests in the workflows root and in the jobs root?

coverbeck: See [this doc](https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag), in particular the second code snippet in that section.

My understanding from that is that by default, jobs are only run for branch commits. I've added an upload_to_s3 job which is configured to only run for tags. Since upload_to_s3 requires a build, that means I need to tell the build jobs to also run for tags, hence this section.

For a branch commit, only build jobs (integration_test_1, etc.) will run.

For a tag, build jobs and upload_to_s3 will run.

PR #665, opened 2019-05-24
Fix null check errors on compiling

wshands:
Add 'null' datatype to variables so errors are not generated when using the --strictnullcheck compiler option

dockstore/dockstore#1793 null checks


src/app/aliases/aliases.component.ts
```diff
@@ -32,20 +32,26 @@ export class AliasesComponent extends Base implements OnInit {
...
+    /*
+    let myType: string | null;
+    myType = this.route.snapshot.paramMap.get('type') ;
+    // Avoid assigning null to this.type
+    this.type = (myType) ? myType : '';
+     */
+    this.type = this.route.snapshot.paramMap.get('type') ;
     this.alias = this.route.snapshot.paramMap.get('alias');
-    this.validType = this.types.includes(this.type);
+    this.validType = (this.type) ? this.types.includes(this.type) : false;
```
coverbeck: Is this necessary? Doesn't `this.types.includes(null)` return just `false`, or does the compiler option flag this? (I don't know)

src/app/aliases/aliases.component.ts
```diff
@@ -32,20 +32,26 @@ export class AliasesComponent extends Base implements OnInit {
   }
 
   ngOnInit() {
-    this.type = this.route.snapshot.paramMap.get('type');
+    /*
```
coverbeck: I think you can remove this block.

src/app/aliases/aliases.component.ts
garyluu: paraphrased: "the function 'include' does not exist on types, your website will now self-destruct"

src/app/aliases/aliases.component.ts
wshands: block removed

PR #667, opened 2019-05-29
Seab 71/google search new

rshreya:
- Added Bioschemas Specification - Tools Schema for both tools and workflows.
[Spec](https://bioschemas.org/specifications/Tool/)
Focused on attributes with Minimum and some Recommended marginalities. 

- The findings of the AWS instance experiment for Google Rich Search Results yielded that it is unlikely that Dockstore is eligible for rich search results. Please see Findings section of the Improving Google Search document [here:](https://wiki.oicr.on.ca/display/DOC/Improving+Google+Search)





































































































































































































































































































































































































































































































**My time at Dockstore has come to an end so I am handing off this task.** 

_Here's what's been done:_
- created a service that creates a SoftwareApplication schema describing tools and workflows using the [Bioschemas Tool specification](https://bioschemas.org/specifications/Tool/)
- added a skeleton test spec for the service
- plugged in the service so it is called in the container and workflow components
- resolved all existing PR comments to date
- this can be tested by running localhost on my branch and accessing a tool or workflow. You can see a SoftwareApplication ld+json inside the html when using inspect

_Here's what still needs to be done:_
- tests for the Bioschema Service
- anything new requested in PR review. 

Thank you :)


src/app/container/container.component.html
```diff
@@ -32,6 +32,7 @@
 </div>
 
 <div class="row m-1">
+  <ngx-json-ld [json]="schema"></ngx-json-ld>
```
coverbeck: Do you have two of these on purpose?

src/app/container/container.component.ts
```diff
@@ -146,10 +146,27 @@ export class ContainerComponent extends Entry implements AfterViewInit {
...
     // will need to be aggregated with a summary page
     this.schema = {
-      '@type': 'ListItem',
-      'position': this.tool.id,
-      'url': this.shareURL
+      '@type': 'SoftwareApplication',
+      'description': this.tool.description,
+      'name': this.tool.name,
+      'softwareVersion': this.defaultVersion,
+      'url': window.location,
```
coverbeck: I'm not sure if this shouldn't be the TRS url, although it's not clear to me if it should be the UI or the backend URL.

Also, this won't be consistent if we decided to go with the UI url; if I bookmark a tab, e.g., [this](https://dockstore.org/containers/quay.io/pancancer/pcawg-dkfz-workflow:2.2.0?tab=launch), it will have a different url than if I go to the info page.

src/app/container/container.component.ts
```diff
@@ -146,10 +146,27 @@ export class ContainerComponent extends Entry implements AfterViewInit {
...
+
+      'audience': 'Bioinformaticians',
+      'dateModified': this.tool.lastUpdated,
```
coverbeck: How does this get rendered? Note that in schema.org the date needs to be specific format.

src/index.html
```diff
@@ -10,6 +10,7 @@
 
 <head>
   <meta charset="utf-8">
+  <meta name="google-site-verification" content="GFLds4kE1pcp99mSh5ipFD49XediZA8PN03qMGmPF8w" />
```
coverbeck: I don't think you want to merge this

src/index.html
```diff
@@ -44,7 +45,22 @@
...
+    });
+    var f = d.getElementsByTagName(s)[0],
+      j = d.createElement(s),
+      dl = l != 'dataLayer' ? '&l=' + l : '';
+    j.async = true;
+    j.src =
+      'https://www.googletagmanager.com/gtm.js?id=' + i + dl;
+    f.parentNode.insertBefore(j, f);
+  })(window, document, 'script', 'dataLayer', '{{ TAG_MANAGER_ID }}');
+  </script>
```
coverbeck: You may need to rebase/merge -- this shouldn't get merged.

src/app/workflow/workflow.component.ts
```diff
@@ -161,10 +161,30 @@ export class WorkflowComponent extends Entry implements AfterViewInit {
     // messy prototype for a carousel https://developers.google.com/search/docs/guides/mark-up-listings
     // will need to be aggregated with a summary page
     this.schema = {
-      '@type': 'ListItem',
-      'position': this.workflow.id,
-      'url': this.shareURL
+      '@type': 'SoftwareApplication',
```
coverbeck: Shouldn't there be a dateModified and audience for this workflows as well?

src/app/container/container.component.ts
```diff
@@ -146,10 +146,27 @@ export class ContainerComponent extends Entry implements AfterViewInit {
     // messy prototype for a carousel https://developers.google.com/search/docs/guides/mark-up-listings
     // will need to be aggregated with a summary page
     this.schema = {
-      '@type': 'ListItem',
-      'position': this.tool.id,
-      'url': this.shareURL
+      '@type': 'SoftwareApplication',
+      'description': this.tool.description,
+      'name': this.tool.name,
+      'softwareVersion': this.defaultVersion,
```
coverbeck: This should not be the default version, it should be the current version.

src/app/shared/date.service.ts
```diff
@@ -73,4 +73,22 @@ export class DateService {
       return 'n/a';
     }
   }
+
+  getISO8601FormatFromDate(timestamp: Date) {
```
garyluu: Return type for these two functions

src/app/shared/modules/workflow.module.ts
```diff
@@ -105,6 +106,7 @@ import { getTooltipConfig } from './../tooltip';
   providers: [
     { provide: TooltipConfig, useFactory: getTooltipConfig },
     DateService,
+    BioschemaService,
```
garyluu: Throw this into the entryModule instead of both workflow and container.

src/index.html
```diff
@@ -8,38 +8,40 @@
...
+  {
+  "@context": "http://schema.org",
+  "@type": "WebSite",
+  "url": "https://dockstore.org",
+  "sameAs": ["https://twitter.com/DockstoreOrg"],
+  "potentialAction": {
+    "@type": "SearchAction",
+    "target": "https://dockstore.org/search?search={search_term_string}",
+    "query-input": "required name=search_term_string"
+  }
```
garyluu: odd formatting, did prettier do this?

emlys: Looks like some changes were made by prettier and some are from one of Shreya's past commits. Should I change back the formatting to the wider indents?

garyluu: what it looks previously is better in my opinion.  if prettier is forcing it to be like it is right now, I'd just leave it as it is and move on for now.

src/app/shared/modules/workflow.module.ts
emlys: moved to entry.module.ts

emlys:
https://developers.google.com/search/docs/data-types/software-app
We will need to add at least 2 more fields to meet the requirements to possibly have rich search results displayed

denis-yuen:
> https://developers.google.com/search/docs/data-types/software-app
> We will need to add at least 2 of `applicationCategory`, `operatingSystem`, `offers`, and `aggregateRating` to meet the basic requirements. `applicationCategory` and `operatingSystem` sound the most relevant. Also, having more recommended/optional fields may make it more likely for rich results to be displayed: https://developers.google.com/search/docs/guides/sd-policies.

Please do, I can't imagine that adding more recommended/optional fields would break anything so feel free to do as many easy/quick ones as you think would be wise.

emlys:
Added fields for `dateCreated`, `applicationSubcategory`, `applicationCategory`, `operatingSystem`, and `downloadUrl. I think the changes were minor enough to keep within this PR

denis-yuen:
> Added fields for `dateCreated`, `applicationSubcategory`, `applicationCategory`, `operatingSystem`, and `downloadUrl`. I think the changes were minor enough to keep within this PR

Cool, I think as @garyluu  you'll probably want to merge in develop (`git hf update` if you're using the hubflow tools) and then get the tests working.

src/app/shared/date.service.ts
```diff
@@ -73,4 +73,22 @@ export class DateService {
...
+    if (timestamp) {
+      const tsAsDate = new Date(timestamp);
+      return tsAsDate.toISOString();
+    } else {
+      return 'n/a';
+    }
+  }
+
+  getISO8601FormatFromNumber(timestamp: number): string {
+    if (timestamp) {
```
coverbeck: This should call `getISO8601FormatFromDate` with a date object so the (small) amount of code is is not duplicated.

src/app/shared/date.service.ts
```diff
@@ -73,4 +73,22 @@ export class DateService {
       return 'n/a';
     }
   }
+
+  getISO8601FormatFromDate(timestamp: Date): string {
+    if (timestamp) {
+      const tsAsDate = new Date(timestamp);
```
coverbeck: Needlessly creating another date

src/app/shared/date.service.ts
```diff
@@ -73,4 +73,22 @@ export class DateService {
...
+      return tsAsDate.toISOString();
+    } else {
+      return 'n/a';
```
coverbeck: Suggest returning null and updating method signature to so indicate.

src/app/workflow/workflow.component.ts
```diff
@@ -430,6 +429,7 @@ export class WorkflowComponent extends Entry implements AfterViewInit {
     }
     this.workflowService.setWorkflowVersion(version);
     this.updateWorkflowUrl(this.workflow);
+    this.schema = this.bioschemaService.getWorkflowSchema(this.workflow, this.selectedVersion);
```
coverbeck: It would be nice if you didn't have to call this method twice.

src/app/shared/bioschema.service.ts
```diff
@@ -0,0 +1,117 @@
...
+  getToolSchema(tool: DockstoreTool, selectedVersion: Tag): BioschemaTool {
+    const results = this.getBaseSchema(tool, selectedVersion);
+    results.applicationSubCategory = 'Tool';
+    if (tool.name) {
+      results.name = tool.name;
+    }
+    if (tool.lastUpdated) {
+      results.dateModified = this.dateService.getISO8601FormatFromDate(tool.lastUpdated);
+    }
+    if (tool.author) {
```
coverbeck: It seems like this block and the block starting on line 94 could go into getBaseSchema

src/app/shared/bioschema.service.ts
```diff
@@ -0,0 +1,117 @@
...
+    private dateService: DateService,
+    private workflowsService: ExtendedWorkflowsService,
+    private containersService: ExtendedToolsService
+  ) {}
+  private getBaseSchema(entry: DockstoreTool | Workflow, selectedVersion: Tag | WorkflowVersion): BioschemaTool {
+    const results: BioschemaTool = {
+      '@context': 'http://schema.org',
+      '@type': 'SoftwareApplication',
+      operatingSystem: ['macOS', 'Linux', 'Windows'],
+      softwareRequirements: 'Docker',
```
coverbeck: Probably OK as is, but you *could* see if the workflow is WDL, CWL, and/or Nextflow, and add as requirement(s).

src/app/shared/bioschema.service.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    if (workflow.author) {
+      results.publisher = {
+        '@type': 'Person',
+        name: workflow.author
+      };
+      if (workflow.email) {
+        results.publisher.email = workflow.email;
+      }
+    } else if (workflow.organization) {
```
coverbeck: Does a tool have an organization? If so (I don't think it does), you could move to getBaseSchema.

src/app/shared/date.service.ts
emlys: changed methods to return `string | null`

src/app/shared/bioschema.service.ts
emlys: I don't see any organization attribute for it

denis-yuen: namespace is the equivalent for tools

src/app/shared/bioschema.service.ts
emlys: added this

src/app/shared/date.service.ts
emlys: here Date seems to just be an alias for a number, the argument has no `toISOString` method until initialized as a new Date. Combined these two methods into one.

src/app/shared/bioschema.service.ts
emlys: moved to `getBaseSchema` using `organization` or `namespace`

src/app/shared/date.service.ts
```diff
@@ -70,7 +69,14 @@ export class DateService {
         return time + (time === 1 ? ' day ago' : ' days ago');
       }
     } else {
-      return 'n/a';
+      return null;
+    }
+  }
+  getISO8601Format(timestamp: Date | number): string | null {
```
coverbeck: You're doing this because the dbCreateDate is actually a number, even though according to the TypeScript model, it should be a Date, right?

It seems a little hacky, but I think the real problem is that dbCreateDate is not the type that it is declared to be (it's a number, but according to src/app/shared/swagger/model/entry.ts, it should be a Date).

@garyluu Do you have any insight into this? Is this a Swagger codegen issue?

garyluu: The swagger.yaml shows:
```
        type: "string"
        format: "date-time"
```
so the UI Swagger codegen seems correct in thinking it's a Date.  The webservice, on the other hand, is returning a number.  So this is ether some serialization issue or that the swagger.yaml is not generated correctly.

emlys: Should I make a separate ticket for that? is this solution ok for the time being?

coverbeck: I think a separate ticket is fine. Please reference this code in the ticket so whoever tackles the ticket will also remember to come back and update this when/if they fix the underlying issue.

emlys: https://ucsc-cgl.atlassian.net/browse/DOCK-832

PR #668, opened 2019-06-03
Links Zenodo credentials in Dockstore

wshands:
Enables the Dockstore user to link their Zenodo credentials to Dockstore so they can create DOIs for workflows through the Dockstore UI
DOCK-763


src/app/shared/dockstore.model.ts
```diff
@@ -55,6 +55,11 @@ export class Dockstore {
   // static GITLAB_SCOPE = 'read_user openid';
   static GITLAB_SCOPE = 'api';
 
+  static ZENODO_AUTH_URL = 'https://zenodo.org/oauth/authorize';
+  static ZENODO_CLIENT_ID = 'fill_this_in';
```
coverbeck: Even though you are following the template of other properties in this file, I suggest you break the pattern and say something like `will be filled in by configuration.service` or `will be filled in` -- the point being that it is no longer necessary to update this file. I should have changed the other ones when I made the change to read the config.json

wshands: done

wshands:
DOCK-763

PR #670, opened 2019-06-04
add text to funding page and update ontario logo

Ldcabansay:
https://github.com/dockstore/dockstore/issues/2493
https://github.com/dockstore/dockstore/issues/2465
Adds text to funding page and updates Ontario logo. Also adds BISTI and STAGE funding info

### DIFF
--- a/src/app/funding/funding.component.html
+++ b/src/app/funding/funding.component.html
@@ -9,11 +9,10 @@
     </div>
   </div>
 </div>
-
 <div class="container">
   <div class="row-fluid mt-3">
     <p class="lead">Support for Dockstore is provided by grants from Genome Canada, the National Institutes of Health, and the Province of Ontario.</p>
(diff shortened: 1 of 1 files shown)


Ldcabansay:
image

image

Btw that extra sized gap before the datastage panel isn't really there, that's just the result of two screenshots.

coverbeck:
When referring to an issue, you need to preface with dockstore/dockstore, because the issues are in a different repo. It would also be good if those issues got into the commit message when you merge.

coverbeck:
In the wording for DataSTAGE, it says "...to create a community of practice...". Should that be "practices"? If it came from them like that, never mind.

denis-yuen:
> When referring to an issue, you need to preface with dockstore/dockstore, because the issues are in a different repo. It would also be good if those issues got into the commit message when you merge.

You can also copy-paste the whole URL if that happens to be easier

PR #671, opened 2019-06-04
Read GTM ID from config.json

coverbeck:
dockstore/dockstore-ui2#2444

Also fix build errors caused by changes to webservice endpoints.


src/app/shared/state/checker-workflow.query.ts
```diff
@@ -31,7 +31,7 @@ export class CheckerWorkflowQuery extends Query<CheckerWorkflowState> {
     map((entry: Entry) => this.isEntryAWorkflow(entry) ? (<Workflow>entry).mode === Workflow.ModeEnum.STUB : false));
   checkerWorkflowPath$: Observable<string> = this.checkerWorkflow$.pipe(
     map(workflow => workflow ? workflow.full_workflow_path : null));
-  parentId$: Observable<number> = this.entry$.pipe(map((entry: Workflow) => entry ? entry.parent_id : null));
+  parentId$: Observable<number> = this.entry$.pipe(map((entry: Workflow) => entry && entry.parentEntry && entry.parentEntry.id));
```
garyluu: Does this even return a number?

coverbeck: The compiler doesn't complain. :)

OTOH, integration tests are failing, so maybe it is wrong.

PR #673, opened 2019-06-07
2470/org member feedback

agduncan94:
There is now helpful text to explain the purpose of the add/edit member box and what the expected outcomes are. Also there is a snackbar saying invite has been sent.


src/app/organizations/state/upsert-organization-member.service.ts
```diff
@@ -70,6 +72,20 @@ export class UpsertOrganizationMemberService {
...
 
+  /**
+   * Get the description based on the mode
+   *
+   * @param {*} data
+   * @returns {string}
+   * @memberof UpsertOrganizationMemberService
+   */
+  getDescription(data: any): string {
+    const mode: TagEditorMode = data.mode;
```
garyluu: null check on data

PR #674, opened 2019-06-07
Changes for workflow language enum refactor

denis-yuen:
See linked webservice PR

### DIFF
--- a/src/app/container/add-tag/add-tag.component.ts
+++ b/src/app/container/add-tag/add-tag.component.ts
@@ -139,7 +139,7 @@ export class AddTagComponent extends Base implements OnInit, AfterViewChecked {
   addTag() {
     this.alertService.start('Adding tag');
     this.containertagsService.addTags(this.tool.id, [this.unsavedVersion]).subscribe((tags: Tag[]) => {
-      this.tool.tags = tags;
+      this.tool.workflowVersions = tags;
       const id = this.tool.id;
       const tagName = this.unsavedVersion.name;
       // Store the unsaved test files if valid and exist
--- a/src/app/shared/descriptor-type-compat.service.ts
+++ b/src/app/shared/descriptor-type-compat.service.ts
@@ -50,9 +50,9 @@ constructor() { }
       return ToolDescriptor.TypeEnum.CWL;
     } else if (descriptorType.toUpperCase() === DescriptorTypeEnum.WDL) {
       return ToolDescriptor.TypeEnum.WDL;
-    } else if (descriptorType.toUpperCase() === DescriptorTypeEnum.NEXTFLOW) {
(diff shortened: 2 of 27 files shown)


src/app/container/add-tag/add-tag.component.ts
```diff
@@ -139,7 +139,7 @@ export class AddTagComponent extends Base implements OnInit, AfterViewChecked {
   addTag() {
     this.alertService.start('Adding tag');
     this.containertagsService.addTags(this.tool.id, [this.unsavedVersion]).subscribe((tags: Tag[]) => {
-      this.tool.tags = tags;
+      this.tool.workflowVersions = tags;
```
agduncan94: A bit weird it is called workflowVersions for tools. Should it just be versions?

denis-yuen: I think this is swagger codegen from the webservice. 
For background, workflows currently have workflowVersions, tools have tags.

There are few options to unify things:
1) Standardize on tags and break users of workflows like older CLIs and DNAstack)
2) Standardize on workflowVersions and break users of tools
3) Standardize on versions and break both

The proposal in these two linked PRs is to do 1 and send back the old json structure with tags when we detect a CLI with an old user agent

denis-yuen:
(actually need to update to new webservice release to run tests)

src/app/workflow/launch-third-party/launch-third-party.component.html
```diff
@@ -1,12 +1,12 @@
-<div *ngIf="workflow?.descriptorType === 'wdl'" class="panel panel-default">
+<div *ngIf="workflow?.descriptorType === 'WDL'" class="panel panel-default">
```
garyluu: Remind me to change this after

src/app/shared/descriptor-type-compat.service.ts
```diff
@@ -14,7 +14,8 @@
  *     limitations under the License.
  */
 import { Injectable } from '@angular/core';
-import { ToolDescriptor } from './swagger';
+import { ToolDescriptor, Workflow } from './swagger';
+import DescriptorTypeEnum = Workflow.DescriptorTypeEnum;
 
 export enum WebserviceDescriptorTypeEnum {
```
garyluu: is this used anywhere anymore?

denis-yuen: Possibly not, I was more fixing the build so that tests would pass as opposed to making any improvements to take advantage of things.

src/app/workflow/launch-third-party/launch-third-party.component.html
denis-yuen: https://github.com/dockstore/dockstore/issues/2566

PR #681, opened 2019-06-17
portion of feature/2458/databasechanges

NatalieEO:
Part of #2548
Changes for renaming lastmodified column in tag table to lastbuild. There are 2 other recommendations. I'd rather split this out and do this while I work on other things for dockstore or put this on hold in general. Thoughts?

### DIFF
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -173,7 +173,6 @@
                 id="versionsTab"
                 [verifiedSource]="(extendedTool$ | async)?.verifiedSources"
                 [selectedVersion]="selectedVersion"
-                [isTool]="true"
               >
               </app-versions-container>
             </div>
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -173,6 +173,7 @@
                 id="versionsTab"
                 [verifiedSource]="(extendedTool$ | async)?.verifiedSources"
                 [selectedVersion]="selectedVersion"
+                [isTool]="true"
               >
               </app-versions-container>
             </div>
@@ -236,7 +237,7 @@
                   (click)="onSelectedVersionChange(sortedVersion)"
                   >{{ sortedVersion.name }}
                 </a>
-                <small>{{ sortedVersion.last_modified | date }}</small>
+                <small>{{ sortedVersion.last_built | date }}</small>
               </p>
               <hr class="m-0" />
             </span>
--- a/src/app/container/versions/versions.component.ts
+++ b/src/app/container/versions/versions.component.ts
@@ -60,8 +60,8 @@ export class VersionsContainerComponent extends Versions implements OnInit {
     super(dockstoreService, dateService, sessionQuery);
   }
 
-  setDefaultSortColumn(): string {
-    return 'last_build';
+  defaultSortColumn(): string {
+    return 'last_built';
   }
 
   ngOnInit() {
--- a/src/app/container/versions/versions.component.ts
+++ b/src/app/container/versions/versions.component.ts
@@ -58,10 +58,7 @@ export class VersionsContainerComponent extends Versions implements OnInit {
     protected sessionQuery: SessionQuery
   ) {
     super(dockstoreService, dateService, sessionQuery);
-  }
-
-  defaultSortColumn(): string {
-    return 'last_built';
+    this.sortColumn = 'last_built';
   }
 
   ngOnInit() {
--- a/src/app/container/versions/versions.component.ts
+++ b/src/app/container/versions/versions.component.ts
@@ -60,6 +60,10 @@ export class VersionsContainerComponent extends Versions implements OnInit {
     super(dockstoreService, dateService, sessionQuery);
   }
 
+  setDefaultSortColumn(): string {
+    return 'last_build';
+  }
+
   ngOnInit() {
     this.publicPageSubscription();
     this.extendedDockstoreToolQuery.extendedDockstoreTool$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((tool: ExtendedDockstoreTool) => {
--- a/src/app/shared/entry.ts
+++ b/src/app/shared/entry.ts
@@ -194,7 +194,7 @@ export abstract class Entry implements OnInit, OnDestroy {
    * @returns {((WorkflowVersion | Tag))}  The version to display to the user
    * @memberof Entry
    */
-  public selectVersion(versions: (Array<WorkflowVersion | Tag>), urlVersion: string, defaultVersion: string): (WorkflowVersion | Tag) | null{
+  public selectVersion(versions: (Array<WorkflowVersion | Tag>), urlVersion: string, defaultVersion: string): (WorkflowVersion | Tag | null) {
     if (!versions || versions.length === 0) {
       return null;
     }
@@ -293,7 +293,7 @@ export abstract class Entry implements OnInit, OnDestroy {
   workflowVersionSorting(a: WorkflowVersion, b: WorkflowVersion): number {
 
     const verifiedSorting = this.verifiedSorting(a, b);
-    if (verifiedSorting !== -2) {
+    if (verifiedSorting !== 0) {
      return verifiedSorting;
     } else {
       if (a.last_modified > b.last_modified) {
--- a/src/app/shared/entry.ts
+++ b/src/app/shared/entry.ts
@@ -195,9 +195,9 @@ export abstract class Entry implements OnInit, OnDestroy {
    * @memberof Entry
    */
   public selectVersion(versions: Array<WorkflowVersion | Tag>, urlVersion: string, defaultVersion: string): WorkflowVersion | Tag | null {
-    if (!versions || versions.length === 0) {
-      return null;
-    }
+    // if (!versions || versions.length === 0) {
+    //   return null;
+    // }
     let foundVersion: WorkflowVersion | Tag;
     if (urlVersion) {
       foundVersion = versions.find((version: WorkflowVersion | Tag) => version.name === urlVersion);
@@ -216,11 +216,17 @@ export abstract class Entry implements OnInit, OnDestroy {
   }
 
   selectTag(versions: Array<Tag>, urlVersion: string, defaultVersion: string): Tag {
+    if (!versions || versions.length === 0) {
+      return null;
+    }
     const selectedTag = this.selectVersion(versions, urlVersion, defaultVersion);
     return selectedTag || versions.reduce((a, b) => (b.last_built > a.last_built ? b : a));
   }
 
   selectWorkflowVersion(versions: Array<WorkflowVersion>, urlVersion: string, defaultVersion: string) {
+    if (!versions || versions.length === 0) {
+      return null;
+    }
     const selectedWorkflowVersion = this.selectVersion(versions, urlVersion, defaultVersion);
     return selectedWorkflowVersion || versions.reduce((a, b) => (b.last_modified > a.last_modified ? b : a));
   }
--- a/src/app/shared/entry.ts
+++ b/src/app/shared/entry.ts
@@ -210,10 +210,26 @@ export abstract class Entry implements OnInit, OnDestroy {
       if (foundVersion) {
         return foundVersion;
       }
+    } else {
+      return null;
     }
 
     // Select newest last_modified version, if it's the same, choose the top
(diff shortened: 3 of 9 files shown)


src/app/container/versions/versions.component.ts
```diff
@@ -54,6 +54,10 @@ export class VersionsContainerComponent extends Versions implements OnInit {
     super(dockstoreService, dateService, sessionQuery);
   }
 
+  setDefaultSortColumn(): string {
```
coverbeck: `set` has the implication that this method modifies state, which it doesn't. Suggest either plain `defaultSortColumn` or `getDefaultSortColumn`.

src/app/shared/entry.ts
```diff
@@ -259,11 +276,51 @@ export abstract class Entry implements OnInit, OnDestroy {
...
+    //   entryBDate = b.lastModified;
+    // }
+
+    if (a.verified && !b.verified) {
+      return -1;
+    } else if (!a.verified && b.verified) {
+      return 1;
+    } else {
+      if (isTool) {
+        this.tagSorting(a, b);
```
coverbeck: You need to return the result: `return this.tagSorting(a, b)`.

src/app/shared/entry.ts
```diff
@@ -259,11 +276,51 @@ export abstract class Entry implements OnInit, OnDestroy {
...
+      } else {
+        this.workflowVersionSorting(a, b);
```
coverbeck: You need to return the result

src/app/shared/entry.ts
```diff
@@ -259,11 +276,51 @@ export abstract class Entry implements OnInit, OnDestroy {
    * @param {Tag|WorkflowVersion} b - version b
    * @returns {number} - indicates order
    */
-  entryVersionSorting(a: Tag | WorkflowVersion, b: Tag | WorkflowVersion): number {
+  entryVersionSorting(a: Tag | WorkflowVersion, b: Tag | WorkflowVersion, isTool: boolean): number {
+    // let entryADate: any;
```
coverbeck: In general, unless you expect to restore this, you shouldn't leave this commented out code

src/app/shared/entry.ts
```diff
@@ -259,11 +276,51 @@ export abstract class Entry implements OnInit, OnDestroy {
...
+    }
+  }
+
+  verifiedSorting(a: Tag | WorkflowVersion, b: Tag | WorkflowVersion): number {
     if (a.verified && !b.verified) {
       return -1;
     } else if (!a.verified && b.verified) {
       return 1;
+    } else {
+      return -2;
```
coverbeck: The convention for compare-type methods is to return 0 if things are equal.

src/app/shared/entry.ts
```diff
@@ -202,10 +203,26 @@ export abstract class Entry implements OnInit, OnDestroy {
...
     }
 
     // Select newest last_modified version, if it's the same, choose the top
-    return versions.reduce((a, b) => b.last_modified > a.last_modified ? b : a);
+    //  return versions.reduce((a, b) => b.last_modified > a.last_modified ? b : a);
+  }
+
+  selectTag(versions: Array<Tag>, urlVersion: string, defaultVersion: string): Tag {
+    const selectedTag = this.selectVersion(versions, urlVersion, defaultVersion);
+    if (selectedTag != null) {
```
coverbeck: This is more of a preference, but I would do this using JavaScript truthiness:

`return selectedTag || versions.reduce((a,b) =>...`

src/app/shared/entry.ts
```diff
@@ -202,10 +203,26 @@ export abstract class Entry implements OnInit, OnDestroy {
       if (foundVersion) {
         return foundVersion;
       }
+    } else {
```
coverbeck: Since you are potentially returning null, the function signature should indicate that it might return null.

Previously, I wouldn't even return null and let it return undefined, but I guess with the null checker we want to to turn on, it probably is better to be explicit.

src/app/shared/entry.ts
```diff
@@ -315,6 +387,18 @@ export abstract class Entry implements OnInit, OnDestroy {
     return recentVersions;
   }
 
+  getSortedWorkflowVersions(versions: Array<WorkflowVersion>, defaultVersion: WorkflowVersion): Array<WorkflowVersion> {
+    let sortedWorkflowVersions: Array<WorkflowVersion> = [];
```
coverbeck: Suggest making this one line and using `const` modifier instead of `let`. Generally, unless you want the variable to be modifiable, you should prefer const.

src/app/shared/entry.ts
```diff
@@ -315,6 +387,18 @@ export abstract class Entry implements OnInit, OnDestroy {
...
+    sortedWorkflowVersions = versions.slice().sort((a, b) => this.workflowVersionSorting(a, b));
+    return this.getSortedVersions(versions, defaultVersion, sortedWorkflowVersions);
+  }
+
+  getSortedTags(versions: Array<Tag>, defaultVersion: WorkflowVersion): Array<Tag> {
+    let sortedTags: Array<Tag> = [];
```
coverbeck: Same as previous comment.

src/app/shared/entry.ts
```diff
@@ -202,10 +203,26 @@ export abstract class Entry implements OnInit, OnDestroy {
       if (foundVersion) {
         return foundVersion;
       }
+    } else {
+      return null;
     }
 
     // Select newest last_modified version, if it's the same, choose the top
-    return versions.reduce((a, b) => b.last_modified > a.last_modified ? b : a);
+    //  return versions.reduce((a, b) => b.last_modified > a.last_modified ? b : a);
```
denis-yuen: Think this comes up later, but commented code should not be checked in unless there is a pretty clear `//TODO` or `//FIXME` that indicates in what condition it should be uncommented

src/app/container/container.component.html
```diff
@@ -211,7 +211,7 @@ <h3>Recent Versions</h3>
                 </a>
               </span>
               <a [ngStyle]="{'font-weight': selectedVersion?.name === sortedVersion?.name ? 'bold': 'normal'}" (click)="onSelectedVersionChange(sortedVersion)">{{sortedVersion.name}} </a>
-              <small>{{sortedVersion.last_modified | date}}</small>
+              <small>{{sortedVersion.last_build | date}}</small>
```
coverbeck: Should this be last_built?

src/app/shared/entry.ts
```diff
@@ -186,7 +187,7 @@ export abstract class Entry implements OnInit, OnDestroy {
    * @returns {((WorkflowVersion | Tag))}  The version to display to the user
    * @memberof Entry
    */
-  public selectVersion(versions: (Array<WorkflowVersion | Tag>), urlVersion: string, defaultVersion: string): (WorkflowVersion | Tag) {
+  public selectVersion(versions: (Array<WorkflowVersion | Tag>), urlVersion: string, defaultVersion: string): (WorkflowVersion | Tag) | null{
```
coverbeck: It looks weird to me with the `null` outside the parens... Can you put it inside (not sure if there's a reason I'm missing)?

src/app/shared/entry.ts
```diff
@@ -259,11 +273,21 @@ export abstract class Entry implements OnInit, OnDestroy {
...
     } else if (!a.verified && b.verified) {
       return 1;
+    } else {
+      return 0;
+    }
+  }
+  workflowVersionSorting(a: WorkflowVersion, b: WorkflowVersion): number {
+
+    const verifiedSorting = this.verifiedSorting(a, b);
+    if (verifiedSorting !== -2) {
```
coverbeck: Since you got rid of `-2` in `verifiedSorting`, I think you should be checking against `0`.

src/app/shared/versions.ts
```diff
@@ -37,12 +38,13 @@ export abstract class Versions extends EntryTab {
   dtOptions;
 
   abstract setNoOrderCols(): Array<number>;
+  abstract defaultSortColumn(): string;
 
   constructor(protected dockstoreService: DockstoreService,
     private dateService: DateService, protected sessionQuery: SessionQuery) {
-    // By default, sort by last_modified, latest first
+    // By default, sort by last_build, latest first
```
coverbeck: last_built?

src/app/shared/entry.ts
```diff
@@ -209,10 +209,23 @@ export abstract class Entry implements OnInit, OnDestroy {
...
-    // Select newest last_modified version, if it's the same, choose the top
-    return versions.reduce((a, b) => (b.last_modified > a.last_modified ? b : a));
+  selectTag(versions: Array<Tag>, urlVersion: string, defaultVersion: string): Tag {
+    const selectedTag = this.selectVersion(versions, urlVersion, defaultVersion);
+    return selectedTag || versions.reduce((a,b) => b.last_built > a.last_built ? b : a);
+  }
+
+  selectWorkflowVersion(versions: Array<WorkflowVersion>, urlVersion: string, defaultVersion: string) {
+    const selectedTag = this.selectVersion(versions, urlVersion, defaultVersion);
+    if (selectedTag != null) {
```
coverbeck: No big deal, but this method and the previous method are doing the same thing, but this one with `if...else` and line 219 "inline".

src/app/shared/entry.ts
```diff
@@ -282,6 +305,21 @@ export abstract class Entry implements OnInit, OnDestroy {
     }
   }
 
+  tagSorting(a: Tag, b: Tag): number {
+    const verifiedSorting = this.verifiedSorting(a, b);
+    if (verifiedSorting !== 0) {
+      return verifiedSorting;
+    } else {
+      if (a.last_built > b.last_built) {
```
coverbeck: This is fine, but you could simplify the whole block to `return b.last_built - a.last_built`.

src/app/shared/versions.ts
```diff
@@ -36,11 +37,12 @@ export abstract class Versions extends EntryTab {
   dtOptions;
 
   abstract setNoOrderCols(): Array<number>;
+  abstract defaultSortColumn(): string;
 
   constructor(protected dockstoreService: DockstoreService, private dateService: DateService, protected sessionQuery: SessionQuery) {
-    // By default, sort by last_modified, latest first
+    // By default, sort by last_built, latest first
```
coverbeck: This comment needs to be updated.

garyluu:
I was here, awaiting CircleCI pass

denis-yuen:
I see, I think this PR relies on the web service one being merged (and then the webservice being pointed at being updated)

src/app/shared/versions.ts
```diff
@@ -28,6 +28,7 @@ import { Tooltip } from './tooltip';
 export abstract class Versions extends EntryTab {
   @Input() versions: Array<Tag | WorkflowVersion>;
   @Input() verifiedSource: Array<any>;
+  @Input() isTool: boolean;
```
garyluu: Is this even used?

src/app/workflow/versions/versions.component.ts
```diff
@@ -50,6 +50,10 @@ export class VersionsWorkflowComponent extends Versions implements OnInit {
     return [4, 5];
   }
 
+  defaultSortColumn(): string {
```
garyluu: This probably can be made into a simple string like my-tool.component.ts's pageName

NatalieEO: Get warnings that abstract property can't be read from constructor so leaving it as is.

garyluu: https://github.com/dockstore/dockstore-ui2/compare/feature/2548/renameColumnInTagTable...feature/2458/suggestion

denis-yuen:
(this needs a 1.7.0-alpha.9)

src/app/shared/versions.ts
NatalieEO: it was being used after all in is container.component.html

garyluu: I see this PR passing in this new `isTool`, but I don't see it used in the .ts or .html

NatalieEO: it's on line 226 in container.component.html

garyluu: Yes, you added it in the versions.ts and container.component.html, but where is it used?

NatalieEO: Ohhh sorry Gary.

denis-yuen:
@NatalieEO  FYI, some tests to fix

PR #682, opened 2019-06-18
Feature/2130/prettier

DockstoreTestUser2:
Add prettier dockstore/dockstore#2130

Mostly maintained default prettier config.  But changed printWidth from the 80 char default to 140 (tslint).
Also changed from double quote default to single quote (tslint)

Using Pre-commit Hook option 1

Also printing more DB info after each test

Prettier doesn't like escaping characters in single string quotes.  It changes it to double quote instead, but then tslint complains.  Using back-tick quotes to maintain compatibility with both.


garyluu:
1. You literally do nothing different.  Proceed as usual.
2. Makes it more obvious when using git blame that it's nothing significant, just formatting.  Seemed odd to heavily inflate my line count changes in this repo.
3. That was a test in case the PR build failed. removed it right after your comment

garyluu:
Remaking as myself otherwise, Dockstoretestuser2 might be pushed to top of publications

PR #684, opened 2019-06-25
fix linking accounts

NatalieEO:
Linking your account wasn't working anymore. Thanks @coverbeck for the fix.


NatalieEO:
@garyluu semicolons for the links.model.ts get added automatically when you commit, but then the tslint complains about them and fails the lint test. Don't know about how you'd want to fix it longterm, but disabled the lint rule for that file

denis-yuen:
(with passing builds)

PR #685, opened 2019-06-27
Add TOS to footer and reorganize

NatalieEO:
image


src/app/footer/footer.component.html
```diff
@@ -85,14 +84,43 @@
           </li>
         </ul>
       </div>
+      <div class="col-sm-4">
+        <ul class="footer-list m-0">
+          <li class="m-0 footer-heading">
+            Legal
+          </li>
+          <li>
+            <a href="https://github.com/ga4gh/dockstore/blob/1.5.4/LICENSE" target="_blank" rel="noopener noreferrer" class="footer-link"
```
coverbeck: It was like this, but this should point to the latest license.

Not sure if it's worth the effort, but ideally this would be dynamic, pointing to the license of the current web service version, whatever it is.

The footer already displays the web service version (I don't know how it gets it), so you could presumably use that to dynamically generate this link.

OTOH, our license probably doesn't change that much, so it may not be worth the effort.

Also, it could be a separate ticket if others think it a good idea rather than part of this PR -- I just happened to notice because this line changed.

src/app/footer/footer.component.css
```diff
@@ -34,3 +28,7 @@ a {
 ul {
   list-style: none;
 }
+
+.footer-heading {
```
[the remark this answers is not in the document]
NatalieEO: Bolds all the list items unfortunately

src/app/footer/footer.component.html
garyluu: `ts-node git.version.ts` writes the `src/app/footer/versions.ts`

NatalieEO: Hm, the links actually don't work I'm realizing now. 404s

coverbeck: I see -- it's the version of the web service the UI was built against.

We could add the web service version to the config.json (or maybe it's already exposed some other way), and detect if there is a difference. Didn't this come up once in the past, like around in the 1.5 timeframe? I have a vague memory of something being out of sync.

I'll log that separately as a wish list item.

garyluu: Oops, the webservice version is gotten from the metadata endpoint, not the versions.ts

coverbeck: I'll close the other ticket then. :)

src/app/footer/footer.component.css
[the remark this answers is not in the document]
garyluu:
```
.footer-list li:first-child {
  font-weight: bold;
}
```

NatalieEO:
image

src/app/footer/footer.component.html
coverbeck: I know this was a long thread :), but can you please either fix this or log a ticket. We shouldn't be linking to a license on an older branch.

NatalieEO:
image

agduncan94:
I'm not sure about the final one with the status up being beside API. If you go to status up there are a variety of things the status is referring too, and not just the webservice.
https://status.dockstore.org/

src/app/footer/footer.component.css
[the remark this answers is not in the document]
garyluu: If it works, footer-heading can be removed from the html

denis-yuen:
> I'm not sure about the final one with the status up being beside API. If you go to status up there are a variety of things the status is referring too, and not just the webservice.
> https://status.dockstore.org/

Rats, true, but this is a big improvement in general, so I'm thinking merge and then possible follow-up ticket.

PR #689, opened 2019-07-04
Feature/2157/service files

agduncan94:
Support for service file viewing. See screenshots.
https://github.com/dockstore/dockstore/issues/2157

Also adds markdown and shell as supported languages for Ace Editor.

service-with-no-versions
service-with-version


agduncan94:
Note to self - waiting on new release of webservice

src/app/search/map-friendly-values.pipe.ts
```diff
@@ -54,11 +54,11 @@ export class MapFriendlyValuesPipe implements PipeTransform {
     [
       'ToolFile.FileTypeEnum',
       new Map([
-        [ToolFile.FileTypeEnum.PRIMARYDESCRIPTOR, 'Primary Descriptor'],
+        [ToolFile.FileTypeEnum.PRIMARYDESCRIPTOR, 'Config'],
```
denis-yuen: Does this affect services only? 
It isn't clear to me in the context and it would be weird describing the "first" CWL file a config even if it is correct for Nextflow

agduncan94: Have to double check, but I think technically it could be used by all. I'll revert to Primary Descriptor and this can be dealt with later.

src/app/shared/versions.ts
```diff
@@ -68,7 +69,10 @@ export abstract class Versions extends EntryTab {
   getIconClass(columnName): IconDefinition {
     return this.dockstoreService.getIconClass(columnName, this.sortColumn, this.sortReverse);
   }
-  convertSorting(): string {
+  convertSorting(mode): string {
+    if (mode && mode === DockstoreTool.ModeEnum.HOSTED) {
+      this.sortColumn = 'id';
+    }
     return this.sortReverse ? '-' + this.sortColumn : this.sortColumn;
```
coverbeck: What happens, and does it matter, when mode is not set? I guess you wouldn't have the versions without workflow being set on line 99 without the workflow...

Also, technically this method could return undefined, so since you're here you should probably update the method signature. And/or set sortColumn to null if the condition on line 73 is not true.

src/app/workflow/workflow.component.html
```diff
@@ -222,18 +222,41 @@ <h3>
...
-              <div *ngIf="entryType === EntryType.Service; else BioWorkflow">
-                <app-entry-file-tab></app-entry-file-tab>
+              <div *ngIf="entryType === EntryType.Service; else FilesTabBioWorkflow" class="p-3">
+                <div *ngIf="(extendedWorkflow$ | async)?.workflowVersions.length > 0; else noVersionsFilesTab">
+                  <app-entry-file-tab></app-entry-file-tab>
+                </div>
+
+                <ng-template #noVersionsFilesTab>
+                  <mat-card class="alert alert-info" role="alert">
+                    <mat-icon>info</mat-icon> This service does not have any versions. Create a release/tag on GitHub with a valid dockstore
```
coverbeck: Instead of `dockstore config file`, may you should say `.dockstore.yml`. Or leave `dockstore config file`, and also add `.dockstore.yml` -- I don't know that new users know what a dockstore config file is.

src/app/shared/versions.ts
agduncan94: In the case mode is not set, it should default it to last_modified/built, which was what this function had previously always set it to. I think that is a sane default since it works with most all non-hosted versions.

PR #691, opened 2019-07-05
Feature/2554/service improvements

garyluu:
For dockstore/dockstore#2554

The main issue was getting rid of the subscribe in the global service because it never gets destroyed and caused issues when going back and forth between MyServices and MyWorkflows

- Removed the mytools successful get alert (consistent with workflows and services)
- Make myworkflowsService not global
- add tslint-config-prettier (even though it doesn't seem to change anything)
- add random null checks
- MyWorkflowsService now has MyBioWorkflowsService and MyServicesService because it's easier to use that way (nothing else should import MyBioWorkflowsService and MyServicesService), MyWorkflowsComponent just import MyWorkflowsService.  Having MyBioWorkflowsService and MyServicesService extend MyWorkflowsService did not seem to help 

- Moved a bunch of component functions into service where it belongs


src/app/myworkflows/myworkflows.service.ts
```diff
@@ -14,39 +14,117 @@
...
+      const orgWorkflowObject: OrgWorkflowObject = {
+        sourceControl: '',
+        organization: '',
+        published: [],
+        unpublished: []
+      };
+      const nsWorkflow: Array<Workflow> = nsWorkflows[i].entries;
+      orgWorkflowObject.sourceControl = nsWorkflows[i].sourceControl;
+      orgWorkflowObject.organization = nsWorkflows[i].organization;
+      orgWorkflowObject.published = nsWorkflow.filter((workflow: Workflow) => {
```
coverbeck: This and the block on line 116 can be written without curly braces and an implicit return. Sort of torn on it myself; I like the ability to set a breakpoint more cleanly this way, but OTOH, my IDE recommends the no block statement when possible. Your call.

src/app/myworkflows/myworkflows.service.ts
```diff
@@ -14,39 +14,117 @@
...
+      if (foundWorkflow) {
+        return foundWorkflow;
+      }
+    }
+    return null;
+  }
+
+  public convertOldNamespaceObjectToOrgEntriesObject(nsWorkflows: Array<any>): Array<OrgWorkflowObject> {
+    const groupEntriesObject: Array<OrgWorkflowObject> = [];
+    for (let i = 0; i < nsWorkflows.length; i++) {
```
coverbeck: You could write this using `map` instead. But OK as is.

src/app/myworkflows/myworkflows.service.ts
```diff
@@ -14,39 +14,117 @@
...
+    } else {
+      this.myServicesService.getMyServices(userId);
+    }
+  }
+
   getGroupIndex(groupEntries: any[], group: string): number {
     return groupEntries.findIndex(orgWorkflow => orgWorkflow.sourceControl + '/' + orgWorkflow.organization === group);
   }
 
+  public findEntryFromPath(path: string, orgWorkflows: Array<OrgWorkflowObject>): ExtendedWorkflow {
```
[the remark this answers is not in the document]
garyluu: Pretty much all of this stuff is merely moved.  Not going to start rewriting things or else I'd never stop.

PR #692, opened 2019-07-05
npm i @types/node@10.14.0

garyluu:
For dockstore/dockstore#2535

Oddly enough, the closest ones to node 10.13.0 are 10.12.30 or 10.14.0.  There isn't anything closer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "@types/jasminewd2": "~2.0.3",
     "@types/jquery": "^3.3.15",
     "@types/jszip": "^3.1.5",
-    "@types/node": "^6.0.60",
+    "@types/node": "^10.14.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -81,7 +81,7 @@
     "@types/jasminewd2": "~2.0.3",
     "@types/jquery": "^3.3.15",
     "@types/jszip": "^3.1.5",
-    "@types/node": "^6.0.60",
+    "@types/node": "^10.14.0",
```
coverbeck: Should this be in sync with the node version in .nvmrc, v10.13.0?

I don't know what this is used for, so comment may not apply.

garyluu: The PR comment.  There  just isn't one

PR #693, opened 2019-07-05
Search organization by all string properties

garyluu:
For dockstore/dockstore#2282

Can't search by collections still, but now can search all string properties of the organizations.

### DIFF
--- a/src/app/organizations/state/organizations.query.ts
+++ b/src/app/organizations/state/organizations.query.ts
@@ -40,8 +40,6 @@ export class OrganizationsQuery extends Query<OrganizationsState> {
             const matchOptions: string[] = [
               organization.description,
               organization.displayName,
-              organization.email,
-              organization.link,
               organization.location,
               organization.name,
               organization.topic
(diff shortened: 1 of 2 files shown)


src/app/organizations/state/organizations.query.ts
```diff
@@ -36,9 +36,10 @@ export class OrganizationsQuery extends Query<OrganizationsState> {
     searchName = searchName.toLowerCase();
     if (organizations) {
       return searchName
-        ? organizations.filter(
-            organization => organization.name.toLowerCase().includes(searchName) || organization.topic.toLowerCase().includes(searchName)
-          )
+        ? organizations.filter(organization => {
+            const matchOptions: any[] = Object.values(organization).filter(objectValue => typeof objectValue === 'string');
```
coverbeck: Don't feel too strongly about this, but this is going to match against the avatarurl as well, which could potentially albeit rarely lead to a slightly confusing experience

denis-yuen: I'd agree, I would not really expect this to match against that (or other values that are not obviously visible as text)

garyluu:
Not checking all strings anymore.  Also removed the `fit`

PR #698, opened 2019-07-11
Feature/use new service endpoints

agduncan94:
Adds gitHubAppInstallationUrl as a config option. This is the base path used for linking to installs of a GitHub App.

Also users userService to pull services for a user instead of using userWorkflows.

dockstore/dockstore#2575

### DIFF
--- a/src/app/shared/session/session.query.ts
+++ b/src/app/shared/session/session.query.ts
@@ -48,6 +48,6 @@ export class SessionQuery extends Query<SessionState> {
   generateGitHubAppInstallationUrl(entryType: EntryType): string {
     let queryParams = new HttpParams();
     queryParams = queryParams.set('state', entryType);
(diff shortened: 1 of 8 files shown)


src/app/shared/session/session.query.ts
```diff
@@ -47,6 +48,6 @@ export class SessionQuery extends Query<SessionState> {
   generateGitHubAppInstallationUrl(entryType: EntryType): string {
     let queryParams = new HttpParams();
     queryParams = queryParams.set('state', entryType);
-    return 'https://github.com/apps/dockstore/installations/new?' + queryParams;
+    return Dockstore.GITHUB_APP_INSTALLATION_URL + '/installations/new?' + queryParams;
```
coverbeck: In src/app/github-name-to-id.pipe.ts you use the `HttpParams.toString()` method; seems clearer and you should probably do it here... I didn't even know this would work like this.

PR #699, opened 2019-07-11
Feature/2606/route tabs

emlys:
https://github.com/dockstore/dockstore/issues/2606
Added query components to accounts page URL so that each tab can be linked individually, e.g. dockstore.org/accounts?tab=profile. This is consistent with how the info, launch, etc. tabs on a workflow's page currently work. The Angular documentation suggests a different strategy giving each tab its own route, e.g. dockstore.org/accounts/profile (https://material.angular.io/components/tabs/overview, tabs and navigation section). Not sure if that method is better in some way, if so maybe it would be a separate ticket to switch over all tabs on the website.
Added tests for the accounts page tabs into cypress/integration/group1/dropdown.ts

### DIFF
--- a/cypress/integration/group1/dropdown.ts
+++ b/cypress/integration/group1/dropdown.ts
@@ -13,7 +13,7 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-import { setTokenUserViewPortCurator } from '../../support/commands';
+import {goToTab, setTokenUserViewPortCurator} from '../../support/commands';
 
 describe('Dropdown test', () => {
   // TODO: GitLab tests are commented out
@@ -76,6 +76,41 @@ describe('Dropdown test', () => {
 
     it('Should show all accounts as linked (except GitLab and Bitbucket)', () => {
       everythingOk();
+    });
+    setTokenUserViewPortCurator();
+    // Check that changing the tab changes the url
+    it('should default to accounts tab', () => {
+      cy.visit('/accounts');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=accounts');
+    });
+    it('Change tab to profiles', () => {
+      goToTab('Profiles');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=profiles');
+    });
+    it('Change tab to dockstore account controls', () => {
+      goToTab('Dockstore Account Controls');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=dockstore%20account%20controls');
+    });
+    it('Change tab to requests', () => {
+      goToTab('Requests');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=requests');
+    });
+    // Check that changing the url changes the tab
+    it('Link to accounts tab', () => {
+      cy.visit('/accounts?tab=accounts');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=accounts');
+    });
+    it('Link to profiles tab', () => {
+      cy.visit('/accounts?tab=profiles');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=profiles');
+    });
+    it('Link to dockstore account controls tab', () => {
+      cy.visit('/accounts?tab=dockstore%20account%20controls');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=dockstore%20account%20controls');
+    });
+    it('Link to requests tab', () => {
+      cy.visit('/accounts?tab=requests');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=requests');
     });
   });
 
--- a/cypress/integration/group1/dropdown.ts
+++ b/cypress/integration/group1/dropdown.ts
@@ -83,6 +83,10 @@ describe('Dropdown test', () => {
       cy.visit('/accounts');
       cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=accounts');
     });
+    it('should default to accounts tab when tab name misspelled', () => {
+      cy.visit('/accounts?tab=abcd');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=accounts');
+    });
     it('Change tab to profiles', () => {
       goToTab('Profiles');
       cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=profiles');
--- a/cypress/integration/group1/dropdown.ts
+++ b/cypress/integration/group1/dropdown.ts
@@ -13,7 +13,7 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-import {goToTab, setTokenUserViewPortCurator} from '../../support/commands';
+import {getTab, goToTab, setTokenUserViewPortCurator} from '../../support/commands';
 
 describe('Dropdown test', () => {
   // TODO: GitLab tests are commented out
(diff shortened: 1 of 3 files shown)


coverbeck:
Please remember to reference GitHub issue in commit message when you squash and merge.

src/app/loginComponents/accounts/accounts.component.ts
```diff
@@ -1,12 +1,53 @@
...
+  validTabs = ['accounts', 'profiles', 'dockstore account controls', 'requests'];
+  constructor(private location: Location, private router: Router) {}
+
   ngOnInit() {
     localStorage.setItem('page', '/accounts');
+    this.parseURL(this.router.url);
+  }
+
+  private parseURL(url: String): void {
+    const decodedUrl: string = decodeURIComponent(url.toString()); // decode any encoded spaces, etc. in the string
```
coverbeck: `url` is already a string; it's redundant to call `toString` again, if you go this route.

But I would suggest instead letting built-in JavaScript classes handle this instead of doing your own parsing with Regex

```
const u = new URL(url);
const tab = u.searchParams.get('tab');
...
```

src/app/loginComponents/accounts/accounts.component.ts
```diff
@@ -1,12 +1,53 @@
...
+    const strippedUrl: String = decodedUrl.endsWith('/') ? decodedUrl.slice(0, -1) : decodedUrl; // remove any trailing slash
+    this.setupTab(strippedUrl);
+  }
+
+  public setupTab(url: String) {
+    const regexp: RegExp = new RegExp('\\?tab=(.*)');
```
coverbeck: If you keep this (see previous comment), you should move this outside of the method, so that the RegExp is compiled only once.

src/app/loginComponents/accounts/accounts.component.ts
```diff
@@ -1,12 +1,53 @@
...
+import { FormControl } from '@angular/forms';
+import { Router } from '@angular/router';
 
 @Component({
   selector: 'app-accounts',
   templateUrl: './accounts.component.html'
 })
 export class AccountsComponent implements OnInit {
+  public currentTab = 'accounts'; // default to the 'accounts' tab
+  protected selected = new FormControl(0);
```
coverbeck: This seems a little mysterious to me... I don't know this part of Angular well enough to know if it is the best way.

Can you just let `selected` get set via parseUrl method? Or are you setting it because it can't be null in the accounts.component.html, on line 11?

src/app/loginComponents/accounts/accounts.component.ts
```diff
@@ -1,12 +1,53 @@
...
+  }
+
+  public setupTab(url: String) {
+    const regexp: RegExp = new RegExp('\\?tab=(.*)');
+    const match = regexp.exec(url.toString());
+    if (match) {
+      // look for a tab name in the url
+      this.currentTab = match[1]; // if found, set the tab accordingly
+    } // if not found, default to the 'accounts' tab
+    this.selectTab(this.validTabs.indexOf(this.currentTab));
```
coverbeck: What happens if the URL contains an invalid tab name? Normally it shouldn't happen, but somebody could type in the url and have a typo. You should probably catch that.

src/app/loginComponents/accounts/accounts.component.ts
emlys: I was running into issues with it being a `String` not a `string`, and `toString` seemed to fix that. Will use `searchParams` instead

src/app/loginComponents/accounts/accounts.component.ts
emlys: I believe it defaults to the accounts tab when that happens, as if you hadn't specified a tab

src/app/loginComponents/accounts/accounts.component.ts
coverbeck: I missed that -- in TypeScript you should be declaring your strings as `string`, not `String`.

emlys: Oh okay

src/app/loginComponents/accounts/accounts.component.ts
coverbeck: Could be worth adding a test that navigates to a non-existent tab and verifies that it behaves like that.

agduncan94: When the url has an invalid tab it goes to the accounts page, but does not update the url to point to accounts. I think that would be nice to have working.

src/app/loginComponents/accounts/accounts.component.ts
```diff
@@ -1,12 +1,53 @@
...
 })
 export class AccountsComponent implements OnInit {
+  public currentTab = 'accounts'; // default to the 'accounts' tab
+  selected = new FormControl(0);
+  validTabs = ['accounts', 'profiles', 'dockstore account controls', 'requests'];
+  constructor(private location: Location, private router: Router) {}
+
   ngOnInit() {
     localStorage.setItem('page', '/accounts');
+    this.parseURL(this.router.url);
```
garyluu: I probably should've mentioned this earlier, but ideally, you would use something like activatedroute queryparams

coverbeck: Ah, much better; was unaware of it.

src/app/loginComponents/accounts/accounts.component.ts
emlys: I was just following along the example of the tabs on workflow pages, but it works just the same without being set to 0. Will take that out.

src/app/loginComponents/accounts/accounts.component.ts
```diff
@@ -1,12 +1,57 @@
...
+  validTabs = ['accounts', 'profiles', 'dockstore account controls', 'requests'];
+  constructor(private location: Location, private router: Router, private activatedRoute: ActivatedRoute) {}
+
   ngOnInit() {
     localStorage.setItem('page', '/accounts');
+    this.parseParam(this.activatedRoute.queryParams);
+  }
+
+  private parseParam(params: Observable<Params>): void {
+    params.subscribe(next => {
```
garyluu: unsubscribe on destroy

src/app/loginComponents/accounts/accounts.component.ts
```diff
@@ -1,12 +1,57 @@
 import { Component, OnInit } from '@angular/core';
-import { Subject } from 'rxjs';
+import { Location } from '@angular/common';
+import { MatTabChangeEvent } from '@angular/material';
+import { FormControl } from '@angular/forms';
+import { ActivatedRoute, Params, Router } from '@angular/router';
+import { Observable } from 'rxjs';
+import Dictionary = cytoscape.Css.Dictionary;
```
garyluu: This is likely not the desired dictionary

cypress/integration/group1/dropdown.ts
```diff
@@ -77,6 +77,45 @@ describe('Dropdown test', () => {
...
+    });
+    it('Change tab to dockstore account controls', () => {
+      goToTab('Dockstore Account Controls');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=dockstore%20account%20controls');
+    });
+    it('Change tab to requests', () => {
+      goToTab('Requests');
+      cy.url().should('eq', Cypress.config().baseUrl + '/accounts?tab=requests');
+    });
+    // Check that changing the url changes the tab
```
garyluu: This doesn't really test whether the tab is active.  Should have:
```
getTab(thing).parent().should('have.class', 'mat-tab-label-active');
```

src/app/loginComponents/accounts/accounts.component.html
```diff
@@ -3,7 +3,15 @@
   Profiles, Accounts and Controls
 </app-header>
 <div class="container separation">
-  <mat-tab-group>
+  <mat-tab-group
+    [selectedIndex]="selected.value"
+    class="ds-tabs"
+    id="workflow_tabs"
+    #entryTabs
```
garyluu: `#entryTabs` and `id` should be changed/removed

cypress/integration/group1/dropdown.ts
emlys: Keep the url check for the first four tests?

agduncan94: I think URL checks are good, a combination of both checks is ideal.

PR #701, opened 2019-07-16
Workflow DOI  UI Implementation

wshands:
Initial implementation of UI for the DOI creation feature

TODO:
Will have to alter when immutable versions implemented?


src/app/workflow/workflow.component.ts
```diff
@@ -352,6 +353,18 @@ export class WorkflowComponent extends Entry implements AfterViewInit {
     }
   }
 
+  requestDOIForWorkflowVersion() {
+    const message = 'Creating DOI';
```
garyluu: Throw majority of the code into the service and add some null checks for this.workflow and this.selectedVersion

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -159,6 +159,7 @@
       <mat-card-subtitle class="m-0">{{ selectedVersion?.name }}</mat-card-subtitle>
     </mat-card-header>
     <mat-card-content class="p-3">
+      <div><strong matTooltip="DOI listed in descriptor">DOI</strong>: {{ selectedVersion.doiURL ? selectedVersion.doiURL : 'n/a' }}</div>
```
denis-yuen: A screenshot in general would be nice. 
This in particular is a bit confusing though, do we ever parse descriptors for DOIs (as opposed to issuing them via zenodo) or is this aspirational?

coverbeck: Agreed a screenshot would help visualize this...

What is `DOI listed in the descriptor`? Which descriptor? Should it just be `DOI`?

Also, more JavaScript-y way to write this would be `{{ selectedVersion.doURL || 'n/a' }}` (still not sure if that's a good TypeScript practice) :)

src/app/workflow/workflow.component.html
```diff
@@ -96,6 +96,15 @@ <h3>
           >
             View Public
           </a>
+          <button 
+            mat-flat-button
+            id="DOIButton"
+            class="mr-2"
+            color="accent"
+           (click)="requestDOIForWorkflowVersion()"
+           [disabled]="(isRefreshing$ | async) || selectedVersion.doiURL"
```
coverbeck: I think this should be `selectedVersion?doiURL`.

src/app/workflow/workflow.component.html
```diff
@@ -96,6 +96,15 @@ <h3>
...
+           [matTooltip]="createdoimessage">Create DOI
```
coverbeck: Depending on how this looks on screen, if there's room, I wonder if it shouldn't say `Create Zenodo DOI`.

wshands: Screen Shot 2019-07-18 at 10 36 49 AM

src/app/workflow/info-tab/info-tab.component.html
wshands: changed matToolTip to "Digital Object Identifier"

wshands: changed matToolTip to "Digital Object Identifier"

src/app/workflow/workflow.component.html
wshands: selectedVersion?doiURL was not accepted when I tried it

coverbeck: Sorry, `selectedVersion?.doiUrl`, which means it won't fail if selectedVersion is null/undefined.

src/app/workflow/workflow.component.ts
wshands: https://ucsc-cgl.atlassian.net/browse/DOCK-830
For a later sprint

src/app/workflow/workflow.component.html
wshands: done

src/app/workflow/workflow.component.html
```diff
@@ -96,6 +96,15 @@ <h3>
           >
             View Public
           </a>
+          <button 
+            mat-flat-button
+            id="DOIButton"
+            class="mr-2"
+            color="accent"
+           (click)="requestDOIForWorkflowVersion()"
+           [disabled]="(isRefreshing$ | async) || selectedVersion?.doiURL"
```
coverbeck: Should this be disabled if the user has not linked their Zenodo account?

Alternatively, we had talked about this redirecting the user to link their Zenodo account.

I suggest disabling for now if the account is not linked, with a tooltip explaining why. Then another ticket for the preferred solution which would be to redirect to link.

src/app/workflow/workflow.component.html
coverbeck: Ok, there's not enough room. :)

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -159,6 +159,7 @@
       <mat-card-subtitle class="m-0">{{ selectedVersion?.name }}</mat-card-subtitle>
     </mat-card-header>
     <mat-card-content class="p-3">
+      <div><strong matTooltip="Digital Object Identifier">DOI</strong>: {{ selectedVersion.doiURL ? selectedVersion.doiURL : 'n/a' }}</div>
```
coverbeck: selectedVersion can be null apparently, based on line159, so `{{ selectedVersion && selectedVersion.doiURL...`. Not sure how the Angular question mark works with the ternary operator, but there could be a more elegant way as well.

src/app/workflow/workflow.component.html
wshands: Need pointer on how to disable...

src/app/workflow/info-tab/info-tab.component.html
wshands: done

src/app/workflow/workflow.component.html
coverbeck: I replied to you on Slack, but you are presumably already doing something similar on the accounts page, which you display Link/Unlink account button based on whether they have Zenodo creds or not.

wshands: Next sprint https://ucsc-cgl.atlassian.net/browse/SEAB-342

PR #703, opened 2019-07-17
Fix navbar

garyluu:
Navbar was not quite working for mobile sizes (missing the my-entries in the dropdown menu).  This just uses one dropdown menu and one login button to avoid repetition of code

Makes the mobile version of the login identical to what it is for the normal size

### DIFF
--- a/src/app/navbar/navbar.component.html
+++ b/src/app/navbar/navbar.component.html
@@ -31,20 +31,7 @@
         </a>
         <span class="spacer"></span>
         <a *ngIf="!isLoggedIn" mat-raised-button color="accent" routerLink="/login">Login</a>
-        <button
-          data-cy="dropdown-main"
-          *ngIf="isLoggedIn"
-          mat-raised-button
-          color="accent"
-          id="dropdown-main"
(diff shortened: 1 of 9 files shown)


src/app/navbar/navbar.component.html
```diff
@@ -103,11 +76,14 @@
...
-      <span *ngIf="isLoggedIn" fxLayout="row">
-        <ng-container [ngTemplateOutlet]="loggedIn"></ng-container>
-      </span>
-    </div>
-  </mat-toolbar-row>
 </mat-toolbar>
+<!-- This is the login button used in both navbar sizes -->
+<ng-template #loginButton>
+  <button data-cy="dropdown-main" *ngIf="isLoggedIn" mat-raised-button color="accent" id="dropdown-main" [matMenuTriggerFor]="menuLarge">
+    <mat-icon class="hidden-sm hidden-md hidden-lg" *ngIf="user?.isAdmin">lock</mat-icon>
```
denis-yuen: Is this saying that admins get a lock icon, curators get an edit icon, and anyone else gets a circle? 
I think we might need a follow-up ticket for more intuitive icons.

PR #707, opened 2019-07-22
Feature/2644/services grab bag

garyluu:
For partial dockstore/dockstore#2644

The sync buttons are placeholders.  They don't actually do anything right now.  I think they're supposed to use the workflow refresh endpoint that was modified like 3 days ago but that can be done later

### DIFF
--- a/src/app/myworkflows/my-workflow/my-workflow.component.html
+++ b/src/app/myworkflows/my-workflow/my-workflow.component.html
@@ -88,13 +88,15 @@
               [refreshMessage]="isRefreshing$ | async"
             ></app-sidebar-accordion>
             <br />
-            <div class="clearfix sidebar-header">
-              <h3 class="sidebar-header">Shared with me</h3>
-            </div>
-            <p *ngIf="groupSharedEntriesObject == undefined || groupSharedEntriesObject == null || groupSharedEntriesObject.length == 0">
-              You have no {{ entryType$ | async }}s shared with you
-            </p>
-            <app-sidebar-accordion [openOneAtATime]="oneAtATime" [groupEntriesObject]="groupSharedEntriesObject"></app-sidebar-accordion>
+            <ng-container *ngIf="entryType !== EntryType.Service">
+              <div class="clearfix sidebar-header">
+                <h3 class="sidebar-header">Shared with me</h3>
+              </div>
+              <p *ngIf="groupSharedEntriesObject == undefined || groupSharedEntriesObject == null || groupSharedEntriesObject.length == 0">
+                You have no {{ entryType$ | async }}s shared with you
+              </p>
+              <app-sidebar-accordion [openOneAtATime]="oneAtATime" [groupEntriesObject]="groupSharedEntriesObject"></app-sidebar-accordion>
+            </ng-container>
           </div>
         </div>
       </mat-sidenav>
--- a/src/app/myworkflows/my-workflow/my-workflow.component.html
+++ b/src/app/myworkflows/my-workflow/my-workflow.component.html
@@ -21,12 +21,11 @@
   <img src="../assets/images/dockstore/dockstore-workflows-green.png" /> {{ myEntryPageTitle$ | async | titlecase }}
 </app-header>
 <button
-  *ngIf="entryType === EntryType.BioWorkflow"
   mat-fab
   color="accent"
   id="registerWorkflowButton"
   class="fab-bottom-right"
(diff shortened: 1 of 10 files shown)


coverbeck:
Screen shot(s) would be nice, if appropriate. :)

src/app/myworkflows/my-workflow/my-workflow.component.html
```diff
@@ -61,20 +60,20 @@
...
-                        rel="noopener noreferrer"
-                        >Manage on GitHub</a
-                      >
-                    </mat-menu>
-                    <button mat-icon-button [matMenuTriggerFor]="appMenu">
-                      <mat-icon>
-                        more_vert
-                      </mat-icon>
-                    </button>
+                      <ng-container *ngIf="entryType === EntryType.Service">
```
coverbeck: This is sort of an out there use case, because I'm not sure if we actually have any users out there like this, but in theory you could have a logged in w/Google user who doesn't have a GitHub account, in which case this button should probably be disabled or hidden.

src/app/myworkflows/my-workflow/my-workflow.component.html
```diff
@@ -88,13 +87,15 @@
...
-            </div>
-            <p *ngIf="groupSharedEntriesObject == undefined || groupSharedEntriesObject == null || groupSharedEntriesObject.length == 0">
-              You have no {{ entryType$ | async }}s shared with you
-            </p>
-            <app-sidebar-accordion [openOneAtATime]="oneAtATime" [groupEntriesObject]="groupSharedEntriesObject"></app-sidebar-accordion>
+            <ng-container *ngIf="entryType !== EntryType.Service">
+              <div class="clearfix sidebar-header">
+                <h3 class="sidebar-header">Shared with me</h3>
+              </div>
+              <p *ngIf="groupSharedEntriesObject == undefined || groupSharedEntriesObject == null || groupSharedEntriesObject.length == 0">
```
coverbeck: I would personally have `!groupSharedEntriesObject` vs. the two explicit comparisons, but you can argue it either way.

garyluu: Hide whitespace changes.  Though I do agree.

src/app/myworkflows/my-workflow/my-workflow.component.html
garyluu: That's actually pretty interesting.  The installation is not tied to the Dockstore's GitHub account at all. It's entirely possible for the user to link an old GitHub account on Dockstore, log out of that GitHub account, login with a new GitHub account, and then click on the "Manage on GitHub" link.  The next page will ask to install on Dockstore on this new GitHub account but it may not be entirely obvious to the user that their old GitHub account is linked and not the new one.

I...do not have a solution for this.

src/app/myworkflows/my-workflow/my-workflow.component.html
coverbeck: I think I lost that setting when I moved from Chrome to FireFox, or at least that's my excuse.

garyluu:
image

garyluu:
Just realized tooltips are odd...

garyluu:
image

coverbeck:
Looks good. One thought -- what if you add the "circular arrows" to the Sync button, to reinforce the button with the word "Sync" does the same thing as the buttons with just the arrows (different scopes of course).

garyluu:
> Looks good. One thought -- what if you add the "circular arrows" to the Sync button, to reinforce the button with the word "Sync" does the same thing as the buttons with just the arrows (different scopes of course).

Ya, I'm not too sure.  What I think should happen is that the left sidebar buttons should all be similar to each other (either all mat-icon-buttons or all mat-buttons with words).  Also, the right entry-specific ones should all be similar to each other (either all mat-icon-buttons or mat-buttons with words).  The left ones currently are icon buttons and I think it's somewhat intuitive enough that it's "good enough".  The right ones, on the other hand, would not be in intuitive enough because I don't think we'll have a button icon for publish and view public entry.

src/app/myworkflows/my-workflow/my-workflow.component.html
denis-yuen: @coverbeck technically, the GitHub app is independent from the user's GitHub login. i.e. a user could login with Google, but then manage the GitHub app without ever letting us know about their GitHub account

PR #709, opened 2019-07-23
#2520 added admin condition

emlys:
`updateCuratorOrganizations()` requires admin permissions. Added a condition so this does not display an error when non-admin users open their accounts page. Also removed unused variable `currentOrgId`.

### DIFF
--- a/src/app/loginComponents/requests/requests.component.ts
+++ b/src/app/loginComponents/requests/requests.component.ts
@@ -82,7 +82,7 @@ export class RequestsComponent extends Base implements OnInit {
       if (isAdmin) {
         this.requestsService.updateCuratorOrganizations(); // requires admin or curator permissions
       } else {
-        this.isCurator$.subscribe(isCurator => {
+        this.isCurator$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(isCurator => {
           if (isCurator) {
             this.requestsService.updateCuratorOrganizations();
           }
--- a/src/app/loginComponents/requests/requests.component.ts
+++ b/src/app/loginComponents/requests/requests.component.ts
@@ -5,6 +5,8 @@ import { Observable } from 'rxjs';
 import { Organization, OrganizationUser } from '../../shared/swagger';
 import { MatDialog, MatDialogRef, MAT_DIALOG_DATA } from '@angular/material';
 import { UserQuery } from '../../shared/user/user.query';
+import { Base } from '../../shared/base';
+import { takeUntil } from 'rxjs/operators';
 
 @Component({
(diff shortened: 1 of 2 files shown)


src/app/loginComponents/requests/requests.component.ts
```diff
@@ -75,6 +73,12 @@ export class RequestsComponent implements OnInit {
     this.isAdmin$ = this.userQuery.isAdmin$;
     this.isCurator$ = this.userQuery.isCurator$;
     this.userId$ = this.userQuery.userId$;
+
+    this.isAdmin$.subscribe(next => {
```
garyluu: unsubscribe

src/app/loginComponents/requests/requests.component.ts
```diff
@@ -75,6 +77,18 @@ export class RequestsComponent implements OnInit {
     this.isAdmin$ = this.userQuery.isAdmin$;
     this.isCurator$ = this.userQuery.isCurator$;
     this.userId$ = this.userQuery.userId$;
+
+    this.isAdmin$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(isAdmin => {
```
garyluu: generally, subscribing within another subscribe is a bad sign.  In this case, you can use something like 
```
combineLatest(this.admin$, this.isCurator$).pipe(takeUntil(this.ngUnsubscribe)).subscribe(([isAdmin, isCurator]:[boolean, boolean]) => {
// do stuff
});} 
```

garyluu: better yet, put the combined observable inside the userQuery as isAdminOrCurator$ and then subscribe to that and do stuff.

src/app/shared/user/user.query.ts
```diff
@@ -11,6 +11,7 @@ export class UserQuery extends Query<UserState> {
   user$: Observable<User> = this.select(state => state.user);
   isCurator$: Observable<boolean> = this.select(state => (state.user ? state.user.curator : null));
   isAdmin$: Observable<boolean> = this.select(state => (state.user ? state.user.isAdmin : null));
+  isAdminOrCurator$: Observable<boolean> = this.select(state => (state.user ? state.user.curator || state.user.isAdmin : null));
```
garyluu:
```
  isAdminOrCurator$: Observable<boolean> = combineLatest(this.isAdmin$, this.isCurator$).pipe(
    map(([isAdmin, isCurator]: [boolean, boolean]) => {
      return isAdmin || isCurator;
    })
  );
```
is one way of doing it without getting the state and avoiding null-checking the user again.  Yours is ok too

PR #711, opened 2019-07-24
Feature/reorganizeTests

garyluu:
There are now two groups of tests that reset the database.
The third group does not reset the DB and should never fail from the DB reset timeout issue

There is a small speedup (~5 mins) because the database does not reset as often. 
1 hosted tool test bug should be fixed
There should be less database reset timeout issues (but still present)

The image URL problem that almost never happens may or may not be fixed

Removed a console.log error from my-workflows

Successful run count 7/9


.circleci/config.yml
```diff
@@ -66,15 +66,7 @@ jobs:
...
-          command: bash -i -c 'npx cypress run --record --config numTestsKeptInMemory=1 --spec cypress/integration/group3/**/*'
-  integration_test_4:
-    executor: integration_test_exec
-    working_directory: ~/repo
-    steps:
-      - setup_integration_test
-      - run:
-          name: Test
-          command: bash -i -c 'npx cypress run --record --config numTestsKeptInMemory=1 --spec cypress/integration/group4/**/*'
+          command: bash -i -c 'npx cypress run --record --config numTestsKeptInMemory=1 --spec cypress/integration/testThatDoNotModifyDatabase/**/*'
```
coverbeck: How does the database get initialized for these tests? Clearly it works, since the tests passed, but I'm just wondering. resetDB() would not only reset the database after a test ran so the next test started with a clean slate, but presumably for the very first test it would set the state.

NatalieEO: Wouldn't it just be going off the state it's left in after the last test in group 2?

coverbeck: I think they run in parallel, not sequentially. Plus the tests might need the "pristine" state, rather than the state left over from other tests.

I'm pretty sure this works, just not getting how. :)

NatalieEO: Ohh right. Is it the second to last run command in this file then?

coverbeck: Ah... I don't think so, but that gave me a hint; it's a few above:

```      - run:
          name: Prepare webservice
          command: bash -i -c 'npm run webservice'
```
That runs a script which does it.

Thanks.

@garyluu I'm good on this.

PR #712, opened 2019-07-24
Feature/home logged in

garyluu:
Basic home page with placeholders for everything except the top banner which was merged in from Emily's branch.

As with most flexbox HTML, simple screenshots do not show the complete picture.  You're going to have to check out the branch (I'm also going to post a video on slack).

The placeholders are using fixed-size which should not be the case.  Expect modifications when the real components are used

All home components are placed inside the home-page module for now.  Maybe in the future, this may need to be flattened again


agduncan94:
Could you attach screenshots.

garyluu:
> Could you attach screenshots.

Read the PR -_-

denis-yuen:
Pending me getting a good internet connection

PR #713, opened 2019-07-24
Snapshot versions in place then refresh for workflows

david4096:
Needs tests and to verify that this is the expected behavior.

Seems like we may want to add an error when someone tries to mutate a frozen version from the webservice. Also, the "upsert version" endpoints might be a better way of doing this instead of modifying a workflow in place.


coverbeck:
Screenshot would be helpful

src/app/workflow/view/view.component.ts
```diff
@@ -89,6 +93,25 @@ export class ViewWorkflowComponent extends View implements OnInit {
...
+    // snapshotted versions.
+    var snapshot = { ...this.version };
+    snapshot['frozen'] = true;
+    this.workflowsService.updateWorkflowVersion(this.workflow.id, [snapshot]).subscribe(
+      result => {
+        this.alertService.detailedSuccess('Snapshot successfully created!');
+        this.refreshService.refreshWorkflow();
+      },
+      error => {
+        if (error) {
```
coverbeck: You should display an error even if `error` is not set -- call `simpleError` if not set.

src/app/workflow/view/view.component.ts
```diff
@@ -89,6 +93,25 @@ export class ViewWorkflowComponent extends View implements OnInit {
...
   }
 
+  snapshotVersion(): void {
+    // Create a new version temporarily with the frozen bit set. We assume the version is
+    // not already a snapshot since the UI controls shouldn't be available for
+    // snapshotted versions.
+    var snapshot = { ...this.version };
+    snapshot['frozen'] = true;
+    this.workflowsService.updateWorkflowVersion(this.workflow.id, [snapshot]).subscribe(
+      result => {
```
coverbeck: `result` should be an array of updated workflow versions.

1. Name should reflect it
2. Is it possible to use those versions to update the store rather than calling refreshWorkflow below?

src/app/workflow/view/view.component.ts
```diff
@@ -89,6 +93,25 @@ export class ViewWorkflowComponent extends View implements OnInit {
     });
   }
 
+  snapshotVersion(): void {
+    // Create a new version temporarily with the frozen bit set. We assume the version is
+    // not already a snapshot since the UI controls shouldn't be available for
+    // snapshotted versions.
+    var snapshot = { ...this.version };
```
coverbeck: This can be a `const`.

Also it's a shame this.version doesn't have a type, but it looks like it would take a bit of work to make it so and is outside the scope of this PR.

src/app/workflow/view/view.component.html
```diff
@@ -17,6 +17,9 @@
 <button type="button" mat-button color="accent" class="btn-block" (click)="showVersionModal()">
   {{ isPublic || !canWrite || (entryType$ | async) === EntryType.Service ? 'View' : 'Edit' }}
 </button>
+<button type="button" mat-button color="accent" class="btn-block" *ngIf="!version.frozen" (click)="snapshotVersion()">
```
coverbeck: At story time, didn't we say this was sort of a big deal, and that there should be a confirmation dialog before proceeding -- otherwise you might accidentally click/freeze something without meaning to? And/or was that for Zenodo?

Also, I forget, did we agree to go with Snapshot as the public facing name? API is calling it frozen, and TRS discussions are apparently using the word immutable.

david4096:
image

src/app/workflow/view/view.component.html
david4096: 1. I will add a dialog.

2. I am +0 on whatever name we choose but figured we might go with frozen or read-only. Can change as necessary but will become more of a hassle once we write docs.

src/app/workflow/view/view.component.ts
david4096: 1. Taking a quick look at the ws it seems like you get all workflowVersions back either way so I changed to `workflowVersions`.
2. I certainly imagine so, is that the usual pattern? I copied the modify/refresh pattern from the versionModal.

coverbeck: Not sure if it's the usual pattern, but it seems like it should be. My only concern is that there is something in the workflow itself that changes when you update a version (just speculating), but hopefully not.

src/app/workflow/view/view.component.ts
coverbeck: For now I would suggest `const snapshot: WorkflowVersion =  {...this.version};`. Then your next line could be `workflowVersion.frozen =`, and in the extremely unlikely event the response model changes, the compiler would catch it.

src/app/workflow/view/view.component.html
coverbeck: I am +0 on name too... Maybe Denis will have a stronger sense after Bosc.

garyluu:
You can make draft PR for WIP stuff

src/app/workflow/view/view.component.ts
```diff
@@ -89,6 +93,35 @@ export class ViewWorkflowComponent extends View implements OnInit {
...
+          this.alertService.simpleError();
+        }
+      }
+    );
+  }
+
+  snapshotVersion(): void {
+    // Create a new version temporarily with the frozen bit set. We assume the version is
+    // not already a snapshot since the UI controls shouldn't be available for
+    // snapshotted versions.
```
garyluu: I can tell you come from Python :smile:

david4096: lol I can think of a few reason. 1) I am trying to comment our way around a mutation operation. I presume there is a way to use the app data model to allow safe mutation operations without needing to create temporary variables? 2) My comments are not blocked. 

In either case let me know! I know that years of Python can have its toll on a person.

garyluu: I initially thought the comments were for the method.  Python usually has the comments below the method name, but Typescript usually has it above.

src/app/workflow/view/view.component.html
```diff
@@ -17,6 +17,16 @@
 <button type="button" mat-button color="accent" class="btn-block" (click)="showVersionModal()">
   {{ isPublic || !canWrite || (entryType$ | async) === EntryType.Service ? 'View' : 'Edit' }}
 </button>
+<button
+  type="button"
+  mat-button
+  color="accent"
+  class="btn-block dockstore-snapshot"
+  *ngIf="!version.frozen && version.referenceType !== 'BRANCH'"
```
garyluu: `!== WorkflowVersion.ReferenceTypeEnum.BRANCH` or similar

src/app/workflow/view/view.component.html
```diff
@@ -17,6 +17,16 @@
 <button type="button" mat-button color="accent" class="btn-block" (click)="showVersionModal()">
   {{ isPublic || !canWrite || (entryType$ | async) === EntryType.Service ? 'View' : 'Edit' }}
 </button>
+<button
+  type="button"
+  mat-button
+  color="accent"
+  class="btn-block dockstore-snapshot"
```
garyluu: If the class is purely used to interact with Cypress, use data-cy (https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements)

src/app/workflow/view/view.component.ts
```diff
@@ -89,6 +91,50 @@ export class ViewWorkflowComponent extends View implements OnInit {
...
   }
 
+  /**
+   * Updates the workflow version and alerts the Dockstore User with success
+   * or failure.
+   *
+   * @private
+   * @memberof ViewWorkflowComponent
+   */
+  private updateWorkflowToSnapshot(version): void {
```
garyluu: All of this should be moved to a service.  The only thing in this file should be:
```
snapshotVersion() {
   this.potatoService.snapshotVersion(parameters);
}
```
and something to get the enum

Also, add type to the parameter

cypress/integration/immutableDatabaseTests/workflowDetails.ts
```diff
@@ -61,13 +61,23 @@ describe('Dockstore Workflow Details', () => {
...
+  it('Change tab to versions and snapshot', () => {
     goToTab('Versions');
     cy
       .get('tbody>tr')
       .should('have.length', 1); // 1 Version and no warning line
     cy.url().should('eq', Cypress.config().baseUrl + '/workflows/github.com/A/l:master?tab=versions');
-  });
+
+    cy
+      .get('.dockstore-snapshot-locked').should('have.length', 0);
```
garyluu: might as well check to see unlocked is there.

src/app/workflow/view/view.component.ts
```diff
@@ -89,6 +91,50 @@ export class ViewWorkflowComponent extends View implements OnInit {
...
+          this.alertService.simpleError();
+        }
+      }
+    );
+  }
+  /**
+   * Opens a confirmation dialog that the Dockstore User can use to
+   * confirm they want a snapshot.
+   *
+   * @private
```
garyluu: lies!

src/app/workflow/view/view.component.ts
```diff
@@ -89,6 +91,50 @@ export class ViewWorkflowComponent extends View implements OnInit {
...
+   * @memberof ViewWorkflowComponent
+   */
+  snapshotVersion(): void {
+    // Create a new version temporarily with the frozen bit set. We assume the version is
+    // not already a snapshot since the UI controls shouldn't be available for
+    // snapshotted versions.
+    const snapshot = { ...this.version };
```
garyluu: try `const snapshot: WorkflowVersion = { ...this.version, frozen: true };`

src/app/workflow/view/view.component.html
```diff
@@ -17,6 +17,16 @@
...
+  *ngIf="!version.frozen && version.referenceType !== 'BRANCH'"
+  (click)="snapshotVersion()"
```
garyluu: disable button upon "isRefreshing" to prevent someone with a really slow internet connection from mashing the button

src/app/workflow/view/view.component.ts
```diff
@@ -89,6 +91,50 @@ export class ViewWorkflowComponent extends View implements OnInit {
...
+    snapshot.frozen = true;
+    const confirmMessage = 'Snapshotting a version cannot be undone. Are you sure you want to snapshot this version?';
```
garyluu: Though it will likely never happen, double-check that the version is not already frozen before proceeding.  Also, make an issue to hook up the new confirmation dialog with this

src/app/workflow/view/view.component.ts
agduncan94: I think this is the kind of thing that would be useful in a tutorial/talk/whatever type of session we have that was brought up by David in retro. Knowing what a service is, when to use it, how it fits in with Angular and Dockstore.

src/app/workflow/versions/versions.component.html
```diff
@@ -81,6 +81,14 @@
         </label>
       </th> -->
 
+          <th>
+            <label
+              placement="top"
+              matTooltip="The metadata and descriptors for a workflow that has been snapshotted will not change over time."
```
coverbeck: ...for a workflow **version** that has been...

or

...for a version that has been...

src/app/workflow/view/view.component.ts
david4096: Thanks both! Going to do as part of https://github.com/dockstore/dockstore/issues/2646

src/app/workflow/view/view.component.html
david4096: Going to do this as part of https://github.com/dockstore/dockstore/issues/2646 will need a little more explanation of the right way to do it.

PR #717, opened 2019-07-29
Add new simple confirmation dialog

garyluu:
Adds a simple confirmation dialog.  Use it by subscribing to the service's openDialog() function.  

Remove members confirmation dialog was replaced by  this


garyluu:
image

garyluu:
> I would also make the close button text optionally user-defined -- I can't think of a specific example right now, but I've seem examples where it might be nice to have more explicit text rather than a generic Cancel.
> 
> Future feature request for really momentous actions: force user to type something as well before Confirm button gets enabled.

Ya, there's a lot of things that can be modified (I mentioned it in the comment).  But since this actually isn't used anywhere right now.  It's best that it's modified when needed.

denis-yuen:
> Ya, there's a lot of things that can be modified (I mentioned it in the method comment). But since this actually isn't used anywhere right now. It's best that it's modified when needed.

Could hook it up with the user self destruct dialogue box to demo how it would actually be used, would kill two birds with one stone

garyluu:
Replaced the member removal dialog with this one.  Had to use innerHTML in order to bold certain things.

PR #718, opened 2019-07-30
Feature/add service test

garyluu:
Adds a new service for testing via generateChangeLog + diff (hopefully there is some better way).  

Uses the 1.7.0-beta.3 webservice


travisci/add_service.xml
```diff
@@ -0,0 +1,409 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
```
denis-yuen: Move this to dockstore-webservice/src/main/resources like dockstore-webservice/src/main/resources/migrations.test.samepaths.xml to keep things organized

Change the name to add_service_1.7.0.xml ala our confidential migrations so we know what version this was for

garyluu:
> Make sure that any docs (wiki) that reference how to do migrations includes this new file and the purpose of it.

This migration file is only used for tests and nowhere else.  CircleCI config.yaml has the commands needed.  Should be obvious

cypress/integration/immutableDatabaseTests/services.ts
```diff
@@ -32,19 +32,19 @@ describe('Dockstore Home', () => {
     });
   });
 
-  // TODO: This is a fake service, need to use a more realistic one later
   describe('services', () => {
-    it('Have no services in /services', () => {
+    it('Have one services in /services', () => {
```
coverbeck: service not services

scripts/run-webservice-script.sh
```diff
@@ -10,5 +10,5 @@ psql -h localhost -c "create user dockstore with password 'dockstore' createdb;"
 psql -h localhost -c "ALTER USER dockstore WITH superuser;" -U postgres
 psql -h localhost -c 'create database webservice_test with owner = dockstore;' -U postgres
 psql -h localhost -f travisci/db_dump.sql webservice_test -U postgres
-java -jar dockstore-webservice.jar db migrate -i 1.5.0,1.6.0,1.7.0 travisci/web.yml
+java -jar dockstore-webservice.jar db migrate -i 1.5.0,1.6.0,1.7.0,add_service_1.7.0 travisci/web.yml
```
denis-yuen: Just double checking, but won't this add the test service to all tests whether they're testing services or not?

garyluu: ya, but it shouldn't really be a problem.  Our initial DB also adds tools/workflows for all tests

denis-yuen: Ok, just making sure it was intentional

PR #719, opened 2019-07-30
WIP: Enable graying of create DOI when zenodo credentials are not linked

wshands:
addresses https://ucsc-cgl.atlassian.net/browse/SEAB-342

Need feedback on implementation...

### DIFF
--- a/src/app/workflow/workflow.component.ts
+++ b/src/app/workflow/workflow.component.ts
@@ -84,7 +84,7 @@ export class WorkflowComponent extends Entry implements AfterViewInit {
   public WorkflowModel = Workflow;
 
   public tokens: Token[];
-  protected zenodoAccountIsLinked = false;
+  public zenodoAccountIsLinked = false;
 
   @Input() user;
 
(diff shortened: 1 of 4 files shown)


garyluu:
Rebase after I merge in my fix

src/app/workflow/workflow.component.ts
```diff
@@ -238,6 +245,24 @@ export class WorkflowComponent extends Entry implements AfterViewInit {
     }
   }
 
+  // Show linked services in the UI
+  private setAvailableTokens(tokens) {
+      const found = tokens.find(token => token.tokenSource === 'zenodo.org');
```
coverbeck: `this.zenodoAccountIsLinked = tokens.some(token => token.tokenSource === 'zenodo.org');`

src/app/workflow/workflow.component.ts
```diff
@@ -253,6 +278,9 @@ export class WorkflowComponent extends Entry implements AfterViewInit {
       }
       this.setUpWorkflow(workflow);
     });
+    this.tokenQuery.tokens$.subscribe((tokens: Token[]) => {
+      this.setTokens(tokens);
```
coverbeck: I don't think you need to save the tokens; just set zenodoAccountIsLinked here.

garyluu:
Could use a simple test for this

cy.get('[data-cy=thing]').should(be.disabled)

as I'm not entirely sure how this works

src/app/workflow/workflow.component.ts
wshands: done

src/app/workflow/workflow.component.ts
wshands: done

wshands:
Closing so I can create a non draft pull request

PR #724, opened 2019-07-31
Feature/disable services

garyluu:
This adds a beta mode that removes all the not quite ready stuff.

Literally copied and pasted the old home page back in as another component and display that as the logged-in homepage.  Need to remember to remove it afterwards (there's a bunch of imports to remove too).

Services is also disabled.

You can however still visit the my-services page manually as well as the new logged-in home page at /beta-homepage

### DIFF
--- a/cypress/integration/immutableDatabaseTests/elasticsearch.ts
+++ b/cypress/integration/immutableDatabaseTests/elasticsearch.ts
@@ -38,7 +38,7 @@ describe('elasticsearch from logged-in homepage', () => {
     it('Should be able to search entries for using text', () => {
       cy.visit('/');
       cy.get('#searchBar')
-        .should('have.attr', 'placeholder', 'Search...')
+        .should('have.attr', 'placeholder', 'Enter Keyword...')
         .type('asdf{enter}');
       cy.url().should('eq', Cypress.config().baseUrl + '/search?search=asdf');
     });
--- a/cypress/integration/immutableDatabaseTests/elasticsearch.ts
+++ b/cypress/integration/immutableDatabaseTests/elasticsearch.ts
@@ -13,7 +13,6 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-import { betaMode } from 'app/shared/constants';
 import { setTokenUserViewPort } from '../../support/commands';
 
 describe('elasticsearch from logged-out homepage', () => {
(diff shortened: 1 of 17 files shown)


denis-yuen:
> This adds a beta mode that removes all the not quite ready stuff.

Or rather, they are inaccessible to a casual user

denis-yuen:
> Literally copied and pasted the old home page back in as another component and display that as the logged-in homepage. Need to remember to remove it afterwards (there's a bunch of imports to remove too).

Can we get a screenshot for this?

garyluu:
image

cypress/integration/immutableDatabaseTests/elasticsearch.ts
```diff
@@ -27,13 +27,19 @@ describe('elasticsearch from logged-out homepage', () => {
...
 describe('elasticsearch from logged-in homepage', () => {
   setTokenUserViewPort();
-  it('Should be able to search entries for using text', () => {
-    cy.visit('/');
-    cy.get('[data-cy=loggedInSearchBar]')
-      .should('have.attr', 'placeholder', 'Search...')
-      .type('asdf{enter}');
-    cy.url().should('eq', Cypress.config().baseUrl + '/search?search=asdf');
-  });
+  // Use this test when beta mode turned off
```
denis-yuen: Can you not access the constant from a test?

garyluu: Tried, didn't work. Gave up

coverbeck:
Just brainstorming here, maybe you've already thought of this:

1. I'm not clear why you need to revert to the old home page. Isn't the only difference the between having services and not having it the dropdown menu choice `My Services`? Probably not, but it's the only visible difference I see on staging right now.
2. Depending on the answer to above, could you just specify the beta feature as a query parameter instead of a new route?
3. It would be nice to know you're in beta mode. And/or what are the beta features. Something like a `beta` label next to `My Services`.

denis-yuen:
> 1. I'm not clear why you need to revert to the old home page. Isn't the only difference the between having services and not having it the dropdown menu choice `My Services`? Probably not, but it's the only visible difference I see on staging right now.

Intention was to have a fallback plan since services should only really be visible to launch-with partners and us to experiment with, we don't seem to have enough hands to push the logged in homepage over the hump. 

Maybe instead of calling it a beta mode, maybe it should be a "labs" mode when services and the new logged in homepage are visible

garyluu:
> Just brainstorming here, maybe you've already thought of this:
> 
> 1. I'm not clear why you need to revert to the old home page. Isn't the only difference the between having services and not having it the dropdown menu choice `My Services`? Probably not, but it's the only visible difference I see on staging right now.
> 2. Depending on the answer to above, could you just specify the beta feature as a query parameter instead of a new route?
> 3. It would be nice to know you're in beta mode. And/or what are the beta features. Something like a `beta` label next to `My Services`.

2. I prefer to reduce the chances of the user accidentally landing on this new page while keeping the code as separate as possible between the two.  Using a query parameter would likely couple the two components more than I would like.
3. This beta feature is pretty much something only we'd access.  And there should be a smooth transition to bring it out of beta.  Adding labels and such complicates things.  I agree maybe this should've been called dev mode or something.

PR #725, opened 2019-07-31
Case insensitive sorting for workflows/services

denis-yuen:
https://github.com/dockstore/dockstore/issues/2685

### DIFF
--- a/src/app/shared/myentries.service.ts
+++ b/src/app/shared/myentries.service.ts
@@ -109,10 +109,10 @@ export abstract class MyEntriesService extends Base {
 
     groupEntries.forEach(groupEntry => {
       groupEntry.entries.sort((a, b) => {
-        if (a[path] < b[path]) {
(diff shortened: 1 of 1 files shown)


src/app/shared/myentries.service.ts
```diff
@@ -109,10 +109,10 @@ export abstract class MyEntriesService extends Base {
 
     groupEntries.forEach(groupEntry => {
       groupEntry.entries.sort((a, b) => {
-        if (a[path] < b[path]) {
+        if (a[path].toLowerCase() < b[path].toLowerCase()) {
```
coverbeck: Is it safe to assume that the value of `path` is never null/undefined?

denis-yuen: `const path = type === EntryType.BioWorkflow || type === EntryType.Service ? 'full_workflow_path' : 'tool_path';` 
I think `path` will always be one of the two values. 
However, `a[path]` might be null/undefined

PR #727, opened 2019-08-02
Feature/2592/shared workflows error

garyluu:
dockstore/dockstore#2592

Very odd way of handling getting user's shared workflows and normal workflows but it works.

What happened was the failure of any of the two endpoints would remove the progress bar immediately. 

This waits for both endpoints even when it fails.  If not individually caught, forkJoin typically abandons everything once there's an error.  This catches the error and continues.  Once both calls finish, it figures out if there's was a caught error.

dockstore/dockstore#2693 also

### DIFF
--- a/src/app/home-page/home-logged-out/home.component.scss
+++ b/src/app/home-page/home-logged-out/home.component.scss
@@ -150,10 +150,6 @@ span {
   }
 }
 
-.youtube {
-  z-index: 1070;
-}
-
 .small-card,
 .small-card > mat-card {
   min-height: 150px;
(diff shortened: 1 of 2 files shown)


src/app/home-page/home-logged-out/home.component.scss
```diff
@@ -150,10 +150,6 @@ span {
   }
 }
 
-.youtube {
```
denis-yuen: Is this https://github.com/dockstore/dockstore/issues/2693 ?

garyluu: ya, forgot to mention

PR #728, opened 2019-08-02
also show org members when not logged in

agduncan94:
Visitors not logged in will be able to see org members.

https://github.com/dockstore/dockstore/issues/2458

### DIFF
--- a/src/app/organizations/state/organization-members.service.ts
+++ b/src/app/organizations/state/organization-members.service.ts
@@ -96,7 +96,7 @@ export class OrganizationMembersService {
             const currentUserId = this.userQuery.getSnapshot().user.id;
             const canEdit = organizationUsers.some(user => user.id.userId === currentUserId && user.accepted);
             const canEditMembers = organizationUsers.some(
-              user => user.id.userId === currentUserId && user.accepted && user.role === 'MAINTAINER'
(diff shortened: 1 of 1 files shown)


src/app/organizations/state/organization-members.service.ts
```diff
@@ -92,12 +92,16 @@ export class OrganizationMembersService {
...
-          const canEdit = organizationUsers.some(user => user.id.userId === currentUserId && user.accepted);
-          const canEditMembers = organizationUsers.some(
-            user => user.id.userId === currentUserId && user.accepted && user.role === 'MAINTAINER'
-          );
-          this.setCanEditState(canEdit, canEditMembers);
+          if (this.userQuery.getSnapshot().user) {
+            const currentUserId = this.userQuery.getSnapshot().user.id;
+            const canEdit = organizationUsers.some(user => user.id.userId === currentUserId && user.accepted);
+            const canEditMembers = organizationUsers.some(
+              user => user.id.userId === currentUserId && user.accepted && user.role === 'MAINTAINER'
```
coverbeck: Using the `OrganizationUser.RoleEnum.MAINTAINER` enum would be preferable to  `'MAINTAINER'`. But was already like that.

garyluu: I didn't notice this because I always hide whitespace changes :smile:

PR #731, opened 2019-08-06
Clarify message

garyluu:
For dockstore/dockstore#2663

If you're wondering why CircleCI didn't run...I have no idea either

### DIFF
--- a/src/app/workflow/permissions/permissions.component.html
+++ b/src/app/workflow/permissions/permissions.component.html
@@ -15,6 +15,10 @@
       During the limited beta of sharing functionality, your Google account must also be registered with
       <a [href]="terraUrl" target="_blank">Terra</a>.
     </p>
+    <p>
+      If your Google account is already registered with Terra, your refresh token may be expired. Please relink your Google account via the
+      Accounts page.
+    </p>
   </div>
   <div style="font-style: italic; font-size: smaller; line-height: 1.5" *ngIf="canViewPermissions">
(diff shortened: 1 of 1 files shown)


denis-yuen:
It's not clear to me (maybe a screenshot would be in order) what this would look like. 
Also, is this a warning message that would appear only when the Google token is expired or all the time?

coverbeck:
> It's not clear to me (maybe a screenshot would be in order) what this would look like.
> Also, is this a warning message that would appear only when the Google token is expired or all the time?

It would be nice if we could display the message only when the refresh token has expired, in response to a failed Ajax call. The problem is that I don't think the Dockstore API easily lets us glean this information.

How about setting a flag in the call to `getWorkflowPermissions` if the response is 401 (or 403), and only displaying this then?

denis-yuen:
> How about setting a flag in the call to `getWorkflowPermissions` if the response is 401 (or 403), and only displaying this then?

That sounds good for now, maybe in a later release we can check on the status of tokens before displaying them in the linked accounts page.

garyluu:
> It's not clear to me (maybe a screenshot would be in order) what this would look like.
> Also, is this a warning message that would appear only when the Google token is expired or all the time?

It's displayed when the getWorkflowPermissions endpoint has not returned a valid response.

garyluu:
image

garyluu:
Basically, if you have a Google account and for whatever reason, permissions cannot be gotten (either due to Terra not being linked or Google refresh token expired), this message will show

PR #732, opened 2019-08-07
Hookup services sync buttons to API

coverbeck:
dockstore/dockstore#2698


src/app/shared/refresh.service.ts
```diff
@@ -139,4 +139,28 @@ export class RefreshService {
...
+      },
+      error => this.alertService.detailedError(error)
+    );
+  }
+
+  syncServicesForOrganziation(organization: string): void {
+    const message = 'Syncing services for organization ' + organization;
+    this.alertService.start(message);
+    this.usersService.syncUserServicesbyOrganization(organization).subscribe(
+      response => {
```
garyluu: rename to something more descriptive, add a response type too

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "1.7.0-beta.1"
+    "webservice_version": "1.7.0-beta.2"
```
garyluu: RIP

coverbeck: I'm guessing this is why my tests are failing. :(

src/app/shared/refresh.service.ts
coverbeck: Pushed fix.

src/app/shared/refresh.service.ts
```diff
@@ -139,4 +139,28 @@ export class RefreshService {
...
     this.containerService.setTools(this.tools);
   }
+
+  syncServices(): void {
+    const message = 'Syncing services';
+    this.alertService.start(message);
+    this.usersService.syncUserServices().subscribe(
+      response => {
+        this.alertService.detailedSuccess();
+        this.workflowService.setWorkflows(response);
```
garyluu: Rename response to something more descriptive and add type.

garyluu:
If you want, you can make an update workflows function that takes in Observable workflow and error message since the two sync functions (as well as others) are surprisingly similar

src/app/shared/refresh.service.ts
coverbeck: Fixed by making a common function as you suggested below.

PR #737, opened 2019-08-14
Feature/2742/hide snapshot public page

garyluu:
For dockstore/dockstore#2742

Hide the snapshot button in the versions tab when in public page
There is now a test for each of them

### DIFF
--- a/src/app/workflow/view/view.component.html
+++ b/src/app/workflow/view/view.component.html
@@ -17,24 +17,26 @@
 <button type="button" mat-button color="accent" class="btn-block" (click)="showVersionModal()">
   {{ isPublic || !canWrite || (entryType$ | async) === EntryType.Service ? 'View' : 'Edit' }}
 </button>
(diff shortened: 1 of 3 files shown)


david4096:
+1

src/app/workflow/view/view.component.html
```diff
@@ -17,24 +17,26 @@
...
->
-  Delete
-</button>
+<ng-container *ngIf="!isPublic">
+  <button
+    type="button"
+    mat-button
+    color="accent"
+    class="btn-block"
+    *ngIf="!version.frozen && version.referenceType !== 'BRANCH'"
```
coverbeck: Was like this, but I wonder if we need `version?...` since line 37 has `workflow?`.

garyluu: I think what should happen is that this component should be entirely hidden if there's no version

PR #738, opened 2019-08-14
Fixed bug, add comment

garyluu:
For dockstore/dockstore#2679

Bandage for the search from homepage.

### DIFF
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -18,7 +18,7 @@ import { MatAccordion } from '@angular/material';
 import { faSort, faSortAlphaDown, faSortAlphaUp, faSortNumericDown, faSortNumericUp } from '@fortawesome/free-solid-svg-icons';
 import { Observable, Subject } from 'rxjs';
 import { debounceTime, distinctUntilChanged, takeUntil } from 'rxjs/operators';
-
+import { AlertService } from '../shared/alert/state/alert.service';
 import { formInputDebounceTime } from '../shared/constants';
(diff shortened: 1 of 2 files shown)


src/app/search/search.component.ts
```diff
@@ -28,8 +28,19 @@ import { ELASTIC_SEARCH_CLIENT } from './elastic-search-client';
 import { QueryBuilderService } from './query-builder.service';
 import { SearchQuery } from './state/search.query';
 import { SearchService } from './state/search.service';
-import { AlertService } from '../shared/alert/state/alert.service';
 
+/**
+ * There are a total of 6 calls per search.
+ * 2 calls are from the tag cloud (1 for tag, 1 for workflow)
```
agduncan94: Do you mean "1 for tool, 1 for workflow"?

src/app/search/search.component.ts
```diff
@@ -28,8 +28,19 @@ import { ELASTIC_SEARCH_CLIENT } from './elastic-search-client';
 import { QueryBuilderService } from './query-builder.service';
 import { SearchQuery } from './state/search.query';
 import { SearchService } from './state/search.service';
-import { AlertService } from '../shared/alert/state/alert.service';
 
+/**
+ * There are a total of 6 calls per search.
```
coverbeck: I always wondered why I saw so many calls in Network Inspector. :)

garyluu: Ya, it's always annoyed me that we have to do 2 calls for the verified/non-verified buckets because the index is not as expected.

PR #740, opened 2019-08-15
Feature/2609/force tos acceptance

NatalieEO:
dockstore/dockstore#2609
needs webservice change/release to pass tests.


scripts/run-webservice-script.sh
```diff
@@ -10,5 +10,5 @@ psql -h localhost -c "create user dockstore with password 'dockstore' createdb;"
 psql -h localhost -c "ALTER USER dockstore WITH superuser;" -U postgres
 psql -h localhost -c 'create database webservice_test with owner = dockstore;' -U postgres
 psql -h localhost -f travisci/db_dump.sql webservice_test -U postgres
-java -jar dockstore-webservice.jar db migrate -i 1.5.0,1.6.0,1.7.0,add_service_1.7.0 travisci/web.yml
+java -jar /Users/natalieperez/Projects/dockstore/dockstore-webservice/target/dockstore-webservice-1.7.0-beta.3-SNAPSHOT.jar db migrate -i 1.5.0,1.6.0,1.7.0,add_service_1.7.0,alter_test_user_1.7.0 travisci/web.yml
```
coverbeck: Don't think you want that in there :)

src/app/policy-update/policy-update.component.html
```diff
@@ -0,0 +1,26 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+<app-header>
+  Policy Agreement Required
+</app-header>
+<div class="container">
+  <div class="column">
+    <p class="lead">
+      You have been logged out because it looks like you need to agree to our latest Terms of Service and Privacy Policy. Please
```
coverbeck: Sort of torn on "looks like". On the one hand it's more polite. On the other hand, it may seem like we don't know. :)

src/app/navbar/navbar.component.ts
```diff
@@ -38,6 +38,8 @@ export class NavbarComponent extends Logout implements OnInit {
   isExtended = false;
   devMode = devMode;
   protected ngUnsubscribe: Subject<{}> = new Subject();
+  currentTOSVersion: User.TosversionEnum = User.TosversionEnum.TOSVERSION1;
```
coverbeck: These should be private.

src/app/policy-update/policy-update.component.html
NatalieEO: Oh yeah, I was going for polite sounding but I can see the other point of view now that you mention it. If others agree for a firmer message, I will change it.

agduncan94: I feel like this should be more generic like "You session has ended" or "You have been signed out because your session has expired"

coverbeck: > I feel like this should be more generic like "You session has ended" or "You have been signed out because your session has expired"

I think we should make it explicit to the user, but maybe that's overkill since the log-in page has the passive acceptance.

garyluu:
Waiting for test pass

PR #742, opened 2019-08-15
Feature/different deploy path

denis-yuen:
* serve images from CDN
* rewrite image paths prior to circle ci upload
* standardize on `../assets/` syntax for images destined for retrieval from CDN

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,9 @@ jobs:
           name: Check dependencies
           command: bash -i -c 'npm ls'
       - run:
+          name: Swap for CDN paths
+          command: bash -i -c 'find src \( -iname '*.html' -o -iname '*.ts' \) -exec sed -i 's/\.\.\/assets\//https:\/\/gui.dockstore.org\/${CIRCLE_TAG:-$CIRCLE_BRANCH}-$(echo $CIRCLE_SHA1 | cut -c -7)\/assets\//g' {} +'
+      - run:
           name: Lint
           command: bash -i -c 'npm run lint'
       - run:
@@ -88,16 +91,6 @@ jobs:
                 arguments: |
                   --recursive \
                   --exclude index.html
-#            - aws-s3/copy:
-#                from: 's3://${AWS_BUCKET}/${CIRCLE_TAG:-$CIRCLE_BRANCH}-$(echo $CIRCLE_SHA1 | cut -c -7)'
-#                to: 's3://${AWS_BUCKET}/${CIRCLE_TAG:-$CIRCLE_BRANCH}-$(echo $CIRCLE_SHA1 | cut -c -7)'
-#                arguments: |
-#                  -exclude '*' \
-#                  --include '*.js' \
-#                  --no-guess-mime-type \
-#                  --content-type="application/javascript" \
-#                  --metadata-directive="REPLACE" \
-#                  --recursive
             - aws-s3/copy:
                 from: dist/index.html
                 to: 's3://${AWS_BUCKET}/${CIRCLE_TAG-$CIRCLE_BRANCH}-$(echo $CIRCLE_SHA1 | cut -c -7)/index.html'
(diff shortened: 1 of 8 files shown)


coverbeck:
@denis-yuen I can't add you as a reviewer, even if remove you as an assignee -- I guess because you created the PR. Please take a look.

denis-yuen:
> @denis-yuen I can't add you as a reviewer, even if remove you as an assignee -- I guess because you created the PR. Please take a look.

LGTM, but GitHub is probably right in this case. 
I'm biased. 

I do see 02e40a8 making it to github properly

.circleci/config.yml
```diff
@@ -164,7 +171,7 @@ workflows:
               only: /.*/
             branches:
               # Note to reviewers: will change to develop before merging, once build works off the branch
-              only: feature/SEAB-314/cd
+              only: /.*/
```
coverbeck: This and line 181 were intended to be `develop`. Line 181 should definitely be develop (I overlooked it when I merged). This one could arguably continue to be its current value; it might be useful to have have builds from all branches on s3.

denis-yuen: This is the distro for cloud.dockstore.org right?
TBH,  I don't know that "develop(no hash)" is useful anymore. 

At least with the current approach, in compose the VM always has a specific hash fragment for a GUI deployment that it is pointing at on gui.dockstore.org. If we were to build a CI/CD, I seems like we should just point to a deployment with a correct branch/tag-hash value.

denis-yuen:
> I'm going to fix line 181. Curious what you are thoughts are on line 174.

To maybe be a bit more clear:
1) I actually like all builds to be available although maybe we'll need to clean them out after a while (maybe all builds that aren't feature branches?)
2) For the eventual CI/CD, I don't think we'll need a IAM role editing cloudfront anymore. 
Rather, compose_setup should just set the UI fragment hash (i.e. `https://gui.dockstore.org/<fragment-hash>/`) correctly and that fragment-hash should be prominently displayed in the UI for debugging rather than a simple "develop" which changes over time

.circleci/config.yml
coverbeck: > This is the distro for cloud.dockstore.org right?

Yes, this is for cloud.dockstore.org, whose whole approach, I just realized, probably doesn't make sense any more given the way we are generating the UI -- why have have an index.html on cloud.dockstore.org in turn load files from gui.dockstore.org?

I do think we should come back to it, so for now I think we should not execute the CF invalidation any more (which is what you say in your other comment).

coverbeck:
>     2. For the eventual CI/CD, I don't think we'll need a IAM role editing cloudfront anymore.
>        Rather, compose_setup should just set the UI fragment hash (i.e. `https://gui.dockstore.org/<fragment-hash>/`) correctly and that fragment-hash should be prominently displayed in the UI for debugging rather than a simple "develop" which changes over time

I was thinking we should eventually (in 1.8/1.9) move to "full Cloudfront", but maybe not. Maybe we discuss in tomorrow's meeting.

.circleci/config.yml
```diff
@@ -79,8 +79,15 @@ jobs:
...
           condition: <<parameters.aws_bucket>>
           steps:
-            - attach_workspace:
-                at: .
+            - get_workspace
+            - install_container_dependencies
+            # Build again to leverage CloudFront
+            - run:
+                name: Swap for CDN paths
+                command: bash -i -c "find src \( -iname '*.html' -o -iname '*.ts' -o -iname '*.json' \) -exec sed -i 's~\.\./assets/~https://gui\.dockstore\.org/${CIRCLE_TAG:-$CIRCLE_BRANCH}-$(echo $CIRCLE_SHA1 | cut -c -7)/assets/~g' {} +"
```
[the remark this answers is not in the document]
denis-yuen: Indeed. 
Or maybe some form of circle-ci variable (don't know if variables survive between steps)

PR #743, opened 2019-08-15
Feature/2661/show modified hosted version immediately

garyluu:
For dockstore/dockstore#2661

Show the modified hosted version immediately

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -42,10 +42,8 @@ export function assertNoTab(tabName: string): any {
 export function resetDB() {
   before(() => {
     cy.exec('java -jar dockstore-webservice.jar db drop-all --confirm-delete-everything travisci/web.yml');
-    cy.exec('PGPASSWORD=dockstore psql -h localhost -f travisci/db_dump.sql webservice_test -U dockstore')
-      .its('stdout').should('contain', 'ALTER TABLE');
-    cy.exec('java -jar dockstore-webservice.jar db migrate -i 1.5.0,1.6.0,1.7.0 travisci/web.yml')
-      .its('stdout').should('contain', 'Successfully released change log lock');
+    cy.exec('PGPASSWORD=dockstore psql -h localhost -f travisci/db_dump.sql webservice_test -U dockstore');
+    cy.exec('java -jar dockstore-webservice.jar db migrate -i 1.5.0,1.6.0,1.7.0 travisci/web.yml');
   });
 }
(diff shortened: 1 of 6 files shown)


src/app/shared/entry.ts
```diff
@@ -220,7 +220,15 @@ export abstract class Entry implements OnInit, OnDestroy {
       return null;
     }
     const selectedTag = this.selectVersion(versions, urlVersion, defaultVersion);
-    return selectedTag || versions.reduce((a, b) => (b.last_built > a.last_built ? b : a));
+    return (
+      selectedTag ||
+      versions.reduce((a, b) => {
+        if (!b.last_built && !a.last_built) {
```
coverbeck: A comment might be nice, e.g., newly created versions don't have a last_built (if that's what's going on).

src/app/shared/entry.ts
```diff
@@ -220,7 +220,15 @@ export abstract class Entry implements OnInit, OnDestroy {
...
+          return b['dbUpdateDate'] > a['dbUpdateDate'] ? b : a;
```
coverbeck: Just curious, why are you using this syntax instead of `b.dbUpdateDate`? Does TypeScript still do type checking this way?

garyluu: was back before the webservice was ready

cypress/support/commands.ts
```diff
@@ -42,10 +42,8 @@ export function assertNoTab(tabName: string): any {
 export function resetDB() {
   before(() => {
     cy.exec('java -jar dockstore-webservice.jar db drop-all --confirm-delete-everything travisci/web.yml');
-    cy.exec('PGPASSWORD=dockstore psql -h localhost -f travisci/db_dump.sql webservice_test -U dockstore')
-      .its('stdout').should('contain', 'ALTER TABLE');
-    cy.exec('java -jar dockstore-webservice.jar db migrate -i 1.5.0,1.6.0,1.7.0 travisci/web.yml')
-      .its('stdout').should('contain', 'Successfully released change log lock');
+    cy.exec('PGPASSWORD=dockstore psql -h localhost -f travisci/db_dump.sql webservice_test -U dockstore');
```
denis-yuen: This seems to be removing tests, intentional?

garyluu: Originally I was using it to debug the reset postgres timeout issue.  Didn't seem to give me any useful information.  Removing it because it's too noisy.

PR #744, opened 2019-08-16
Feature/2487/home page fixes

garyluu:
For dockstore/dockstore#2487

Fixes some home page bugs.

Also does dockstore/dockstore#2753

Every route change scrolls back to the top, I wonder if I should just limit to search page instead.

### DIFF
--- a/src/app/footer/footer.component.html
+++ b/src/app/footer/footer.component.html
@@ -57,6 +57,11 @@
               <small><mat-icon>rss_feed</mat-icon>RSS</small>
             </a>
           </li>
+          <li class="m-0">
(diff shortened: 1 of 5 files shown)


src/app/footer/footer.component.html
```diff
@@ -57,6 +57,11 @@
               <small><mat-icon>rss_feed</mat-icon>RSS</small>
             </a>
           </li>
+          <li class="m-0">
+            <a routerLink="/sitemap" target="_blank" rel="noopener noreferrer" class="footer-link">
+              <small>Sitemap</small>
```
coverbeck: 👍 I hated manually typing `sitemap` in the address bar

PR #745, opened 2019-08-20
Fix descriptor types for STUBs

garyluu:
A simple fix for the descriptor type in STUB

dockstore/dockstore#2657


denis-yuen:
Actually, it's unclear to me. What happens when the workflow is in a stub and doesn't have a language yet or do we always have a default to CWL in the DB?

garyluu:
> Actually, it's unclear to me. What happens when the workflow is in a stub and doesn't have a language yet or do we always have a default to CWL in the DB?

We always have CWL in the DB it appears

PR #746, opened 2019-08-21
Fix refreshing Google tokens

coverbeck:
dockstore/dockstore#2745

Problem was we weren't getting a refresh token back in the
initial authentication dance.

Looks like a dependabot update of ng2-ui-auth caused this.

Changed AuthConfig in auth.model to have a type (when I changed
it, compilation errors happened), then adjusted the code. Note
the defaultHeaders gave a compilation error as well, so I removed
it.

Regarding, changing the value of prompt to "consent", this page,
https://developers.google.com/identity/protocols/OpenIDConnect#prompt,
says the only allowed values are none, consent, and select_account.
I went with consent, because you may not get a refresh token if
you already requested an access token only before. See
https://github.com/googleapis/google-api-python-client/issues/213.

Had to use TypeScript User Defined Type Guards for the first time in
configuration.service.ts to fix compilation errors.

### DIFF
--- a/src/app/configuration.service.ts
+++ b/src/app/configuration.service.ts
@@ -3,6 +3,7 @@ import { Config, MetadataService } from './shared/swagger';
 import { Dockstore } from './shared/dockstore.model';
 import { ConfigService } from 'ng2-ui-auth';
 import { AuthConfig } from './shared/auth.model';
+import { IOauth1Options, IOauth2Options } from 'ng2-ui-auth/lib/config-interfaces';
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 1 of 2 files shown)


denis-yuen:
> Looks like a dependabot update of ng2-ui-auth caused this.

:(

coverbeck:
> Another option would be to define your own interfaces:
> 
> ```
> export interface IProviders2 {
>   [provider: string]: IOauth2Options;
> }
> 
> export interface IPartialConfigOptions2 extends IPartialConfigOptions {
>   providers?: IProviders2;
> }
> ```

Interesting... Hadn't thought that would be possible, but it is. Will file away for future reference, but since I've tested things the way they are now, I'm going to keep the current approach.

PR #748, opened 2019-08-21
feature/2755/service-versions-view-edit

Ldcabansay:
https://github.com/dockstore/dockstore/issues/2755

Requested changes from #2755
1. Changed card title from 'Edit Version Tag' to 'Edit/View Service Version'
2. Make non-editable fields into normal text (remove textbox)
3. Change 'Workflow Path' to 'Service Path'
4. Remove 'test parameter file(s)' field entirely.
5. Make hidden box editable (and work like it does in entries)
6. Change action link text from 'View' to 'Edit'

Added service_path to form validations so now you can make changes and save them (currently only hidden checkbox)

image

For now just made changes in existing components (largely grouped with workflows), but we should move services related things to their own places in the future.

### DIFF
--- a/src/app/shared/validationMessages.model.ts
+++ b/src/app/shared/validationMessages.model.ts
@@ -29,6 +29,7 @@ export const formErrors = {
   reference: '',
   versionTag: '',
   workflow_path: '',
+  service_path: '',
   workflowName: '',
   amazonDockerRegistryPath: '',
   sevenBridgesDockerRegistryPath: ''
@@ -53,6 +54,7 @@ export const validationDescriptorPatterns = {
   versionTag: '^[a-zA-Z0-9]+([-_.]*[a-zA-Z0-9]+)*$',
   reference: '[\\w-]+((/|.)[\\w-]+)*',
   workflowDescriptorPath: '^/([^\\/?:*|<>]+/)*[^\\/?:*|<>]+.(cwl|wdl|yaml|yml|config)',
+  serviceDescriptorPath: '^.([^\\/?:*|<>]+/)*[^\\/?:*|<>]+.(yml)',
   workflowName: '[a-zA-Z0-9]+([-_][a-zA-Z0-9]+)*',
   cwlTestParameterFilePath: '^/([^/?:*|<>]+/)*[^/?:*|<>]+.(json|yml|yaml)$',
   wdlTestParameterFilePath: '^/([^/?:*|<>]+/)*[^/?:*|<>]+.(json|yml|yaml)$',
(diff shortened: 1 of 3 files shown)


coverbeck:
Based on screenshot alone,  the git reference and service path should not be inputs, even though disabled, unless they will ever be enabled (which I don't think they will). The should be text.

src/app/shared/validationMessages.model.ts
```diff
@@ -142,6 +144,12 @@ export const validationMessages = {
     maxlength: 'Workflow Path is too long. (Max 256 characters.)',
     pattern: `Must begin with '/' and end with '*.cwl', '*.yml', '*.yaml', '*.config', or'*.wdl' ` + 'depending on the descriptor type.'
   },
+  service_path: {
```
coverbeck: I don't think the service_path is ever editable, so I don't think you need this section.

src/app/workflow/version-modal/version-modal.component.html
```diff
@@ -35,17 +35,17 @@ <h4 mat-dialog-title>{{ isPublic || !canWrite ? 'View' : 'Edit' }} Version Tag</
...
-              name="workflow_path"
+              name="{{ (isService$ | async) ? 'service_path' : 'workflow_path' }}"
               [(ngModel)]="version.workflow_path"
               minlength="3"
               maxlength="128"
-              [pattern]="validationPatterns.workflowDescriptorPath"
+              [pattern]="(isService$ | async) ? validationPatterns.serviceDescriptorPath : validationPatterns.workflowDescriptorPath"
               required
               [matTooltip]="tooltip.workflowPath"
               placeholder="e.g. CancerCollaboratory/dockstore-tool-liftover"
```
coverbeck: This comment for line 52, which I can comment on because you didn't change it. Following up on my comment on the screenshot, I think you should just not have the input at all if a service, and display plain text instead. A disabled input implies there's some way of enabling it.

src/app/shared/validationMessages.model.ts
Ldcabansay: changed input to text and so removed all the service path stuff from form validation

src/app/workflow/version-modal/version-modal.component.html
Ldcabansay: Changed the input to text

src/app/workflow/version-modal/version-modal.component.html
```diff
@@ -13,7 +13,7 @@
   ~    See the License for the specific language governing permissions and
   ~    limitations under the License.
   -->
-<h4 mat-dialog-title>{{ isPublic || !canWrite ? 'View' : 'Edit' }} Version Tag</h4>
+<h4 mat-dialog-title>{{ isPublic || !canWrite ? 'View' : 'Edit' }} {{ (isService$ | async) ? 'Service Version' : 'Version Tag' }}</h4>
```
garyluu: Should be "Service Version" and "Workflow Version" in my opinion

denis-yuen: I agree to avoid "Tag", we've been using that term in the context of tools and Docker container "tags" exclusively

Ldcabansay: Changed to "Workflow Version"

PR #749, opened 2019-08-22
add fullscreen overlay so the tooltips for the 2 dag toggles show

NatalieEO:
dockstore/dockstore#1993
I don't think this was being covered by the sidebar anymore? Anyway, fixed the two tooltips not showing in fullscreen mode by adding the overlaycontainer.


NatalieEO:
> Interesting, so you do that and it just works? I guess the old days of modifying CSS directly are going away. :)

Yeah I wasn't successful messing with the z-index, but found this https://github.com/angular/components/issues/1628 after some googling

PR #750, opened 2019-08-22
lazy load first tab as well

agduncan94:
https://github.com/dockstore/dockstore/issues/2354

* Change first tab for files to be lazy loaded

### DIFF
--- a/src/app/container/files/files.component.html
+++ b/src/app/container/files/files.component.html
@@ -15,8 +15,11 @@
   -->
 
 <mat-tab-group mat-stretch-tabs>
+  <!-- Keep this as lazy loaded -->
   <mat-tab label="Dockerfile">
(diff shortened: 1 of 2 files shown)


src/app/container/files/files.component.html
```diff
@@ -15,8 +15,11 @@
   -->
 
 <mat-tab-group mat-stretch-tabs>
+  <!-- Keep this as lazy loaded -->
```
denis-yuen: Why? (in comments)

src/app/workflow/files/files.component.html
```diff
@@ -15,8 +15,11 @@
   -->
 
 <mat-tab-group mat-stretch-tabs>
+  <!-- Keep this as lazy loaded -->
```
denis-yuen: Ditto

src/app/container/files/files.component.html
agduncan94: According to Gary this section has been flip flopping between lazy and eager, so this is just to emphasize in the future not to change it back.

PR #752, opened 2019-08-22
Allow same page anchor links

garyluu:
I don't know why it stopped working, but this fixes it.

There's no ticket for this as far as I know.  But the issue was the "Start Here" button of the home page required clicking twice.  This PR enables anchor links to the same page.

https://angular.io/api/router/ExtraOptions#anchorScrolling


denis-yuen:
What is this for? 
If there is no ticket, please elaborate in the title and PR description

PR #753, opened 2019-08-22
clear sourcefiles from versions to shrink request size

agduncan94:
* Clear sourcefiles for update workflow request to snapshot so that request is smaller (issue was a lot of sourcefiles made the request huge, exceeding 50K limit)

https://github.com/dockstore/dockstore/issues/2762

### DIFF
--- a/src/app/workflow/view/view.component.ts
+++ b/src/app/workflow/view/view.component.ts
@@ -100,6 +100,9 @@ export class ViewWorkflowComponent extends View implements OnInit {
    * @memberof ViewWorkflowComponent
    */
   private updateWorkflowToSnapshot(version): void {
+    // Clear sourcefiles to shrink version
+    version.sourceFiles = [];
+
     this.workflowsService.updateWorkflowVersion(this.workflow.id, [version]).subscribe(
       workflowVersions => {
         this.alertService.detailedSuccess('Snapshot successfully created!');


src/app/workflow/view/view.component.ts
```diff
@@ -100,6 +100,9 @@ export class ViewWorkflowComponent extends View implements OnInit {
    * @memberof ViewWorkflowComponent
    */
   private updateWorkflowToSnapshot(version): void {
+    // Clear sourcefiles to shrink version
+    version.sourceFiles = [];
```
coverbeck: I think you should create a new object; if the operation fails for whatever reason, sourceFiles would be empty and would for example affect the Files tab. Unless this already is a copy... Not sure.

agduncan94: According to Gary, it is a clone. I will do some more testing locally before I merge this.

agduncan94: So this is the value that is passed to the function
`const snapshot: WorkflowVersion = { ...this.version, frozen: true };`

Here is the explanation of the spread operator in this case https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals

Here is the relevant line about the spread

> It copies own enumerable properties from a provided object onto a new object

PR #766, opened 2019-08-28
Add CGC bbutton for CWL workflows

StefanijaO:
Here's our pull request for the CGC button. It's intended for all CWL workflows.


QuicksilverMachine:
Hello, I'm the responsible engineer on this feature, I'll try to answer you questions:

1. It's not possible, but since everything is ready on our end we could walk you through our registration process, and you could try it with a free account for testing purposes.
2. We support only relative file imports for now (files referenced in the run of a step).
3. We currently do not support importing of custom container files, so the problem is on our end. We might revisit this on a later date.

QuicksilverMachine:
1. We've added staging.dockstore.org to our whitelist for imports. It should work now.
2. We've updated the pull request with the http import check and the image url.
3. As I've said we will surely revisit this later. It's an ok start for us to just support workflows for now.

As for the workflow, could you tell us which workflow did you try that didn't succeed? 
We also created a simple one for testing purposes: https://dockstore.org/workflows/dockstore.org/QuicksilverMachine/test_cwl:14

denis-yuen:
Bit of a heads-up, after importing the above. 
https://cgc.sbgenomics.com/u/dyuen/test/apps/#q?archived=false no longer seems to work. Obviously can't be fixed as part of this PR, but thought the heads-up might be useful. 
Linking to specific apps still works

Screenshot from 2019-08-29 11-56-43

QuicksilverMachine:
Thank you for the info, we’ll check it out, it seems to be a problem on our end anyways.

coverbeck:
I think the one I tried, purely coincidentally, was your Quicksilver one; I picked it because it was the first one at the top of my list; this is what I have in my history. That's an older version, so maybe that was the issue.

https://cgc.sbgenomics.com/integration/trs/import/?trs=https://dockstore.org/api/api/ga4gh/v2/tools/%23workflow%2Fdockstore.org%2FQuicksilverMachine%2Ftest_cwl/versions/2

QuicksilverMachine:
@coverbeck Yes, that version didn't work, only 14 works right.

QuicksilverMachine:
@coverbeck Also, we fixed the image url in our second commit, could you remove the change request?

PR #771, opened 2019-08-30
Feature/2813/tooltip workflow mode

garyluu:
For dockstore/dockstore#2813

Fixes the tooltip for the build mode.  Removed the bolding because tooltips are supposed to be plain.

Side note: Material recommends short tooltips


garyluu:
> Screenshot might be helpful, will the breaks disappear too?

image

Taking this screenshot took longer than fixing the bug.

PR #772, opened 2019-08-31
Detect http(s) imports in CWL

coverbeck:
dockstore/dockstore#2819

Use regexes to test for CWL http(s) imports, and disable
the CGC button if any are detected.

This is an admittedly crude test, but until dockstore/dockstore#2834,
it should do.

I couldn't find any examples of CWL http imports/mixins/includes, so I deduced them from non-https examples, hopefully correctly.

### DIFF
--- a/src/app/shared/file.service.spec.ts
+++ b/src/app/shared/file.service.spec.ts
@@ -20,10 +20,12 @@ import {
   cwlSourceFileWithIncludeImport,
   cwlSourceFileWithMixinImport,
   cwlSourceFileWithNoImport,
+  cwlSourceFileWithSomeHttpLinks,
   emptyWdlSourceFile,
   sampleSourceFile,
   sampleTag,
   wdlSourceFile,
+  wdlSourceFileWithCommentedHttpImport,
   wdlSourceFileWithHttpImport
 } from '../test/mocked-objects';
 import { ga4ghPath } from './constants';
@@ -62,6 +64,7 @@ describe('FileService', () => {
 
   it('should detect http import in WDL', inject([FileService], (fileService: FileService) => {
     expect(fileService.hasHttpImport(wdlSourceFileWithHttpImport)).toBeTruthy();
+    expect(fileService.hasHttpImport(wdlSourceFileWithCommentedHttpImport)).toBeFalsy();
     expect(fileService.hasHttpImport(emptyWdlSourceFile)).toBeFalsy();
     expect(fileService.hasHttpImport(wdlSourceFile)).toBeFalsy();
   }));
@@ -72,5 +75,6 @@ describe('FileService', () => {
     expect(fileService.hasHttpImport(cwlSourceFileWithMixinImport)).toBeTruthy();
     expect(fileService.hasHttpImport(cwlSourceFileWithCommentedMixinImport)).toBeFalsy();
     expect(fileService.hasHttpImport(cwlSourceFileWithIncludeImport)).toBeTruthy();
+    expect(fileService.hasHttpImport(cwlSourceFileWithSomeHttpLinks)).toBeFalsy();
   }));
 });
--- a/src/app/shared/file.service.spec.ts
+++ b/src/app/shared/file.service.spec.ts
@@ -14,7 +14,18 @@
  *    limitations under the License.
  */
 import { inject, TestBed } from '@angular/core/testing';
-import { sampleSourceFile, sampleTag } from '../test/mocked-objects';
+import {
+  cwlSourceFileWithCommentedMixinImport,
+  cwlSourceFileWithHttpsImport,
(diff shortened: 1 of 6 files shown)


src/app/workflow/launch-third-party/launch-third-party.component.html
```diff
@@ -53,19 +53,22 @@ <h3>Launch with</h3>
...
           <div fxFlex="100%" *ngIf="workflow?.descriptorType === 'CWL'">
-            <div *ngIf="hasHttpImports$ | async" class="import-warning">
-              Warning: this version of the CWL has http imports, which are not supported by the CGC. Make sure to select a version without http
-              imports in the CGC.
-            </div>
             <a
               mat-raised-button
-              [matTooltip]="!(hasContent$ | async) ? 'The CWL has no content.' : ''"
+              [matTooltip]="
+                !(hasContent$ | async)
```
garyluu: just make a tooltip an observable

src/app/shared/file.service.spec.ts
```diff
@@ -48,4 +59,18 @@ describe('FileService', () => {
         '/api/ga4gh/v2/tools/%23workflow%2Fgithub.com%2FHumanCellAtlas%2Fskylab%2FHCA_SmartSeq2/versions/dockstore/PLAIN-WDL/descriptor/HISAT2.wdl'
     );
   }));
+
+  it('should detect http import in WDL', inject([FileService], (fileService: FileService) => {
```
denis-yuen: Think we could use some negative tests for something like this. 
For example, if people added hyperlinks in the middle of  CWL `doc:` fields or CWL+WDL `#` comments  (watch out for https://github.com/broadinstitute/cromwell/issues/2870),  and WDL descriptions 
https://docs.dockstore.org/docs/publisher-tutorials/wdl-best-practices/

src/app/shared/file.service.ts
```diff
@@ -20,6 +20,9 @@ import { DescriptorTypeCompatService } from './descriptor-type-compat.service';
 import { Dockstore } from './dockstore.model';
 import { SourceFile, Tag, ToolDescriptor, WorkflowVersion } from './swagger';
 
+const wdlImportHttpRegEx: RegExp = new RegExp(/^\s*import\s+"?https?/, 'm');
```
denis-yuen: I think you're safe for the above, but just in case.

src/app/workflow/launch-third-party/launch-third-party.component.html
```diff
@@ -53,19 +53,22 @@ <h3>Launch with</h3>
...
+                  ? 'The CWL has no content.'
+                  : (hasHttpImports$ | async)
+                  ? 'This version of the CWL has http imports, which are not supported by the CGC. Select a version without http imports.'
```
denis-yuen: `http(s)` to be perfectly clear

coverbeck: Fixed in upcoming push

src/app/shared/file.service.spec.ts
coverbeck: I'm not handling the multiline doc case and I don't think I easily can with a simple regex. I think I should go back to a warning message instead of disabling the button, just in case there is a false positive.

denis-yuen: I meant that a simple link (say to an image or to a lab page) in one of the above situations shouldn't trigger your regex, but you should add a test to verify. 

I don't think you need to worry about the case that someone literally writes `import http://foo.com` in a comment, multi-line or now if that's what you meant.

coverbeck:
Notes for second review:

1. Added some falsey tests
2. Made tooltip an observable.
    * If there's a more elegant way to handle the map/mergemap combo, let me know
    * Couldn't use `this.hasContent$` because of the `share`, in case you're wondering why I used that instead of `this.descriptorsQuery.hasContent$`. It would not get invoked.... Did not figure out exactly what is going on.

src/app/workflow/launch-third-party/launch-third-party.component.ts
```diff
@@ -146,6 +146,22 @@ export class LaunchThirdPartyComponent extends Base implements OnChanges, OnInit
    */
   workflowPathAsQueryValue: string;
 
+  cgcTooltip$: Observable<string> = this.descriptorsQuery.hasContent$.pipe(
```
garyluu: (Combinelatest or zip) + pipe + map seems easier.  What was the benefit to doing it this way (2 maps, 2 pipes, and mergemap?

coverbeck: I'm used to using just map and flatmap (mergemap in TS) in Java/Scala, and don't know all the RXJS operators very well... Will update

PR #779, opened 2019-09-04
Turn back on tests for snapshotting

david4096:
This PR is meant to bring the snapshot/doi UI features into a working state by changing the logic of refresh and turning the tests back on.

Address SEAB-430

### DIFF
--- a/cypress/integration/immutableDatabaseTests/workflowDetails.ts
+++ b/cypress/integration/immutableDatabaseTests/workflowDetails.ts
@@ -72,8 +72,7 @@ describe('Dockstore Workflow Details', () => {
     cy.get('[data-cy=dockstore-snapshot-unlocked]')
       .its('length')
       .should('be.gt', 0);
-        cy.get('[data-cy=dockstore-snapshot]')
-          .should('not.be.visible');
+
    });
 
   describe('Change tab to files', () => {
--- a/cypress/integration/immutableDatabaseTests/workflowDetails.ts
+++ b/cypress/integration/immutableDatabaseTests/workflowDetails.ts
@@ -67,8 +67,8 @@ describe('Dockstore Workflow Details', () => {
       .get('tbody>tr')
       .should('have.length', 1); // 1 Version and no warning line
     cy.url().should('eq', Cypress.config().baseUrl + '/workflows/github.com/A/l:master?tab=versions');
-    // Buttons to create snapshots are visible
-    cy.get('[data-cy=dockstore-snapshot]').its('length').should('be.gt', 0);
+    // Buttons to create snapshots are hidden on public
+    cy.get('[data-cy=dockstore-snapshot]').should('not.be.visible');
     cy.get('[data-cy=dockstore-snapshot-locked]').should('have.length', 0);
 
     cy.get('[data-cy=dockstore-snapshot-unlocked]')
--- a/cypress/integration/immutableDatabaseTests/workflowDetails.ts
+++ b/cypress/integration/immutableDatabaseTests/workflowDetails.ts
@@ -61,12 +61,14 @@ describe('Dockstore Workflow Details', () => {
     cy.url().should('eq', Cypress.config().baseUrl + '/workflows/github.com/A/l:master?tab=launch');
   });
 
-  it('Change tab to versions and not see snapshot', () => {
+  it('Change tab to versions and not see snapshotted version', () => {
     goToTab('Versions');
     cy
       .get('tbody>tr')
       .should('have.length', 1); // 1 Version and no warning line
     cy.url().should('eq', Cypress.config().baseUrl + '/workflows/github.com/A/l:master?tab=versions');
+    // Buttons to create snapshots are visible
+    cy.get('[data-cy=dockstore-snapshot]').its('length').should('be.gt', 0);
     cy.get('[data-cy=dockstore-snapshot-locked]').should('have.length', 0);
 
     cy.get('[data-cy=dockstore-snapshot-unlocked]')
--- a/src/app/workflow/view/view.service.ts
+++ b/src/app/workflow/view/view.service.ts
@@ -52,13 +52,14 @@ export class ViewService {
   private updateWorkflowToSnapshot(workflow: Workflow, version: WorkflowVersion, cb: Function): void {
     const snapshot: WorkflowVersion = { ...version, frozen: true };
     this.workflowsService.updateWorkflowVersion(workflow.id, [snapshot]).subscribe(
-      workflowVersions => {
+      (workflowVersions: Array<WorkflowVersion>) => {
         const selectedWorkflow = { ...this.workflowQuery.getActive() };
-        selectedWorkflow.workflowVersions = workflowVersions;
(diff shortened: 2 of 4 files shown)


src/app/workflow/view/view.service.ts
```diff
@@ -53,10 +53,10 @@ export class ViewService {
     const snapshot: WorkflowVersion = { ...version, frozen: true };
     this.workflowsService.updateWorkflowVersion(workflow.id, [snapshot]).subscribe(
       workflowVersions => {
+        const selectedWorkflow = { ...this.workflowQuery.getActive() };
```
coverbeck: 1. Since line 54 is asynchronous, is there the risk that `this.workflowQuery.getActive()` returns a different workflow than the one represented by `workflow.id` on line 54?
2. Why can't you just use the `workflow` parameter from line 52?

Just asking; not sure what the answers are....

Also, could you explicitly specify the type of `selectedWorkflow`?

src/app/workflow/view/view.service.ts
```diff
@@ -130,8 +130,9 @@ export class ViewService {
    * @memberof ViewService
    */
   private requestDOISuccess(version: WorkflowVersion, workflowVersions: Array<WorkflowVersion>): void {
-    const newSelectedVersion = workflowVersions.find(v => v.id === version.id);
-    this.workflowService.setWorkflowVersion(newSelectedVersion);
+    const workflow = { ...this.workflowQuery.getActive() };
```
coverbeck: Same questions/comment as on line 56.

src/app/workflow/view/view.service.ts
david4096: 1. I think you're describing a situation where someone navigates away from this workflow while the snapshot is in progress. In that case the workflowVersions might get set from a different version and display some mangled data. 
2. If we use workflow from 52 then if you snapshot, browse away, it will cause the old workflow to reappear.

I think its a good catch, though rare since snapshotting should be fast. I think the expectation would be to silently succeed when someone browses away. This could be done by guaranteeing the returned workflow id matches the one gotten via querying the store, if I understand correctly.

src/app/workflow/view/view.service.ts
david4096: Yeah, almost same answer except this one can take longer to eventually fire and so could lead to problems more often.

src/app/workflow/view/view.service.ts
coverbeck: >     1. I think you're describing a situation where someone navigates away from this workflow while the snapshot is in progress. In that case the workflowVersions might get set from a different version and display some mangled data.

 Yes, I was guessing at something like that, even though in practice it's a stretch to happen.

>     2. If we use workflow from 52 then if you snapshot, browse away, it will cause the old workflow to reappear.

Even if you clone it, set its workflowVersions, and call this.workflowService.setWorkflow(<cloned workflow>)?  Basically changing line 56 to `const selectedWorkflow = {...workflow}`? 

If you do that, then the first point becomes moot... Assuming it's possible.

cypress/integration/immutableDatabaseTests/workflowDetails.ts
```diff
@@ -72,8 +72,7 @@ describe('Dockstore Workflow Details', () => {
     cy.get('[data-cy=dockstore-snapshot-unlocked]')
       .its('length')
       .should('be.gt', 0);
-        cy.get('[data-cy=dockstore-snapshot]')
-          .should('not.be.visible');
```
garyluu: :eyes:

denis-yuen: Elaborate

garyluu: The name of the test is blah blah blah "not see snapshot".
The not see snapshot part is commented out.

src/app/workflow/view/view.service.ts
```diff
@@ -153,9 +156,12 @@ export class ViewService {
       confirmationButtonText: 'Snapshot Version',
       cancelButtonText: 'Cancel'
     };
-    this.confirmationDialogService.openDialog(dialogData, bootstrap4mediumModalSize).subscribe(confirmationResult => {
+    this.confirmationDialogService.openDialog(dialogData, bootstrap4mediumModalSize).subscribe((confirmationResult: Boolean) => {
       if (confirmationResult) {
-        this.updateWorkflowToSnapshot(workflow, version, () => this.alertService.detailedSuccess('Snapshot successfully created!'));
+        this.updateWorkflowToSnapshot(workflow, version, () => {
+          this.alertService.detailedSuccess(`A Snapshot was successfully requested for workflow
```
coverbeck: This is nit-picky, but:

1. I would not have a capital S in snapshot
2. I would say created or issued instead of requested -- "successfully requested" means, to me, that the request was accepted, but not necessarily that it was fulfilled.
3. I'd probably remove "successfully".

src/app/workflow/view/view.service.ts
```diff
@@ -129,10 +129,13 @@ export class ViewService {
...
-  private requestDOISuccess(version: WorkflowVersion, workflowVersions: Array<WorkflowVersion>): void {
-    const newSelectedVersion = workflowVersions.find(v => v.id === version.id);
-    this.workflowService.setWorkflowVersion(newSelectedVersion);
-    this.alertService.simpleSuccess();
+  private requestDOISuccess(workflow: Workflow, version: WorkflowVersion): void {
+    const selectedWorkflow = { ...this.workflowQuery.getActive() };
+    if (selectedWorkflow.id === workflow.id) {
+      this.workflowService.setWorkflow(workflow);
+    }
+    this.alertService.detailedSuccess(`A Digital Object Identifier (DOI) was successfully requested for workflow
```
coverbeck: See other comment on "successfully requested".

src/app/workflow/view/view.service.ts
```diff
@@ -52,13 +48,16 @@ export class ViewService {
   private updateWorkflowToSnapshot(workflow: Workflow, version: WorkflowVersion, cb: Function): void {
     const snapshot: WorkflowVersion = { ...version, frozen: true };
     this.workflowsService.updateWorkflowVersion(workflow.id, [snapshot]).subscribe(
-      workflowVersions => {
+      (workflowVersions: Array<WorkflowVersion>) => {
+        this.alertService.detailedSuccess(`A Snapshot was successfully requested for workflow
```
coverbeck: See comment on "successfully requested".

src/app/workflow/view/view.service.ts
```diff
@@ -49,7 +49,7 @@ export class ViewService {
     const snapshot: WorkflowVersion = { ...version, frozen: true };
     this.workflowsService.updateWorkflowVersion(workflow.id, [snapshot]).subscribe(
       (workflowVersions: Array<WorkflowVersion>) => {
-        this.alertService.detailedSuccess(`A Snapshot was successfully requested for workflow
+        this.alertService.detailedSuccess(`A Snapshot was created for workflow
```
coverbeck: snapshot not Snapshot

PR #780, opened 2019-09-05
Set sourceFiles to empty array on workflowVersions endpoint

coverbeck:
dockstore/dockstore#2833
dockstore/dockstore#2836


src/app/interceptors/workflow-versions.interceptor.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * An interceptor that ensures that PUT requests on the workflowVersions
+ * endpoint have the sourceFiles property set to an empty array.
+ *
+ * The sourceFiles property is ignored by the web service endpoint, but can be
+ * sizeable. This affects performance and more importantly, can run into a max
+ * request limit size that we have on our nginx server.
+ *
+ * See dockstore/dockstore#2833
```
garyluu: The full URL might be more useful

coverbeck: Updated.

PR #781, opened 2019-09-05
Use cloudflare

garyluu:
For dockstore/dockstore#2759

Staging is directing to https://staging.dockstore.org/worker-json.js instead of https://gui.dockstore.org/2.4.0-rc.0-4930df6/worker-json.js  .  Too much of a pain to direct it to the right location.  Using the publicly available cloudflare CDN version of it instead.  

Our build is smaller and I actually trust the cloudflare one more.

Our ace version is 1.4.4, so using 1.4.4.  If we update ace-editor, we will need to update the URL too sadly.

### DIFF
--- a/src/app/shared/code-editor/code-editor.component.ts
+++ b/src/app/shared/code-editor/code-editor.component.ts
@@ -47,6 +47,7 @@ export class CodeEditorComponent implements AfterViewInit {
 
   ngAfterViewInit() {
     const aceMode = 'ace/mode/' + this.mode;
+    ace.config.set('workerPath', 'https://cdnjs.cloudflare.com/ajax/libs/ace/1.4.4/');
     this.editor = ace.edit(this.aceId, {
       mode: aceMode,
       readOnly: this.readOnly,
--- a/angular.json
+++ b/angular.json
@@ -19,12 +19,7 @@
             "assets": [
               "src/assets",
               "src/favicon.ico",
-              "src/manifest.json",
-              {
-                "glob": "worker-json.js",
-                "input": "./node_modules/ace-builds/src-min-noconflict",
-                "output": "/"
-              }
+              "src/manifest.json"
             ],
             "styles": [
               "node_modules/bootstrap/dist/css/bootstrap.css",


denis-yuen:
Any idea what it does/wasn't doing when it was missing?

garyluu:
> Any idea what it does/wasn't doing when it was missing?

No idea, it seems to work completely fine without it

garyluu:
Also RIP S3 access key

coverbeck:
> Our ace version is 1.4.4, so using 1.4.4. If we update ace-editor, we will need to update the URL too sadly.

Yikes, I wish there were a way to put a comment in the JSON next to the ace dependency to remind for the next update of the editor, but there's not. :(

src/app/shared/code-editor/code-editor.component.ts
```diff
@@ -47,6 +47,7 @@ export class CodeEditorComponent implements AfterViewInit {
 
   ngAfterViewInit() {
     const aceMode = 'ace/mode/' + this.mode;
+    ace.config.set('workerPath', 'https://cdnjs.cloudflare.com/ajax/libs/ace/1.4.4/');
```
coverbeck: As an alternative, you could make this a relative path so it works on dev builds, then change .circleci/config.yml to update this for gui.dockstore.org. Or was this already the approach you tried when you said it was too much of a pain?

Advantage: Less risk of an update to Ace breaking if also we forget to update this line.
Disadvantage: Makes code between prod/local a little more different than it is already and I'm wild about changing the source code like are doing (and we may move away from that if we go full bore CloudFront).

garyluu: I find the sed to be a lot more fragile than this because we'd change file names, change how this command looks, etc. Different versions of the worker-json.js actually work, so I'm not too worried about it even if we do update ace at some point.

garyluu:
> Is it possible to write a test to see when the workerPath breaks?

This error only happens on staging.dockstore.org because of the CloudFront setup. I can possibly make a test, but it wouldn't really tell me if it's broken on CloudFront or not.

denis-yuen:
Oh, I mean the proposed one.
Like if 1.4.4 in  `https://cdnjs.cloudflare.com/ajax/libs/ace/1.4.4/worker-json.js`  doesn't match the dependency version

PR #782, opened 2019-09-05
Features/2828/tools project links

ByteMap:
Displayed Quay.io link and other links

dockstore/dockstore#2828

Set up the ImageProviderService for search.service.ts allowing the
tools data to merge with the images (quay.io and other links) before
pushing the data into the Akita 'store' state.

Modified the filterEntry method in search.service.ts to return a tuple
in order to unit test without side effects.

### DIFF
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -384,7 +384,7 @@ export class SearchComponent implements OnInit, OnDestroy {
     })
       .then(hits => {
         this.hits = hits.hits.hits;
-        const filteredHits: [Array<any>, Array<any>] = this.searchService.filterEntry(this.hits, this.query_size);
+        const filteredHits: [Array<Hit>, Array<Hit>] = this.searchService.filterEntry(this.hits, this.query_size);
         this.searchService.setHits(filteredHits[0], filteredHits[1]);
         if (this.values.length > 0 && hits) {
           this.searchTerm = true;
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -27,6 +27,7 @@ import { ELASTIC_SEARCH_CLIENT } from './elastic-search-client';
 import { QueryBuilderService } from './query-builder.service';
 import { SearchQuery } from './state/search.query';
 import { SearchService } from './state/search.service';
+import { Hit } from './state/search.service';
 
 /**
  * There are a total of 5 calls per search.
@@ -61,7 +62,7 @@ export class SearchComponent implements OnInit, OnDestroy {
    */
   /*TODO: Bad coding...change this up later (init)..*/
   private setFilter = false;
-  public hits: Object[];
+  public hits: Hit[];
 
   // Possibly 100 workflows and 100 tools (extra +1 is used to see if there are > 200 results)
   public readonly query_size = 201;
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -383,7 +383,8 @@ export class SearchComponent implements OnInit, OnDestroy {
     })
       .then(hits => {
         this.hits = hits.hits.hits;
-        this.searchService.filterEntry(this.hits, this.query_size);
+        const filteredHits: [Array<any>, Array<any>] = this.searchService.filterEntry(this.hits, this.query_size);
+        this.searchService.setHits(filteredHits[0], filteredHits[1]);
         if (this.values.length > 0 && hits) {
           this.searchTerm = true;
         }
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -26,8 +26,7 @@ import { AdvancedSearchService } from './advancedsearch/advanced-search.service'
 import { ELASTIC_SEARCH_CLIENT } from './elastic-search-client';
 import { QueryBuilderService } from './query-builder.service';
 import { SearchQuery } from './state/search.query';
-import { SearchService } from './state/search.service';
-import { Hit } from './state/search.service';
+import { Hit, SearchService } from './state/search.service';
 
 /**
  * There are a total of 5 calls per search.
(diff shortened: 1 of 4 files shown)


garyluu:
Waiting for test pass

src/app/test/mocked-objects.ts
```diff
@@ -456,3 +456,87 @@ export const testSourceFiles: Array<SourceFile> = [
     }
   }
 ];
+
+export const elasticSearchResponse = [
```
coverbeck: Even though I worked with you on this and suggested this, so it's my fault, but I think your mock response should also contain a workflow

src/app/search/state/search.service.spec.ts
```diff
@@ -49,4 +51,14 @@ describe('SearchService', () => {
   it('should not crash on null advancedSearchObject', inject([SearchService], (service: SearchService) => {
     expect(service.hasSearchText(null, null, null)).toEqual(false);
   }));
+
+  it('should create image provider', inject([SearchService], (service: SearchService) => {
+    const filtered: [Array<any>, Array<any>] = service.filterEntry(elasticSearchResponse, 201);
+    const tools = filtered[0];
+    expect(filtered[1].length).toBe(0);
```
coverbeck: In conjunction with my other comment, should test that the workflow shows up here, and perhaps make sure it doesn't have imgProvider and imgProviderUrl

garyluu:
Unrelated but interesting, the hit['source'] isn't actually a tool or workflow as certain fields are missing

src/app/search/state/search.service.ts
```diff
@@ -108,11 +110,11 @@ export class SearchService {
    * @param {number} query_size
    * @memberof SearchService
    */
-  filterEntry(hits: Array<any>, query_size: number) {
+  filterEntry(hits: Array<any>, query_size: number): [Array<any>, Array<any>] {
```
garyluu: This is of type:
```
Array<{
            _index: string;
            _type: string;
            _id: string;
            _score: number;
            _source: T;
            _version?: number;
            _explanation?: Explanation;
            fields?: any;
            highlight?: any;
            inner_hits?: any;
            sort?: string[];
        }>
```
Might want to define this somewhere

src/app/search/state/search.service.ts
```diff
@@ -108,11 +110,11 @@ export class SearchService {
...
     const workflowHits = [];
     const toolHits = [];
     hits.forEach(hit => {
-      hit['_source'] = this.providerService.setUpProvider(hit['_source']);
+      hit['_source'] = this.providerService.setUpProvider(this.imageProviderService.setUpImageProvider(hit['_source']));
```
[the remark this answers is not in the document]
ByteMap: Moved the imageProviderService down to the if statement where it is determining if the hit is a tool.

src/app/search/state/search.service.ts
ByteMap: Defined an interface Hit and replaced Array<Any> with Array<Hit> in the filterEntry method and all associated methods.

src/app/test/mocked-objects.ts
ByteMap: Included a workflow in the mocked-objects

src/app/search/state/search.service.spec.ts
ByteMap: Included a test for the workflows to insure that it does not have imageProvider or imageProviderURL

src/app/search/search.component.ts
```diff
@@ -383,7 +384,8 @@ export class SearchComponent implements OnInit, OnDestroy {
     })
       .then(hits => {
         this.hits = hits.hits.hits;
-        this.searchService.filterEntry(this.hits, this.query_size);
+        const filteredHits: [Array<any>, Array<any>] = this.searchService.filterEntry(this.hits, this.query_size);
```
coverbeck: Should be `Array<Hit>, Array<Hit>`

src/app/search/state/search.service.spec.ts
```diff
@@ -49,4 +59,16 @@ describe('SearchService', () => {
   it('should not crash on null advancedSearchObject', inject([SearchService], (service: SearchService) => {
     expect(service.hasSearchText(null, null, null)).toEqual(false);
   }));
+
+  it('should create image provider', inject([SearchService], (service: SearchService) => {
+    const filtered: [Array<any>, Array<any>] = service.filterEntry(elasticSearchResponse, 201);
```
coverbeck: Array<Hit>, Array<Hit>

src/app/search/state/search.service.ts
```diff
@@ -108,20 +125,21 @@ export class SearchService {
...
         } else if (hit['_type'] === 'workflow') {
           workflowHits.push(hit);
         }
       }
     });
-    this.setHits(toolHits, workflowHits);
+    return [toolHits, workflowHits];
   }
 
   setHits(toolHit: any, workflowHit: any) {
```
coverbeck: Change this to take `Array<Hit>, Array<Hit>` as params.

src/app/search/search.component.ts
ByteMap: Done.

src/app/search/state/search.service.spec.ts
ByteMap: Done.

src/app/search/state/search.service.ts
ByteMap: Changed from (toolHit: any, workflowHit:any) to (toolHit: Array<Hit>, workflowHit: Array<Hit>)

src/app/search/search.component.ts
```diff
@@ -27,6 +27,7 @@ import { ELASTIC_SEARCH_CLIENT } from './elastic-search-client';
 import { QueryBuilderService } from './query-builder.service';
 import { SearchQuery } from './state/search.query';
 import { SearchService } from './state/search.service';
+import { Hit } from './state/search.service';
```
coverbeck: This should be part of the previous line: `import { Hit, SearchService } from './state/search.service`.

If you let WebStorm import instead of manually adding, this will typically just work.

ByteMap: Done!

PR #786, opened 2019-09-10
updated to google recommendation

emlys:
* removed json-ld from index.html because as stated here https://www.searchenginejournal.com/google-do-not-put-organization-schema-markup-on-every-page/289981/#close and here https://developers.google.com/search/docs/data-types/sitelinks-searchbox it should not be on every page, just the homepage
* removed twitter links because as stated here https://developers.google.com/search/docs/data-types/social-profile including social media links is deprecated
* the search bar part appears to be set up right. it's up to google to decide to display it or not.
* added a few more fields from the schema.org Website and Organization categories


coverbeck:
Your commit message should reference the issue. You can either amend the commit or make sure you update the message when you merge.

emlys:
> Your commit message should reference the issue. You can either amend the commit or make sure you update the message when you merge.

will update it in the squashed message

PR #787, opened 2019-09-12
Check to see if zenodo account is linked

david4096:
Opening as draft. Currently you will be redirected back to the workflow but not the versions tab itself after linking your zenodo account. I think it's probably ok for now.

### DIFF
--- a/src/app/workflow/view/view.service.ts
+++ b/src/app/workflow/view/view.service.ts
@@ -27,10 +27,13 @@ import { ConfirmationDialogData } from '../../confirmation-dialog/confirmation-d
 import { AccountsService } from '../../loginComponents/accounts/external/accounts.service';
 import { TokenQuery } from '../../shared/state/token.query';
 import { bootstrap4mediumModalSize } from '../../shared/constants';
+import { takeUntil } from 'rxjs/operators';
 
 @Injectable()
 export class ViewService {
   version: Subject<WorkflowVersion> = new BehaviorSubject<WorkflowVersion>(null);
+  protected ngUnsubscribe: Subject<{}> = new Subject();
+
   constructor(
     private alertService: AlertService,
     private accountsService: AccountsService,
@@ -40,6 +43,11 @@ export class ViewService {
     private workflowService: WorkflowService,
     private workflowsService: WorkflowsService
   ) {}
+
+  ngOnDestroy() {
+    this.ngUnsubscribe.next();
+    this.ngUnsubscribe.complete();
+  }
 
   /**
    * Updates the workflow version and alerts the Dockstore User with success
@@ -212,7 +220,7 @@ export class ViewService {
    */
   requestDOIForWorkflowVersion(workflow: Workflow, version: WorkflowVersion): void {
     // Set the dialog to open a snapshot if its not frozen
-    this.tokenQuery.hasZenodoToken$.subscribe((hasZenodoToken: boolean) => {
+    this.tokenQuery.hasZenodoToken$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((hasZenodoToken: boolean) => {
       if (hasZenodoToken) {
         if (!version.frozen) {
           this.showSnapshotBeforeDOIDialog(workflow, version);
--- a/src/app/workflow/view/view.service.ts
+++ b/src/app/workflow/view/view.service.ts
@@ -88,13 +88,16 @@ export class ViewService {
       cancelButtonText: 'Cancel'
     };
 
-    this.confirmationDialogService.openDialog(dialogData, bootstrap4mediumModalSize).subscribe(confirmationResult => {
-      if (confirmationResult) {
-        this.accountsService.link(TokenSource.ZENODO);
-      } else {
-        this.alertService.detailedSuccess('You cancelled DOI creation.');
-      }
-    });
+    this.confirmationDialogService
+      .openDialog(dialogData, bootstrap4mediumModalSize)
+      .pipe(first())
+      .subscribe(confirmationResult => {
+        if (confirmationResult) {
+          this.accountsService.link(TokenSource.ZENODO);
+        } else {
+          this.alertService.detailedSuccess('You cancelled DOI creation.');
+        }
+      });
   }
 
   /**
@@ -113,13 +116,16 @@ export class ViewService {
       cancelButtonText: 'Cancel'
     };
 
-    this.confirmationDialogService.openDialog(dialogData, bootstrap4mediumModalSize).subscribe(confirmationResult => {
-      if (confirmationResult) {
-        this.updateWorkflowToSnapshot(workflow, version, () => this.showRequestDOIDialog(workflow, version));
-      } else {
-        this.alertService.detailedSuccess('You cancelled DOI creation.');
-      }
-    });
+    this.confirmationDialogService
+      .openDialog(dialogData, bootstrap4mediumModalSize)
+      .pipe(first())
+      .subscribe(confirmationResult => {
+        if (confirmationResult) {
+          this.updateWorkflowToSnapshot(workflow, version, () => this.showRequestDOIDialog(workflow, version));
+        } else {
+          this.alertService.detailedSuccess('You cancelled DOI creation.');
+        }
+      });
   }
 
   /**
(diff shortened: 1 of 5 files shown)


denis-yuen:
FYI, changing base branch @david4096

david4096:
Thanks @denis-yuen this is now open for review!

src/app/workflow/view/view.service.ts
```diff
@@ -176,11 +212,16 @@ export class ViewService {
    */
   requestDOIForWorkflowVersion(workflow: Workflow, version: WorkflowVersion): void {
     // Set the dialog to open a snapshot if its not frozen
-    // TODO check to see if they have a zenodo token
-    if (!version.frozen) {
-      this.showSnapshotBeforeDOIDialog(workflow, version);
-    } else {
-      this.showRequestDOIDialog(workflow, version);
-    }
+    this.tokenQuery.hasZenodoToken$.subscribe((hasZenodoToken: boolean) => {
```
coverbeck: You need to either unsubscribe or probably better yet do something like `pipe(first())` or `pipe(take(1)`.

I'm also a little unclear if each new subscription gets the current value, although it probably does... Did you try for more than one version on the same page?

src/app/workflow/view/view.service.ts
```diff
@@ -134,8 +166,11 @@ export class ViewService {
     if (selectedWorkflow.id === workflow.id) {
       this.workflowService.setWorkflow(workflow);
     }
+
+    const workflowName: string = workflow.workflowName == null ? workflow.repository : workflow.workflowName;
```
coverbeck: Others may not agree, but my personal preference is to write this as `const workflowName: string = workflowName.workflowName || workflow.repository`. Feel free to ignore.

src/app/workflow/view/view.service.ts
```diff
@@ -67,6 +70,31 @@ export class ViewService {
...
+                Zenodo to make it easy to request Digital Object Identifiers (DOIs) for your workflow and only takes a moment.
+                Would you like to link a Zenodo account now?`,
+      title: 'Request DOI (Link Zenodo Account)',
+      confirmationButtonText: 'Link Zenodo Account',
+      cancelButtonText: 'Cancel'
+    };
+
+    this.confirmationDialogService.openDialog(dialogData, bootstrap4mediumModalSize).subscribe(confirmationResult => {
+      if (confirmationResult) {
+        this.accountsService.link('zenodo.org');
```
coverbeck: There is an enum for this.

src/app/workflow/view/view.service.ts
garyluu: unsubscribe

src/app/workflow/view/view.service.ts
```diff
@@ -67,6 +70,31 @@ export class ViewService {
     );
   }
 
+  /**
+   * Opens a confirmation dialog that the asks the Dockstore User if they
+   * would like to associate their Zenodo account for requesting DOIs
+   *
+   * @memberof ViewService
+   */
+  private showLinkZenodoDialog(workflow: Workflow, version: WorkflowVersion): void {
```
coverbeck: You're not using either of these parameters, so you should remove them.

src/app/workflow/view/view.service.ts
david4096: I prefer yours.

garyluu: Just a reminder that it's not the same.  Although in this case, the coalesce (falsey check) probably works better.

src/app/workflow/view/view.service.ts
```diff
@@ -67,6 +72,31 @@ export class ViewService {
...
+    const dialogData: ConfirmationDialogData = {
+      message: `It looks like you have not yet linked your Zenodo and Dockstore accounts. Dockstore integrates with
+                Zenodo to make it easy to request Digital Object Identifiers (DOIs) for your workflow and only takes a moment.
+                Would you like to link a Zenodo account now?`,
+      title: 'Request DOI (Link Zenodo Account)',
+      confirmationButtonText: 'Link Zenodo Account',
+      cancelButtonText: 'Cancel'
+    };
+
+    this.confirmationDialogService.openDialog(dialogData, bootstrap4mediumModalSize).subscribe(confirmationResult => {
```
garyluu: ditto

src/app/workflow/view/view.service.ts
```diff
@@ -150,17 +187,18 @@ export class ViewService {
...
     const dialogData: ConfirmationDialogData = {
-      message: `Snapshotting a version will make it so it can no longer be edited and cannot be undone. <p>Are
-                you sure you would like to snapshot version <b>${version.name}</b>?`,
+      message: `Snapshotting a version will make it so it <b>can no longer be edited and cannot be undone</b>. <p>Are
+                you sure you would like to snapshot version <b>${version.name}</b> <p><b>Warning: This CANNOT be undone!</b></p>?`,
       title: 'Snapshot',
       confirmationButtonText: 'Snapshot Version',
       cancelButtonText: 'Cancel'
     };
     this.confirmationDialogService.openDialog(dialogData, bootstrap4mediumModalSize).subscribe((confirmationResult: boolean) => {
```
garyluu: here too while you're at it

src/app/workflow/view/view.service.ts
```diff
@@ -134,8 +168,11 @@ export class ViewService {
     if (selectedWorkflow.id === workflow.id) {
       this.workflowService.setWorkflow(workflow);
     }
+
+    const workflowName: string = workflow.workflowName || workflow.repository;
```
agduncan94: I'm wondering if it should be either repository + '/' + workflowName or repository depending on if the name is present. Not sure if that takes up too much space in the confirm message.

src/app/workflow/view/view.service.ts
```diff
@@ -150,17 +187,18 @@ export class ViewService {
...
+      message: `Snapshotting a version will make it so it <b>can no longer be edited and cannot be undone</b>. <p>Are
+                you sure you would like to snapshot version <b>${version.name}</b> <p><b>Warning: This CANNOT be undone!</b></p>?`,
       title: 'Snapshot',
       confirmationButtonText: 'Snapshot Version',
       cancelButtonText: 'Cancel'
     };
     this.confirmationDialogService.openDialog(dialogData, bootstrap4mediumModalSize).subscribe((confirmationResult: boolean) => {
       if (confirmationResult) {
         this.updateWorkflowToSnapshot(workflow, version, () => {
+          const workflowName: string = workflow.workflowName || workflow.repository;
```
agduncan94: Same here

src/app/workflow/view/view.service.ts
david4096: Added a note to https://github.com/dockstore/dockstore/issues/2846

src/app/workflow/view/view.service.ts
david4096: dockstore/dockstore#2846

david4096:
Added pipe first and Andrew's comments to the other ticket for this. dockstore/dockstore#2846

src/app/workflow/view/view.service.ts
```diff
@@ -102,25 +138,32 @@ export class ViewService {
...
+      .openDialog(dialogData, bootstrap4mediumModalSize)
+      .pipe(first())
+      .subscribe(confirmationResult => {
+        if (confirmationResult) {
+          this.alertService.start(`A Digital Object Identifier (DOI) is being requested for workflow
+                                       "${workflowName}" version "${version.name}"!`);
+          this.workflowsService
+            .requestDOIForWorkflowVersion(workflow.id, version.id)
+            .subscribe(
+              (versions: Array<WorkflowVersion>) => this.requestDOISuccess({ ...workflow, workflowVersions: versions }, version),
```
coverbeck: It wasn't easy to tell here that a success message is displayed, and that the start message "state" was exited. If it's not the indeterminate progress bar will go on forever. You do it in `this.requestDOISuccess`, but in the future I suggest the start/success/error invocations all be in the same method, that way it's easier to tell if it's being properly "terminated".

I don't think it's worth changing for this PR, as this code will probably be replaced with the wizard we've talked about, so just FYI,

src/app/workflow/view/view.service.ts
```diff
@@ -134,8 +177,11 @@ export class ViewService {
     if (selectedWorkflow.id === workflow.id) {
       this.workflowService.setWorkflow(workflow);
     }
+
+    const workflowName: string = workflow.workflowName || workflow.repository;
```
coverbeck: This could have been a method, so if/when you implement Andrew's suggestion, you only have to do it once.

PR #789, opened 2019-09-12
Feature/verification source and platform

garyluu:
Adds verification source and platform

image

Requires the webservice

No tests :cry: 

https://github.com/dockstore/dockstore/pull/2887


denis-yuen:
@garyluu can you link the webservice PR that this shouldn't be merged before?

PR #792, opened 2019-09-13
Files visible after direct navigation to workflow

coverbeck:
dockstore/dockstore#2854

When you navigate within the app, e.g., from home page to My Workflows, the `NavigationEnd` event fires (line 132), which leads to `selectEntry` being called.

If you navigate directly to a workflow, by reloading, bookmark, etc., there is no NavigationEnd event, so `selectEntry` is never called.

Changed the code that was calling `setWorkflow` to instead call `selectEntry` (selectEntry does call setWorkflow, as well as do other stuff), which leads to a validations Ajax call, which both gets the file contents and should also fix Natalie's issue with validation messages.

@NatalieEO would you be able to test the issue you were running into with this branch?

This could probably use a test, but discovered it at the end of my day, so opening for initial review and thoughts.

### DIFF
--- a/src/app/myworkflows/my-workflow/my-workflow.component.ts
+++ b/src/app/myworkflows/my-workflow/my-workflow.component.ts
@@ -171,13 +171,16 @@ export class MyWorkflowComponent extends MyEntry implements OnInit {
             // If a user navigates directly to an unpublished workflow on their my-workflows page (via bookmark, refresh),
             // the url needs to be used to set the workflow onInit.
             // Otherwise, the select - tab.pipe results in really strange behaviour. Not entirely sure why.
-            // Changed this to invoke this.selectEntry to fix https://github.com/dockstore/dockstore/issues/2854
-            this.selectEntry(
-              this.findEntryFromPath(
-                this.urlResolverService.getEntryPathFromUrl(),
-                this.groupEntriesObject.concat(this.groupSharedEntriesObject)
-              )
+            const entry = this.findEntryFromPath(
+              this.urlResolverService.getEntryPathFromUrl(),
+              this.groupEntriesObject.concat(this.groupSharedEntriesObject)
             );
+            if (entry) {
+              // Url could have non-existent entry
+              this.workflowService.setWorkflow(entry);
+              // Call this.selectEntry to fix https://github.com/dockstore/dockstore/issues/2854
+              this.selectEntry(entry);
+            }
             // Only select initial entry if there current is no selected entry.  Otherwise, leave as is.
             if (!this.workflow) {
               if (this.workflows.length > 0) {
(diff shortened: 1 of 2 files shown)


garyluu:
Did you check that select entry is always only called once?

NatalieEO:
Yeah, I changed the same line on my branch and the messages are loading correctly for services and workflows.

garyluu:
Then my next question is, does tools have this same problem?

NatalieEO:
At least for the messages, my-tools seems to be working fine.

coverbeck:
Needs further investigation; removing reviewers for now.

coverbeck:
> Did you check that select entry is always only called once?

It was called more than once, but it no longer is with new push.

src/app/myworkflows/my-workflow/my-workflow.component.ts
```diff
@@ -171,12 +171,16 @@ export class MyWorkflowComponent extends MyEntry implements OnInit {
...
-              )
+            const entry = this.findEntryFromPath(
+              this.urlResolverService.getEntryPathFromUrl(),
+              this.groupEntriesObject.concat(this.groupSharedEntriesObject)
             );
+            if (entry) {
+              // Url could have non-existent entry
+              this.workflowService.setWorkflow(entry);
+              // Call this.selectEntry to fix https://github.com/dockstore/dockstore/issues/2854
+              this.selectEntry(entry);
```
garyluu: this is strange because this.selectEntry most likely calls `this.workflowService.setWorkflow(entry);` resulting in it being called twice

coverbeck: Ah yes, what happens is that `this.workflowsService.setWorkflow` from `this.selectEntry`, but as a result of an Ajax request; the test on line 181/185 executes before the request. Will restructure.

coverbeck:
@denis-yuen I meant to target release/2.4.0 when I created this PR, but targeted develop by mistake. I just changed the target, but double-checking if it's OK to merge to the release branch.

agduncan94:
@coverbeck I think if you rerun the tests and they pass then it is ok to merge.

coverbeck:
I relaunched the tests.

But I'm wondering, does the target branch actually matter in CircleCI? In Travis, there is the push build which runs with the branch merged against develop, but in CircleCI isn't it only building the PR?

I think the tests should be run regardless to be safe, but just wanted to make sure I understand.

garyluu:
> I relaunched the tests.
> 
> But I'm wondering, does the target branch actually matter in CircleCI? In Travis, there is the push build which runs with the branch merged against develop, but in CircleCI isn't it only building the PR?
> 
> I think the tests should be run regardless to be safe, but just wanted to make sure I understand.

https://github.com/dockstore/dockstore/issues/2475

PR #794, opened 2019-09-13
Feature/seab 433/zenodo linking

david4096:
Update develop to current state of release with release changes. Includes some other work. Should that be done in a separate PR?


denis-yuen:
@david4096 what was this, given that we have zenodo linking now is this an obsolete experiment now?

PR #798, opened 2019-09-17
SEAB-434 lighthouse audit quick fixes

emlys:
Addressed easy-to-fix items in the Google Lighthouse audit of staging.dockstore.org. A lot of these things seem to be more complicated than I expected:

## Performance
Some of these are only flagged as problems when using 4G throttling in the audit.
* Ensure text remains visible during web font load: Addressed for local Gotham fonts and Google font. Not sure if this fix can be done for Fontawesome font as it's in a library.
* Preload key requests: fix for this is commented out in index.html. While this fix addresses the audit warning, it doesn't seem to actually help, as it introduces a console warning that the preloaded files aren't used. They end up being downloaded twice. Not sure if this is fixable.
* Serve static assets with an efficient cache policy: Made a new ticket for this.
* Properly size images: Images are properly sized for a typical laptop view. https://developers.google.com/web/tools/lighthouse/audits/oversized-images?utm_source=lighthouse&utm_medium=devtools The recommendation is to have multiple versions of each image in different sizes and select a version based on the dimensions of where it will be displayed. I think this is outside the scope of this PR.
* Serve images in next gen formats: Google recommends using a combination of JPEG 2000, JPEG XR, and WebP depending on the browser. None of them is super widely supported so this would be kind of complicated.
* Remove unused CSS: None of the CSS loaded is completely unused. Could maybe use one of the tools listed here https://developers.google.com/web/tools/lighthouse/audits/unused-css?utm_source=lighthouse&utm_medium=devtools for deferring uncritical css.

## Best Practices
* Links to cross-origin destinations are unsafe: Fixed

## Accessibility
* <html> element does not have a [lang] attribute: Fixed
* Image elements do not have [alt] attributes: Fixed
* Form elements do not have associated labels: Fixed

## SEO
* Document does not have a meta description: Fixed
* Image elements do not have [alt] attributes: Fixed
* robots.txt is not valid: Made a new ticket for this

## Progressive Web App
Not applicable?

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ Run `ng g component component-name` to generate a new component. You can also us
 Optionally override the webservice version using `npm config set dockstore-ui2:webservice_version ${WEBSERVICE_VERSION}`
 Run `npm run build` to build the project. The build artifacts will be stored in the `dist/` directory. 
 
-### Production Build
+### Angular Production Build
 
 For an Angular Production Build, you will need to install Nginx. 
 Replace your nginx.conf file (location depends on your installation) with this template, filling in the two paths:
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Run `npm run build` to build the project. The build artifacts will be stored in 
 
 ### Production Build
 
-For a production build, you will need to install Nginx. 
+For an Angular Production Build, you will need to install Nginx. 
 Replace your nginx.conf file (location depends on your installation) with this template, filling in the two paths:
 ```
 events {
@@ -189,10 +189,7 @@ http {
     }
 }
 ```
-Use `npm run build.prod` for a production build and start it with `nginx`. localhost:4200 should be available immediately.
-
-
-
+Use `npm run build.prod` for an Angular Production Build and start it with `nginx`. localhost:4200 should be available immediately.
 
 
 ## Running unit tests
--- a/README.md
+++ b/README.md
@@ -124,7 +124,76 @@ Run `ng g component component-name` to generate a new component. You can also us
 ## Build
 
 Optionally override the webservice version using `npm config set dockstore-ui2:webservice_version ${WEBSERVICE_VERSION}`
-Run `npm run build` to build the project. The build artifacts will be stored in the `dist/` directory. Use the `npm run build.prod` for a production build.
+Run `npm run build` to build the project. The build artifacts will be stored in the `dist/` directory. 
+
+### Production Build
+
+For a production build, you will need to install Nginx. 
+Replace your nginx.conf file (location depends on your installation) with this template, filling in the two paths:
+```
+events {
+}
+http {
+  include       /usr/local/etc/nginx/mime.types;
+  server {
+    listen 4200;
+      location = /swagger.json {
+        proxy_pass  http://localhost:8080/swagger.json;
+      }
+    location /api/ {
+          rewrite ^ $request_uri;
+          rewrite ^/api/(.*) $1 break;
+          return 400;
+          proxy_pass     http://localhost:8080/$uri;
+    }
+    location / {
+        root  /<path to>/dockstore-ui2/dist; # FILL IN
+        index  index.html index.htm;
+        try_files $uri $uri/ /index.html =404;
+    }
+  }
+  server {
+    listen 5200;
+    location / {
+        root  /<path to>/dockstore-ui2/dist; # FILL IN
+        index  index.html index.htm;
+        try_files $uri $uri/ /index.html =404;
+        if ($request_method = 'OPTIONS') {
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -124,7 +124,76 @@ Run `ng g component component-name` to generate a new component. You can also us
...
 Optionally override the webservice version using `npm config set dockstore-ui2:webservice_version ${WEBSERVICE_VERSION}`
-Run `npm run build` to build the project. The build artifacts will be stored in the `dist/` directory. Use the `npm run build.prod` for a production build.
+Run `npm run build` to build the project. The build artifacts will be stored in the `dist/` directory. 
+
+### Production Build
+
+For a production build, you will need to install Nginx. 
+Replace your nginx.conf file (location depends on your installation) with this template, filling in the two paths:
+```
+events {
```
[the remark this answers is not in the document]
coverbeck: Sorry, this was my suggestion -- I had run the Angular prod build like that before, and thought it was easier to just use a new one for running on localhost. For the one in you have to run `install_bootstrap` which I don't think works on a Mac,  assumes that the webservice is accessible at http://webservice, etc.. This is a fallback for a less complex setup for a local build.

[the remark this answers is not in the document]
coverbeck: @denis-yuen Was my argument at all convincing or do you still feel we shouldn't have this?

README.md
```diff
@@ -124,7 +124,76 @@ Run `ng g component component-name` to generate a new component. You can also us
 ## Build
 
 Optionally override the webservice version using `npm config set dockstore-ui2:webservice_version ${WEBSERVICE_VERSION}`
-Run `npm run build` to build the project. The build artifacts will be stored in the `dist/` directory. Use the `npm run build.prod` for a production build.
+Run `npm run build` to build the project. The build artifacts will be stored in the `dist/` directory. 
+
+### Production Build
```
coverbeck: Assuming Denis agrees to keeping this section, I would say `Angular Production Build`

README.md
```diff
@@ -124,7 +124,76 @@ Run `ng g component component-name` to generate a new component. You can also us
...
+
+For a production build, you will need to install Nginx.
```
coverbeck: Ditto.

README.md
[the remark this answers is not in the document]
denis-yuen: Oh sorry, forgot. 
I think there's a couple issues, but I may not fully understand. 
Assuming the idea here is to create an nginx that can run locally for a developer but pass the audit checks

1) This configuration will/may have(?) diverge from compose_setup
2) It's not clear to me what needs to be taken from this and moved over the compose_setup and it is compose_setup which "counts" since that is what is running on staging.dockstore.org and dockstore.org
3) If the problem with running compose_setup is that is doesn't run on a Mac, we should fix that rather than have this workaround. 

That said, I'm fine with this as a temporary workaround, with the knowledge that we will need a follow-up ticket to get this into the "real" nginx config in compose_setup.

src/index.html
```diff
@@ -1,14 +1,25 @@
...
     <link rel="manifest" href="manifest.json" />
     <meta name="theme-color" content="#21335B" />
+    <meta
+      name="Description"
+      content="Dockstore, developed by the Cancer Genome Collaboratory, is an open platform used
+     by the GA4GH for sharing Docker-based tools described with the Common Workflow Language (CWL), the Workflow
+     Description Language (WDL), or Nextflow (NFL)"
+    />
+
+    <!--As suggested by Google Lighthouse audit, preload fonts-->
```
denis-yuen: In my tests a while back, this looked like the most "bang-for-the-buck" so this deserves a longer look/investigation.

emlys: There's probably something i'm missing, as preloading and `font-display: swap` are supposed to go nicely together, but i'm out of ideas

README.md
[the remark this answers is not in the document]
coverbeck: Do you run compose_setup with the web service running via the IDE's debugger?  I would think that would be tricky, but maybe I don't know docker-compose well enough.

It would be worth the teams comparing our local dev setups at some point; we might be doing some things wrong or inefficiently on our side.

README.md
```diff
@@ -124,7 +124,73 @@ Run `ng g component component-name` to generate a new component. You can also us
 ## Build
 
 Optionally override the webservice version using `npm config set dockstore-ui2:webservice_version ${WEBSERVICE_VERSION}`
-Run `npm run build` to build the project. The build artifacts will be stored in the `dist/` directory. Use the `npm run build.prod` for a production build.
+Run `npm run build` to build the project. The build artifacts will be stored in the `dist/` directory. 
+
+### Production Build
```
coverbeck: Can you change this title too

PR #799, opened 2019-09-18
Rearrange logos

coverbeck:
dockstore/dockstore#2911

Separate languages and partners into their own rows.

Screen Shot 2019-09-18 at 11 50 08 AM

### DIFF
--- a/src/app/sponsors/sponsors.component.html
+++ b/src/app/sponsors/sponsors.component.html
@@ -26,7 +26,18 @@
       </div>
     </div>
   </div>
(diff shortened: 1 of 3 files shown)


coverbeck:
> The logo changes will have to be made on the documentation too :'(

We requested that Seven Bridges do a doc PR, so I think we can wait for that. If they don't submit it, I think it's OK to leave as is (or do you think we should change it regardless, without the SB logo?

agduncan94:
@coverbeck I think it is fine to just leave it as is for now, just need to remember to update.

PR #806, opened 2019-09-23
Feature/2893/angular8

garyluu:
Many updates in order to get Angular 8 compiling.

Angular 8 by default builds for ES6. However, https://github.com/angular/angular-cli/issues/15493 prevents CircleCI from doing so. Reverting it back to ES5 :cry:


src/app/confirmation-dialog/confirmation-dialog.component.ts
```diff
@@ -1,6 +1,13 @@
 import { Component, Inject, OnInit } from '@angular/core';
 import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material';
 
+export class ConfirmationDialogData {
```
coverbeck: Just curious, why did you move this? Convention, or is something enforcing that exports be at the top? (I actually prefer it this way).

garyluu: Cannot be used before declared or something like that.

garyluu:
> What is static in this context?

https://angular.io/api/core/ViewChild

It was auto-converted by `ng update` in this PR

garyluu:
I tested every dependency I updated except json-ld (wasn't entirely sure the best way to)

browserslist
```diff
@@ -0,0 +1,12 @@
...
+# https://github.com/browserslist/browserslist#queries
+
+# You can see what browsers were selected by your queries by running:
+#   npx browserslist
+
+> 0.5%
+last 2 versions
+Firefox ESR
+not dead
+not IE 9-11 # For IE 9-11 support, remove 'not'.
```
denis-yuen: Unfortunately, we probably should support IE 9-11 till that group gets a bit smaller. 
Looking at https://angular.io/guide/browser-support it seems like they should be supported by default

PR #811, opened 2019-09-26
Support display of the concept DOI (DOI for all workflow versions)

wshands:
https://ucsc-cgl.atlassian.net/browse/SEAB-253


src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -145,6 +145,10 @@
               {{ 'descriptor_type' | mapFriendlyValue: workflow?.descriptorType }}
             </div>
           </li>
+          <div>
+            <strong matTooltip="Digital Object Identifier for all workflow versions">DOI</strong>:
+            {{ workflow && workflow.conceptDoi ? workflow.conceptDoi : 'n/a' }}
```
garyluu: my new way `(workflow || {}).conceptDoi || 'n/a'`

garyluu:
Possibly not show that in STUBs, others should chime in

src/app/workflow/info-tab/info-tab.component.html
coverbeck: Or (`workflow && workflow.conceptDoi) || 'n/a')`. :) 

What's the TS 3.7 feature?

garyluu: the way I mentioned is because it scales well when heavily nested.

so it would be `(((((layer1 || {}).layer2 || {}).layer3 || {}).layer4 || {})`
instead of `layer1 || layer1.layer2 || layer1.layer2.layer3 || layer1.layer2.layer3.layer4`

garyluu: https://github.com/microsoft/TypeScript/issues/16#issuecomment-515160784 for TS 3.7

wshands: Fixed; used (workflow && workflow.conceptDoi) || 'n/a')

denis-yuen:
Restarted build on circle-ci, seems like build was stuck or something

PR #814, opened 2019-10-01
Merge base on CircleCI before tests

garyluu:
For dockstore/dockstore#2475

I created this branch kind of off develop but a few commits short (before all the Angular updates).

If you go to https://github.com/dockstore/dockstore-ui2/tree/feature/2475/testMergeCircleCI, you'll see that it's 7 commits behind.

You can tell this PR works because if you go into CircleCI, the Angular version is printed out and it correctly shows version 8.x.x .

Printing the Angular version is for testing purposes only, I'll be reverting it after approval.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,9 @@ jobs:
           name: Check dependencies
           command: bash -i -c 'npm ls'
       - run:
+          name: Check Angular Version
+          command: bash -i -c 'npx ng version'
+      - run:
           name: Lint
           command: bash -i -c 'npm run lint'
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -17,13 +17,24 @@ jobs:
...
+              git checkout -qf FETCH_HEAD
+            fi
       - install_container_dependencies
       - run:
           name: Install dependencies
           command: bash -i -c 'npm ci'
       - run:
           name: Check dependencies
           command: bash -i -c 'npm ls'
+      - run:
```
coverbeck: What's this for?

garyluu: See PR comments

coverbeck: D'oh!

PR #815, opened 2019-10-02
Feature/epam builder fix

garyluu:
Update pipeline-builder to latest dev.

Also adds a simple test to check if the library (not our code) is broken.

See failed first commit to confirm that it's broken.

See last two commit for passing tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "ngx-bootstrap": "^3.0.1",
     "ngx-clipboard": "^12.2.1",
     "ngx-markdown": "^8.1.0",
-    "pipeline-builder": "^0.3.10-dev.298",
+    "pipeline-builder": "^0.3.10-dev.313",
     "rxjs": "^6.5.3",
     "ts-md5": "^1.2.6",
     "zone.js": "~0.9.1"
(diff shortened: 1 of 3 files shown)


garyluu:
> Does the upgrade have any bearing on [dockstore/dockstore#2461](https://github.com/dockstore/dockstore/issues/2461)?

Doesn't look like it.  Seems like the Angular 8 update may have fixed it though.

package.json
```diff
@@ -58,7 +58,7 @@
     "ngx-bootstrap": "^3.0.1",
     "ngx-clipboard": "^12.2.1",
     "ngx-markdown": "^8.1.0",
-    "pipeline-builder": "^0.3.10-dev.298",
+    "pipeline-builder": "^0.3.10-dev.313",
```
denis-yuen: Do they have non-dev releases?
(Is this the equivalent on a SNAPSHOT in Maven?)

src/app/workflow/dag/wdl-viewer/state/wdl-viewer.service.spec.ts
```diff
@@ -0,0 +1,96 @@
+/*
+ *    Copyright 2017 OICR
```
denis-yuen: Getting a bit dated

package.json
garyluu: They do, but it hasn't been updated in forever

denis-yuen: :disappointed:

PR #817, opened 2019-10-08
Feature/2827/project links icons

ByteMap:
Replaced project links with project icons

dockstore/dockstore#2827

Implemented providerIcon field to ExtendedWorkflow.ts, and providerIcon and imgProviderIcon fields to ExtendedDockstoreTool.ts

Created custom icons for Dockstore and Quay.io listed in custom-icons.ts

ProviderIcon and imgProviderIcon fields are filled in when setUpProvider() and setUpImageProvider() are called in search.service.ts

Added Seven Bridges to service-stubs.ts

Readjusted column widths after replacing project links with project icons

Implemented new tests and a new mocked object for the icons
Screen Shot 2019-10-07 at 12 18 10 PM
Screen Shot 2019-10-07 at 12 18 18 PM
Screen Shot 2019-10-07 at 12 18 41 PM
Screen Shot 2019-10-07 at 12 18 48 PM


coverbeck:
@ByteMap Can you please attach screenshots to the PR?

Also, please address the lgtm alert. Thanks.

src/app/shared/provider.service.ts
```diff
@@ -24,9 +27,30 @@ export class ProviderService {
...
     tool.providerUrl = this.getProviderUrl(gitUrl, tool.provider);
+    tool.providerIcon = this.getProviderIcon(tool.providerUrl);
     return tool;
   }
 
+  private getProviderIcon(providerUrl: string): IconDefinition {
+    if (!providerUrl) {
+      return null;
+    }
+    if (providerUrl.startsWith('https://github.com/')) {
```
coverbeck: There is a method right below that already determines the registry based on the gitUrl. Instead of having your own logic to determine the registry, I suggest using that method, something like:

```
const provider = getProvider(gitUrl)
if (provider === 'GitHub') return faGitHub
etc.
```

ByteMap: Implemented using the said method.

src/app/shared/provider.service.ts
```diff
@@ -24,9 +27,31 @@ export class ProviderService {
...
+    tool.providerIcon = this.getProviderIcon(tool.gitUrl);
     return tool;
   }
 
+  private getProviderIcon(providerUrl: string): IconDefinition {
+    const provider = this.getProvider(providerUrl);
+    if (!provider) {
+      return null;
+    }
+    if (provider === 'GitHub') {
```
coverbeck: Should have mentioned this in my previous comment, but ideally you would create constants out of these strings so they are not duplicated here and in the getProvider method.

src/app/shared/image-provider.service.ts
```diff
@@ -49,10 +52,28 @@ export class ImageProviderService {
     tool.imgProvider = friendlyRegistryName;
     if (registry) {
       tool.imgProviderUrl = this.getImageProviderUrl(tool.path, registry);
+      tool.imgProviderIcon = this.getImageProviderIcon(tool.registry);
     }
     return tool;
   }
 
+  private getImageProviderIcon(registry: RegistryEnum): IconDefinition {
```
garyluu: `| null` for both RegistryEnum and IconDefinition

src/app/shared/image-provider.service.ts
```diff
@@ -49,10 +52,28 @@ export class ImageProviderService {
...
+    }
+    switch (registry) {
+      case RegistryEnum.QUAYIO:
+        return faQuay;
+      case RegistryEnum.DOCKERHUB:
+        return faDocker;
+      case RegistryEnum.GITLAB:
+        return faGitlab;
+      case RegistryEnum.SEVENBRIDGES: // TODO: Get SevenBridges Icon
+      case RegistryEnum.AMAZONECR: // TODO: Get AMAZON_ECR Icon
```
agduncan94: Are there issues for these?

denis-yuen: In particular, the ECR icon is pretty easy to get from https://aws.amazon.com/architecture/icons/

ByteMap: Ive tried looking through the downloaded folder for the ECR icon that Amazon provided, but I cannot find the icon.

denis-yuen:
give one of these a shot, might need to play around with transparency or poke around for files with names like these

[images.zip](https://github.com/dockstore/dockstore-ui2/files/3719292/images.zip)

src/app/shared/image-provider.service.ts
coverbeck: I do see it in there; look for the file Amazon-EC2-Container-Registry_Image*

src/app/shared/provider.service.ts
```diff
@@ -24,25 +35,46 @@ export class ProviderService {
 
     tool.provider = this.getProvider(gitUrl);
     tool.providerUrl = this.getProviderUrl(gitUrl, tool.provider);
+    tool.providerIcon = this.getProviderIcon(tool.gitUrl);
     return tool;
   }
 
+  private getProviderIcon(providerUrl: string | null): IconDefinition | null {
```
coverbeck: You already have the provider from line 25/36. Instead of calling `getProvider` again on line 43, just pass in the provider as a parameter here.

src/app/shared/provider.service.ts
```diff
@@ -16,6 +16,17 @@
 
 import { ExtendedDockstoreTool } from './models/ExtendedDockstoreTool';
 import { ExtendedWorkflow } from './models/ExtendedWorkflow';
+import { IconDefinition } from '@fortawesome/fontawesome-common-types';
+import { faBitbucket, faGithub, faGitlab } from '@fortawesome/free-brands-svg-icons';
+import { faDockstore, faWhale } from './custom-icons';
+
+const GITHUB = 'GitHub';
+
```
coverbeck: Real nitpicky, I would probably remove the blank lines in between the const's.

Seeing it the way it is now, I would be tempted to make an enum, but that can be done at some other point.

src/app/shared/provider.service.ts
ByteMap: Done.

src/app/shared/provider.service.ts
ByteMap: Done

PR #819, opened 2019-10-11
Feature/2741/my entries object state

garyluu:
For dockstore/dockstore#2741

Same PR as before except with more types and removed the unused pipe.

More stable when opening a ton of tabs but marginally slower.

Still two state changes like before. Ideally should be one.

Also:
- Better sort (in order to determine initial entry)


src/app/mytools/my-tool/my-tool.component.ts
```diff
@@ -233,9 +170,14 @@ export class MyToolComponent extends MyEntry implements OnInit {
...
 
-export interface OrgToolObject {
+/**
+ * When using this, T should always be DockstoreTool
+ *
+ * @export
+ * @interface OrgToolObject
+ * @template T
+ */
+export interface OrgToolObject<T> extends OrgEntryObject<T> {
```
coverbeck: I think you can make this `export interface OrgToolObject extends OrgEntryObject<DockstoreTool>`, then you don't need the comment on line 174, nor run the risk of someone using this interface with the wrong T.

src/app/myworkflows/my-workflow/my-workflow.component.ts
```diff
@@ -283,13 +215,26 @@ export class MyWorkflowComponent extends MyEntry implements OnInit {
...
-export interface OrgWorkflowObject {
+
+/**
+ * When using this, T should always be Workflow
+ *
+ * @export
+ * @interface OrgWorkflowObject
+ * @template T
+ */
+export interface OrgWorkflowObject<T> extends OrgEntryObject<T> {
```
coverbeck: Same as comment previous comment.

src/app/mytools/my-tool/my-tool.component.ts
garyluu: Nope

src/app/mytools/sidebar-accordion/sidebar-accordion.component.html
```diff
@@ -1,8 +1,8 @@
 <mat-accordion>
-  <mat-expansion-panel *ngFor="let orgObj of groupEntriesObject" [expanded]="orgObj | expandPanel: (toolId$ | async)">
+  <mat-expansion-panel *ngFor="let orgObj of groupEntriesObject" [expanded]="orgObj.expanded">
     <mat-expansion-panel-header>
       <mat-panel-title class="org-accordion-header">
-        {{ orgObj?.namespace }}
+        {{ orgObj.registry + '/' + orgObj?.namespace }}
```
agduncan94: Missing ?

garyluu: It's impossible for it to be undefined. The 2nd one is redundant, but oh well.

PR #824, opened 2019-10-22
Feature/workflow register wizard

agduncan94:
* First pass of the workflow whitelist wizard (missing things like filter)
* Text is not finalised so please comment on it
https://github.com/dockstore/dockstore/issues/2939

page-1
page-2
page-3

Disabled options are those which cannot be deleted (for now non-stub)


agduncan94:
There seems to be a lot of tests that sometimes fail.

cypress/integration/group2/myworkflows.ts
```diff
@@ -120,6 +120,66 @@ describe('Dockstore my workflows', () => {
...
+        .route({
+          method: 'GET',
+          url: 'api/users/registries/github.com/organizations',
+          response: [ 'foobar', 'lorem' ]
+        })
+        .route({
+          method: 'GET',
+          url: 'api/users/registries/github.com/organizations/foobar',
+          response: [
+            {'organization': 'foobar', 'repositoryName': 'canDeleteMe', 'gitRegistry': 'github.com', 'present': true, 'canDelete': true, 'path': 'foobar/canDeleteMe'},
```
garyluu: Try adding type here

coverbeck:
Comments based on screenshots:

1. Suggest a "Repositories" label
2. What will be the behavior (and maybe this is clear in the code I haven't looked at yet), if you change the organization or registry? For example, I select GitHub/coverbeck, toggle a couple of repos, then change org to DataBiosphere -- will my changes to coverbeck repos still be applied?

coverbeck:
>     2. What will be the behavior (and maybe this is clear in the code I haven't looked at yet), if you change the organization or registry? For example, I select GitHub/coverbeck, toggle a couple of repos, then change org to DataBiosphere -- will my changes to coverbeck repos still be applied?

Ah, I see, you apply immediately and have a message that so indicates. So you can ignore this.

src/app/shared/entry-wizard/entry-wizard.component.html
```diff
@@ -0,0 +1,37 @@
...
+      <mat-label>Organization</mat-label>
+      <mat-select [(ngModel)]="selectedGitOrganization" name="organization" (ngModelChange)="getRepositories($event)">
+        <mat-option *ngFor="let organization of gitOrganizations$ | async" [value]="organization">
+          {{ organization }}
+        </mat-option>
+      </mat-select>
+    </mat-form-field>
+
+    <div fxLayout="column" fxLayoutAlign="start start" fxLayoutGap="20px">
+      <div *ngFor="let repo of gitRepositories$ | async">
```
coverbeck: I put this in my screenshot comment already, but I think there should be a mat-label here

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -0,0 +1,141 @@
...
+import { AlertService } from '../alert/state/alert.service';
+import { Repository } from '../openapi/model/repository';
+import { HttpErrorResponse } from '@angular/common/http';
+import { finalize } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class EntryWizardService {
+  constructor(private entryWizardStore: EntryWizardStore, private defaultService: DefaultService, private alertService: AlertService) {}
```
coverbeck: Is DefaultService an intentional name, or was some annotation left out on the webservice side?

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -0,0 +1,141 @@
...
+          this.entryWizardStore.update({ gitOrganizations: organizations });
+        },
+        (error: HttpErrorResponse) => {
+          this.alertService.detailedError(error);
+        }
+      );
+  }
+
+  /**
+   * Updates the list of repositories based on the given git registry and organization
```
coverbeck: I find the word "update" in this comment and the function names a little confusing in comparison to "addWorkflow" -- this is only updating the Akita store, but addWorkfow is updating the backend, if I'm understanding correctly.

Maybe not a big deal, but when I saw `updateGitRepositories`, I was expecting to see something like `this.defaultService.update...`.

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -0,0 +1,141 @@
...
+    const registryEnum = this.convertSourceControlStringToEnum(registry);
+    this.defaultService
+      .addWorkflow(registryEnum, organization, repositoryName)
+      .pipe(finalize(() => this.entryWizardStore.setLoading(false)))
+      .subscribe(
+        (workflow: BioWorkflow) => {
+          this.alertService.detailedSuccess('Workflow ' + registry + '/' + organization + '/' + repositoryName + ' has been added');
+        },
+        (error: HttpErrorResponse) => {
+          this.alertService.detailedError(error);
```
coverbeck: If this fails, don't you need to update the UI? Won't the toggle show the repo enabled, but it actually won't be?

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -0,0 +1,141 @@
...
+    const registryEnum = this.convertSourceControlStringToEnum(registry);
+    this.defaultService
+      .deleteWorkflow(registryEnum, organization, repositoryName)
+      .pipe(finalize(() => this.entryWizardStore.setLoading(false)))
+      .subscribe(
+        (workflow: BioWorkflow) => {
+          this.alertService.detailedSuccess('Workflow ' + registry + '/' + organization + '/' + repositoryName + ' has been deleted');
+        },
+        (error: HttpErrorResponse) => {
+          this.alertService.detailedError(error);
```
coverbeck: If this fails, don't you need to update the UI? Won't the toggle show the repo disabled, but it actually won't be?

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -0,0 +1,141 @@
...
+   */
+  addWorkflow(registry: string, organization: string, repositoryName: string) {
+    this.entryWizardStore.setLoading(true);
+    const registryEnum = this.convertSourceControlStringToEnum(registry);
+    this.defaultService
+      .addWorkflow(registryEnum, organization, repositoryName)
+      .pipe(finalize(() => this.entryWizardStore.setLoading(false)))
+      .subscribe(
+        (workflow: BioWorkflow) => {
+          this.alertService.detailedSuccess('Workflow ' + registry + '/' + organization + '/' + repositoryName + ' has been added');
```
coverbeck: Does the store need to be updated?

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -0,0 +1,141 @@
...
+   */
+  removeWorkflow(registry: string, organization: string, repositoryName: string) {
+    this.entryWizardStore.setLoading(true);
+    const registryEnum = this.convertSourceControlStringToEnum(registry);
+    this.defaultService
+      .deleteWorkflow(registryEnum, organization, repositoryName)
+      .pipe(finalize(() => this.entryWizardStore.setLoading(false)))
+      .subscribe(
+        (workflow: BioWorkflow) => {
+          this.alertService.detailedSuccess('Workflow ' + registry + '/' + organization + '/' + repositoryName + ' has been deleted');
```
coverbeck: Does the store need to be updated?

agduncan94:
@coverbeck as soon as you toggle, workflows are added/deleted. Clicking the finish button (at the bottom) will fetch workflows and update the my workflows listing.

src/app/shared/state/entry-wizard.service.ts
agduncan94: Probably an annotation left out.

src/app/shared/state/entry-wizard.service.ts
agduncan94: By store do you mean the list of workflows on the page (my-workflows)? This is updated when you click finish. I'm open to more changes in the next round of updates once people get the chance to try it out.

src/app/shared/state/entry-wizard.service.ts
agduncan94: Looks like maybe this would solve it
https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Annotations#tag

src/app/shared/state/entry-wizard.service.ts
agduncan94: Yes this is a good point.

src/app/shared/state/entry-wizard.service.ts
coverbeck: I meant the Akita store:

1. You get list of repos from web service and put them into entity-wizard.store
2. The UI displays the (filtered) repos
3. User toggles a repo
4. This code is called to toggle repo on backend
5. Doesn't store on step 1 need to have new state of repo?

src/app/shared/state/entry-wizard.service.ts
agduncan94: I've made an issue to update the endpoints to have tags.

src/app/shared/state/entry-wizard.service.ts
agduncan94: Good point, I've made it pull changes on error so that the toggles are correct.

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -0,0 +1,145 @@
...
+  addWorkflowToDatabase(registry: string, organization: string, repositoryName: string) {
+    this.entryWizardStore.setLoading(true);
+    const registryEnum = this.convertSourceControlStringToEnum(registry);
+    this.defaultService
+      .addWorkflow(registryEnum, organization, repositoryName)
+      .pipe(finalize(() => this.entryWizardStore.setLoading(false)))
+      .subscribe(
+        (workflow: BioWorkflow) => {
+          this.alertService.detailedSuccess('Workflow ' + registry + '/' + organization + '/' + repositoryName + ' has been added');
+          this.updateGitRepositoryStore(registry, organization);
```
coverbeck: Could you just update the store with this one item instead of fetching everything again?

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -0,0 +1,145 @@
...
+      );
+  }
+
+  /**
+   * Removes a workflow for the given git repository
+   * @param registry git registry
+   * @param organization git organization
+   * @param repositoryName git repository name
+   */
+  removeWorkflowToDatabase(registry: string, organization: string, repositoryName: string) {
```
coverbeck: From instead of To in function name

src/app/workflow/register-workflow-modal/register-workflow-modal.component.ts
```diff
@@ -62,12 +65,16 @@ export class RegisterWorkflowModalComponent implements OnInit, AfterViewChecked,
   };
   public options = [
     {
-      label: 'Use CWL, WDL or Nextflow from GitHub, Bitbucket, etc.',
+      label: 'Quickly register workflows from GitHub, Bitbucket, and GitLab',
       value: 0
     },
     {
-      label: 'Create and save CWL, WDL, or Nextflow on Dockstore.org',
+      label: 'Create non-standard workflows from GitHub, Bitbucket, and GitLab',
```
coverbeck: What is a non-standard workflow? This sounds odd to me.

src/app/shared/state/entry-wizard.service.ts
agduncan94: The store isn't a list of workflows but rather a list of repository objects.

src/app/shared/state/entry-wizard.service.ts
agduncan94: Oops

src/app/workflow/register-workflow-modal/register-workflow-modal.component.ts
agduncan94: I was trying to word it in such a way that it was for workflows that don't use the default descriptor path or that have multiple workflows in one repository.

src/app/shared/state/entry-wizard.service.ts
agduncan94: Oh realized this is in the success section, I think I can remove this and keep it only in the error section.

agduncan94: Talked with Gary, have a possible solution.

src/app/shared/entry-wizard/entry-wizard.component.html
```diff
@@ -0,0 +1,38 @@
...
+    <mat-form-field class="w-100">
+      <mat-label>Git Registry</mat-label>
+      <mat-select [(ngModel)]="selectedGitRegistry" name="registry" (ngModelChange)="getOrganizations($event)">
+        <mat-option *ngFor="let registry of gitRegistries$ | async" [value]="registry">
+          {{ registry }}
+        </mat-option>
+      </mat-select>
+    </mat-form-field>
+  </div>
+  <div *ngIf="selectedGitRegistry !== null && (gitOrganizations$ | async)">
```
garyluu: truthy check instead?

src/app/shared/entry-wizard/entry-wizard.component.ts
```diff
@@ -0,0 +1,58 @@
...
+  selector: 'entry-wizard',
+  templateUrl: './entry-wizard.component.html',
+  styleUrls: ['./entry-wizard.component.scss']
+})
+export class EntryWizardComponent implements OnInit {
+  isLoading$: Observable<boolean>;
+  gitRegistries$: Observable<string[]>;
+  gitOrganizations$: Observable<string[]>;
+  gitRepositories$: Observable<Repository[]>;
+  selectedGitRegistry: string = null;
```
garyluu: I think undefined works better

src/app/shared/entry-wizard/entry-wizard.component.html
```diff
@@ -0,0 +1,38 @@
+<mat-progress-bar *ngIf="isLoading$ | async" mode="indeterminate"></mat-progress-bar>
+<div>
+  <p>Toggle repositories on to create workflows from the associated Git repository. Once added, these workflows can be customized.</p>
+  <p>This approach is best for Git repositories with only one workflow.</p>
```
denis-yuen: Slightly more broadly, is it just best for the first workflow in a repo? 
Or do we really mean that if a repo has more than one, this wizard should be avoided?

src/app/shared/org-schema.service.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { inject, TestBed } from '@angular/core/testing';
-import { Organization } from './openapi';
+import { Organization } from './swagger';
```
denis-yuen: Is this swapping the "wrong" way on purpose?

src/app/shared/org-schema.service.ts
```diff
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { Organization } from './openapi';
```
denis-yuen: Ditto

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -0,0 +1,177 @@
...
+      }
+      case BioWorkflow.SourceControlEnum.GitlabCom: {
+        return BioWorkflow.SourceControlEnum.GitlabCom;
+      }
+      case BioWorkflow.SourceControlEnum.BitbucketOrg: {
+        return BioWorkflow.SourceControlEnum.BitbucketOrg;
+      }
+      case BioWorkflow.SourceControlEnum.DockstoreOrg: {
+        return BioWorkflow.SourceControlEnum.DockstoreOrg;
+      }
```
garyluu: default return null with an error message seems better imo

src/app/workflow/register-workflow-modal/register-workflow-modal.component.ts
```diff
@@ -234,6 +244,17 @@ export class RegisterWorkflowModalComponent implements OnInit, AfterViewChecked,
     }
   }
 
+  /**
+   * Updates list of entries for the logged in user
+   */
+  getMyEntries() {
+    combineLatest([this.userQuery.user$, this.sessionQuery.entryType$]).subscribe(([user, entryType]) => {
```
garyluu: unsubscribe

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -0,0 +1,177 @@
...
+    const updatedGitRepos = Object.assign([], gitRepos);
+    updatedGitRepos[index] = newRepository;
+    this.entryWizardStore.update({ gitRepositories: updatedGitRepos });
+  }
+
+  /**
+   * Converts a string source control to a Source Control enum
+   * @param sourceControl source control string
+   */
+  convertSourceControlStringToEnum(sourceControl: string): BioWorkflow.SourceControlEnum {
```
denis-yuen: Seems like there should be a simpler way of writing this, not familiar enough with Typescript to help though

src/app/shared/org-schema.service.spec.ts
agduncan94: Yes, because I don't think it was actually working before.

src/app/shared/entry-wizard/entry-wizard.component.html
agduncan94: The idea is that if you have only one workflow in a repository, you probably don't need to specify things like the workflow name.

If you have multi workflows in one repository, each workflow will have a workflow name. Chances are there will not be a main repository that doesn't use a workflow name.

src/app/shared/state/entry-wizard.service.ts
agduncan94: Tried looking but couldn't find a better approach that worked.

PR #825, opened 2019-10-22
Feature/2561/cache npm

ByteMap:
Cache NPM Node_Modules

dockstore/dockstore#2561

Modified config.yml to cache dependencies from package-lock.json to
the node_modules directory.

Implemented a bash script that checks if the node_modules directory
exists. If it does not exist, it will run a npm ci. If it does exist,
it will do nothing and fetch its dependencies from node_modules.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,10 +27,10 @@ jobs:
             fi
       - install_container_dependencies
       - restore_cache:
-          keys:
-            - dep-cache-{{ checksum "./package-lock.json" }}
+          key: dep-cache-{{ checksum "./package-lock.json" }}
       - run:
           name: Install dependencies
+          #Only run 'npm ci' if node_modules was not restored from cache
           command: |
               if [ ! -d "./node_modules" ]
               then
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -26,9 +26,20 @@ jobs:
               git checkout -qf FETCH_HEAD
             fi
       - install_container_dependencies
+      - restore_cache:
+          keys:
```
coverbeck: This could just be `key`, since we don't have multiple keys.

ByteMap: Done.

.circleci/config.yml
```diff
@@ -26,9 +26,20 @@ jobs:
...
+            - dep-cache-{{ checksum "./package-lock.json" }}
       - run:
           name: Install dependencies
-          command: bash -i -c 'npm ci'
+          command: |
+              if [ ! -d "./node_modules" ]
```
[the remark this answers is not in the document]
ByteMap: Done

PR #826, opened 2019-10-30
dockstore/dockstore#2295

emlys:
dockstore/dockstore#2295


emlys:
> Could use an explanation on why package-lock.json was changed

was a mistake, undid those changes

PR #827, opened 2019-11-07
Support for workflow version aliases

wshands:
UI portion that supports workflow version aliases
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-670


src/app/aliases/aliases.component.ts
```diff
@@ -17,13 +17,14 @@ export class AliasesComponent extends Base implements OnInit {
...
   workflow$: Observable<Workflow | null>;
+  workflowVersionPathInfo$: Observable<WorkflowVersionPathInfo | null>;
   tool$: Observable<DockstoreTool | null>;
 
   public type: string | null;
   public alias: string | null;
   public validType: boolean;
   // Types contains resource types that support aliases
-  public types = ['organizations', 'collections', 'workflows', 'tools', 'containers'];
+  public types = ['organizations', 'collections', 'workflows', 'tools', 'containers', 'workflow-versions'];
```
coverbeck: We could use consts for the array values; then the if statements starting at line 42 could at least compare against the consts, and avoid the possibility of a typo. Or there might even be a more elegant solution with enums or something like [this](https://stackoverflow.com/questions/44497388/typescript-array-to-string-literal-type)

src/app/aliases/aliases.component.ts
```diff
@@ -55,6 +56,14 @@ export class AliasesComponent extends Base implements OnInit {
           this.router.navigate(['/organizations', collection.organizationName, 'collections', collection.name]);
         }
       });
+    } else if (this.type === 'workflow-versions' && this.alias) {
```
coverbeck: Every if statement tests `this.alias`. I would have one `if (this.alias)` at the top so you don't have to keep repeating the test.

PR #831, opened 2019-11-14
dockstore/dockstore#2942

emlys:
dockstore/dockstore#2942

### DIFF
--- a/src/app/search/search-entry-table.ts
+++ b/src/app/search/search-entry-table.ts
@@ -49,6 +49,11 @@ export abstract class SearchEntryTable extends Base implements OnInit {
       this.dataSource.paginator.pageSize = pageSize;
       this.privateNgOnInit();
     });
+    this.dataSource.sortData = (data: DockstoreTool[], sort: MatSort) => {
+      return data.slice().sort((a, b) => {
+        return this.searchService.compareAttributes(a, b, sort.active, sort.direction);
+      });
+    };
   }
 
   updatePageSize($event: any) {
--- a/src/app/search/search-entry-table.ts
+++ b/src/app/search/search-entry-table.ts
@@ -48,7 +48,7 @@ export abstract class SearchEntryTable extends Base implements OnInit {
       this.dataSource.paginator.pageSize = pageSize;
       this.privateNgOnInit();
     });
-    this.dataSource.sortData = (data: DockstoreTool[], sort: MatSort) => {
+    this.dataSource.sortData = (data: DockstoreTool[] | Workflow[], sort: MatSort) => {
       return data.slice().sort((a, b) => {
         return this.searchService.compareAttributes(a, b, sort.active, sort.direction);
       });
--- a/src/app/search/search-entry-table.ts
+++ b/src/app/search/search-entry-table.ts
@@ -33,6 +33,32 @@ export abstract class SearchEntryTable extends Base implements OnInit {
 
   public readonly displayedColumns = ['name', 'verified', 'author', 'descriptorType', 'projectLinks', 'starredUsers'];
   abstract dataSource: MatTableDataSource<Workflow | DockstoreTool>;
+
+  static compareAttributes(a: DockstoreTool | Workflow, b: DockstoreTool | Workflow, attribute: string, direction: string) {
+    let aVal = a[attribute];
+    let bVal = b[attribute];
+    console.log(attribute);
+
+    // regardless of sort direction, null or empty values go at the end
+    if (!aVal) {
+      return 1;
+    }
+    if (!bVal) {
+      return -1;
+    }
+
+    // ignore case when sorting by author or name
+    if (attribute === 'author' || attribute === 'name') {
+      aVal = aVal.toLowerCase();
+      bVal = bVal.toLowerCase();
+    }
+    if (direction === 'asc') {
+      return aVal < bVal ? -1 : 1;
+    } else if (direction === 'desc') {
+      return aVal < bVal ? 1 : -1;
+    }
+  }
+
   abstract privateNgOnInit(): void;
 
   constructor(protected dateService: DateService, protected searchQuery: SearchQuery, protected searchService: SearchService) {
--- a/src/app/search/search-entry-table.ts
+++ b/src/app/search/search-entry-table.ts
@@ -34,31 +34,6 @@ export abstract class SearchEntryTable extends Base implements OnInit {
   public readonly displayedColumns = ['name', 'verified', 'author', 'descriptorType', 'projectLinks', 'starredUsers'];
   abstract dataSource: MatTableDataSource<Workflow | DockstoreTool>;
 
-  static compareAttributes(a: DockstoreTool | Workflow, b: DockstoreTool | Workflow, attribute: string, direction: string) {
-    let aVal = a[attribute];
-    let bVal = b[attribute];
-    console.log(attribute);
-
-    // regardless of sort direction, null or empty values go at the end
-    if (!aVal) {
-      return 1;
-    }
-    if (!bVal) {
-      return -1;
-    }
-
-    // ignore case when sorting by author or name
-    if (attribute === 'author' || attribute === 'name') {
-      aVal = aVal.toLowerCase();
-      bVal = bVal.toLowerCase();
-    }
-    if (direction === 'asc') {
-      return aVal < bVal ? -1 : 1;
-    } else if (direction === 'desc') {
-      return aVal < bVal ? 1 : -1;
-    }
-  }
-
   abstract privateNgOnInit(): void;
 
   constructor(protected dateService: DateService, protected searchQuery: SearchQuery, protected searchService: SearchService) {
--- a/src/app/search/search-entry-table.ts
+++ b/src/app/search/search-entry-table.ts
@@ -33,7 +33,6 @@ export abstract class SearchEntryTable extends Base implements OnInit {
 
   public readonly displayedColumns = ['name', 'verified', 'author', 'descriptorType', 'projectLinks', 'starredUsers'];
   abstract dataSource: MatTableDataSource<Workflow | DockstoreTool>;
-
   abstract privateNgOnInit(): void;
 
   constructor(protected dateService: DateService, protected searchQuery: SearchQuery, protected searchService: SearchService) {
--- a/src/app/search/search-tool-table/search-tool-table.component.ts
+++ b/src/app/search/search-tool-table/search-tool-table.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, HostListener, Input, OnInit } from '@angular/core';
 import { MatTableDataSource } from '@angular/material/table';
 import { takeUntil } from 'rxjs/operators';
 import { DateService } from '../../shared/date.service';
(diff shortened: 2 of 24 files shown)


emlys:
Fixed alphabetization in the tool and workflow search results tabs as well as in the facets because it was a very similar issue. Tool/workflow name alphabetization in the results is still weird because sorting is done by `name`, while the displayed name is prefixed with the `namespace`. See https://github.com/dockstore/dockstore/issues/3014

src/app/search/search-entry-table.ts
```diff
@@ -33,6 +33,32 @@ export abstract class SearchEntryTable extends Base implements OnInit {
 
   public readonly displayedColumns = ['name', 'verified', 'author', 'descriptorType', 'projectLinks', 'starredUsers'];
   abstract dataSource: MatTableDataSource<Workflow | DockstoreTool>;
+
+  static compareAttributes(a: DockstoreTool | Workflow, b: DockstoreTool | Workflow, attribute: string, direction: string) {
```
coverbeck: I feel like this belongs in search.service.ts

garyluu: Should have a unit test for this

emlys:
On staging, null or empty string author names are sorted higher than defined names but displayed as n/a which is sort of confusing. I went with placing them at the end of the results, regardless of sort direction. This makes sense to me because if you're sorting by author, you probably don't care about results with no author, and some people seem to agree here [https://ux.stackexchange.com/questions/45479/where-to-put-empty-values-in-sorted-list]. But you could also make a case for sorting them last when ascending and first when descending, or putting them with the N's because they are displayed as "n/a".

src/app/search/search-entry-table.ts
```diff
@@ -33,6 +33,32 @@ export abstract class SearchEntryTable extends Base implements OnInit {
...
+    let aVal = a[attribute];
+    let bVal = b[attribute];
+    console.log(attribute);
+
+    // regardless of sort direction, null or empty values go at the end
```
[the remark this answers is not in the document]
emlys: Sure. Added a comment below about it

src/app/search/search-entry-table.ts
```diff
@@ -33,6 +33,32 @@ export abstract class SearchEntryTable extends Base implements OnInit {
...
+      return -1;
+    }
+
+    // ignore case when sorting by author or name
+    if (attribute === 'author' || attribute === 'name') {
+      aVal = aVal.toLowerCase();
+      bVal = bVal.toLowerCase();
+    }
+    if (direction === 'asc') {
+      return aVal < bVal ? -1 : 1;
```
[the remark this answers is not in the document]
emlys: I figured the results would be the same because even if they're equal they still have to get put in some order, so it wouldn't matter? But yes returning 0 would be more technically correct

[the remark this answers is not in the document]
coverbeck: In looking around I've seen it both ways. so it might be OK. But looking at [the Mozilla doc on sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), their examples return 0 (while having a note about certain browsers not honoring it equally). I think it best to return 0.

Depending on the implementation of the sort, and this is just a wild guess on my part, I would think 0 could be marginally faster, albeit probably not noticeably so. If you return 0, you know you don't have to sort the item any more, if you return non-zero, then you might need to sort the item against another item.

src/app/search/search-entry-table.ts
coverbeck: FWIW, it will be easier to unit test if it's in the service, at least it is for me. I dislike writing unit tests for components because there's so much more going on, whereas with a service you can just have a function you can invoke and test the outputs.

src/app/search/state/search.service.ts
```diff
@@ -75,6 +76,38 @@ export class SearchService {
...
 
+  /**
+   * Return -1 if a sorts before b, 1 if b sorts before a, and 0 if they are the same,
+   * comparing based on the given attribute and direction
+   * @param a: DockstoreTool or Workflow
+   * @param b: DockstoreTool or Workflow
+   * @param attribute: workflow or tool property to sort by
+   * @param direction: 'asc' or 'desc'
+   */
+  static compareAttributes(a: DockstoreTool | Workflow, b: DockstoreTool | Workflow, attribute: string, direction: string) {
```
coverbeck: I don't think this should be static. We don't have static methods in other services that I can find.

garyluu: direction: SortDirection

src/app/search/state/search.service.spec.ts
```diff
@@ -103,4 +104,46 @@ describe('SearchService', () => {
       expect(workflowSource.imgProviderUrl).toBe(undefined);
     }
   ));
+  it('should sort workflows correctly', inject([SearchService], (service: SearchService) => {
+    const a: Workflow = {
+      author: 'a',
+      gitUrl: 'https://giturl',
+      mode: 'FULL',
```
garyluu: use enum for mode and descriptor type.

src/app/search/state/search.service.spec.ts
```diff
@@ -103,4 +104,46 @@ describe('SearchService', () => {
...
+      organization: '',
+      repository: '',
+      sourceControl: '',
+      descriptorType: 'CWL',
+      workflow_path: '',
+      defaultTestParameterFilePath: ''
+    };
+
+    const b: Workflow = {
```
garyluu:
```
    const b: Workflow = { ...a, author: 'B' };

    const c: Workflow = { ...a, author: null };
```

src/app/search/state/search.service.ts
```diff
@@ -75,6 +76,38 @@ export class SearchService {
...
+      return -1;
+    }
+
+    // ignore case when sorting by author or name
+    if (attribute === 'author' || attribute === 'name') {
+      aVal = aVal.toLowerCase();
+      bVal = bVal.toLowerCase();
+    }
+    if (direction === 'asc') {
+      return aVal < bVal ? -1 : aVal > bVal ? 1 : 0;
```
coverbeck: I missed this the first time, but I think you should be using [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare). Something like `aVal.localCompare(bVal)`, and then multiple by -1 if desc

src/app/search/state/search.service.ts
```diff
@@ -271,10 +304,17 @@ export class SearchService {
...
+      if (!a.key) {
+        return 1;
+      } // always put null/empty values at the end regardless of sort direction
+      if (!b.key) {
+        return -1;
+      }
+
       if (orderMode) {
-        return a.key > b.key ? 1 : -1;
+        return a.key.toLowerCase() > b.key.toLowerCase() ? 1 : -1; // ascending
```
coverbeck: Why did you change this one? Won't this change the behavior of existing sorts?

Not saying you shouldn't have just trying to understand -- as I see it your compareAttributes is handling the case for author and name. What is the case for?

src/app/search/state/search.service.spec.ts
```diff
@@ -103,4 +104,46 @@ describe('SearchService', () => {
...
+      mode: 'FULL',
+      organization: '',
+      repository: '',
+      sourceControl: '',
+      descriptorType: 'CWL',
+      workflow_path: '',
+      defaultTestParameterFilePath: ''
+    };
+
+    expect(SearchService.compareAttributes(a, b, 'author', 'asc')).toEqual(-1);
```
coverbeck: Since  compareAttributes has branching logic based on the attribute name, your tests should cover an attribute that doesn't go into that code branch.

src/app/search/state/search.service.ts
emlys: Okay. IntelliJ highlights it saying it should be static, I don't see a way to make it use anything specific to the instance, but its just a warning not an error

coverbeck: I think this is all there is:

1. Remove static here
2. In other files, instead of `SearchService.compareAttributes`, invoke `searchService.compareAttributes`, where `searchService` is an instance of the service.

src/app/search/state/search.service.ts
emlys: ah okay, that does look cleaner

src/app/search/state/search.service.ts
emlys: I thought we would want to ignore case in other sorts, but you're right this isn't needed for this PR. No problem to undo it

src/app/search/state/search.service.spec.ts
emlys: added 2 cases for descriptorType attribute

src/app/search/search-tool-table/search-tool-table.component.ts
```diff
@@ -34,6 +35,11 @@ export class SearchToolTableComponent extends SearchEntryTable implements OnInit
         this.dataSource.data = entries;
       }
     });
+    this.dataSource.sortData = (data: DockstoreTool[], sort: MatSort) => {
```
coverbeck: Can you just move this to search-entry-table.ts, and then you don't have to repeat it here and in search-workflow-table.component.ts? Sorry I didn't notice before.

src/app/search/search-workflow-table/search-workflow-table.component.ts
```diff
@@ -49,6 +50,11 @@ export class SearchWorkflowTableComponent extends SearchEntryTable implements On
         this.dataSource.data = entries;
       }
     });
+    this.dataSource.sortData = (data: Workflow[], sort: MatSort) => {
```
coverbeck: See previous comment.

src/app/search/search-tool-table/search-tool-table.component.ts
emlys: ah good point

src/app/search/search-entry-table.ts
```diff
@@ -48,6 +48,11 @@ export abstract class SearchEntryTable extends Base implements OnInit {
       this.dataSource.paginator.pageSize = pageSize;
       this.privateNgOnInit();
     });
+    this.dataSource.sortData = (data: DockstoreTool[], sort: MatSort) => {
```
garyluu: try `DockstoreTool[] | Workflow[]`

src/app/search/state/search.service.ts
```diff
@@ -271,10 +314,16 @@ export class SearchService {
...
     orderedArray = orderedArray.sort((a, b) => {
+      if (!a.key) {
+        return 1;
+      }
+      if (!b.key) {
+        return -1;
+      }
       if (orderMode) {
-        return a.key > b.key ? 1 : -1;
+        return a.key.toLowerCase() > b.key.toLowerCase() ? 1 : -1;
```
coverbeck: I know it was already like this, but since you use it in the other method, I think you should `localeCompare` here as well.

src/app/search/state/search.service.ts
```diff
@@ -271,10 +314,16 @@ export class SearchService {
...
       } else {
-        return a.key < b.key ? 1 : -1;
+        return a.key.toLowerCase() < b.key.toLowerCase() ? 1 : -1;
```
coverbeck: Ditto

PR #836, opened 2019-11-22
Featured content

Ldcabansay:
Adds featured content widget to logged-in homepage. The content is pulled from an external static html on an s3 bucket. The html file is updated through the [dockstore/extra-content](https://github.com/dockstore/extra-content) repo, with changes automatically uploaded to s3 via circleCI. A webservice config `featuredContentUrl` is used to set the s3 url for either staging, production, or local builds. 
image


src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -19,10 +19,11 @@
     <app-entries *ngIf="user$ | async" class="w-100"></app-entries>
     <app-organizations *ngIf="user$ | async" class="w-100"></app-organizations>
   </div>
-
   <div fxLayout="column" fxLayoutAlign="start start" fxLayoutGap="20px" class="px-4 w-100" fxFlex="1 1 100%" fxFlexAlign="stretch">
-    <div class="widget-section">
-      <h3>Explore Featured Content</h3>
+    <div class="widget-section-featured-content">
+      <h3 class="feat-content-h3">Explore Featured Content</h3>
```
coverbeck: I do not see `feat-content-h3` defined anywhere

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -19,10 +19,11 @@
     <app-entries *ngIf="user$ | async" class="w-100"></app-entries>
     <app-organizations *ngIf="user$ | async" class="w-100"></app-organizations>
   </div>
-
   <div fxLayout="column" fxLayoutAlign="start start" fxLayoutGap="20px" class="px-4 w-100" fxFlex="1 1 100%" fxFlexAlign="stretch">
-    <div class="widget-section">
-      <h3>Explore Featured Content</h3>
+    <div class="widget-section-featured-content">
```
coverbeck: I don't see this class either

src/styles.css
```diff
@@ -99,3 +99,41 @@ markdown pre code {
 .starring-button {
   margin-top: 20px;
 }
+
```
coverbeck: These following styles are presumably referenced by the external content, right? If so, I think you should do one of the following:

1. You should have comments before and after this section stating that it used by the external content, or someday somebody like me will come in and delete it  thinking it's not being used because I can't find usages in grep.
2. More generally, I wonder if this styling should instead live in the external content itself, perhaps as inline css. It feels slightly off to have it separated from HTML. What if we decide one day that we don't want the external content to be cards? You'd either have to change this code, requiring a new UI build, which is what we're trying to avoid, or you change the external HTML to not use these classes, in which case, you "orphan" this CSS. OTOH, there are libraries of CSS that live independently from the HTML. And there's the risk that the CSS in the external HTML might clobber our CSS.

@garyluu thoughts on this?

Ldcabansay: For #2 I was thinking about that, but it gots a bit unruly to have everything inline. I was thinking cards could be our basic default, but if we did want to do a one-off styling for that specific html we could do in-line styles.

coverbeck: After further verbal discussion, I don't think option 2 makes sense. We're really loading a snippet of HTML, not an HTML page, where I had been thinking you could you have the CSS in the `<head>` section. So option 1 is the way to go.

garyluu: A few things:
- I didn't realize we still had a style.css file
- If possible, move this into the scss file and then have one class style that contains a bunch of nested styles using scss. This way u can just comment on the entire thing.
- I thought we agreed to ban 'px' for most things
- I don't mind having the style in this repo as long as it's obvious (comments and such)

src/app/shared/dockstore.model.ts
```diff
@@ -69,6 +69,7 @@ export class Dockstore {
   static GITHUB_APP_INSTALLATION_URL = 'will be filled in by configuration.service';
 
   static DOCUMENTATION_URL = 'https://docs.dockstore.org';
+  static FEATURED_CONTENT_URL = 'will be filled in by configuration.service';
```
agduncan94: Make sure to update the compose setup with this new variable.

Ldcabansay: Was merged yesterday 👍

PR #837, opened 2019-11-22
Feature/2966/search routing

garyluu:
For dockstore/dockstore#2966 

So many changes:
This fixes routing so that the back button works
This remembers pageSize and pageIndex
Basic search and advanced search are now mutually exclusive. If the URL has stuff from both, basic search is prioritized.
Added manual test because we don't have elasticsearch running in tests yet.
Search for tool and workflow names too in basic search

See https://garyluu.github.io/dockstore-ui2 for a slightly up to date commit of this branch.

There's a pretty high chance that something broke while fixing this. Everyone should extensively test that link before this should be merged.


denis-yuen:
Was here, kicked integration test 2

src/app/search/search.component.ts
```diff
@@ -151,82 +153,69 @@ export class SearchComponent implements OnInit, OnDestroy {
...
   parseParams() {
-    let useAdvSearch = false;
     const paramMap: ParamMap = this.activatedRoute.snapshot.queryParamMap;
     if (!paramMap) {
       return;
     }
+    const newAdvancedSearchObject: AdvancedSearchObject = initialAdvancedSearchObject;
+    let newFilters: Map<string, Set<string>> = new Map<string, Set<string>>();
+    // URL is gospel, if it doesn't have a search term, then there's no search term
+    if (!paramMap.has('search')) {
```
coverbeck: 'search' could be a constant; it's used 5 times. In general it would be nice if there were an interface with all the difference queryParams we honor.

PR #842, opened 2019-11-29
Use new restful starring endpoints

garyluu:
Use the new restful starring endpoint

For dockstore/dockstore#2395


garyluu:
Need webservice change

PR #845, opened 2019-12-06
Recent entries does not show the registry

agduncan94:
Will now show b/c/d instead of a/b/c/d. I think the registry usually doesn't matter, so it makes sense to hide it. This is how search currently works.

Depends on https://github.com/dockstore/dockstore/pull/3061


agduncan94:
@coverbeck Yes that is strange, I'll try with a new release before I merge this.

PR #847, opened 2019-12-13
Feature/homepage ellipsis

agduncan94:
Max width for two sidebars, also ellipsis for my entries and my orgs.
Screen Shot 2019-12-13 at 9 15 46 AM

On smaller screens right two columns become one column.
Screen Shot 2019-12-13 at 9 16 08 AM

### DIFF
--- a/src/app/home-page/home-logged-in/home-logged-in.component.scss
+++ b/src/app/home-page/home-logged-in/home-logged-in.component.scss
@@ -1,6 +1,6 @@
 .widget-section {
   min-height: 300px;
-  // width: 100%;
+  width: 100%;
 }
 
 .twitter {
--- a/src/app/home-page/home-logged-in/home-logged-in.component.scss
+++ b/src/app/home-page/home-logged-in/home-logged-in.component.scss
@@ -1,6 +1,6 @@
 .widget-section {
   min-height: 300px;
-  width: 100%;
+  // width: 100%;
 }
 
 .twitter {
@@ -13,13 +13,14 @@ mat-sidenav-container {
   height: 100%;
 }
 
-.side-column {
+.left-side-column {
   background-color: #f6f8fa;
 }
 
 @media screen and (min-width: 992px) {
   .side-column {
     max-width: 350px !important;
+    min-width: 350px !important;
   }
 }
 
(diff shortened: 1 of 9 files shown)


src/app/home-page/home-logged-in/home-logged-in.component.scss
```diff
@@ -13,10 +13,17 @@ mat-sidenav-container {
...
 }
 
-.side-column {
+.left-side-column {
   background-color: #f6f8fa;
 }
 
+@media screen and (min-width: 992px) {
+  .side-column {
+    max-width: 350px !important;
```
garyluu: If it's max 350 and min 350, that's pretty much just width: 350px.  Additionally, I thought we banned pixels.

src/app/home-page/widget/entries/entries.component.scss
```diff
@@ -0,0 +1,11 @@
+.entry-item {
+  white-space: nowrap;
```
garyluu: This is used so often that we should put this somewhere general

src/app/home-page/home-logged-in/home-logged-in.component.scss
agduncan94: What was the replacement to px?

denis-yuen:
(passing build)

src/app/home-page/home-logged-in/home-logged-in.component.scss
agduncan94: Used REM

agduncan94:
@coverbeck Do you think simply the repository + optional entry name is enough to identify?

agduncan94:
I am in favour of a more human readable name that can be displayed around the site, though I am not sure if they lack of full path will confuse people.

denis-yuen:
> I am in favour of a more human readable name that can be displayed around the site, though I am not sure if they lack of full path will confuse people.

Depends on the context. 
One concern that users had was that one could intentionally make a workflow that is named similarly to another one. That's why we added namespace to workflows in search (i.e. two workflows with the same name next to each other in search)

coverbeck:
> @coverbeck Do you think simply the repository + optional entry name is enough to identify?

Not sure what the answer is at this point; just wanted to point out what I felt what is a small problem.

PR #848, opened 2019-12-16
Add tooltip to disabled slider

garyluu:
For dockstore/dockstore#3052

### DIFF
--- a/src/app/shared/entry-wizard/entry-wizard.component.html
+++ b/src/app/shared/entry-wizard/entry-wizard.component.html
@@ -30,7 +30,7 @@
           [name]="repo.repositoryName"
           [checked]="repo.present"
           [disabled]="!repo.canDelete && repo.present"
-          [matTooltip]="!repo.canDelete && repo.present ? 'This workflow is full so it cannot be deleted.' : ''"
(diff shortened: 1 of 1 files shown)


src/app/shared/entry-wizard/entry-wizard.component.html
```diff
@@ -30,6 +30,7 @@
           [name]="repo.repositoryName"
           [checked]="repo.present"
           [disabled]="!repo.canDelete && repo.present"
+          [matTooltip]="!repo.canDelete && repo.present ? 'This workflow is full so it cannot be deleted.' : ''"
```
coverbeck: I don't understand the tool tip. What is a full workflow?

garyluu: No idea, just wrote what Andrew put in the issue.

denis-yuen: The description we give users is near the bottom of https://docs.dockstore.org/en/develop/getting-started/dockstore-workflows.html#create-workflow-stubs-from-github-bitbucket-and-gitlab

agduncan94: A full workflow is simply not a stub, though in the future ideally there will be no stubs.

PR #850, opened 2019-12-16
Feature/585/notifications ui

ByteMap:
SEAB#585

Implemented Akita store to retrieve and store notifications from
the webservice

Implemented an interface to store notification id and expiration of
dismissed notifications

Displayed all notifications contained in the Observable,
allNotifications$, that was fetched from the notifications query

Created a method to store the notification id and expiration that is
contained within the created interface when the dismissed button
is pressed

Removed any outdated notifications in the localstorage upon
initialization

Created unit tests for notifications.component.ts,
notification.store.ts, notification.query.ts,and notification.service.ts

Created CircleCI integration tests and a fixture for the tests

Package.json has its webservice_version currently pointing to 
feature/SEAB-563/notifications-backend instead of 1.8.0-alpha.1
so the automated CircleCI tests will not pass. However, the tests will
pass if they were ran locally


cypress/integration/immutableDatabaseTests/notifications.ts
```diff
@@ -0,0 +1,97 @@
...
+        cy.route({
+          method: 'POST',
+          url: '/*',
+          response: json
+        });
+      });
+
+      verifyNotificationExist();
+
+      cy.visit('/search');
```
garyluu: array and loop through here and below

src/app/notifications/notifications.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  public dismissedNotifications: Array<DismissedNotification> = JSON.parse(localStorage.getItem(this.storageKey) || '[]');
+
+  ngOnInit() {
+    this.notificationsService.getNotifications();
+    this.allNotifications$ = this.notificationsQuery.allNotifications$;
+    this.removeExpiredDisabledNotifications(this.dismissedNotifications);
+  }
+
+  removeExpiredDisabledNotifications(notificationsStored: Array<DismissedNotification>) {
+    const today = formatDate(new Date(), 'yyyy-MM-dd', 'en');
```
coverbeck: It would be simpler and quicker just to compare numbers. And it won't do any rounding off (your code only compares down to the day).

```
const now = Date.now()
this.dismissedNotifications = this.dismissedNotifications.filter(notification => notification.expiration > now)
```

cypress/integration/immutableDatabaseTests/notifications.ts
```diff
@@ -0,0 +1,97 @@
+ /*
+ *    Copyright 2019 OICR
```
denis-yuen: It's 2020, this should maybe also include UCSC. But does it even matter with open-source? That's probably something for other people to look into, just the 2020 bit for now.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "1.8.0-alpha.1"
+    "webservice_version": "feature/SEAB-563/notifications-backend"
```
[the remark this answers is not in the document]
denis-yuen: beta.4 even

src/app/notifications/notifications.component.css
```diff
@@ -0,0 +1,19 @@
+.notification-bar {
+  background-color: #3b4045;
+  width: 100%;
+}
+
+.LOW {
```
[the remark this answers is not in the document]
denis-yuen: That might be my fault, when I saw the rendered example, it seemed a bit much.  
I think HIGH should be deleted from the backend and LOW, MEDIUM, CRITICAL seems enough (or alternatively L,M,H)

src/app/notifications/notifications.component.html
```diff
@@ -0,0 +1,10 @@
+<div *ngFor="let notification of allNotifications$ | async">
+  <div class="container text-center notification-bar {{ notification.priority }}">
+    <div class="container-fluid text-center" *ngIf="!notificationDismissed(notification)" data-cy="notification-banner">
+      {{ notification.message }}
```
coverbeck: You should use the markdown component so that any markdown in the message will be rendered correctly.

src/app/notifications/notifications.component.html
```diff
@@ -0,0 +1,10 @@
+<div *ngFor="let notification of allNotifications$ | async">
+  <div class="container text-center notification-bar {{ notification.priority }}">
```
Ldcabansay: Since we're using the notifications endpoint for both sitewide banner stuff and news/updates pieces, you should filter the notifications by type.  After you loop through with ngFor, you can add an ngIf to the next div to filter out by type.

src/app/notifications/notifications.component.css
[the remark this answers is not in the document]
coverbeck: I logged an issue for the backend for this: https://github.com/dockstore/dockstore/issues/3109

src/app/notifications/notifications.component.html
coverbeck: Alternatively, you could have `bannerNotifications$` in notifications.query.ts and then you don't need the filtering logic in the HTML.

ByteMap: Added a *ngIf to check if the notification is SITEWIDE. For now, if the notification is not SITEWIDE, then it does nothing.

agduncan94:
@ByteMap The package json is referencing a branch that doesn't exist. It should be changed to a release that has the notifications backend.

src/app/notifications/notifications.component.html
ByteMap: Done.

src/app/notifications/notifications.component.html
```diff
@@ -0,0 +1,16 @@
+<div *ngFor="let notification of allNotifications$ | async">
```
garyluu: should be `notDismissedNotifications` instead of `allNotifications` here because your `*ngIf` down there is currently being executed a bajillion times

src/app/notifications/notifications.component.ts
```diff
@@ -0,0 +1,48 @@
...
+interface DismissedNotification {
+  id: number;
+  expiration: Date;
+}
+
+@Component({
+  selector: 'app-notifications',
+  templateUrl: './notifications.component.html',
+  styleUrls: ['./notifications.component.css'],
+  encapsulation: ViewEncapsulation.None
```
coverbeck: This was my suggestion as we were trying to figure out why you're styling wasn't working for the generated HTML, but I think you should remove it, and instead move the paragraph styling to the global css. The problem with that all of your component's CSS may "pollute" things.

@garyluu Don't know if you have a better idea on how to approach this? The issue is that the `<markdown>` component converts the markdown to a `<p>`, and because that paragraph element is not in the template, the component CSS doesn't affect it.

@garyluu

src/app/notifications/notifications.component.html
```diff
@@ -0,0 +1,16 @@
...
+  <div class="container text-center notification-bar {{ notification.priority?.toLowerCase() }}">
+    <div *ngIf="notification.type == 'SITEWIDE'">
+      <div class="container-fluid text-center" *ngIf="!notificationDismissed(notification)" data-cy="notification-banner">
```
coverbeck: To elaborate on Gary's comment, it's generally not good to have a function invocation in an ngIf, because it constantly gets evaluated.

garyluu: https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Onboarding#pitfalls-and-best-practices  number 3

src/app/notifications/notifications.component.ts
ByteMap: I have followed your suggestion of moving the paragraph styling to the global scss file and removed the encapsulation line.

coverbeck:
Can you please attach a screenshot with a notification.

src/app/notifications/notifications.component.ts
```diff
@@ -0,0 +1,49 @@
...
+
+interface DismissedNotification {
+  id: number;
+  expiration: Date;
+}
+
+@Component({
+  selector: 'app-notifications',
+  templateUrl: './notifications.component.html',
+  styleUrls: ['./notifications.component.css']
```
coverbeck: Ideally this would be scss instead of css.

src/styles.scss
```diff
@@ -1292,6 +1292,11 @@ mat-icon.star {
   text-overflow: ellipsis;
 }
 
+.notification-display p {
+  text-align: center;
+  font-size: 18px;
```
coverbeck: When we were looking at this, wasn't there an issue with margin-bottom? Do you need to add something for that?

src/app/test/mocked-objects.ts
```diff
@@ -713,3 +714,19 @@ export const validTool: ExtendedDockstoreTool = {
...
+  priority: 'LOW',
+  expiration: null
+};
+
+export const expiredMockNotification: Notification = {
+  id: 121,
+  message: 'Testing123',
+  type: 'SITEWIDE',
+  priority: 'LOW',
+  expiration: new Date('2018-11-25T00:00:00')
```
coverbeck: @garyluu Confused by what is going on; when we debugged this with the UI invoking the web service API, the expiration was a number, the timestamp. But the generated Swagger code has the type as a Date. I think for the tests this should be a number, but then the compile complains.

Have you run into this?

garyluu: You can probably do new Date(29301802481203)

garyluu: Doesn't sound too familiar

denis-yuen:
(thought I heard in stand-up, Louise was going to try to help with this)

Ldcabansay:
Current state
image

notifications_banner_ui

src/app/test/mocked-objects.ts
coverbeck: I logged this in #3127

src/styles.scss
```diff
@@ -1292,7 +1292,26 @@ mat-icon.star {
...
+
+.notification-display p {
+  text-align: center;
+  display: inline-block;
+}
+
+//removes the 10px margin added to p element on markdown component
+.markdown-remove-pmargin p {
+  margin-bottom: 0px;
+}
```
garyluu: Is this duplicated?

Ldcabansay: yes 😢

src/app/notifications/notifications.component.html
```diff
@@ -0,0 +1,14 @@
+<div *ngFor="let notification of activeNotifications$ | async">
+  <div class="container text-center notification-bar {{ notification.priority?.toLowerCase() }}">
+    <div *ngIf="notification.type == 'SITEWIDE'">
```
garyluu: === and there's probably an enum somewhere

Ldcabansay: Yeah there's an enum, I have it set as an enum in my newsbody PR, so I'll fix it there.

PR #853, opened 2019-12-19
give better success message for saving hosted workflow versions

NatalieEO:
dockstore/dockstore#2905

### DIFF
--- a/src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
+++ b/src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
@@ -64,8 +64,12 @@ export class WorkflowFileEditorComponent extends FileEditing {
     if (!this.versions || this.versions.length === 0) {
       return true;
     }
-    const mostRecentId = this.versions.reduce((max, n) => Math.max(max, n.id), this.versions[0].id);
+    const mostRecentId = this.getIdOfNewestVersion(this.versions);
     return this._selectedVersion.id === mostRecentId;
+  }
+
+  getIdOfNewestVersion(versions: Array<WorkflowVersion>): number {
+    return versions.reduce((max, n) => Math.max(max, n.id), this.versions[0].id);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
```diff
@@ -107,7 +107,14 @@ export class WorkflowFileEditorComponent extends FileEditing {
           this.workflowsService.getWorkflow(editedWorkflow.id).subscribe(
             (newlyGottenWorkflow: Workflow) => {
               this.workflowService.setWorkflow(newlyGottenWorkflow);
-              this.alertService.detailedSuccess();
+              const mostRecentId = newlyGottenWorkflow.workflowVersions.reduce(
+                (max, n) => Math.max(max, n.id),
+                newlyGottenWorkflow.workflowVersions[0].id
+              );
+              const updatedVersion = newlyGottenWorkflow.workflowVersions.find(v => v.id === mostRecentId);
```
agduncan94: You can also do something like stream().max(Comparator.comparing(WorkflowVersion::getId)) to get the version with the max id. Not sure if it is any better.

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
```diff
@@ -107,7 +107,14 @@ export class WorkflowFileEditorComponent extends FileEditing {
           this.workflowsService.getWorkflow(editedWorkflow.id).subscribe(
             (newlyGottenWorkflow: Workflow) => {
               this.workflowService.setWorkflow(newlyGottenWorkflow);
-              this.alertService.detailedSuccess();
+              const mostRecentId = newlyGottenWorkflow.workflowVersions.reduce(
```
coverbeck: This `reduce` looks the same as the code on line 67. Could you create a common function?

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
coverbeck: I think that's Java :) Is there a corresponding TypeScript way?

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
```diff
@@ -64,10 +64,14 @@ export class WorkflowFileEditorComponent extends FileEditing {
     if (!this.versions || this.versions.length === 0) {
       return true;
     }
-    const mostRecentId = this.versions.reduce((max, n) => Math.max(max, n.id), this.versions[0].id);
+    const mostRecentId = this.getIdOfNewestVersion(this.versions);
     return this._selectedVersion.id === mostRecentId;
   }
 
+  getIdOfNewestVersion(versions: Array<WorkflowVersion>): number {
+    return versions.reduce((max, n) => Math.max(max, n.id), this.versions[0].id);
```
coverbeck: This code assumes at least one version. I guess you're safe, since the previous invocation checks for it, and the new invocation will be invoked only after a successful save.

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
agduncan94: :man_facepalming: Still adjusting back to timezone

PR #854, opened 2019-12-19
Feature/remove circular dependency

agduncan94:
https://github.com/dockstore/dockstore/issues/3063

Now on modal close if true is returned when registering workflows, a fetch of all workflows will be done.

### DIFF
--- a/src/app/myworkflows/myworkflows.service.ts
+++ b/src/app/myworkflows/myworkflows.service.ts
@@ -106,7 +106,7 @@ export class MyWorkflowsService extends MyEntriesService<Workflow, OrgWorkflowOb
         .subscribe(reloadEntries => {
           if (reloadEntries) {
             const user = this.userQuery.getValue().user;
-            if (user && entryType) {
+            if (user) {
               this.getMyEntries(user.id, entryType);
             }
           }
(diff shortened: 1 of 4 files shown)


src/app/myworkflows/myworkflows.service.ts
```diff
@@ -99,7 +99,18 @@ export class MyWorkflowsService extends MyEntriesService<Workflow, OrgWorkflowOb
...
     if (entryType === EntryType.BioWorkflow) {
-      this.matDialog.open(RegisterWorkflowModalComponent, { width: '600px' });
+      const dialogRef = this.matDialog.open(RegisterWorkflowModalComponent, { width: '600px' });
+      dialogRef
+        .afterClosed()
+        .pipe(takeUntil(this.ngUnsubscribe))
+        .subscribe(reloadEntries => {
+          if (reloadEntries) {
+            const user = this.userQuery.getValue().user;
+            if (user && entryType) {
```
coverbeck: You already verified that entryType is not null on line 101.

src/app/myworkflows/myworkflows.service.ts
```diff
@@ -99,7 +99,18 @@ export class MyWorkflowsService extends MyEntriesService<Workflow, OrgWorkflowOb
 
   registerEntry(entryType: EntryType | null) {
     if (entryType === EntryType.BioWorkflow) {
-      this.matDialog.open(RegisterWorkflowModalComponent, { width: '600px' });
+      const dialogRef = this.matDialog.open(RegisterWorkflowModalComponent, { width: '600px' });
+      dialogRef
+        .afterClosed()
+        .pipe(takeUntil(this.ngUnsubscribe))
```
garyluu: This service is a singleton, it's never destroyed. So this subscription is never closed. I believe this causes a memory leak. In general, there should never be subscriptions within services (except the HTTP ones).

garyluu: Apparently, it auto closes: https://stackoverflow.com/a/54923792

Just remove the pipe take until.  also remove `extends Base` from the service as it's a trap

PR #855, opened 2019-12-20
Update karma and @angular-devkit/build-angular

garyluu:
Fix some GitHub security alerts.

Just updated those 2 to latest


garyluu:
> Lots of changes in package-lock.json, confirm intended?

Should be, it gets cascaded down

PR #856, opened 2020-01-03
Use fixed tweet limit to avoid constant polling

garyluu:
For dockstore/dockstore#1630

Can't seem to stop polling on other pages, disabling polling altogether by grabbing a fixed amount of tweets.

Functionality lost:
- height is no longer fixed, no scrollbar, height auto-adjusts based on the height of the specified amount of tweets
- can no longer view more tweets when scrolling down


denis-yuen:
:thinking:

garyluu:
Logged out
image

Logged in
image

PR #862, opened 2020-01-07
search page tool tips

NatalieEO:
dockstore/dockstore#2962
Screen Shot 2020-01-07 at 2 26 25 PM

### DIFF
--- a/src/app/search/state/search.service.ts
+++ b/src/app/search/state/search.service.ts
@@ -484,6 +484,8 @@ export class SearchService {
 
   initializeToolTips() {
     return new Map([
+      // Git hook auto fixes from single quotes with an escaped 's but linter complains about double quotes.
+      /* tslint:disable-next-line:quotemark*/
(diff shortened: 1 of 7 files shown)


src/app/search/state/search.service.ts
```diff
@@ -482,6 +482,20 @@ export class SearchService {
     ]);
   }
 
+  initializeToolTips() {
+    return new Map([
+      ['private_access', "A private tool requires authentication to view on Docker's registry website and to pull the Docker image."],
+      ['verified', 'Indicates that at least one version of a tool or workflow has been successfuly ran by our team or an outside party.'],
```
coverbeck: run instead of ran

src/app/test/service-stubs.ts
```diff
@@ -190,6 +191,22 @@ export class SearchStubService {
     ]);
   }
 
+  initializeToolTips() {
```
coverbeck: It's unfortunate this has to be duplicated, but I see code was like that already and probably beyond the scope of this PR.

PR #863, opened 2020-01-08
Feature/2966/back button fix

garyluu:
Data must be set after pagination it seems.

dockstore/dockstore#2966

Quote from some other guy
"
This took me hours to finally track down and understand why my table wasn’t working. Placing some console.logs() helped me figure out the order of events and why it didn’t work consistently. My scenario was similar to the original issue above with a dynamic data source, but slightly different. For me, when I would first refresh my angular app, the paginator state would be set and then the data for my table would be set. When these events would happen in the this order the paginator worked just as expected.

Because I was using ReplaySubjects for getting the data for my tables, my paginator state would be set in ngAfterViewInit and then the table data would come in from my subscription (it depends on user ID so I don’t have an initial value and that is why I didn’t use BehaviorSubjects). My problem was that, when I would navigate to another component in my app, and return to my dynamic table data source, the table data would be set before the paginator state. This would make the paginator show page x, but the displayed data would always be the first page of data.
"

### DIFF
--- a/cypress/integration/manualTests/search.ts
+++ b/cypress/integration/manualTests/search.ts
@@ -52,13 +52,12 @@ describe('Admin UI', () => {
       cy.get('a')
         .contains('Organizations')
         .click();
-      cy.get('a')
-        .contains('Search')
-        .click();
+      cy.go('back');
       cy.contains('21 - 40');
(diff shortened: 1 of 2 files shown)


agduncan94:
If possible would be interesting to find out exactly why this fixes it, maybe it will help with other search issues.

PR #867, opened 2020-01-08
my org tooltip is now the display name

agduncan94:
My org tooltip is now the same as the text shown.

### DIFF
--- a/src/app/home-page/widget/organizations/organizations.component.html
+++ b/src/app/home-page/widget/organizations/organizations.component.html
@@ -16,7 +16,7 @@
     </mat-form-field>
 
     <div *ngFor="let organization of myItems" class="w-100 ellipsis-overflow">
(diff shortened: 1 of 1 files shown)


src/app/home-page/widget/organizations/organizations.component.html
```diff
@@ -16,7 +16,7 @@ <h3>My Organizations</h3>
     </mat-form-field>
 
     <div *ngFor="let organization of myItems" class="w-100 ellipsis-overflow">
-      <span [matTooltip]="organization.name">
+      <span [matTooltip]="organization.displayName">
```
coverbeck: Is displayName always set? If not, should it be organization.displayName || organization.name?

agduncan94: It is a required field so it will always be set.

PR #870, opened 2020-01-08
discourse is now forum; help desk added; no bullets

agduncan94:
* discourse link now says forum
* added a help desk link
* no more list bullets
no-icons

Open to critique on the list bullet/no bullet.


denis-yuen:
> Aww, "View Requests" is all lonely now

It does look a bit weird

agduncan94:
I think the whitespace in the image plus the whitespace on github makes it look a lot worse than it actually is.

PR #871, opened 2020-01-09
facet buttons

NatalieEO:
dockstore/dockstore#2963
Current. Button defaults to count. 
Screen Shot 2020-01-09 at 9 05 45 AM
Option 1
Screen Shot 2020-01-09 at 9 08 11 AM
Option 2
Screen Shot 2020-01-09 at 9 08 58 AM
Option 3
Screen Shot 2020-01-09 at 1 02 16 PM
Option 4
Screen Shot 2020-01-09 at 2 01 09 PM
Option 5
Screen Shot 2020-01-09 at 1 33 44 PM

### DIFF
--- a/src/app/search/search.component.scss
+++ b/src/app/search/search.component.scss
@@ -87,8 +87,8 @@
   font-size: 12px;
   background-color: #ffffff;
   &.active {
-    background-color: #f4f4f4;
-    color: #3b769a;
+    background-color: #487980;
+    color: #ffffff;
     font-weight: bold;
   }
   &:not(.active) {
--- a/src/app/search/search.component.scss
+++ b/src/app/search/search.component.scss
@@ -89,10 +89,14 @@
   &.active {
     background-color: #f4f4f4;
     color: #3b769a;
+    font-weight: bold;
   }
   &:not(.active) {
     &:hover {
       background-color: #f8f8f8;
+    }
+    .ng-fa-icon {
+      display: none;
     }
   }
 }


coverbeck:
Are we using standard Material colors for this? I ask because the active tab has always looked like it's disabled/not active to me, due to the background color. Does anybody else see that?

It distorts things enough for me that I'm not sure I can pick either of your options. :)

NatalieEO:
> Are we using standard Material colors for this? I ask because the active tab has always looked like it's disabled/not active to me, due to the background color. Does anybody else see that?
> 
> It distorts things enough for me that I'm not sure I can pick either of your options. :)

After looking at some examples, it seems like they generally grey out the not active option or underline and or highlight with a vibrant color the active option. So yeah, can see how the grey is confusing. Can change to try those out and provide more options.

Ldcabansay:
I like option 3

garyluu:
Was also here, as long as "Count" is actually above the count, i'm ok with it.

coverbeck:
I lean towards 3, I still feel like it should be something as shown in [material guidlines](https://material.io/components/tabs/#states), where the both the background and text for the non-active tab are darker shade of the active tab. But if nobody else feels strongly, I won't belabor the point.

agduncan94:
Option 3 or 4

NatalieEO:
Option 3

denis-yuen:
For the record, I'm actually :man_shrugging:

garyluu:
There other option is just have a sort by dropdown :smile: 

https://i.stack.imgur.com/a5kAB.png

PR #873, opened 2020-01-09
on medium and smaller screens change to rows

agduncan94:
https://github.com/dockstore/dockstore/issues/2432
smaller-screen-orgs

Could not figure out how to deal with overflow :cry:. I think that the current design is bad anyways (I designed it) so I moved it to using rows instead of columns (on screens below large).


agduncan94:
@NatalieEO I agree. I think the other information can be grouped together, and then stars could be somewhere on its own.

PR #880, opened 2020-01-10
Set user to be initially undefined

garyluu:
So the problem is...

Initially user is null (logged out home page appears)
User is gotten by webservice (logged out home page still shown)
Then user gets set (logged in home page appears)

Changed it so that `user` is initially undefined. This is interesting because async piping in an ngIfElse statement appears to cause neither branches to execute. So in this fix, nothing appears until user is set (either null or actually something). 

End result is it's only set once and logged in homepage or logged out homepage only appears once.

Kind of a dangerous change, leaving out of 1.8.0.

For https://github.com/dockstore/dockstore/issues/3151


agduncan94:
@garyluu does this solve https://github.com/dockstore/dockstore/issues/3121 point 6?

garyluu:
It does, the other way would be to use the loading thing instead (which would probably be less disruptive)

coverbeck:
This is fixing dockstore/dockstore#3121, point 6, right? But that ticket is marked as closed. Confused...

denis-yuen (approved):
Seems worth a try in 1.11

garyluu:
> This is fixing [dockstore/dockstore#3121](https://github.com/dockstore/dockstore/issues/3121), point 6, right? But that ticket is marked as closed. Confused...

Oh ya, I guess there was a ticket for this after all.

NatalieEO (approved):

coverbeck (approved):
Still a little confused why we need this if the related issue has already been fixed, but we can try it and see.

garyluu:
> Still a little confused why we need this if the related issue has already been fixed, but we can try it and see.

It has not been fixed. notice how the last checkbox was not checked off

PR #881, opened 2020-01-13
footer is column on small screens and below

agduncan94:
On small screens (phones) the footer content switches from rows to column.
Previously
row-footer

Now
column-footer


garyluu:
Doesn't seem much different
image

PR #892, opened 2020-01-17
Removed ngx-bootstrap from search page tabs

ByteMap:
dockstore/dockstore#2747

Re-purposed ticket to simply removing ngx-bootstrap from search table
as removing ngx-bootstrap from package.json will affect over 20 files

Migrated from using ngx-bootstrap to material for the search page
tabs

Removed 'tabset.homeComponent' from styles.scss as it is no longer
necessary

Removed fields 'browseToolsTab' and 'browseWorkflowsTab' from
home.component.ts as they are no longer necessary

Created a new variable, selectedIndex, that gets the index of the active
tab on initialization

Added styling in search-results.component.scss to match the styling of
the search table before ngx-bootstrap was removed as closely as possible


coverbeck:
Please attach screen shot(s)

src/app/search/search-results/search-results.component.ts
```diff
@@ -52,6 +54,7 @@ export class SearchResultsComponent implements OnInit {
   ngOnInit() {
     this.createTagCloud('tool');
     this.createTagCloud('workflow');
+    this.activeToolTab$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(toolsActive => (this.selectedIndex = toolsActive ? 0 : 1));
```
coverbeck: Avoid "magic" numbers -- create const's like TOOLS_TAB_INDEX and WORKFLOWS_TAB_INDEX and use them here and on line 37 instead of 0 and 1.

src/app/search/search-results/search-results.component.scss
```diff
@@ -14,6 +14,19 @@
  *    limitations under the License.
  */
 
+mat-tab-group.homeComponent {
+  border: 1px solid #cfd1d7;
+  border-bottom-right-radius: 15px;
```
coverbeck: Avoid pixels in css. See [10th item or so](https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide).

ByteMap:
Screenshots of how the new search table would look like.
Screen Shot 2020-01-17 at 11 35 55 AM
Screen Shot 2020-01-17 at 11 36 07 AM

src/app/search/search-results/search-results.component.scss
ByteMap: Done.

src/app/search/search-results/search-results.component.ts
ByteMap: Done.

ByteMap:
New screenshots of how the search table looks
Screen Shot 2020-01-21 at 3 58 54 PM
Screen Shot 2020-01-21 at 3 59 09 PM

coverbeck:
Can you please attach an updated screenshot? Curious if others like the removal of "Browse" from the tabs, which I suggested because it seemed repetitive.

coverbeck:
> Can you please attach an updated screenshot? Curious if others like the removal of "Browse" from the tabs, which I suggested because it seemed repetitive.

Oops, I see it is there already. As we just discussed, it would be nice if there were some extra space between the icons and "Tools" and "Workflows".

denis-yuen:
> Can you please attach an updated screenshot? Curious if others like the removal of "Browse" from the tabs, which I suggested because it seemed repetitive.

Sleek

denis-yuen:
Pending success on builds

src/app/search/search-results/search-results.component.ts
```diff
@@ -42,16 +43,23 @@ export class SearchResultsComponent implements OnInit {
...
 
+  private readonly TOOLS_TAB_INDEX = 0;
+  private readonly WORKFLOWS_TAB_INDEX = 1;
+
   ngOnInit() {
     this.createTagCloud('tool');
     this.createTagCloud('workflow');
+    this.activeToolTab$
+      .pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe(toolsActive => (this.selectedIndex = toolsActive ? this.TOOLS_TAB_INDEX : this.WORKFLOWS_TAB_INDEX));
```
garyluu: Trying making selectedIndex an observable and pipe mapping activeToolTab to it.  This should let you get rid of the extends Base unsubscribe hassle.

src/app/search/search-results/search-results.component.scss
```diff
@@ -14,6 +14,23 @@
...
  */
 
+mat-tab-group.homeComponent {
+  border: 0.0625rem solid #cfd1d7;
+  border-bottom-right-radius: 1rem;
+  border-bottom-left-radius: 1rem;
+}
+
+::ng-deep .mat-tab-header-pagination {
+  display: none !important;
```
garyluu: See if you can get rid of the ng-deep and !important.

src/app/search/search-results/search-results.component.ts
coverbeck: The problem we were running into was this HTML snippet: `[(selectedIndex)]="selectedIndex"` -- and Angular doesn't let you specify an observable nor use async there.

Can you clarify how to get around that?

garyluu: `[(selectedIndex)]` => `[seletedIndex]`

src/app/search/search-results/search-results.component.scss
ByteMap: ng-deep and !important was the better workaround that I can find in removing the '< >' on the side of the tab headers. I have attached a screenshot to better depict the issue. The issue only shows up when the user is resizing the browser window vertically. 
Screen Shot 2020-01-22 at 1 12 04 PM

The other workaround would be to insert:
`document.getElementsByClassName('mat-tab-header-pagination-before')[0].remove();
  document.getElementsByClassName('mat-tab-header-pagination-after')[0].remove();`
into ngOnInit()

Apparently, a user in stackoverflow said that this is a known issue that exists in angular material tabs.
https://stackoverflow.com/questions/57888846/set-horizontal-scroll-in-angular-material-tabs

garyluu: Stick:
```
mat-tab-group.no-pagination {
  .mat-tab-header-pagination {
    display: none;
  }
}
```

garyluu:
Ugh, my prettier may have caused some conflicts

src/app/search/search-results/search-results.component.scss
ByteMap: Done.

src/app/search/search-results/search-results.component.ts
ByteMap: Made selectedIndex an observable, then pipe mapped activeToolTab to it.

src/app/search/state/search.query.ts
```diff
@@ -68,11 +70,20 @@ export class SearchQuery extends Query<SearchState> {
       return true;
     } else if (workflows.length === 0 && param === 'workflow') {
       return false;
+    } else if (workflows.length > 0 && tools.length > 0) {
+      this.savedTabIndex$.subscribe(index => {
```
garyluu: Don't need to subscribe, you can do this.getValue().currentTabIndex

src/app/search/state/search.store.ts
```diff
@@ -31,6 +31,7 @@ export interface SearchState {
   pageIndex: number;
   advancedSearch: AdvancedSearchObject;
   showModal: boolean;
+  currentTabIndex: boolean;
```
garyluu: Might be easier to just set the index as...the index (number)

src/app/search/state/search.query.ts
ByteMap: Done, thanks for the tip!

src/app/search/search-results/search-results.component.ts
```diff
@@ -20,28 +20,31 @@ import { ELASTIC_SEARCH_CLIENT } from '../elastic-search-client';
...
-  public browseToolsTab = 'browseToolsTab';
-  public browseWorkflowsTab = 'browseWorkflowsTab';
-  public activeToolTab$: Observable<boolean>;
+export class SearchResultsComponent extends Base implements OnInit {
+  public activeToolTab$: Observable<number>;
   public noToolHits$: Observable<boolean>;
   public noWorkflowHits$: Observable<boolean>;
   public showWorkflowTagCloud$: Observable<boolean>;
   public showToolTagCloud$: Observable<boolean>;
+  public selectedIndex$: Observable<any>;
```
coverbeck: This would ideally not be `any`. Or were you forced to do it due to some compiler issues?

src/app/search/search-results/search-results.component.ts
```diff
@@ -52,6 +55,11 @@ export class SearchResultsComponent implements OnInit {
   ngOnInit() {
     this.createTagCloud('tool');
     this.createTagCloud('workflow');
+    this.selectedIndex$ = this.searchQuery.activeToolTab$.pipe(
+      map(activeToolTab => {
```
coverbeck: This is stylistic that I don't think we have an official rule on, so feel free to ignore, but I prefer to use curly braces only when you have statements. You could have instead, if I'm not getting confused with Java, something like:

```
map(activeToolTab => {active: activeToolTab})
```

src/app/search/search-results/search-results.component.ts
ByteMap: IntelliJ would complain if I got rid of any as selectedIndex$ is type Observable<{active: number;}>. I believe that it might be best to leave it as type any for now.

src/app/search/search-results/search-results.component.ts
ByteMap: Ive tried getting rid of the return statement, but that resulted in the label active being unnecessary and unused, which caused some errors.

src/app/search/search-results/search-results.component.ts
coverbeck: Hmm, that works for me. Where did it complain?

ByteMap: Line 58. Once you change the type from any to number or anything else, it will complain.

coverbeck: I tried it and did a build, and it worked. Might be an IDE configuration thing.

PR #895, opened 2020-01-21
Feature/versions table

garyluu:
This were mostly done during the visit to UCSC based on what people mentioned.

For partial dockstore/dockstore#3075 and dockstore/dockstore#3081

image

Just for workflows for now

image

TODO: Need to fix styling.  Need to figure out a way to handle the first column being too long.  Scroll? Truncate + ellipses? Wrap?
TODO: Tooltips for disabled actions dropdown

Also, I'm having some incredibly hard time cancelling a mat-menu. Any idea anyone?

Yes, I know I misspelt Angular in the comments, will fix it along with more PR changes. 

Hide whitespace changes for review.


src/app/workflow/versions/versions.component.html
```diff
@@ -15,170 +15,164 @@
...
-            <mat-icon *ngIf="!version.valid">close</mat-icon>
-          </td>
+        <a
+          href="https://doi.org/{{ version.doiURL }}"
+          *ngIf="version.doiURL"
+          data-cy="workflow-version-DOI-badge"
+          matTooltip="Click to view this version's DOI entry information."
+          target="_blank"
+          rel="noopener noreferrer"
+          ><img src="{{ zenodoUrl }}/badge/DOI/{{ version.doiURL }}.svg" alt="DOI"
```
coverbeck: It was like this and super minor and I wouldn't necessarily bother, but I think the alt should contain the DOI URL as well.

src/app/workflow/versions/versions.component.html
```diff
@@ -15,170 +15,164 @@
...
-        <tr
-          (click)="setVersion(version)"
-          *ngFor="let version of versions | orderBy: convertSorting(workflow?.mode)"
-          [ngClass]="{ info: _selectedVersion != null && version != null && _selectedVersion.id == version.id }"
+    <ng-container matColumnDef="snapshot">
+      <th
+        scope="col"
+        mat-header-cell
+        *matHeaderCellDef
+        matTooltip="The metadata and descriptors for a version that has been snapshotted will not change over time."
```
coverbeck: It was also like this, but the metadata, or at least some of it, can change after it's been snapshotted.

cypress/integration/immutableDatabaseTests/services.ts
```diff
@@ -13,13 +13,7 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-import {
```
denis-yuen: Shouldn't this kind of thing be caught by the git hook prettier?

cypress/integration/immutableDatabaseTests/services.ts
```diff
@@ -153,7 +151,7 @@ describe('Dockstore Home', () => {
     cy.contains('tr', 'Git Reference');
     cy.contains('td', '1.3');
     cy.contains('tr', 'Date Modified');
-    cy.contains('td', 'Jul 19, 2019, 1:13:48 PM');
```
denis-yuen: I think we can lose the seconds, but lets keep minutes. 
When debugging, I did often upload a few versions rapidly and it can be good to distinguish between them. 
Also while at it, I prefer YYYY-MM-DD

src/app/workflow/versions/versions.component.html
```diff
@@ -15,170 +15,164 @@
...
-            <fa-icon class="pull-right" [icon]="getIconClass('last_modified')" (click)="clickSortColumn('last_modified')"></fa-icon>
-            <label placement="top" matTooltip="Date of last update to Git reference">Date Modified</label>
-          </th>
+    <ng-container matColumnDef="valid">
+      <th
+        scope="col"
+        mat-header-cell
+        *matHeaderCellDef
+        mat-sort-header
+        matTooltip="A version is valid if it has at least one valid descriptor file."
```
denis-yuen: This wasn't changed in this PR, but is this true anymore? 
@agduncan94  should this actually be "A version is valid if it has no validation messages"?

src/app/workflow/versions/versions.component.html
```diff
@@ -15,170 +15,164 @@
...
-            <label placement="top" matTooltip="A hidden version is only visible here and not publicly.">
-              Hidden
-            </label>
-          </th>
+    <ng-container matColumnDef="verified">
+      <th
+        scope="col"
+        mat-header-cell
+        *matHeaderCellDef
+        matTooltip="A version is verified if it has been verified to work by a person/group."
```
denis-yuen: There's a `href="{{ Dockstore.DOCUMENTATION_URL }}/faq.html#what-is-a-default-version-of-a-tool-or-workflow"` above, can we get that here too but linked to https://docs.dockstore.org/en/develop/advanced-topics/verification.html

src/app/workflow/versions/versions.component.html
agduncan94: I think this sentence makes more sense for tools, since a workflow would only have one descriptor. I think this meant if a tool version had a valid CWL but invalid WDL, it would be considered valid.

Your suggestion is definitely better. Maybe something like
"A version is valid if the descriptor file(s) have been successfully validated."

cypress/integration/immutableDatabaseTests/services.ts
garyluu: Prettier currently only does the stuff in src, not cypress

denis-yuen: Can we expand it to cypress?

garyluu: Could, make an issue for it.  PR getting bloated already.

src/app/workflow/versions/versions.component.html
garyluu: Tooltips with links aren't a thing

src/app/workflow/versions/versions.component.html
```diff
@@ -15,170 +15,164 @@
...
+          >
+          <a target="_blank" rel="noopener noreferrer" [href]="version.commitID | commitUrl: workflow?.providerUrl">
+            <mat-icon
+              *ngIf="version.commitID && (version.commitID | commitUrl: workflow?.providerUrl)"
+              [matTooltip]="version.commitID ? 'Commit ID: ' + version.commitID : ''"
+              >open_in_new</mat-icon
+            ></a
+          >
+        </span>
+        <ng-template #alreadySelected>
```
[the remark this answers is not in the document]
garyluu: How do you confirm it makes a difference?

cypress/integration/immutableDatabaseTests/services.ts
denis-yuen: Good idea, can do one bit formatting commit/PR that way, please create and link here for context

src/app/workflow/versions/versions.component.html
denis-yuen: Yeah, I may not be commenting on the right line. 
But the Git Reference column uses a bubble, the verified column seems to link the checkmark. 
Let's make it consistent by using the bubble in both places and removing the link

src/app/workflow/versions/versions.component.html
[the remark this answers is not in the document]
coverbeck: Maybe the number of rows isn't significant enough to matter here. In the video, I believe he said they were looking at the amount of purple (rendering) in the performance tab.

In any case, this was just fresh on my mind after having seen the video as a possible anti-pattern, but assuming it is even true, is probably not going to matter for the number of versions we have, and given that our bottleneck is most like the time of Ajax requests, not the rendering.

cypress/integration/immutableDatabaseTests/services.ts
garyluu: dockstore/dockstore#3185

src/app/workflow/versions/versions.component.html
garyluu: Splitting this out into separate ticket because the Git Reference column's bubble has disappeared because setting default version is now done in the dropdown and not on the column anymore.  Not sure if we want the lone verified bubble anymore

garyluu: Ok, not splitting it out. It does look weird though

PR #896, opened 2020-01-21
getting started widget for logged-in homepage

Ldcabansay:
Filling up logged-in homepage with content from on-boarding until events widget is ready. 
* added getting started widget to logged-in homepage
* added minor styling changes for featured content widget

screenshot:
image

### DIFF
--- a/src/featured-content.scss
+++ b/src/featured-content.scss
@@ -9,7 +9,8 @@ The following classes are used style the featured content external html
 }
 
 .feat-content-card {
-  box-shadow: 0 0.4rem 0.8rem 0 rgba(0, 0, 0, 0.2); /* this adds the "card" effect */
+  border-radius: 0.5rem;
+  box-shadow: 0 0.4rem 0.5rem 0 rgba(0, 0, 0, 0.2); /* this adds the "card" effect */
   text-align: center;
   justify-content: center;
   min-width: 20rem;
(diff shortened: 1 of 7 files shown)


src/featured-content.scss
```diff
@@ -30,10 +31,13 @@ a.cardClick:hover {
   text-decoration: none;
 }
 
-h3.feat-content-title {
+.feat-content-title {
   color: #487980;
   margin-top: 1rem;
 }
+.feat-content-card:hover {
+  box-shadow: 0 0.4rem 0.5rem 0 rgba(0, 0, 0, 0.5);
```
garyluu: Can you explain the rationale behind adding box-shadow (aka increasing elevation) when hovering?

Ldcabansay: Just for pretty-ing, the box gets highlighted when you however. 
hover-feat-content

Ldcabansay:
> Approving because I don't care enough to request changes but... https://material.io/design/environment/elevation.html#depicting-elevation states that:
> 
> image
> 
> Rest of the link says when to use it.

But its a clickable card, the same link does the same thing on hover for their clickable cards

garyluu:
> > Approving because I don't care enough to request changes but... https://material.io/design/environment/elevation.html#depicting-elevation states that:
> > image
> > Rest of the link says when to use it.
> 
> But its a clickable card, the same link does the same thing on hover for their clickable cards

It actually doesn't, the style applied to cards is the obvious grey'ing (overlay) of the entire component (similar to hovering over most components).  The subtle elevation increase is secondary and optional to the overlay.  Hence (if you read the material design doc), it says " Elevation, combined with an overlay, can signify state on some components.".  Right now, a combination of the lack of the overlay + the already increased elevation makes it look like I can drag and drop the component on top over others.

PR #900, opened 2020-01-23
Fix PUT error

garyluu:
For dockstore/dockstore#3180

Appears to be an error when trying to PUT /containers/{containerId} 

The starredUsers property from the UI is not understood by the webservice.

Solution: Removing ALL unused properties

Also added an alert to notify users when PUT has failed.


coverbeck:
> The starredUsers property from the UI is not understood by the webservice.

I think the problem is that Swagger CodeGen does not know about the `@JsonSerialize` annotation; looking at the generated DockstoreTool.ts in the UI, it has `starredUsers?: Array<User>`, but the response from the webservice is an `Array<number>`. I will  create an issue for that.

PR #906, opened 2020-01-29
Fix merge conflicts from master

garyluu:
Weird stuff with master having extra code that 2.5.0 doesn't.  Removing that code which is old anyways.


garyluu:
Manually tested that the autocomplete still works

PR #907, opened 2020-01-31
Fix lint warnings

garyluu:
Fix lint warnings

@coverbeck See if src/app/workflow/dag/cwl-viewer/cwl-viewer.service.spec.ts is correctly fixed.


garyluu:
> cwl-viewer.service.spec.ts LGTM.
> 
> Why new Observable instead of Observable.create? They've deprecated?

Ya, most of these lint fixes are because of deprecation

PR #908, opened 2020-01-31
Update openapi generator

garyluu:
https://ucsc-cgl.atlassian.net/browse/SEAB-764

Simply updating open api generator

Skipping validation because there's multiple endpoints with the same operation id

### DIFF
--- a/scripts/generate-openapi-script.sh
+++ b/scripts/generate-openapi-script.sh
@@ -4,7 +4,7 @@ set -o pipefail
 set -o nounset
 set -o xtrace
 
-GENERATOR_VERSION="3.3.4"
+GENERATOR_VERSION="4.2.3"
 BASE_PATH="https://raw.githubusercontent.com/dockstore/dockstore/$npm_package_config_webservice_version"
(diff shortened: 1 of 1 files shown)


garyluu:
> > Skipping validation because there's multiple endpoints with the same operation id
> 
> this seems bad to me, in the webservice, swagger codegen should be using operationid to name methods. If there are duplicates, this should affect that as well.
> 
> Feel free to open a new ticket, but let's get unique operationids first then update this without disabling the validation

Multiple tickets were already opened. The previous openapi-generator version didn't do any validation.  Nothing is lost here with this update. Sitting on this PR for an extended period of time doesn't add any value.

denis-yuen:
> Multiple tickets were already opened. The previous openapi-generator version didn't do any validation. Nothing is lost here with this update. Sitting on this PR for an extended period of time doesn't add any value.

I find the PR a useful reminder and the PR isn't linked to anything.
 
Looking at your validation results it looks like the validation errors for openapi3 are fairly simple, let's just get them fixed so we don't have to turn off validation for openapi3.

garyluu:
Closing PR for now because validation errors are not realisticly going to be fixed for a long time.

PR #909, opened 2020-01-31
Feature/fix some unused locals

garyluu:
tslint `"no-use-before-declare": true` is deprecated. Turning on the tsconfig `"noUnusedLocals": true` causes A LOT of warnings.  Most of the warnings are from swagger/openapi which I can't seem to disable right now (new openapi generator helps).  This PR fixes most of our written code's problems

### DIFF
--- a/src/app/container/descriptors/descriptors.component.ts
+++ b/src/app/container/descriptors/descriptors.component.ts
@@ -15,8 +15,6 @@
  */
 import { Component, Input } from '@angular/core';
 import { Observable } from 'rxjs';
-
-import { ContainerService } from '../../shared/container.service';
 import { DescriptorService } from '../../shared/descriptor.service';
 import { FileService } from '../../shared/file.service';
 import { GA4GHFilesQuery } from '../../shared/ga4gh-files/ga4gh-files.query';
@@ -46,7 +44,6 @@ export class DescriptorsComponent extends EntryFileSelector {
   protected entryType: 'tool' | 'workflow' = 'tool';
 
   constructor(
-    private containerService: ContainerService,
     private descriptorsService: DescriptorService,
     protected gA4GHService: GA4GHService,
     private toolQuery: ToolQuery,
--- a/src/app/container/dockerfile/dockerfile.component.ts
+++ b/src/app/container/dockerfile/dockerfile.component.ts
@@ -19,7 +19,6 @@ import { Observable } from 'rxjs';
 import { finalize, first } from 'rxjs/operators';
 
 import { ga4ghPath } from '../../shared/constants';
-import { ContainerService } from '../../shared/container.service';
 import { Dockstore } from '../../shared/dockstore.model';
 import { FileService } from '../../shared/file.service';
 import { ContainersService } from '../../shared/swagger';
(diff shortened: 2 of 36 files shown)


src/app/container/descriptors/descriptors.component.ts
```diff
@@ -25,8 +23,8 @@ import { EntryFileSelector } from '../../shared/selectors/entry-file-selector';
 import { GA4GHService, ToolDescriptor, ToolFile } from '../../shared/swagger';
 import { Tag } from '../../shared/swagger/model/tag';
 import { ToolQuery } from '../../shared/tool/tool.query';
-import { FilesService } from '../../workflow/files/state/files.service';
 import { FilesQuery } from '../../workflow/files/state/files.query';
+import { FilesService } from '../../workflow/files/state/files.service';
```
denis-yuen: Why is this moving around, does prettier handle import ordering?

src/app/container/dockerfile/dockerfile.component.ts
```diff
@@ -48,12 +47,7 @@ export class DockerfileComponent {
   public customDownloadHREF: SafeUrl;
   public customDownloadPath: string;
   public loading = true;
-  constructor(
-    public fileService: FileService,
-    private toolQuery: ToolQuery,
-    private containerService: ContainerService,
-    private containersService: ContainersService
-  ) {
+  constructor(public fileService: FileService, private toolQuery: ToolQuery, private containersService: ContainersService) {
```
denis-yuen: Ditto, looks like just a formatting change

src/app/workflow/version-modal/version-modal.service.ts
```diff
@@ -59,6 +59,7 @@ export class VersionModalService {
    */
   saveVersion(workflowVersion: WorkflowVersion, originalTestParameterFilePaths, newTestParameterFiles, workflowMode: String) {
     const message1 = 'Saving workflow version';
+    // TODO: Use or delete this second message
```
denis-yuen: :+1:

src/app/container/info-tab/info-tab.component.ts
```diff
@@ -67,7 +65,7 @@ export class InfoTabComponent implements OnInit, OnChanges {
     if (this.selectedVersion && this.tool) {
       this.currentVersion = this.selectedVersion;
       const found = this.validVersions.find((version: Tag) => version.id === this.selectedVersion.id);
-      this.isValidVersion = found ? true : false;
+      this.isValidVersion = !!found;
```
coverbeck: You could just use `some` instead of `find` on line 69/67 and assign directly to isValidVersion.

src/app/container/descriptors/descriptors.component.ts
garyluu: It's a linter thing, not prettier. I have it turned on for mine because it's the "correct" way to do imports. If my changes touch imports, mine will correct it. Our previous co-ops didn't use an IDE that supports it so it seemed unreasonable for me to request changes every time their imports weren't in the correct order.

src/app/container/dockerfile/dockerfile.component.ts
garyluu: This is actually prettier, since the length is shorter than 400 characters, it will put it on one line now

PR #912, opened 2020-02-11
New error message for DAG tab

wshands:
Instead of saying 'A valid descriptor could not be found' when the WDL file is malformed and a DAG cannot be created print the message 'Please ensure that the Descriptor is a valid WDL workflow.' 

The error message from the parser will be printed in the File tab area so the user can determine what is wrong with the WDL file. https://github.com/dockstore/dockstore/pull/3225

Addresses https://ucsc-cgl.atlassian.net/browse/DOCK-1137

For dockstore/dockstore#3118

### DIFF
--- a/src/app/workflow/dag/dag.component.html
+++ b/src/app/workflow/dag/dag.component.html
@@ -25,7 +25,7 @@
       <mat-card class="alert alert-warning" role="alert" *ngIf="!(dagResult$ | async) && !selectedVersion?.valid">
         <mat-icon>warning</mat-icon>&nbsp;
         <span *ngIf="(descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL">
-          A Descriptor associated with this workflow could not be found.
+          Please ensure that the Descriptor is a valid WDL workflow.
         </span>
         <span *ngIf="(descriptorType$ | async) === ToolDescriptor.TypeEnum.CWL">
           Please ensure that the Descriptor is a valid CWL 1.0 workflow.
--- a/src/app/workflow/dag/dag.component.html
+++ b/src/app/workflow/dag/dag.component.html
@@ -24,8 +24,13 @@
     <div fxFlex *ngIf="dagType === 'classic' && !(isNFL$ | async) && !(dagResult$ | async)">
       <mat-card class="alert alert-warning" role="alert" *ngIf="!(dagResult$ | async) && !selectedVersion?.valid">
         <mat-icon>warning</mat-icon>&nbsp;
-        <span *ngIf="(descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL">
-          Please ensure that the Descriptor is a valid WDL workflow.
+        <span *ngIf="(descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL && isPublic">
+          The DAG cannot be displayed because the descriptor is either missing or not valid.
+          Click on the Files tab for more information.
+        </span>
+        <span *ngIf="(descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL && !isPublic">
+          The DAG cannot be displayed because the descriptor missing or is not a valid WDL workflow.
+          Ensure the descriptor exists and is valid. Click on the Files tab for more information.
         </span>
         <span *ngIf="(descriptorType$ | async) === ToolDescriptor.TypeEnum.CWL">
           Please ensure that the Descriptor is a valid CWL 1.0 workflow.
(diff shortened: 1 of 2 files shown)


src/app/workflow/dag/dag.component.html
```diff
@@ -25,7 +25,7 @@
       <mat-card class="alert alert-warning" role="alert" *ngIf="!(dagResult$ | async) && !selectedVersion?.valid">
         <mat-icon>warning</mat-icon>&nbsp;
         <span *ngIf="(descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL">
-          A Descriptor associated with this workflow could not be found.
+          Please ensure that the Descriptor is a valid WDL workflow.
```
denis-yuen: These two error messages, do they get displayed to both users and the author of the workflow?
Only the author actually has any ability to ensure that the WDL/CWL 1.0 workflow is valid.

coverbeck:
As I recall when I wrote this up, the call to get the DAG was already returning an error message, but we weren't displaying it. Is that message different than the message you are now displaying on the Files tab?

src/app/workflow/dag/dag.component.html
coverbeck: What happens if there is no descriptor associated with the workflow, as opposed to an invalid descriptor? Can it get in here, and if so, is that the message we want to display?

Also, it was like this already, but I don't think "Descriptor" should start with a capital.

wshands: Not sure; I think maybe both.

wshands: Definitely displayed to the author

wshands: Screen Shot 2020-02-11 at 9 28 16 AM

denis-yuen: So to elaborate, I think the message makes sense to show to the author. 
It makes less sense when shown to general users. 

Charles comment might be nice to tackle as well. i.e. the old message make sense if there are no descriptors, the new one makes sense if there are descriptors but they are invalid

wshands:
The message that the DAG API call returns is the same as the message now displayed on the File tab. So perhaps we can make the DAG message: 
'Please ensure that the workflow descriptor exists and is valid WDL."

src/app/workflow/dag/dag.component.html
wshands: How about "The workflow descriptor either does not exist or is not valid WDL."

coverbeck:
> 'The workflow descriptor either does not exist or is not valid WDL."

Can you easily tell the difference in code? If you know, it would be better to display a more precise error message. If you can't tell the difference, then that message sounds good to me.

wshands:
Added ability to determine if page is public based on what I see in view.component.ts; I'll need feedback on whether this is recommended way...

src/app/workflow/dag/dag.component.html
```diff
@@ -24,8 +24,13 @@
     <div fxFlex *ngIf="dagType === 'classic' && !(isNFL$ | async) && !(dagResult$ | async)">
       <mat-card class="alert alert-warning" role="alert" *ngIf="!(dagResult$ | async) && !selectedVersion?.valid">
         <mat-icon>warning</mat-icon>&nbsp;
-        <span *ngIf="(descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL">
-          Please ensure that the Descriptor is a valid WDL workflow.
+        <span *ngIf="(descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL && isPublic">
+          The DAG cannot be displayed because the descriptor is either missing or not valid.
+          Click on the Files tab for more information.
+        </span>
+        <span *ngIf="(descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL && !isPublic">
```
garyluu: Should only do the (descriptorType$ | async) === ToolDescriptor.TypeEnum.WDL once.  Nest the isPublic check inside

wshands: Fixed

src/app/workflow/dag/dag.component.ts
```diff
@@ -70,6 +71,7 @@ export class DagComponent extends EntryTab implements OnInit, OnChanges, AfterVi
   public refreshCounter = 1;
   public dagLoading$: Observable<boolean>;
   public wdlViewerResult$: Observable<boolean>;
+  public isPublic: Boolean = false;
```
coverbeck: Should be lower case `boolean`, not `Boolean`: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

garyluu:
didn't notice more than 1 files has changed.  use async pipe for the isPublic

src/app/workflow/dag/dag.component.ts
wshands: Done

src/app/workflow/dag/dag.component.html
wshands: Comment changes based on whether page is public or the My Workflows page

PR #916, opened 2020-02-14
Feature/cypress update

garyluu:
Update Cypress which makes tests failures seemingly more consistent.

Increase timeout because locally my computer is slower (not running prod build).

Explanation of some fixes:
- Check that the page has mostly loaded (by looking at tool_path existence) before clicking the sidebar accordion
- Wait 😢  5 seconds after opening the dialog before typing in values (I don't know why this is broken, it's not waiting for api call either)
- Added a loading bar when getting test parameter files to prevent form change while user is typing

Ran group 2 tests 14 times.  failed 1 time because of db timeout issue

Also seems I forgot to manually run fixes for the Cypress folder


coverbeck:
@ByteMap, Heads up, this is going to cause you some merge issues for your work on dockstore/dockstore#2562.

cypress/integration/group2/organizations.ts
```diff
@@ -15,6 +15,8 @@
  */
 import { approvePotatoMembership, resetDB, setTokenUserViewPort } from '../../support/commands';
 
+const imageURL =
```
denis-yuen: I don't think the image data is being used (I could be wrong). 
But it is tempting for someone to rely on this someday and then for our tests to die if cloudinary is down
Just check this potato into Github (our tests would die anyway if GitHub was down)

garyluu: We don't actually have a test that checks whether the image is displayed or not. Meaning even if cloudinary is down, it would still pass.

denis-yuen: I'm aware, please review.

PR #920, opened 2020-02-21
Version update only

denis-yuen:
For debugging just the webservice update


denis-yuen:
Turns out the migrations need to be specified in two places

PR #921, opened 2020-02-24
Feature/2562/install cypress once

ByteMap:
Installed Cypress Once

dockstore/dockstore#2562

Added Cypress to package.json

Removed 'Install Cypress' from .circleci/config.yml

Implemented a Cypress binary cache in .circleci/config.yml and had each integration test restore that cache

Caching the .cache directory as that is where the Cypress Binary is located which would resolve Cypress tests complaining that "although node_modules is cached, cypress binary is not."

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "@types/jquery": "^3.3.31",
     "@types/node": "^10.14.19",
     "codelyzer": "^5.0.1",
+    "cypress": "^3.4.1",
     "husky": "^3.0.5",
     "jasmine-core": "^3.5.0",
     "jasmine-spec-reporter": "^4.2.1",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "@types/jquery": "^3.3.31",
     "@types/node": "^10.14.19",
     "codelyzer": "^5.0.1",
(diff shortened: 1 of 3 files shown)


ByteMap:
Before Installing Cypress once
Screen Shot 2020-02-12 at 3 17 32 PM

After install Cypress once
Screen Shot 2020-02-24 at 3 12 47 PM

package.json
```diff
@@ -78,6 +78,7 @@
     "@types/jquery": "^3.3.31",
     "@types/node": "^10.14.19",
     "codelyzer": "^5.0.1",
+    "cypress": "^3.4.1",
```
coverbeck: Not the right Cypress version. See #916 that i pinged you on.

ByteMap: Ah shoot, forgot about that. I thought that doing a rebase on develop then a force push would bring everything up to date.

ByteMap: Fixed.

denis-yuen:
> How much of a difference does this make time-wise?

Looking at the image, a couple minutes for integration test 2, maybe a minute for integration test 1. 
The whole thing is slower, maybe because the build queue was busy? 

Its unfortunately less significant than hoped

PR #922, opened 2020-02-25
Link checking cypress tests

david4096:
This PR adds some simple link checking tests and a `cypress-remote.json` that can be used to test against a remote resource. It adds data-cy elements in a few places.

Cypress doesn't allow you to have tests that act on more than one domain so the external link checks just ensure the element is there and has the appropriate URL.

### DIFF
--- a/cypress-remote.json
+++ b/cypress-remote.json
@@ -1,7 +1,0 @@
-{
-  "projectId": "1ya4vj",
-  "baseUrl": "https://staging.dockstore.org",
-  "viewportHeight": 900,
-  "viewportWidth": 1440,
-  "defaultCommandTimeout": 10000
-}
--- a/cypress-remote.json
+++ b/cypress-remote.json
@@ -1,0 +1,7 @@
+{
+  "projectId": "1ya4vj",
+  "baseUrl": "https://staging.dockstore.org",
+  "viewportHeight": 900,
+  "viewportWidth": 1440,
+  "defaultCommandTimeout": 10000
+}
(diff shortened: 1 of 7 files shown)


src/app/home-page/home-logged-out/home.component.html
```diff
@@ -174,16 +181,17 @@ <h3 class="text-center">Next steps</h3>
...
       >
         Developer tutorial <mat-icon>open_in_new</mat-icon>
       </a>
       <a mat-stroked-button href="{{ Dockstore.DOCUMENTATION_URL }}/launch-with/launch.html" target="_blank" rel="noopener noreferrer">
         Launch workflows <mat-icon>open_in_new</mat-icon>
       </a>
-      <button mat-stroked-button routerLink="/organizations" (click)="scrollToTop()">
-        Discover Organizations
+      <button mat-stroked-button routerLink="/organizations" (click)="scrollToTop()" data-cy="homepage-organizations-button">
+        Discover Organii dont want to zations
```
NatalieEO: 🤔

denis-yuen:
> @denis-yuen Thoughts on number 2? Should we give up on smoke tests for the hotfix, or do you think it is worth redoing this on a hotfix branch for a first iteration?

I'm ok either way, whether we manually test these links or add in the new data-cy attributes (which doesn't seem super disruptive).

cypress/integration/manualTests/monitor.ts
```diff
@@ -0,0 +1,57 @@
...
+      });
+    });
+  });
+
+});
+
+
+before(() => {
+  cy.visit('');
+});
```
garyluu: Comment on why there's two of these

cypress-remote.json
```diff
@@ -0,0 +1,7 @@
+{
+  "projectId": "1ya4vj",
+  "baseUrl": "https://staging.dockstore.org",
```
garyluu: If there are multiple baseUrls to test, then there should really just be the one cypress.json with baseUrl overridden when running the command.

PR #929, opened 2020-03-02
Feature/3281/workflow path length

garyluu:
For dockstore/dockstore#3281

Lots of our max-length file paths were 128 which is different from the database.

Changing as many as I can find to match.

Let me know if there are other places I haven't thought of


garyluu:
> Think we can be generous with 280 just in case

Won't this cause a 500 error when a user puts in something longer than database?

NatalieEO:
Should change the validation messages in `validationMessages.model.ts`?

PR #932, opened 2020-03-04
Feature/3258/link orcid

emlys:
dockstore/dockstore#3258


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "1.9.0-alpha.3"
+    "webservice_version": "feature/2421/link-orcid"
```
denis-yuen: Will need another alpha release, can create one when the other PR is merged

src/app/loginComponents/accounts/external/accounts.component.ts
```diff
@@ -84,6 +84,14 @@ export class AccountsExternalComponent implements OnInit, OnDestroy {
       message: 'Zenodo credentials are used for creating Digital Object Identifiers (DOIs) on Zenodo.',
       show: false,
       logo: 'zenodo.jpg'
+    },
+    {
+      name: 'ORCID',
+      source: TokenSource.ORCID,
+      bold: '',
+      message: 'ORCID credentials are used for linking ORCID IDs to workflows.',
```
coverbeck: I don't think this message is quite right, but on the other hand, we're not sure what level of support we're going to add for 1.9. Maybe for now `ORCID credentials are used for linking ORCID IDs to workflows published on Zenodo`?

Or maybe this is fine. Anybody else have an opinion?

garyluu: :man_shrugging:

denis-yuen: `ORCID credentials are used for linking ORCID IDs to workflows published on Zenodo` makes sense. I also saw the ticket on promoting it on user profiles (stargazers, organization membership) which makes sense to me

src/app/loginComponents/accounts/external/accounts.component.ts
```diff
@@ -89,7 +89,7 @@ export class AccountsExternalComponent implements OnInit, OnDestroy {
       name: 'ORCID',
       source: TokenSource.ORCID,
       bold: '',
-      message: 'ORCID credentials are used for linking ORCID IDs to workflows.',
+      message: 'ORCID credentials are used for linking ORCID IDs to workflows published on Zenodo.',
```
coverbeck, line 92: They'll be used for more than that, as we discussed at storytime, but I don't know how comprehensive you want to be. If you can think of something more general better that would be good, but don't need to block the PR for it.

denis-yuen: Maybe `ORCID credentials are used for vouching for workflows and tools plus linking ORCID IDs to workflows published on Zenodo.`

agduncan94:
Not blocking for this but I think we are reaching a point where our account linking page is outgrowing the format. We should consider how to better organize it now that we will have 8 account cards.

PR #933, opened 2020-03-05
Feature/differential loading take2

garyluu:
For dockstore/dockstore#3047

Angular update did the trick


garyluu:
Transferred over network (according to Chrome)
Before: 17.4 MB 
After: 16.4 MB

PR #934, opened 2020-03-05
Removed ngx-bootstrap from hostedTools

ByteMap:
dockstore/dockstore#3251

Migrated from ngx-bootstrap tabs to Material tabs

Removed ngx-bootstrap module from spec file

Changed Cypress tests to work for the new Material tabs for
hostedTools

### DIFF
--- a/src/app/container/tool-file-editor/tool-file-editor.component.html
+++ b/src/app/container/tool-file-editor/tool-file-editor.component.html
@@ -14,8 +14,8 @@
   ~    limitations under the License.
   -->
 
-<tabset [justified]="true">
-  <tab heading="Dockerfile" id="dockerfileTab">
+<mat-tab-group mat-stretch-tabs>
+  <mat-tab label="Dockerfile" id="dockerfileTab">
     <app-code-editor-list
       [sourcefiles]="dockerFile"
       [editing]="editing"
@@ -24,8 +24,8 @@
       [entryType]="'tool'"
       [entrypath]="entrypath"
     ></app-code-editor-list>
-  </tab>
-  <tab heading="Descriptor Files" id="descriptorFilesTab">
+  </mat-tab>
+  <mat-tab label="Descriptor Files" id="descriptorFilesTab">
     <mat-form-field>
       <mat-select [(value)]="selectedDescriptorType" placeholder="Descriptor Type">
         <mat-option value="{{ ToolDescriptor.TypeEnum.CWL }}">{{ ToolDescriptor.TypeEnum.CWL }}</mat-option>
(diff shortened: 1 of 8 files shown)


ByteMap:
How the new Material Tabs would look.
Screen Shot 2020-03-05 at 1 24 11 PM 1
Screen Shot 2020-03-05 at 1 24 31 PM
Screen Shot 2020-03-05 at 1 24 40 PM

coverbeck:
Can you please attach a screenshot?

ByteMap:
Also it seems that the Cypress test seems a little bit flaky especially at commands such as line 83. When getting the 'tool-path', the data is slow at updating itself. For instance, when executing the command `cy.get('#tool-path').contains('quay.io/hosted-tool/ht:1');`, ht does not update to ht1 until a few seconds later, which causes the test to fail. The only solution that I can think of is to extend `cy.wait()`, but that should be a last resort. Does anyone have any solution in mind besides extending `cy.wait()`?

agduncan94:
@ByteMap Looks good but I would add some padding so that the dropdown for descriptor type isn't so far left.

ByteMap:
Newest preview for the slight button and dropdown adjustments

For the tool-file-editor:
Screen Shot 2020-03-20 at 3 09 49 PM
Screen Shot 2020-03-20 at 3 10 01 PM

For the workflow-file-editor:
Screen Shot 2020-03-20 at 3 12 47 PM
Screen Shot 2020-03-20 at 3 12 56 PM

src/app/container/tool-file-editor/tool-file-editor.component.html
```diff
@@ -14,54 +14,60 @@
...
+        [sourcefiles]="dockerFile"
+        [editing]="editing"
+        [fileType]="'dockerfile'"
+        [selectedVersion]="currentVersion"
+        [entryType]="'tool'"
+        [entrypath]="entrypath"
+      ></app-code-editor-list>
+    </ng-template>
+  </mat-tab>
+  <mat-tab label="Descriptor Files" id="descriptorFilesTab">
```
garyluu: unused id here and elsewhere

src/app/container/tool-file-editor/tool-file-editor.component.scss
```diff
@@ -5,3 +5,7 @@
 .editor-button-container {
   margin-top: 10px;
 }
+
+.mat-form-field {
+  padding-left: 0.5rem;
```
garyluu: this is just pl-2

src/app/shared/code-editor-list/code-editor-list.component.scss
```diff
@@ -5,3 +5,7 @@
 .link-button {
   color: inherit;
 }
+
+.add-file-button {
+  left: 0.5rem;
```
garyluu: just use pl-2 if possible

ByteMap: The button is not affected by pl-2, so instead I have used ml-2.

src/app/container/tool-file-editor/tool-file-editor.component.scss
ByteMap: Done

src/app/container/tool-file-editor/tool-file-editor.component.html
ByteMap: Removed the unused IDs

PR #935, opened 2020-03-10
Removing refresh all 

denis-yuen:
Finishing removal of refresh all from the UI

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "1.9.0-alpha.4"
+    "webservice_version": "1.9.0-alpha.5"
   },
   "scripts": {
     "ng": "npx ng",
--- a/package.json
+++ b/package.json
@@ -64,8 +64,8 @@
     "zone.js": "~0.9.1"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "^0.803.21",
-    "@angular/cli": "^8.3.6",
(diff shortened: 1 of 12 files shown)


denis-yuen:
FYI, differential loading died three times in a row with the message shown in Slack

package.json
```diff
@@ -64,8 +64,8 @@
     "zone.js": "~0.9.1"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "^0.803.25",
```
coverbeck: Comment for line 6 -- if you update to alpha.4, the compiler should catch any usages of the endpoints you have removed (although you probably already caught them all)

package.json
```diff
@@ -64,8 +64,8 @@
...
-    "@angular/cli": "^8.3.25",
+    "@angular-devkit/build-angular": "^0.803.21",
```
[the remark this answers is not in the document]
denis-yuen: yup

package.json
denis-yuen: Worth a shot

denis-yuen:
`ci/circleci: integration_test_3` died three times with 
Screenshot from 2020-03-11 18-14-32

coverbeck:
> `ci/circleci: integration_test_3` died three times with

Probably my fault for suggesting you upgrade to the latest web service :( Although it probably represents a real issue.

denis-yuen:
Note for self, there is a 500 error visible in the cypress video for `/api/workflows/published`

garyluu:
I looked into this a bit, updating the webservice to the current develop will automatically fix the test failure.

PR #936, opened 2020-03-12
Display missing file validation for services

kathy-t:
If a file is missing in the .dockstore.yml, the validation message appears when viewing the Files > Files tab and the Files > Primary Descriptor tab.

In Files > Files tab:
Screenshot from 2020-03-12 11-03-30

In Files > Primary Descriptor tab:
Screenshot from 2020-03-12 11-05-09

While I was testing the validation for services, I noticed that when I selected a new version from the drop down menu at the top, the previous version's validation message still remained. This was fixed in my [second commit](https://github.com/dockstore/dockstore-ui2/commit/6909076c3cc79327438021f7ff358b881e0f468e). The problem was that previous validation message wasn't being cleared if you changed the version when you were in the Files > Files tab.

I also noticed that if you were in the Files > Primary Descriptor tab and you selected a new version from the drop down, it would load the file contents for Files > Files. This was fixed in my [third commit](https://github.com/dockstore/dockstore-ui2/commit/09145a3c1f02c89e48cc65091a17cead2c65aa2b). The problem was that the Files > Files content was loaded for the new version selected no matter what your current selectedFileType was.

### DIFF
--- a/src/app/entry/entry-file-tab/state/entry-file-tab.service.ts
+++ b/src/app/entry/entry-file-tab/state/entry-file-tab.service.ts
@@ -187,7 +187,8 @@ export class EntryFileTabService extends Base {
    */
   private setAlmostEverything(unfilteredFiles: ToolFile[]) {
     const fileTypes = this.getFileTypes(unfilteredFiles);
-    const selectedFileType: ToolFile.FileTypeEnum = fileTypes[0];
+    const previousFileType = this.entryFileTabQuery.getValue().selectedFileType;
+    const selectedFileType: ToolFile.FileTypeEnum = previousFileType ? previousFileType : fileTypes[0];
     const files = this.filterFiles(selectedFileType, unfilteredFiles);
     const file = files[0];
     this.entryFileTabStore.update(state => {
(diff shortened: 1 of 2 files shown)


src/app/entry/entry-file-tab/state/entry-file-tab.service.ts
```diff
@@ -186,7 +187,8 @@ export class EntryFileTabService extends Base {
    */
   private setAlmostEverything(unfilteredFiles: ToolFile[]) {
     const fileTypes = this.getFileTypes(unfilteredFiles);
-    const selectedFileType: ToolFile.FileTypeEnum = fileTypes[0];
+    const previousFileType = this.entryFileTabQuery.getValue().selectedFileType;
+    const selectedFileType: ToolFile.FileTypeEnum = previousFileType ? previousFileType : fileTypes[0];
```
garyluu: Looked into this a bit (about previousFileType not applicable to this new version select). So it does indeed assign the wrong value to selectedFileType and cascades down to files and file, but some other method is able to correct this.  Regardless, it should not be doing this. It needs to check that the previousFileType is valid first.

PR #939, opened 2020-03-17
Feature/3312/disable launch for other languages

garyluu:
For dockstore/dockstore#3312

Adds the ability to disable launch support for other languages.

It's just a simple bootstrap-styled warning alert

### DIFF
--- a/src/app/workflow/workflow.component.html
+++ b/src/app/workflow/workflow.component.html
@@ -170,16 +170,25 @@
           </mat-tab>
           <div *ngIf="entryType === EntryType.BioWorkflow">
             <mat-tab id="launchTab" label="Launch">
-              <app-launch
-                *ngIf="workflow?.workflowVersions.length > 0; else noVersions"
-                [selectedVersion]="selectedVersion"
-                [basePath]="workflow?.path"
-                [path]="workflow?.full_workflow_path"
-                [canRead]="canRead"
-                [canWrite]="canWrite"
-                [isOwner]="isOwner"
-                [mode]="workflow?.mode"
-              ></app-launch>
+              <div *ngIf="!(launchSupport$ | async); else launchSupported">
+                <div class="p-3">
+                  <mat-card class="alert alert-warning" role="alert">
+                    <mat-icon>warning</mat-icon> Dockstore does not support launching with this language.
+                  </mat-card>
+                </div>
+              </div>
+              <ng-template #launchSupported>
+                <app-launch
+                  *ngIf="workflow?.workflowVersions.length > 0; else noVersions"
+                  [selectedVersion]="selectedVersion"
+                  [basePath]="workflow?.path"
+                  [path]="workflow?.full_workflow_path"
+                  [canRead]="canRead"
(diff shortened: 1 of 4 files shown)


coverbeck:
Could we get a screenshot?

garyluu:
> Could we get a screenshot?

image

src/app/workflow/workflow.component.html
```diff
@@ -170,16 +170,25 @@ <h3>
...
-                [path]="workflow?.full_workflow_path"
-                [canRead]="canRead"
-                [canWrite]="canWrite"
-                [isOwner]="isOwner"
-                [mode]="workflow?.mode"
-              ></app-launch>
+              <div *ngIf="!(launchSupport$ | async); else launchSupported">
+                <div class="p-3">
+                  <mat-card class="alert alert-warning" role="alert">
+                    <mat-icon>warning</mat-icon> Dockstore does not support launching with this language.
```
denis-yuen: Seems a bit too final. 
Maybe `Dockstore does not yet have customized launch-with instructions for X language.` Ideas welcome

src/app/workflow/workflow.component.html
```diff
@@ -170,16 +170,25 @@ <h3>
...
-                [path]="workflow?.full_workflow_path"
-                [canRead]="canRead"
-                [canWrite]="canWrite"
-                [isOwner]="isOwner"
-                [mode]="workflow?.mode"
-              ></app-launch>
+              <div *ngIf="!(launchSupport$ | async); else launchSupported">
+                <div class="p-3">
+                  <mat-card class="alert alert-warning" role="alert">
+                    <mat-icon>warning</mat-icon> Dockstore does not yet have customized launch-with instructions for this language.
```
coverbeck: As in Denis' suggestion, I think it should mention the specific language rather than `this language`. Or is that going to be hard to do generically because we say `gxformat2` instead of Galaxy?

src/app/workflow/workflow.component.html
Ldcabansay: "Launch-with support for X language coming soon see [link to documentation?] for more details"

It can lead to the documentation page where we talk about the roadmap or how to manually download and upload to galaxy in the interim?

src/app/workflow/workflow.component.html
garyluu: I think mentioning the language itself adds little to no value to the user.  And yes, the backend provides "gxformat2", Nextflow provides "NFL".  All of which we have and would have overrided in the frontend

src/app/workflow/workflow.component.html
garyluu: Documentation page for the other languages is not going to work.

src/app/workflow/workflow.component.html
denis-yuen: Looking at TRS again in a coincidence, I'm tempted to switch it to GALAXY anyway. 
For example, we don't treat WDL draft-2 differently from WDL version 1

PR #940, opened 2020-03-18
Feature/workflow gh apps link

agduncan94:
Option for installing GitHub App on users repos.

register-workflow-step-1
register-workflow-step-2

### DIFF
--- a/src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
+++ b/src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
@@ -49,20 +49,15 @@
     </mat-step>
     <mat-step>
       <ng-template matStepLabel>Create a workflow</ng-template>
-      <div *ngIf="selectedOption?.value == 0">
+      <div *ngIf="selectedOption?.value === 0">
         <p>
           Navigate to GitHub to install our GitHub app on your repositories/organizations. See our
-          <a
-            href="{{ Dockstore.DOCUMENTATION_URL }}/getting-started/github-apps.html"
-            target="_blank"
-            rel="noopener noreferrer"
-            data-cy="homepage-getting-started-link"
-          >
+          <a href="{{ Dockstore.DOCUMENTATION_URL }}/getting-started/github-apps.html" target="_blank" rel="noopener noreferrer">
             documentation <mat-icon>open_in_new</mat-icon>
           </a>
           for how to use GitHub Apps to automatically sync your workflows with GitHub.
         </p>
-        <div fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="20px" class="mb-2">
+        <div fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="2rem" class="mb-2">
           <a
             mat-flat-button
             type="button"
@@ -81,7 +76,7 @@
           </button>
         </div>
       </div>
-      <div *ngIf="selectedOption?.value == 1">
+      <div *ngIf="selectedOption?.value === 1">
         <entry-wizard></entry-wizard>
         <div class="modal-footer">
           <button mat-button color="secondary" matStepperPrevious>Back</button>
@@ -93,7 +88,7 @@
       <form
         fxLayout="column"
         fxLayoutAlign="space-evenly stretch"
-        *ngIf="selectedOption?.value == 2"
+        *ngIf="selectedOption?.value === 2"
         #registerWorkflowForm="ngForm"
         name="registerWorkflowForm"
         (submit)="registerWorkflow()"
@@ -209,7 +204,7 @@
       <form
         fxLayout="column"
         fxLayoutAlign="space-evenly stretch"
-        *ngIf="selectedOption?.value == 3"
+        *ngIf="selectedOption?.value === 3"
         #registerHostedWorkflowForm="ngForm"
         name="registerHostedWorkflowForm"
         (ngSubmit)="registerHostedWorkflow()"
--- a/src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
+++ b/src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
@@ -51,8 +51,16 @@
       <ng-template matStepLabel>Create a workflow</ng-template>
       <div *ngIf="selectedOption?.value == 0">
         <p>
-          Navigate to GitHub to install our GitHub app on your repositories/organizations. See our documentation for how to use GitHub Apps
-          to automatically sync your workflows with GitHub.
+          Navigate to GitHub to install our GitHub app on your repositories/organizations. See our
+          <a
+            href="{{ Dockstore.DOCUMENTATION_URL }}/getting-started/github-apps.html"
+            target="_blank"
+            rel="noopener noreferrer"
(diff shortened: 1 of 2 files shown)


src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
```diff
@@ -50,6 +50,38 @@ <h4 mat-dialog-title>Register Workflow</h4>
     <mat-step>
       <ng-template matStepLabel>Create a workflow</ng-template>
       <div *ngIf="selectedOption?.value == 0">
+        <p>
+          Navigate to GitHub to install our GitHub app on your repositories/organizations. See our
+          <a
+            href="{{ Dockstore.DOCUMENTATION_URL }}/getting-started/github-apps.html"
+            target="_blank"
+            rel="noopener noreferrer"
+            data-cy="homepage-getting-started-link"
```
garyluu: data-cy

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
```diff
@@ -50,6 +50,38 @@ <h4 mat-dialog-title>Register Workflow</h4>
...
+          >
+        </div>
+        <div class="modal-footer">
+          <button mat-button color="secondary" matStepperPrevious>Back</button>
+          <button type="button" mat-flat-button color="primary" [mat-dialog-close]="true" (click)="clearWorkflowRegisterError()">
+            Finish
+          </button>
+        </div>
+      </div>
+      <div *ngIf="selectedOption?.value == 1">
```
garyluu: `===` all of the options

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
```diff
@@ -50,6 +50,38 @@ <h4 mat-dialog-title>Register Workflow</h4>
...
+            href="{{ Dockstore.DOCUMENTATION_URL }}/getting-started/github-apps.html"
+            target="_blank"
+            rel="noopener noreferrer"
+            data-cy="homepage-getting-started-link"
+          >
+            documentation <mat-icon>open_in_new</mat-icon>
+          </a>
+          for how to use GitHub Apps to automatically sync your workflows with GitHub.
+        </p>
+        <div fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="20px" class="mb-2">
```
garyluu: 2 rem

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
```diff
@@ -49,7 +49,34 @@ <h4 mat-dialog-title>Register Workflow</h4>
     </mat-step>
     <mat-step>
       <ng-template matStepLabel>Create a workflow</ng-template>
-      <div *ngIf="selectedOption?.value == 0">
+      <div *ngIf="selectedOption?.value === 0">
+        <p>
+          Navigate to GitHub to install our GitHub app on your repositories/organizations. See our
+          <a href="{{ Dockstore.DOCUMENTATION_URL }}/getting-started/github-apps.html" target="_blank" rel="noopener noreferrer">
```
coverbeck: I'm sure your doc covers it, but I wonder if it's worth stating here that a .dockstore.yml will be required? People may have an existing workflow in pipeline, come into Dockstore, pick the recommended way to add their workflow, and then... nothing -- because they don't have .dockstore.yml.

We do want to force people to go this way, and your doc will cover it, so maybe I'm overthinking it.

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
```diff
@@ -49,7 +49,34 @@ <h4 mat-dialog-title>Register Workflow</h4>
...
+        <div fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="2rem" class="mb-2">
+          <a
+            mat-flat-button
+            type="button"
+            color="primary"
+            [href]="gitHubAppInstallationLink$ | async"
+            target="_blank"
+            rel="noopener noreferrer"
+            matTooltip="Manage on GitHub"
+            ><mat-icon>add</mat-icon> Manage Installations on GitHub</a
```
coverbeck: "Installations" sounds odd to me. Is it because you install the app into your GitHub orgs/repos? No alternative suggestions at this point though.

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
```diff
@@ -49,7 +49,34 @@ <h4 mat-dialog-title>Register Workflow</h4>
...
+            documentation <mat-icon>open_in_new</mat-icon>
+          </a>
+          for how to use GitHub Apps to automatically sync your workflows with GitHub.
+        </p>
+        <div fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="2rem" class="mb-2">
+          <a
+            mat-flat-button
+            type="button"
+            color="primary"
+            [href]="gitHubAppInstallationLink$ | async"
```
coverbeck: Is this correct -- if I click the link the GitHub app configuration will open in a new tab, I do whatever I need to on that tab, save it and close it -- and then I come back here to the same dialog? Where I then click Finish to dismiss? If so, doesn't seem quite right.

How about:

1. Instead of being an href, clicking on the button dismisses the dialog before bringing up the GitHub app in a new tab?
2. And/or replace the Finish button with a Manage on GitHub
3. ?

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
agduncan94: Ya, that was my logic. I agree there is probably a better way to phrase it but unsure what that would be.

denis-yuen: `Installations` is used pretty consistently in the GItHub documentation. Perhaps what is needed here is to clarify that it is "our" app being installed to their repos/orgs?

agduncan94: Manage Connected Repositories on GitHub?

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
agduncan94: Would it make sense saying 'close' instead of finish? Not sure if I like automatically closing when the used clicks manage on GitHub but I can be convinced otherwise.

agduncan94:
@denis-yuen What about "Manage Dockstore Installations on GitHub" or "Manage Connected Repositories on GitHub"? For reference, travis-ci.com uses "Manage Repositories on GitHub".

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
agduncan94: I'll add a note that a .dockstore.yml is required.

src/app/workflow/register-workflow-modal/register-workflow-modal.component.html
coverbeck: Let's leave it as is for now -- I don't have an overwhelmingly better alternative.

PR #942, opened 2020-03-24
Feature/3051/register tool wizard

garyluu:
For dockstore/dockstore#3051

Video of it was posted in Slack, not much has changed since

https://oicr.slack.com/archives/C16ET3CF4/p1585079984399200


src/app/container/refresh-wizard/refresh-wizard.component.html
```diff
@@ -0,0 +1,72 @@
...
+            '] ' +
+            error.statusText +
+            ': ' +
+            (error.error && error.error.message ? error.error.message : error.error)
+        }}
+      </p>
+    </mat-card>
+    <div *ngIf="organizations$ | async as organizations">
+      <div *ngIf="organizations === undefined">
+        I don't know how you got passed the loading bar
```
coverbeck, line 28: `past` not `passed`, but I would just get rid of this ngIf and change the one on line 30 to be `*ngIf="!organizations"`

src/app/container/state/refresh-wizard.service.ts
```diff
@@ -0,0 +1,120 @@
...
+    });
+  }
+
+  refreshOrganization(organization: string) {
+    this.refreshWizardStore.setLoading(true);
+    this.swaggerUsersService
+      .refreshToolsByOrganization(this.userQuery.getValue().user.id, organization, TokenSource.QUAY)
+      .pipe(finalize(() => this.refreshWizardStore.setLoading(false)))
+      .subscribe(thing => {
+        console.log('I have no idea what this returns');
```
coverbeck, line 59: I don't understand this method. You're refreshing tools for an org, but then not looking for the result. In the next refreshRepository method, you do call `getMyEntries` after it's been refreshed. Should you be doing that here?

src/app/container/state/refresh-wizard.service.ts
```diff
@@ -0,0 +1,120 @@
...
+@Injectable()
+export class RefreshWizardService {
+  constructor(
+    private refreshWizardStore: RefreshWizardStore,
+    private dockerRegistriesService: UsersService,
+    private swaggerUsersService: SwaggerUsersService,
+    private userQuery: UserQuery,
+    private refreshWizardQuery: RefreshWizardQuery,
+    private matSnackBar: MatSnackBar,
+    private thing: MytoolsService,
```
coverbeck, line 25: mytoolsService instead of thing

src/app/container/refresh-wizard/refresh-wizard.component.html
garyluu: Nope, It's purposely split into two for a reason.  Null is an actually problem returned by the call.  Undefined is something wrong with loading bar not showing

src/app/container/state/refresh-wizard.service.ts
garyluu: This is currently unused, not fully implemented. I guess I'll remove this for now.

denis-yuen:
> Video of it was posted in Slack, not much has changed since
 
Can you link to it (in the OP) in case we need to understand the squash commit in the future.

src/app/container/state/refresh-wizard.service.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.sessionService.setLoadingDialog(true);
+    const userId = this.userQuery.getValue().user.id;
+    const entryType = this.sessionQuery.getValue().entryType;
+    const selectedOrganization = this.refreshWizardQuery.getValue().selectedOrganization;
+    this.swaggerUsersService
+      .refreshToolsByOrganization(userId, selectedOrganization, repository)
+      .pipe(finalize(() => this.sessionService.setLoadingDialog(false)))
+      .subscribe(
+        () => {
+          this.matSnackBar.open('Synchronizing tool succeeded');
```
denis-yuen, line 63: Is there a reason to use the term sync here and below instead of refresh?

src/app/container/refresh-wizard/refresh-wizard.component.html
```diff
@@ -0,0 +1,72 @@
...
+            </mat-option>
+          </mat-select>
+        </mat-form-field>
+        <div *ngIf="repositories$ | async as repositories">
+          <div *ngIf="repositories">
+            <div *ngFor="let repository of repositories">
+              <button
+                mat-icon-button
+                color="primary"
+                aria-label="Example icon-button with a heart icon"
```
denis-yuen, line 51: Is this heart icon label left over from somewhere?

src/app/container/refresh-wizard/refresh-wizard.component.html
```diff
@@ -0,0 +1,72 @@
...
+                (click)="refreshRepository(repository)"
+                [disabled]="(loading$ | async) || (repositoryLoading$ | async)"
+                matTooltip="Sync with Dockstore"
```
denis-yuen, line 54: Sync here too but the wizard is called a refresh wizard

src/app/container/state/refresh-wizard.service.ts
garyluu: I think I called it sync instead of refresh because it also adds tools.  But I guess it would apply to refresh too.

src/app/container/refresh-wizard/refresh-wizard.component.html
```diff
@@ -0,0 +1,72 @@
...
+        The webservice encountered an error.
+      </p>
+      <p class="error-output">
+        {{
+          '[HTTP ' +
+            error.status +
+            '] ' +
+            error.statusText +
+            ': ' +
+            (error.error && error.error.message ? error.error.message : error.error)
```
coverbeck, line 22: If `error.error` is falsey, e.g., null, then this will display `...: null`, if it's a concern.

agduncan94: I know we do this elsewhere, but I feel like we shouldn't display the error code to users, and instead just use an error message.

src/app/container/refresh-wizard/refresh-wizard.component.html
```diff
@@ -0,0 +1,72 @@
...
+            <div *ngFor="let repository of repositories">
+              <button
+                mat-icon-button
+                color="primary"
+                aria-label="Refresh button"
+                (click)="refreshRepository(repository)"
+                [disabled]="(loading$ | async) || (repositoryLoading$ | async)"
+                matTooltip="Refresh on Dockstore"
+              >
+                <mat-icon>sync</mat-icon>
```
agduncan94, line 56: The sync icon doesn't have the same state information that exists for workflows in the form of the slider. What was the reasoning for using the refresh icon over the slider?

src/app/container/refresh-wizard/refresh-wizard.component.html
garyluu: This is actually copied from the refresh component (because I don't think it's worth warranting creating another shared component). The error code discussion has been brought up before, but I think Denis wanted it there?

src/app/container/refresh-wizard/refresh-wizard.component.html
garyluu: workflow deregisters, tool doesn't.  this refreshes an existing one or adds a new one.

src/app/container/refresh-wizard/refresh-wizard.component.html
denis-yuen: Yeah, I actually like displaying more information about errors than less. 
I feel it gives more technical users a chance at understanding the problem, ourselves a quicker source of information (what error code did you get?), and there's not much cost aside from less savvy users seeing a more technical error message.

src/app/container/refresh-wizard/refresh-wizard.component.html
agduncan94: But if it refreshes an existing one, that doesn't make sense since it is a register tool modal and not a refresh tool modal.

garyluu: this is actually called the refresh-wizard component.

PR #943, opened 2020-03-25
Enable Terra for workflows with file imports

coverbeck:
dockstore/dockstore#3342

Only do it for GitHub based-workflows, as Terra is getting
the imported files directly from GitHub.


src/app/workflow/launch-third-party/launch-third-party.component.ts
```diff
@@ -237,12 +237,16 @@ export class LaunchThirdPartyComponent extends Base implements OnChanges, OnInit
     return `Export this workflow version to ${platform}.`;
   }
 
+  private isGitHubWorkflow(): boolean {
+    return this.workflow && this.workflow.gitUrl && this.workflow.gitUrl.startsWith('git@github.com');
+  }
+
   private terraTooltip(hasContent: boolean, hasFileImports, platform: string): string {
```
denis-yuen: platform is a variable in the methodname and text, but the method is called terraTooltip

PR #945, opened 2020-03-30
Add custom X-Dockstore-UI header

kathy-t:
For https://github.com/dockstore/dockstore/issues/3113

Added the following custom headers to track API calls from the Dockstore UI:
- X-Dockstore-UI: Dockstore UI version
- X-Request-ID: UUID for each request
- X-Session-ID: UUID for each browser session

Here's an example of what the headers look like:
Screen Shot 2020-04-02 at 1 52 13 PM

### DIFF
--- a/src/app/interceptors/custom-header.interceptor.ts
+++ b/src/app/interceptors/custom-header.interceptor.ts
@@ -18,7 +18,7 @@ import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
 import { versions } from '../footer/versions';
 
-const uiVersion = versions.tag.match(/\d+\.\d+\.\d+/)[0];
+const uiVersion = versions.tag;
 
 /**
  * An interceptor that ensures that every request has a custom X-Dockstore-UI header with the UI version as
--- a/src/app/interceptors/custom-header.interceptor.ts
+++ b/src/app/interceptors/custom-header.interceptor.ts
@@ -1,0 +1,31 @@
+/**
+ *    Copyright 2019 OICR
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+import { HttpEvent, HttpHandler, HttpInterceptor, HttpRequest } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { version } from '../../../package.json';
+
+/**
+ * An interceptor that ensures that every request has a custom X-Dockstore-UI header with the UI version as
+ * the value so we can track API calls coming from the Dockstore UI.
+ */
+@Injectable()
+export class CustomHeaderInterceptor implements HttpInterceptor {
+  intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
+    req = req.clone({ headers: req.headers.set('X-Dockstore-UI', version) });
+    return next.handle(req);
+  }
+}
--- a/src/app/interceptors/custom-header.interceptor.ts
+++ b/src/app/interceptors/custom-header.interceptor.ts
@@ -1,5 +1,5 @@
 /**
- *    Copyright 2019 OICR
+ *    Copyright 2020 OICR
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
  *    you may not use this file except in compliance with the License.
@@ -16,7 +16,9 @@
 import { HttpEvent, HttpHandler, HttpInterceptor, HttpRequest } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
-import { version } from '../../../package.json';
+import { versions } from '../footer/versions';
+
+const uiVersion = versions.tag.match(/\d+\.\d+\.\d+/)[0];
 
 /**
  * An interceptor that ensures that every request has a custom X-Dockstore-UI header with the UI version as
@@ -25,7 +27,7 @@ import { version } from '../../../package.json';
 @Injectable()
 export class CustomHeaderInterceptor implements HttpInterceptor {
   intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
-    req = req.clone({ headers: req.headers.set('X-Dockstore-UI', version) });
+    req = req.clone({ headers: req.headers.set('X-Dockstore-UI', uiVersion) });
     return next.handle(req);
   }
 }
--- a/src/app/interceptors/custom-header.interceptor.ts
+++ b/src/app/interceptors/custom-header.interceptor.ts
@@ -16,18 +16,25 @@
 import { HttpEvent, HttpHandler, HttpInterceptor, HttpRequest } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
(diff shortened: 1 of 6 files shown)


tsconfig.json
```diff
@@ -22,6 +22,7 @@
       "jszip": [
         "../node_modules/jszip/dist/jszip.min.js"
       ]
-    }
+    },
+    "resolveJsonModule": true
```
garyluu: is this actually necessary?

src/app/interceptors/custom-header.interceptor.ts
```diff
@@ -0,0 +1,31 @@
...
+import { version } from '../../../package.json';
+
+/**
+ * An interceptor that ensures that every request has a custom X-Dockstore-UI header with the UI version as
+ * the value so we can track API calls coming from the Dockstore UI.
+ */
+@Injectable()
+export class CustomHeaderInterceptor implements HttpInterceptor {
+  intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
+    req = req.clone({ headers: req.headers.set('X-Dockstore-UI', version) });
```
coverbeck: I don't think we keep the package.json version up to date, although we should. Right now it's at 2.3.0 in the package.json, but our UI version is 2.5.1. Can you get the version from the same place the footer gets it instead?

src/app/interceptors/custom-header.interceptor.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ *    Copyright 2019 OICR
```
coverbeck: 2020 :) Although we should probably do a global search and replace of these.

src/app/interceptors/custom-header.interceptor.ts
kathy-t: The footer gets it from ``dockstore-ui/src/app/footer/versions.ts``, but I don't see a UI version there, just the webservice version. The webservice version is set in ``git.version.ts`` using ``process.env.npm_package_config_webservice_version``, but I'm not sure where this is set in order to do a similar thing for the ui version

kathy-t: Ignore my previous comment, I misunderstood the file. The UI version is there

src/app/interceptors/custom-header.interceptor.ts
```diff
@@ -0,0 +1,33 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+import { HttpEvent, HttpHandler, HttpInterceptor, HttpRequest } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { versions } from '../footer/versions';
+
+const uiVersion = versions.tag.match(/\d+\.\d+\.\d+/)[0];
```
coverbeck: Why are you doing this instead of just using `versions.tag` as is? What's "beyond" what you're capturing, and why don't you want to include it? Not saying it's wrong; just asking for clarification.

kathy-t: An example of versions.tag is ``2.5.1-rc.1-29-g06e87643`` so I just captured ``2.5.1`` as the version based on your previous comment about the current UI version being 2.5.1. I wasn't sure if the info after that was needed. Should I keep the whole thing?

coverbeck: Hmm, not sure... Does it cause issues for metrics if you use the long name? For staging, I think we would at least want the `-rc.1` portion of the name.

kathy-t: I can use regular expressions when I create the metric filters, so shouldn't cause any issues for metrics if I use the long name since I can still get what I need. I'll use the full name then

src/app/interceptors/custom-header.interceptor.ts
```diff
@@ -0,0 +1,40 @@
...
+ * An interceptor that ensures that every request has a custom X-Dockstore-UI header with the UI version as
+ * the value and a custom UUID header so we can track API calls coming from the Dockstore UI.
+ */
+@Injectable()
+export class CustomHeaderInterceptor implements HttpInterceptor {
+  intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
+    req = req.clone({
+      setHeaders: {
+        'X-Dockstore-UI': uiVersion,
+        'X-UUID': uuid
```
coverbeck, line 35: 1. I prefer something like X-Dockstore-UUID, or maybe [X-Request-Id](https://stackoverflow.com/questions/25433258/what-is-the-x-request-id-http-header). But don't feel too strongly about it.
2. Alternatively, you could generate a different uuid per request, but that might be slow -- this is definitely a major improvement as is.

denis-yuen: Ideally (in a world where we don't care about privacy), we would want one UUID per browser. So we know what a person is up to and what problems they encounter. 
In a more realistic world, one UUID per browser session would be enough to preserve privacy but let us know what the user path through the site looks like. 
And having said that, I guess you (@coverbeck ) were more thinking about tracing single REST requests as they made their way through the LB, nginx, and webservice weren't you ..

coverbeck: > And having said that, I guess you (@coverbeck ) were more thinking about tracing single REST requests as they made their way through the LB, nginx, and webservice weren't you ..

Yes, I was thinking of that case.

kathy-t: The way I implemented it, there would be one UUID per browser session. If the user goes to Dockstore in a new tab, or refreshed Dockstore on their current tab, a new UUID would be generated.

@coverbeck If you want, I could add a X-Request-ID header as well and generate a UUID for each request? In which case, @denis-yuen would you still want a X-Dockstore-UUID header that has a UUID for each browser session?

denis-yuen: Yes, I think the X-Request-ID header would be a good idea and can/should co-exist with the UUID header you have already added. Thanks!

coverbeck: Despite my lobbying for it, I am slightly concerned there might be a performance hit generating a UUID on every request. Can you please at least informally measure how long it takes to generate one? Thanks!

kathy-t: So I used performance.now() before and after the call that generates a UUID and calculated the difference to get an idea of how long it takes.

I logged the results in the console:
Screen Shot 2020-04-02 at 12 41 26 PM

Looks like it's roughly 0.03 - 0.05 ms to generate a UUID.

coverbeck: Neat and TIL: `performance.now()` -- I'd always used `Date.now()`. :)

But performance.now() returns a value in milliseconds, so how are you calculating that it completes in less than a millisecond? Is is actually 3-5 milliseconds? Probably OK if so, but as a further data point, are you running on a powerful machine?

kathy-t: I'm just doing ``endTime - startTime`` :) 

According to [this](https://developers.google.com/web/updates/2012/08/When-milliseconds-are-not-enough-performance-now), ``performance.now()`` is a double with microseconds in the fractional, so that's probably why the elapsed time is less than a millisecond. Just a note in case you use ``performance.now()`` in the future, [this](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) says that the precision of ``performance.now()`` is also dependent on browser settings, so for example Firefox would round it to the closest millisecond by default (I used Chrome and it didn't round).

I don't think I'm running on a particularly powerful machine. Here are the specs if you want to judge for yourself:
Screen Shot 2020-04-02 at 1 19 36 PM

coverbeck: Ah yes, I was trying it in FF and reading [this](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now).

kathy-t: So a UUID for each request should be okay then?

coverbeck: I'm OK with it.

PR #948, opened 2020-04-01
lgtm recommended changes

NatalieEO:
Tried to go through the alerts one by one and addressed the ones that made sense to me/didn't seem like I'd inadvertently break something 
Alerts here: https://lgtm.com/projects/g/dockstore/dockstore-ui2/?mode=list


src/app/workflow/permissions/permissions.component.ts
```diff
@@ -100,9 +97,8 @@ export class PermissionsComponent implements OnInit {
     }
   }
 
-  private handleError(e: HttpErrorResponse, defaultMessage: string) {
+  private handleError(e: HttpErrorResponse) {
```
coverbeck, line 100: Why did LGTM suggest you change this one? Oh, because we weren't using `message` from line 105, which means we were ignoring the `defaultMessage` parameter?

NatalieEO: Yeah, we were never making use of the default message so got rid of it

PR #951, opened 2020-04-03
Display ORCID for organization members

emlys:
dockstore/dockstore#3371
Screen Shot 2020-04-03 at 2 47 41 PM

No change in the UI if the user has no ORCID id stored.

Capitalized it like 'ORCID iD' because that's how it is on orcid.org, but it looks kind of strange. thoughts?


src/app/organizations/organization-members/organization-members.component.ts
```diff
@@ -125,4 +125,9 @@ export class OrganizationMembersComponent extends Base implements OnInit {
         }
       });
   }
+
+  getUserORCIDLink(organizationUser: OrganizationUser) {
+    const orcid = organizationUser.user.orcid;
```
coverbeck, line 130: If you put a console.log in here, I think you'll see this method is getting invoked a lot. If so, don't use a method in the HTML.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "1.9.0-alpha.5"
+    "webservice_version": "feature/organizations-orcid"
```
denis-yuen, line 6: From the build log, this won't work (only actual artifacts uploaded to artifacts.oicr.on.ca will) 
i.e. need a release from webservice develop to point at

denis-yuen:
The little green icon in this implementation looks nice 
https://reactome.org/userguide/claim-your-work

Icon shows up here too https://zenodo.org/record/3727255

Maybe we can sidestep the `iD` looking strange by replacing it with the icon

emlys:
Here's how it looks in bold without a colon:
Screen Shot 2020-04-06 at 9 31 14 AM

coverbeck:
> Here's how it looks in bold without a colon:

I think the colon looks better. :( Maybe because the label isn't distinct enough -- maybe the label would need to be even bolder...

That said, I shouldn't have sent you down this road; we should decide product-wide if we want to have colons or not. I think you should just stick with what you had (except trying Denis' comment).

emlys:
If it's inline with the text it's too small to see:
Screen Shot 2020-04-06 at 10 13 22 AM

denis-yuen:
Probably does need to be bigger, are the Zenodo or Reactome examples using a larger font size or is the icon just bigger than the text?

emlys:
Larger id logo:
Screen Shot 2020-04-06 at 10 25 28 AM

emlys:
Zenodo example looks to be bigger font size, reactome example has the logo bigger than the text

emlys:
IMO either of these is pretty good:
Screen Shot 2020-04-06 at 10 35 33 AM
Screen Shot 2020-04-06 at 10 35 57 AM

coverbeck:
> IMO either of these is pretty good:

If you're going with the icon, which I do like, then I think the colon does look particularly out of place. I would put a little more space between the id logo and the ORCID; right now the 3 items look like they are spaced evenly apart.

emlys:
With more space:
Screen Shot 2020-04-06 at 11 31 08 AM

src/app/organizations/organization-members/organization-members.component.ts
emlys: I'm only seeing it get invoked once per user with an ORCID

PR #952, opened 2020-04-03
Feature/3173/galaxy validation

garyluu:
For dockstore/dockstore#3173

Surfacing the validations on the frontend.  Webservice currently only has the descriptor validation but not the test parameter file


src/app/entry/entry-file-tab/state/entry-file-tab.service.ts
```diff
@@ -32,6 +32,48 @@ export class EntryFileTabService extends Base {
...
+   * @returns {(Object | null)}
+   * @memberof EntryFileTabService
+   */
+  public static getValidationMessageObject(toolFile: ToolFile | null, version: WorkflowVersion | null): Object | null {
+    if (version && version.validations && toolFile) {
+      const validations = version.validations;
+      const supportedValidationTypeEnum: Validation.TypeEnum[] = EntryFileTabService.toolFileToValdationTypeEnums(toolFile.file_type);
+      const foundValidation: Validation = validations.find(validation => supportedValidationTypeEnum.includes(validation.type));
+      if (foundValidation) {
+        const validationObject = JSON.parse(foundValidation.message);
```
coverbeck, line 51: Is it safe to assume this is always valid JSON? Otherwise an exception will get thrown.

PR #953, opened 2020-04-06
Feature/no refresh dockstore yml

agduncan94:
* remove sync button from dockstore yml workflows
* add refresh version option to action list
* basic test for dockstore yml workflow and refresh button

https://github.com/dockstore/dockstore/issues/3301


src/app/shared/refresh.service.ts
```diff
@@ -90,6 +90,28 @@ export class RefreshService {
...
   }
 
+  /**
+   * Refresh an individual version of a workflow and optionally updates the GA4GH files
+   * @param workflowId
+   * @param versionName
+   * @memberof RefreshService
+   */
+  refreshWorkflowVersion(workflowId: string, versionName: string): void {
+    const message = 'Refreshing ' + this.workflow.full_workflow_path + ' version ' + versionName;
```
garyluu, line 100: It was a mistake for this service to subscribe to another service. Avoid using `this.workflow` so that it's easier to remove, just pass it in from the component or workflowQuery.getValue().

agduncan94: Do you think it is worth creating an issue to fix this service subscribing to a service?

garyluu: wouldn't hurt I guess

PR #954, opened 2020-04-07
Feature/1265/inform users of migration

agduncan94:
First part of https://github.com/dockstore/dockstore/issues/3322

I thought the discussion could continue here. Currently have an always visible bar on GitHub legacy workflows. Only shows on my workflows.

Screen Shot 2020-04-07 at 2 39 01 PM

### DIFF
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -14,23 +14,22 @@
   ~    limitations under the License.
   -->
 <div class="p-3">
-  <mat-card class="alert alert-info">
-    <span
-      *ngIf="
-        workflow?.source_control_provider === 'GITHUB' &&
-        (workflow?.mode === WorkflowType.ModeEnum.FULL || workflow?.mode === WorkflowType.ModeEnum.STUB)
-      "
+  <mat-card
+    class="alert alert-info"
+    *ngIf="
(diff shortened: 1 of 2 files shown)


src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -14,6 +14,24 @@
...
+    class="alert alert-info"
+    *ngIf="
+      !isPublic &&
+      workflow?.source_control_provider === 'GITHUB' &&
+      (workflow?.mode === WorkflowType.ModeEnum.FULL || workflow?.mode === WorkflowType.ModeEnum.STUB)
+    "
+  >
+    Keep your workflow automatically in sync with GitHub with our new registration process. Click
+    <a
+      href="{{ Dockstore.DOCUMENTATION_URL }}/faq.html#why-should-i-migrate-my-existing-workflow-to-use-github-apps-and-a-dockstore-yml"
```
denis-yuen, line 27: Seems like a big enough page that it should get its own page, but this can be changed simply enough later :man_shrugging:

agduncan94: Yeah, that makes sense as it can be pretty important. Would that go in the getting started guide or advanced developer topics?

denis-yuen: Getting started , ideally we want this to (eventually?) be easy enough for all developers

PR #960, opened 2020-04-13
Add short friendly name to extended descriptor languages

garyluu:
For dockstore/dockstore#3292

Originally going to use a short friendly name from the webservice. but realized merging the frontend descriptor language with the backend descriptor language bean is required first

Just making a frontend-only shortFriendlyName field for now.

### DIFF
--- a/src/app/entry/extendedDescriptorLanguage.ts
+++ b/src/app/entry/extendedDescriptorLanguage.ts
@@ -91,7 +91,7 @@ const superService: ExtendedDescriptorLanguage = {
 
 const superGalaxy: ExtendedDescriptorLanguage = {
   value: 'gxformat2',
-  shortFriendlyName: 'Galaxyv2',
+  shortFriendlyName: 'GalaxyV2',
   friendlyName: 'Galaxy Workflow Format 2',
   descriptorPathPattern: '^/([^/?:*|<>]+/)*[^/?:*|<>]+.(ga|yaml|yml)',
   descriptorPathPlaceholder: 'e.g. /Dockstore.yml',
--- a/src/app/entry/extendedDescriptorLanguage.ts
+++ b/src/app/entry/extendedDescriptorLanguage.ts
@@ -1,9 +1,10 @@
 import { DescriptorLanguageBean, SourceFile, ToolDescriptor, Workflow } from 'app/shared/swagger';
 
 /**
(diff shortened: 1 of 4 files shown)


src/app/entry/extendedDescriptorLanguage.ts
```diff
@@ -86,6 +91,7 @@ const superService: ExtendedDescriptorLanguage = {
 
 const superGalaxy: ExtendedDescriptorLanguage = {
   value: 'gxformat2',
+  shortFriendlyName: 'Galaxyv2',
```
coverbeck, line 94: I vote for `GalaxyV2`

src/app/entry/extendedDescriptorLanguage.ts
```diff
@@ -86,6 +91,7 @@ const superService: ExtendedDescriptorLanguage = {
 
 const superGalaxy: ExtendedDescriptorLanguage = {
   value: 'gxformat2',
+  shortFriendlyName: 'GalaxyV2',
```
denis-yuen, line 94: Unfortunately, the shortFriendlyName might have to be "Galaxy"
The friendlyName "Galaxy Workflow Format"

For example, the path *.ga would be format 1, *.yml or *.yaml would be format 2. 
But we don't really track this and that not tracking it in this bean would seem to be consistent with WDL which only gets one enum value regardless of whether it is WDL draft-2 or WDL v1

src/app/search/map-friendly-values.pipe.ts
```diff
@@ -29,10 +37,27 @@ export class MapFriendlyValuesPipe implements PipeTransform {
...
+        ['nfl', this.shortFriendlyNFLName],
+        ['NFL', this.shortFriendlyNFLName],
+        [Workflow.DescriptorTypeEnum.Gxformat2, this.shortFriendlyGalaxyName]
+      ])
+    ],
     [
       'descriptor_type',
-      new Map([['CWL', 'CWL'], ['WDL', 'WDL'], ['cwl', 'CWL'], ['wdl', 'WDL'], ['nfl', 'Nextflow'], ['NFL', 'Nextflow']])
+      new Map([
+        ['CWL', this.shortFriendlyCWLName],
```
denis-yuen, line 53: Nitpick: pairs of CWL and WDL are interwoven, NFL is separate

PR #961, opened 2020-04-20
Automate daily tests

emlys:
Automated tests on dev.dockstore.net are going to fail for now because they depend on `data-cy` attributes which I added to the UI on this branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,11 +247,11 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: "40 * * * *" # This is 4am PDT / 3am PST
+          cron: "0 11 * * *" # This is 4am PDT / 3am PST
           filters:
             branches:
               only:
-                - feature/SEAB-815/daily-tests
+                - develop
     jobs:
       - uptime_monitor
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,45 @@ executors:
           POSTGRES_PASSWORD: postgres
           POSTGRES_HOST_AUTH_METHOD: trust
 jobs:
+
+  uptime_monitor:
+    working_directory: ~/repo
+    docker:
+      - image: circleci/buildpack-deps:18.04-browsers
+    steps:
+      - checkout
+      - run:
+          name: Checkout merge commit (PRs only)
+          command: |
+            if [[ -n "${CIRCLE_PULL_REQUEST}" ]]
+            then
+              git fetch origin +refs/pull/${CIRCLE_PULL_REQUEST##*/}/merge:
+              git checkout -qf FETCH_HEAD
+            fi
+      - install_container_dependencies
+      - restore_cache:
+          key: dep-cache-{{ checksum "./package-lock.json" }}
+      - run:
+          name: Install dependencies
+          #Only run 'npm ci' if node_modules was not restored from cache
+          command: |
+            if [ ! -d "./node_modules" ]
+            then
+              bash -i -c 'npm ci'
+            fi
+      - run:
+          name: Build
+          command: bash -i -c 'npm run build.prod'
+      - run:
+          name: Install cypress dependencies
+          command: sudo apt install libgtk2.0-0 -yq
+      - run:
+          name: Install cypress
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -204,6 +243,18 @@ workflows:
...
               only: /^develop/
+
+  nightly:
+    triggers:
+      - schedule:
+          cron: "0 11 * * *" # This is 4am PDT / 3am PST
+          filters:
+            branches:
+              only:
+                - feature/SEAB-815/daily-tests
```
coverbeck, line 254: This needs to be develop

cypress/integration/manualTests/monitor.ts
```diff
@@ -50,3 +52,27 @@ describe('Monitor homepage links', () => {
...
+      .click();
+    cy.get('[data-cy=workflowColumn]')
+      .first()
+      .within(() => {
+        cy.get('a').click(); // click on the link to the first workflow
+      });
+  });
+
+  it('get the svg icons', () => {
+    cy.get('[data-cy=dnanexusIcon]').should('exist');
```
coverbeck, line 74: Have you tested if this catches the case when the icon is not present?

You can make the network request fail locally by blocking it.

I'm concerned this element may be present if the request to get the icon fails; successfully fetching the icon appends SVG DOM elements, which I think you may need to look for.

cypress/integration/manualTests/monitor.ts
```diff
@@ -50,3 +52,27 @@ describe('Monitor homepage links', () => {
     });
   });
 });
+
+describe('Test launch button icons', () => {
+  it('find a WDL workflow', () => {
+    cy.visit('/search');
+    cy.wait(10000); // wait for the tabs to become clickable
```
garyluu, line 59: should wait for something specific instead.  like the tab has content (which may imply that it is clickable) or something else.

cypress/integration/manualTests/monitor.ts
```diff
@@ -50,3 +52,27 @@ describe('Monitor homepage links', () => {
...
+    goToTab('Workflows');
+
+    // click to sort by descriptor type descending so WDL is at the top
+    cy.get('[data-cy=descriptorType]')
+      .click()
```
garyluu, line 64: what's the purpose of the double click?

package.json
```diff
@@ -16,7 +16,8 @@
     "compodoc": "npx compodoc -p src/tsconfig.compodoc.json --output docs",
     "test": "npx ng test",
     "lint": "npx ng lint",
-    "e2e": "npx ng e2e"
+    "e2e": "npx ng e2e",
+    "test-dev": "npx cypress run -c baseUrl=https://dev.dockstore.net -s \"cypress/integration/manualTests/monitor.ts\""
```
coverbeck, line 20: Can you please add test-staging and test-prod as well. Even though they won't be automated at this point, it will be nice to run the tests manually against staging/prod when we deploy a new release.

cypress/integration/manualTests/monitor.ts
emlys: the first click sorts by descriptor type ascending, the second click reverses the sort so that the first workflow is WDL

cypress/integration/manualTests/monitor.ts
emlys: You're right, should've caught that. Changed it to check for a svg element

package.json
```diff
@@ -79,7 +82,7 @@
     "@types/jquery": "^3.3.31",
     "@types/node": "^10.14.19",
     "codelyzer": "^5.0.1",
-    "cypress": "4.0.1",
+    "cypress": "^4.0.1",
```
coverbeck, line 85: Why did you change this? It didn't end up changing the package-lock.json (did you verify it didn't?), so it probably doesn't matter, but generally best to avoid unnecessary changes if possible.

emlys: I didn't change it, I thought it was changed when I rebuilt it or something.  Can change it back if it's not important

coverbeck: I feel like you had to have changed it :) but it's fine

PR #963, opened 2020-04-20
Feature/file tab refactor

garyluu:
Refactoring the file tab to be more correct and more intuitive (though more lines):

See https://docs.google.com/spreadsheets/d/1tDsm87okoWwaeO9Ae4cc5vapWbTSbXI5eqNSWEOrUcE/edit#gid=762878948  for changes

This is a prerequesite for fixing hosted workflows for plugin languages + services (if there ever is one)

Also added a simple alert when there's no files in the new component

### DIFF
--- a/src/app/workflow/workflow.component.html
+++ b/src/app/workflow/workflow.component.html
@@ -234,65 +234,88 @@
           </mat-tab>
           <mat-tab id="fileTab" label="Files">
             <ng-template matTabContent>
-              <!-- For old languages, we use the old files component because it's well tested, for new unknown languages, use the new plugin -->
-              <div
-                *ngIf="
-                  workflow.descriptorType !== 'CWL' && workflow.descriptorType !== 'WDL' && workflow.descriptorType !== 'NFL';
-                  else FilesTabBioWorkflow
-                "
-                class="p-3"
-              >
-                <div *ngIf="(extendedWorkflow$ | async)?.workflowVersions.length > 0; else noVersionsFilesTab">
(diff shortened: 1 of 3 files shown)


src/app/workflow/workflow.component.html
```diff
@@ -233,72 +233,97 @@ <h3>
...
+              <ng-template #noVersionsFilesTab>
+                <div [ngSwitch]="workflow.mode" class="p-3">
+                  <mat-card *ngSwitchCase="WorkflowModel.ModeEnum.STUB" class="alert alert-info" role="alert">
+                    <mat-icon>info</mat-icon> To see files, please refresh the {{ entryType }}.
+                  </mat-card>
+                  <div *ngSwitchCase="WorkflowModel.ModeEnum.HOSTED">
+                    <ng-container *ngTemplateOutlet="hostedComponent"></ng-container>
+                  </div>
+                  <mat-card *ngSwitchCase="WorkflowModel.ModeEnum.DOCKSTOREYML" class="alert alert-info" role="alert">
+                    <mat-icon>info</mat-icon> This {{ entryType }} does not have any versions. Create a release/tag on GitHub with a valid
```
Ldcabansay, line 277: Out of curiosity, when something isn't a stub, how does no versions happen? 

Also in comparison with the first noVersionFilesTab case, does this mean files were found (and displayed) but somehow has no versions? Or is this no files and no versions?

garyluu: Some of these cases are impossible, but showing something instead of nothing.

garyluu: If there's no versions, it's assumed to have no files.

PR #964, opened 2020-04-21
Feature/seab 1036/licenses

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-1036

As noted in the readme, there isn't a way to permanently change a dependency's license from unknown to something accurate using this package. Which makes it more difficult to detect real changes when dependencies/licenses are updated. It doesn't seem like the package.json gets updated very often, so I feel like it's okay for this to be a manual process for now?

Also, there's still an unknown license listed in the csv. The corresponding repo, did not contain license information when I checked.


README.md
```diff
@@ -116,6 +116,15 @@ Run `npm run start -- --host=<host>` for a dev server. Navigate to `http://<host
 
 Run `npm update`. This will automatically update package.json and package-lock.json.
 
+When you update a dependency in the package.json, make sure to update the third party licenses file `THIRD-PARTY-LICENSES.csv`. 
+To update, run the command `license-checker  --excludePackages dockstore-ui2@2.3.0 --csv > UPDATED-LICENSES.csv`.
```
coverbeck, line 120: I think there should a script in the package.json that does this, so that we can just run `npm run license', and we don't have to type in those options.

Also, I think it should just write to THIRD-PARTY-LICENSES.csv, and git will detect if there are differences or not; no need to run diff manually.

The downside to this is handling the unknown/unlicensed cases, but in our discussion you said that one that was unknown, you couldn't determine even by going to the repo, so I think we just leave it like that.

package.json
```diff
@@ -54,6 +54,7 @@
     "elasticsearch-browser": "^15.0.0",
     "font-awesome": "^4.7.0",
     "jquery": "^3.4.0",
+    "license-checker": "^25.0.1",
```
garyluu, line 57: seems more like a devDep to me

README.md
denis-yuen: Hmmm, it seems that this way there will be no way of permanently saying a certain dependency is X license. 
Not sure how to work around that. 
One option that might help might be to run the license checker on travis/circle and fail the build if the license checker seems to be changing the number of lines in `UPDATED-LICENSES.csv`

THIRD-PARTY-LICENSES.csv
```diff
@@ -0,0 +1,1376 @@
+"module name","license","repository"
```
denis-yuen, line 1: Has this file been manually updated?

NatalieEO: No, the file has not been manually updated, but might have to be in the future.

package.json
NatalieEO: Moved to devDependencies

README.md
NatalieEO: added a test on circleci to check if the package.json and package-lock.json have been changed in comparison to what's on develop. If they have, check the third-party-licenses.csv is also updated compared to develop. Fail if not.

.circleci/config.yml
```diff
@@ -169,24 +176,31 @@ workflows:
...
+          filters:
+            tags:
+              only: /.*/
       - integration_test_1:
           filters:
             tags:
               only: /.*/
           requires:
             - lint_unit_test_coverage
+            - licenses_test
```
coverbeck, line 189: Doesn't this only need to be run once? Perhaps after unit tests?

package.json
```diff
@@ -13,6 +13,7 @@
     "prebuild.prod": "npm run prebuild",
     "build.prod": "npx ng build --progress false --prod",
     "build": "npx ng build --progress false",
+    "license": "license-checker  --excludePackages dockstore-ui2@2.3.0 --csv > THIRD-PARTY-LICENSES.csv",
```
coverbeck, line 16: Does this work if you do `npm run license`? I could be wrong, but unless you've installed license globally, I think you need to precede this with npx.

.circleci/config.yml
garyluu: This is how it looks like: https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/2848/workflows/756bf50c-2c4e-4ba1-babf-2b46f6a9154b

package.json
NatalieEO: Yeah, just tried it again and it worked with `npm run license`

.circleci/config.yml
NatalieEO: Yeah, this runs once and must pass before the unit tests are able to run.

package.json
coverbeck: I don't understand why, but I tried it as well and it worked. I guess package.json scripts have an implied npx or something.

garyluu: The way I see it, if there's a global license-checker and a local one, npx will ensure the local one runs.

coverbeck: Hmm, interesting, when you do `npm run`, it apparently automatically adds `node_modules/.bin` to your path. Never knew. Ideally you would still use npx for Gary's case, but not crucial.

PR #968, opened 2020-04-24
Feature/support hosted galaxy

garyluu:
Support hosted Galaxy workflows.

Kept logic mostly the same.
Moved functions into a service.
Add tests (~92% coverage).
Ran strictNullChecks on the file (which is why it looks so complicated)
A lot of the existing logic is odd. but for better or worse, keeping it the same for now.

### DIFF
--- a/src/app/shared/code-editor-list/code-editor-list.component.spec.ts
+++ b/src/app/shared/code-editor-list/code-editor-list.component.spec.ts
@@ -10,6 +10,7 @@ import { MatToolbarModule } from '@angular/material/toolbar';
 import { MatTooltipModule } from '@angular/material/tooltip';
 import { ClipboardModule } from 'ngx-clipboard';
 import { WorkflowService } from '../state/workflow.service';
+import { ToolDescriptor } from '../swagger';
 import { PrivateFileDownloadPipe } from './../../shared/entry/private-file-download.pipe';
 import { PrivateFilePathPipe } from './../../shared/entry/private-file-path.pipe';
 import { PublicFileDownloadPipe } from './../../shared/entry/public-file-download.pipe';
--- a/src/app/shared/code-editor-list/code-editor-list.component.spec.ts
+++ b/src/app/shared/code-editor-list/code-editor-list.component.spec.ts
@@ -10,7 +10,6 @@ import { MatToolbarModule } from '@angular/material/toolbar';
 import { MatTooltipModule } from '@angular/material/tooltip';
 import { ClipboardModule } from 'ngx-clipboard';
 import { WorkflowService } from '../state/workflow.service';
-import { ToolDescriptor } from '../swagger';
 import { PrivateFileDownloadPipe } from './../../shared/entry/private-file-download.pipe';
 import { PrivateFilePathPipe } from './../../shared/entry/private-file-path.pipe';
 import { PublicFileDownloadPipe } from './../../shared/entry/public-file-download.pipe';
--- a/src/app/shared/code-editor-list/code-editor-list.component.ts
+++ b/src/app/shared/code-editor-list/code-editor-list.component.ts
@@ -69,7 +69,7 @@ export class CodeEditorListComponent {
    * @memberof CodeEditorListComponent
    */
   isPrimaryDescriptor(path: string): boolean {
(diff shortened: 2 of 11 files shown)


src/app/shared/code-editor-list/code-editor-list.component.spec.ts
```diff
@@ -10,6 +10,7 @@ import { MatToolbarModule } from '@angular/material/toolbar';
 import { MatTooltipModule } from '@angular/material/tooltip';
 import { ClipboardModule } from 'ngx-clipboard';
 import { WorkflowService } from '../state/workflow.service';
+import { ToolDescriptor } from '../swagger';
```
coverbeck, line 13: Why did this get added? It's not being used that I can see.

garyluu:
Hmm...I have a few unused parameters.

src/app/shared/code-editor-list/code-editor-list.component.ts
```diff
@@ -76,73 +61,20 @@ export class CodeEditorListComponent {
...
-   * @return {boolean}      Is path for primary descriptor
+   *
+   * @static
+   * @param {string} path Path to check
+   * @returns {boolean} Is path for primary descriptor
+   * @memberof CodeEditorListComponent
    */
   isPrimaryDescriptor(path: string): boolean {
-    return path === '/Dockstore.cwl' || path === '/Dockstore.wdl' || path === this.NEXTFLOW_CONFIG_PATH || path === this.NEXTFLOW_PATH;
+    return CodeEditorListService.isPrimaryDescriptor(path);
```
coverbeck, line 73: You said in your PR comment that a lot of this logic is odd and you're keeping it that way for now, so it is probably beyond the scope of this, but without grokking the whole flow, what happens if the primary descriptor is 'alignment.wdl'?

src/app/shared/code-editor-list/code-editor-list.service.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  static isPrimaryDescriptor(path: string | null): boolean {
+    if (!path) {
+      return false;
+    }
+    const primaryDescriptors = [
+      '/Dockstore.cwl',
+      '/Dockstore.wdl',
+      CodeEditorListService.NEXTFLOW_CONFIG_PATH,
+      CodeEditorListService.NEXTFLOW_PATH
```
coverbeck, line 69: What about Galaxy?

src/app/shared/code-editor-list/code-editor-list.service.ts
```diff
@@ -0,0 +1,291 @@
...
+   * Determines whether to show the current sourcefile based on the descriptor type and tab
+   *
+   * @static
+   * @param {(SourceFile.TypeEnum | null | undefined)} sourcefileType
+   * @param {FileCategory} fileCategory
+   * @param {ToolDescriptor.TypeEnum} descriptorType
+   * @returns {boolean} Whether to show sourcefile or not
+   * @memberof CodeEditorListService
+   */
+  static showSourcefile(
```
coverbeck, line 29: Just curious, because I've wondered about which way to tackle this before and I'm not clear on when to choose which, what is your reasoning for having static methods instead of injecting an instance of the service into the constructor of classes using this?

src/app/shared/code-editor-list/code-editor-list.component.ts
garyluu: Everything is messed up, webservice too

src/app/shared/code-editor-list/code-editor-list.service.ts
garyluu: Doesn't really seem to affect it for some reason.

src/app/shared/code-editor-list/code-editor-list.service.ts
garyluu: If it can be made static, make it static. This file will likely change in the future because I'll be trying to merge the descriptor language bean from the backend with the frontend which means I'll need to inject a few things into this service.

src/app/shared/code-editor-list/code-editor-list.service.ts
garyluu: or maybe it does...hmm

PR #969, opened 2020-04-24
Always use new component for Dockstore.YML workflows

garyluu:


denis-yuen:
To be clear, this seems to contradict the code comment "For old languages, we use the old files component because it's well tested, for new unknown languages, use the new component "
This will make old languages use the new component if they use GitHub apps, are any problems anticipated/expected with this change?

garyluu:
> To be clear, this seems to contradict the code comment "For old languages, we use the old files component because it's well tested, for new unknown languages, use the new component "
> This will make old languages use the new component if they use GitHub apps, are any problems anticipated/expected with this change?

Apparently when Dockstore.yml was first added, it was already using the newComponent. my previous PR changed it to the old component.  So I'm simply reverting my PR previous change.

PR #971, opened 2020-04-28
Don't reselect entry when myworkflows change

garyluu:
Kind of for dockstore/dockstore#3287

So the `recomputeWhatEntryToSelect` function was getting executed everytime this.workflows, this.sharedworkflows, route was update.  In most cases, this was fine.  However, there's weird cases in timings when the order of update didn't match the recomputeWhatEntryToSelect resolve order which causes some odd behaviour.  This would cause the frontend to rapidly select two different entries and then end up with the wrong entry selected (with the correct route).

This PR removes a recomputeWhatEntryToSelect  so that it's only automatically resolved when there's a route change.  What this means is that any this.workflows or this.sharedworkflows changes resulting from other actions will not cause the entry selection, you will have to trigger reselection manually.

I couldn't reproduce the original issue, it'd be great if people tested this out locally or dev.dockstore.net (if it gets merged).


src/app/shared/state/workflow.service.ts
```diff
@@ -17,6 +17,20 @@ export class WorkflowService {
   copyBtn$ = this.copyBtnSource.asObservable();
   constructor(private workflowStore: WorkflowStore, private extendedWorkflowService: ExtendedWorkflowService) {}
 
+  /**
+   * Converts the mapping of roles to workflows to a concatentation of all the workflows
+   * @param workflows mapping returned by shared workflows endpoint
+   */
+  static convertSharedWorkflowsToWorkflowsList(workflows: Array<any>): Array<Workflow> {
```
coverbeck, line 24: Add `| null` for return type :)

PR #972, opened 2020-04-28
remove sync org and sync my services

agduncan94:
Removes the Sync org and Sync my services buttons.

https://github.com/dockstore/dockstore/issues/3416


garyluu:
Are we sure there's no use case for it?

agduncan94:
Maybe we can brainstorm. Previously it was used to create stubs for services that should exist (have a github app install) but are not yet on dockstore. They also add the user to any github app they should have access to.

The first issue no longer exists (no more service stubs), but the second one kind of does. Having refresh in the same place as it exists for workflows may be confusing. Perhaps this is better as a button on the account page (something like sync with GitHub).

denis-yuen:
What solution is there in 1.9.0 for the second use case right now?

agduncan94:
No solution yet for 1.9.0, but I'll make an issue for it.

agduncan94:
@denis-yuen Is this ok to review then since it seems like the solution would be a different location in the UI?

PR #975, opened 2020-04-30
Add logout page

coverbeck:
dockstore/dockstore#3363

Verbiage warns user that they may not be as logged out as they think.

In looking at code to understand how to do this, noticed an unused field
in track-login.service.ts, so removed it. Also took IDE recommendation
to shorten an import.

Screen Shot 2020-04-30 at 1 13 51 PM

### DIFF
--- a/src/app/logout/logout.component.html
+++ b/src/app/logout/logout.component.html
@@ -20,8 +20,7 @@
   <div class="column">
     <p class="lead">
       You have logged out of Dockstore. Note that if you are still logged into GitHub or Google, a user may be able to log back into
-      Dockstore without being prompted for credentials. If on a public computer, we recommend also signing out of GitHub and Google. Click
-      <a routerLink="/login">here</a> to login again.
+      Dockstore without being prompted for credentials. If on a public computer, we recommend also signing out of GitHub and Google.
     </p>
   </div>
 </div>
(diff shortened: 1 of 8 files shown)


src/app/logout/logout.component.html
```diff
@@ -0,0 +1,27 @@
...
+-->
+<app-header>
+  Logged Out
+</app-header>
+<div class="container">
+  <div class="column">
+    <p class="lead">
+      You have logged out of Dockstore. Note that if you are still logged into GitHub or Google, a user may be able to log back into
+      Dockstore without being prompted for credentials. If on a public computer, we recommend also signing out of GitHub and Google. Click
+      <a routerLink="/login">here</a> to login again.
```
Ldcabansay, line 24: Probably the log-in again part is unnecessary

coverbeck: Re-requesting review just to make sure I understand you. Screenshot at top updated.

coverbeck:
> Bonus would be to avoid bootstrap 3 or even bootstrap 4

Just blindly copied some existing HTML and didn't even notice there were Bootstrap classes being used.

coverbeck:
@garyluu Switched to flex, which is new to me. Let me know what you think.

Screen Shot 2020-04-30 at 4 20 44 PM

PR #977, opened 2020-05-07
Feature/3410/change dockstore yml file tab

agduncan94:
Services use the new TRS component (matches prod) and Workflows with a dockstore.yml use the existing component. Modified the descriptor component to also work with OTHER files. In this case, OTHER is the dockstore.yml.

https://github.com/dockstore/dockstore/issues/3410

### DIFF
--- a/src/app/workflow/descriptors/descriptors.component.ts
+++ b/src/app/workflow/descriptors/descriptors.component.ts
@@ -35,6 +35,7 @@ import { FilesService } from '../files/state/files.service';
 export class DescriptorsWorkflowComponent extends EntryFileSelector {
   @Input() id: number;
   @Input() entrypath: string;
+  @Input() otherFiles: boolean;
   @Input() set selectedVersion(value: WorkflowVersion) {
     this.onVersionChange(value);
     this.checkIfValid(true, value);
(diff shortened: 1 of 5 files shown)


src/app/workflow/workflow.component.html
```diff
@@ -248,7 +248,12 @@ <h3>
                     <ng-container *ngTemplateOutlet="hostedComponent"></ng-container>
                   </div>
                   <div *ngSwitchCase="WorkflowModel.ModeEnum.DOCKSTOREYML">
-                    <ng-container *ngTemplateOutlet="newComponent"></ng-container>
+                    <div *ngIf="workflow.descriptorType === WorkflowModel.DescriptorTypeEnum.Service">
```
garyluu, line 251: What's so special about services?

src/app/workflow/descriptors/descriptors.component.ts
```diff
@@ -35,6 +35,7 @@ import { FilesService } from '../files/state/files.service';
 export class DescriptorsWorkflowComponent extends EntryFileSelector {
   @Input() id: number;
   @Input() entrypath: string;
+  @Input() otherFiles: boolean;
```
garyluu, line 38: so descriptors is going to show OTHER files then?

agduncan94: If true, shows only other files, if false, shows descriptor files. I just didn't want to clone an entire component just for the other files.

src/app/workflow/workflow.component.html
agduncan94: It is a bit of a different scenario, as a service is a dockstore.yml plus an array of files. The dockstore.yml is both a dockstore.yml and a descriptor (kind of). So I thought for simplicity to just keep it with the new component, since that is what we currently do and it works fine.

agduncan94:
@garyluu You can test workflow with https://github.com/agduncan94/ghapps-single-workflow and service with https://github.com/agduncan94/test-service

agduncan94:
Upon further review, there is an issue with this PR so I am going to close it and try to fix.

PR #980, opened 2020-05-08
My Workflows

garyluu:
For frontend part of dockstore/dockstore#3434

The CollectionEntry changes surprisingly isn't typed here so it doesn't fail (not a good thing).  
MyWorkflows type has been left unchanged so there's no changes here.

I tested this locally, but there will need to be another PR to update the package.json to 1.8.6 once that's released (and not use the CircleCI jar).

### DIFF
--- a/scripts/generate-openapi-script.sh
+++ b/scripts/generate-openapi-script.sh
@@ -8,11 +8,14 @@ GENERATOR_VERSION="3.3.4"
 BASE_PATH="https://raw.githubusercontent.com/dockstore/dockstore/$npm_package_config_webservice_version"
 # DOCKSTORE-2428 - demo how to add new workflow language, generate from local copy of swagger
 # BASE_PATH="../dockstore"
+CIRCLE_CI_PATH="https://3855-33383826-gh.circle-artifacts.com/0/tmp/artifacts"
 
 wget --no-verbose https://repo.maven.apache.org/maven2/org/openapitools/openapi-generator-cli/${GENERATOR_VERSION}/openapi-generator-cli-${GENERATOR_VERSION}.jar -O openapi-generator-cli.jar
 rm -Rf src/app/shared/swagger
 rm -Rf src/app/shared/openapi
-java -jar openapi-generator-cli.jar generate -i ${BASE_PATH}/dockstore-webservice/src/main/resources/swagger.yaml -l typescript-angular -o src/app/shared/swagger -c swagger-config.json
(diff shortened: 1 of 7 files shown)


scripts/generate-openapi-script.sh
```diff
@@ -8,11 +8,14 @@ GENERATOR_VERSION="3.3.4"
 BASE_PATH="https://raw.githubusercontent.com/dockstore/dockstore/$npm_package_config_webservice_version"
 # DOCKSTORE-2428 - demo how to add new workflow language, generate from local copy of swagger
 # BASE_PATH="../dockstore"
+#CIRCLE_CI_PATH="https://3855-33383826-gh.circle-artifacts.com/0/tmp/artifacts"
```
denis-yuen, line 11: Intentionally left in? 
If so, might need to code comment `3855-33383826` (is that unique to "our" repo on circle?)

scripts/run-webservice-script.sh
```diff
@@ -4,7 +4,10 @@ set -o pipefail
 set -o nounset
 set -o xtrace
 
-wget -O dockstore-webservice.jar --no-verbose --tries=10 https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-webservice/${npm_package_config_webservice_version}/dockstore-webservice-${npm_package_config_webservice_version}.jar
+#JAR_PATH="https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-webservice/${npm_package_config_webservice_version}/dockstore-webservice-${npm_package_config_webservice_version}.jar"
+JAR_PATH="https://3855-33383826-gh.circle-artifacts.com/0/tmp/artifacts/dockstore-webservice-1.8.6-SNAPSHOT.jar"
```
denis-yuen, line 8: Ditto

scripts/generate-openapi-script.sh
garyluu: It's intentionally left in for when I or others need to use a JAR for CircleCI.  The URL is apparent when you already uploaded a JAR via CircleCI

scripts/run-webservice-script.sh
garyluu: This is going to be replaced by the next 1.8.6 package.json PR

scripts/generate-openapi-script.sh
denis-yuen: Sounds good, add that as a code comment linking to https://wiki.oicr.on.ca/display/DOC/Dockstore+Releases#DockstoreReleases-PreDeploy and add a screenshot of where to look in CircleCI, thanks!

PR #984, opened 2020-05-21
add button for sync with existing workflows

agduncan94:
Screen Shot 2020-05-21 at 2 41 59 PM
No idea where to add this or what icon. This one looks too much like refresh. The gist of what the button does it add the user to existing workflows on Dockstore that others have registered but that the user should have access to.

I was gonna put where the message  "Don't see an organisation..." on the left sidebar, but it looks too cluttered.

I've also added it to the note when the user has no workflows.

Making this a draft as it both relies on the new web service release and I don't know where to put the button.

https://github.com/dockstore/dockstore/issues/3486


garyluu:
I'm...not sure what I'm looking at in the image.

agduncan94:
@garyluu sorry I should've been more clear. The refresh all button was removed in 1.8.0, so this new button where the refresh used to be is the new button I am adding.

NatalieEO:
idk if I like my own ideas but...
1. what if that my workflows sections was an accordion like the others and had 'register workflows +' and the 'add yourself to workflows... (icon)`?
2. An actions dropdown with the options?

coverbeck:
>     2. An actions dropdown with the options?

I like this one.

agduncan94:
So the + and sync buttons in the image would be replaced by [Actions v], then clicking on actions would show "register workflows" and "add me to existing Dockstore workflows" or something like that?

NatalieEO:
> So the + and sync buttons in the image would be replaced by [Actions v], then clicking on actions would show "register workflows" and "add me to existing Dockstore workflows" or something like that?

Yeah something like that

agduncan94:
Would we want the word "Action" or would the "..." icon be better?

agduncan94:
Some options
1) Actions with a dropdown which includes add workflow(s) and add a user to existing workflows. https://material.angular.io/components/menu/overview
2) Have two icon buttons as seen in the screenshot in this PR.
3) Context menu. Similar to option 1 but uses vertical ... icon (https://lh3.googleusercontent.com/djK1rThQ30yU32azhWve0J0rMUkJLzPnitUbEwYf5Hnf5MZgmmFlY-LbGj6n51UeyroPUywoqirccgyzAPAaNNCl0gKMqmqCsnA5Dg=w1064-v0). This is used in Gmail for example.
4) Floating action button - we moved away from this so probably not (https://material.io/components/buttons-floating-action-button)

agduncan94:
We decided to use the Slack approach, main action icon shown (ex. Add workflows), vertical ... is used for extra actions.

agduncan94:
Screen Shot 2020-05-29 at 10 41 39 AM

garyluu:
I personally think it's better to leave the icon in the menu out.  If we add a new item to the menu, we won't have to worry about finding a unique icon for it.

src/app/myworkflows/my-workflow/my-workflow.component.ts
```diff
@@ -210,6 +212,16 @@ export class MyWorkflowComponent extends MyEntry implements OnInit {
     this.myWorkflowsService.registerEntry(this.entryType);
   }
 
+  addToExistingWorkflows(): void {
+    this.userQuery.user$
```
garyluu, line 216: don't subscribe, if the user in the state changes (for w/e reason), then you'd have two addUserToWorkflows calls.  

Just do userQuery.user$.getValue() to get the current user.  disable button when user is falsey.

src/app/myworkflows/my-workflow/my-workflow.component.html
```diff
@@ -129,6 +138,21 @@ <h3 class="sidebar-header">Shared with me</h3>
                 >
                   Register {{ entryType$ | async | titlecase }}
                 </button>
+
+                <li *ngIf="entryType === EntryType.BioWorkflow">
```
garyluu, line 142: Wrap both ngIf things in one tag

src/app/shared/user/user.service.ts
```diff
@@ -38,6 +40,17 @@ export class UserService {
     });
   }
 
+  addUserToWorkflows(user: User): void {
+    this.alertService.start('Adding user to existing workflows on Dockstore');
+    this.usersService.addUserToDockstoreWorkflows(user.id).subscribe(
```
garyluu, line 45: Hmm...surprised this worked, thought swagger was disabled for new endpoints.  In any case, use the openapi service and model, cast the models for compatibility.

src/app/myworkflows/my-workflow/my-workflow.component.html
```diff
@@ -54,6 +54,20 @@
...
                         </button>
+                        <button mat-icon-button [matMenuTriggerFor]="myWorkflowsActions"><mat-icon>more_vert</mat-icon></button>
+                        <mat-menu #myWorkflowsActions="matMenu">
+                          <button
+                            mat-menu-item
+                            color="primary"
+                            type="button"
+                            (click)="addToExistingWorkflows()"
+                            [disabled]="isRefreshing$ | async"
+                            matTooltip="Discover workflows added by others using your linked source control accounts"
```
coverbeck, line 65: I think an explicit question mark button would be better. I've found that many users don't know to look for tooltips.

That could wait for a future PR, as it's a bit more work to add.

src/app/myworkflows/my-workflow/my-workflow.component.html
```diff
@@ -54,6 +54,20 @@
...
+                            matTooltipPosition="after"
+                          >
+                            <span>Find Existing Workflows</span>
```
coverbeck, line 68: I prefer "Discover" over "Find", like you have in the tooltip.

I also think this is a little vague... where is it going to find them?

Discover Existing Dockstore Workflows?

src/app/myworkflows/my-workflow/my-workflow.component.html
```diff
@@ -129,6 +138,21 @@ <h3 class="sidebar-header">Shared with me</h3>
...
+                  If there are already workflows on Dockstore added by other users that you should have access to, press the button below.
+                </li>
+                <button
+                  *ngIf="entryType === EntryType.BioWorkflow"
+                  mat-raised-button
+                  color="primary"
+                  type="button"
+                  (click)="addToExistingWorkflows()"
+                  [disabled]="isRefreshing$ | async"
+                  matTooltip="Add yourself to existing Dockstore workflows that you should have access to on remote sites."
```
coverbeck, line 152: Suggest removing "should".

Instead of "remote sites", suggest "source control sites".

src/app/shared/user/user.service.ts
agduncan94: I couldn't get patch to work without having both swagger and OpenAPI3 annotations, but ya I'll change to openapi.

agduncan94: I cannot add both the usersService swagger and usersService openAPI in the same file. Other endpoints being called are not implemented for openapi, so I am going to keep this as swagger.

garyluu: Not sure how I didn't notice the swagger PATCH 

`import { UsersService as OpenAPIUsersService } from '../../../shared/openapi';`

We use OpenAPI for a lot of new endpoints.

agduncan94:
For future reference, I tried to get working with OpenAPI. https://github.com/dockstore/dockstore/issues/3517

denis-yuen:
(on build success)

PR #987, opened 2020-05-27
Refresh individual repos instead of org

coverbeck:
Still need some more tests; waiting to assign reviewers until then.

dockstore/dockstore#3485

When user clicks Refresh Organization button, refresh each repo
individually instead of invoking the soon to be disappearing endpoint
that refreshes the whole org.

Need access to orgEntryObject, so passing that to component instead
of org/namespace.

Did not do tools, although did some of the prep work for it.

Added new method in WorkflowService to update a workflow, sort of
confusing (to me) in that there are two Akita stores that need to get
updated.

Also changed RefreshOrganizationComponent to extend Base and eliminate
some code duplication.

Test to ensures that clicking refresh org refreshes the individual
workflow.

Noticed that only one test was running in myworkflows.ts due
to an it.only; removed that and then added skip to all tests
that were failing. Created a followup ticket, dockstore/dockstore#3500
for followup on addressing skip tests.


src/app/shared/state/workflow.service.ts
```diff
@@ -106,6 +106,16 @@ export class WorkflowService {
     }
   }
 
+  /**
+   * Updates an existing workflow by updating the current list of workflows
+   * and updating extendededWorkfllowService
+   * @param workflow the workflow to be updated
+   */
+  updateWorkflow(workflow: BioWorkflow | Service) {
+    this.upsertWorkflowToWorkflow(workflow);
```
garyluu, line 115: What happens when you remove this line (don't update the list of workflows)?

coverbeck: Good catch; I think I'd started out with this one, and when it didn't work, added the other line, not noticing this isn't necessary.

In testing this, I noticed that the next line is causing the currently selected repo to change, without changing the url, which leads to confusion.

coverbeck: Removed it and it works.

src/app/shared/state/extended-workflow.service.ts
```diff
@@ -49,6 +51,18 @@ export class ExtendedWorkflowService {
...
 
+  /**
+   * Updates the extendedWorkflow if has the same id as workflow
+   *
+   * @param workflow
+   */
+  updateIfActive(workflow: Workflow) {
+    const extendedWorkflow = this.extendedWorkflowQuery.getValue();
+    if (extendedWorkflow && extendedWorkflow.id === workflow.id) {
+      this.update(workflow);
```
garyluu, line 62: You're updating the extendedworkflow store but not the original workflow in the normal workflowstore.  They're going to be out of sync.

coverbeck: > You're updating the extendedworkflow store but not the original workflow in the normal workflowstore. They're going to be out of sync.

This is true, but it's how I interpreted [your comment here](https://github.com/dockstore/dockstore-ui2/pull/987#discussion_r431871770) that it wasn't necessary. I read into it thinking your point was that when you switch workflows, the UI fetches the latest anyways.

My original commit was updating both stores, but then you had that comment, so I'm not clear exactly what you want. Although in hindsight maybe your comment was just a question and not a request for change.

Are you saying:

1. If it's the active workflow, update the workflowstore and the extendedworkflow store; don't update the workflow store if it's not the active workflow?
2. Or, just update both stores?

garyluu: Basically #1

I was expecting a check for active and then `workflowService.setWorkflow()` if it is

coverbeck: > I was expecting a check for active and then `workflowService.setWorkflow()` if it is

Why not update the workflowstore if it isn't active? It can become active.

denis-yuen:
Please create a ticket to follow up and remove the underlying endpoint afterwards

coverbeck:
> Please create a ticket to follow up and remove the underlying endpoint afterwards

dockstore/dockstore#3523

src/app/shared/state/extended-workflow.service.ts
garyluu: because if you swap to another workflow, it's going to trigger a more complete GET request, with validations and aliases.

coverbeck: > because if you swap to another workflow, it's going to trigger a more complete GET request, with validations and aliases. 

You approved, but as justification, I feel like the store might as well be up to date; there might be some flow some day that relies on it.

> Uhoh, should the workflow be using the response for any workflow then?

I don't understand; can you elaborate?

garyluu: " it's going to trigger a more complete GET request, with validations and aliases."

The current responses you're getting are incomplete.

PR #990, opened 2020-05-28
Feature/3447/sometimes refresh workflow

garyluu:
For dockstore/dockstore#3447

### DIFF
--- a/src/app/workflow/version-modal/version-modal.service.ts
+++ b/src/app/workflow/version-modal/version-modal.service.ts
@@ -78,6 +78,7 @@ export class VersionModalService {
           this.modifyTestParameterFiles(workflowVersion, originalTestParameterFilePaths, newTestParameterFiles).subscribe(
             success => {
               // Checks if there was a test parameter file added/removed
+              // The this.modifyTestParameterFiles function returns a {} observable if nothing was done, this is a way of checking for it
               if (!(Object.keys(success).length === 0 && success.constructor === Object)) {
                 toRefresh = true;
               }
--- a/src/app/workflow/version-modal/version-modal.service.ts
+++ b/src/app/workflow/version-modal/version-modal.service.ts
@@ -57,10 +57,20 @@ export class VersionModalService {
    * @param {any} newTestParameterFiles
    * @memberof VersionModalService
    */
-  saveVersion(workflowVersion: WorkflowVersion, originalTestParameterFilePaths, newTestParameterFiles, workflowMode: String) {
+  saveVersion(
+    originalVersion: WorkflowVersion,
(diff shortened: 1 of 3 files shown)


src/app/workflow/version-modal/version-modal.service.ts
```diff
@@ -47,34 +47,51 @@ export class VersionModalService {
...
+    // Checks if the workflow path was changed
+    if (originalVersion.workflow_path !== workflowVersion.workflow_path) {
+      toRefresh = true;
+    }
     this.alertService.start(message1);
     if (workflowMode !== 'HOSTED') {
       this.workflowsService.updateWorkflowVersion(workflowId, [workflowVersion]).subscribe(
         response => {
           this.alertService.start(message2);
           this.modifyTestParameterFiles(workflowVersion, originalTestParameterFilePaths, newTestParameterFiles).subscribe(
```
denis-yuen, line 78: Not clear to me, will this refresh if the test file path is changing but the number of files is constant?

garyluu: `4. Refresh workflow again if workflow path has changed or if a file was added/removed` and `// Checks if there was a file added/removed`

garyluu: hmm...maybe i should say specifically say test parameter file

src/app/workflow/version-modal/version-modal.service.ts
```diff
@@ -45,36 +45,54 @@ export class VersionModalService {
...
+    newTestParameterFiles,
+    workflowMode: String
+  ) {
     const message1 = 'Saving workflow version';
     const message2 = 'Modifying test parameter files';
     const workflowId = this.workflowQuery.getActive().id;
+    let toRefresh = false;
+    // Checks if the workflow path was changed
+    if (originalVersion.workflow_path !== workflowVersion.workflow_path) {
+      toRefresh = true;
```
coverbeck, line 71: I'm wondering if this shouldn't be handled in the web service? The web service should know whether the workflow needs to be refreshed after an update.

Arguments against:

1. There are two API calls, so there could be two refreshes
2. Maybe there is value to have the workflow updated but not refreshed.

garyluu: Ya, I made this argument years ago.  At the time, it was number 2.

denis-yuen: Could be wrong, but I think at the time we were working on a hotfix so we wanted to limit changes. This is for 1.9, so I think web service could be ok. 

Two refreshes would suck though

PR #995, opened 2020-06-03
partial github app feedback

NatalieEO:
dockstore/dockstore#3490
https://docs.google.com/document/d/1cJDSe4TSBdkZ-rt6nlxGGHPuh-W8t39BQlDw487Uzhg/edit

Addressed some of the comments Charles made
- Don't open GitHub installations page in new tab
- only ever display /.dockstore.yml as its path
- We decided the dialog pop up didn't make sense so not doing that

I didn't see an obvious way to change the default files tab, but I can look more into it if others agree with the comment.

### DIFF
--- a/src/app/entry/entry-file-tab/entry-file-tab.component.html
+++ b/src/app/entry/entry-file-tab/entry-file-tab.component.html
@@ -19,7 +19,7 @@
           <div class="w-100" fxLayout="row" fxLayoutAlign="space-between center">
             <mat-form-field *ngIf="selectedFile$ | async as selected" class="w-50">
               <mat-select [value]="selected" (selectionChange)="matSelectChange($event)">
-                <mat-option [value]="file" *ngFor="let file of files$ | async"> {{ file.path }} </mat-option>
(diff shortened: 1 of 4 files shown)


garyluu:
Out of memory, might need Charle's change.  Wonder what changed.

NatalieEO:
> > I didn't see an obvious way to change the default files tab, but I can look more into it if others agree with the comment.
> 
> I agree with my own comment. :) I forget, do we show it on the public view? If not, then not as big a deal.

We do show it on the public view

coverbeck:
These are good improvements and I suggest you go ahead and merge, and create ticket for the order. It does feel wrong to me that went I first look at a published workflow the file I would most interested in seeing would be the primary descriptor, so I do feel it's worth pursuing at some point.

NatalieEO:
#

> These are good improvements and I suggest you go ahead and merge, and create ticket for the order. It does feel wrong to me that went I first look at a published workflow the file I would most interested in seeing would be the primary descriptor, so I do feel it's worth pursuing at some point.

dockstore/dockstore#3528

PR #996, opened 2020-06-03
validation messages

NatalieEO:
dockstore/dockstore#3400

### DIFF
--- a/src/app/shared/validationMessages.model.ts
+++ b/src/app/shared/validationMessages.model.ts
@@ -76,6 +76,18 @@ export const validationMessages = {
     maxlength: 'Descriptor Path is too long. (Max 1000 characters.)',
     pattern: `Invalid Descriptor Path format. Descriptor Path must begin with '/' and end with '*.wdl'.`
   },
+  nflPath: {
+    required: 'This field cannot be empty.',
+    minlength: 'Descriptor Path is too short. (Min. 3 characters.)',
+    maxlength: 'Descriptor Path is too long. (Max 1000 characters.)',
+    pattern: `Invalid Descriptor Path format. Descriptor Path must begin with '/' and end with '*.config'.`
+  },
+  galaxyPath: {
+    required: 'This field cannot be empty.',
+    minlength: 'Descriptor Path is too short. (Min. 3 characters.)',
+    maxlength: 'Descriptor Path is too long. (Max 1000 characters.)',
+    pattern: `Invalid Descriptor Path format. Descriptor Path must begin with '/' and end with '*.ga', '*.yml', or '*.yaml'.`
+  },
   dockerfilePath: {
     required: 'This field cannot be empty.',
     minlength: 'Dockerfile Path is too short. (Min. 3 characters.)',
--- a/src/app/shared/validationMessages.model.ts
+++ b/src/app/shared/validationMessages.model.ts
@@ -66,110 +66,110 @@ export const validationDescriptorPatterns = {
 export const validationMessages = {
   cwlPath: {
     required: 'This field cannot be empty.',
-    minlength: 'Descriptor Path is too short. (Min. 3 characters.)',
-    maxlength: 'Descriptor Path is too long. (Max 1000 characters.)',
+    minlength: 'Descriptor Path is too short. (Minimum 3 characters)',
+    maxlength: 'Descriptor Path is too long. (Max 1000 characters)',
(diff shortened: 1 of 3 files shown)


src/app/shared/validationMessages.model.ts
```diff
@@ -76,6 +76,18 @@ export const validationMessages = {
     maxlength: 'Descriptor Path is too long. (Max 1000 characters.)',
     pattern: `Invalid Descriptor Path format. Descriptor Path must begin with '/' and end with '*.wdl'.`
   },
+  nflPath: {
+    required: 'This field cannot be empty.',
+    minlength: 'Descriptor Path is too short. (Min. 3 characters.)',
```
coverbeck, line 81: 1. Not 100% sure, but `Min. 3 characters` is not a sentence, so I wouldn't end it with a period.
2. Unless space is short, I would spell out `Minimum`. Max is OK because it's a word on its own, but I would vote against abbreviations.

Point 1 also applies to lines 82, 87, and 88.
Point 2 also applies to line 88.

src/app/workflow/register-workflow-modal/register-workflow-modal.component.ts
```diff
@@ -134,6 +135,7 @@ export class RegisterWorkflowModalComponent implements OnInit, AfterViewChecked,
       }
     });
     this.selectInitialSourceControlRepository();
+    this.workflowPathError = validationMessages.cwlPath.pattern;
```
coverbeck, line 138: What is this for? Seems wrong off-hand to pick a default that may not match the descriptor type. Can't you just leave it null?

If you can leave it null, then you should change line 57 to `public workflowPathError: String | null;`

src/app/workflow/register-workflow-modal/register-workflow-modal.component.ts
```diff
@@ -237,6 +239,15 @@ export class RegisterWorkflowModalComponent implements OnInit, AfterViewChecked,
     this.descriptorValidationPattern = DescriptorLanguageService.workflowDescriptorTypeEnumToExtendedDescriptorLanguageBean(
       descriptorType
     ).descriptorPathPattern;
+    if (descriptorType === Workflow.DescriptorTypeEnum.CWL) {
```
coverbeck, line 242: I think a switch statement would be better; if we ever 

1. Turn on strictnullchecks
2. Add a new language

Then the compiler would catch a missing language.

src/app/workflow/register-workflow-modal/register-workflow-modal.component.ts
NatalieEO: Actually the line ended up being unnecessary. Removed

src/app/workflow/register-workflow-modal/register-workflow-modal.component.ts
```diff
@@ -237,6 +238,23 @@ export class RegisterWorkflowModalComponent implements OnInit, AfterViewChecked,
...
+      case Workflow.DescriptorTypeEnum.WDL:
+        this.workflowPathError = validationMessages.wdlPath.pattern;
+        break;
+      case Workflow.DescriptorTypeEnum.NFL:
+        this.workflowPathError = validationMessages.nflPath.pattern;
+        break;
+      case Workflow.DescriptorTypeEnum.Gxformat2:
+        this.workflowPathError = validationMessages.galaxyPath.pattern;
+        break;
+      default:
```
coverbeck, line 254: I don't think you should have a default. That way the compiler can catch a missing one.

OTOH, we're a long way from turning on strictnullchecks. And if we don't have it, then maybe default is OK.

Will let others call it.

garyluu: Wasn't aware that the compiler could catch the missing one

coverbeck: On further reflection, go with default. We're too far away from having strictnullchecks.

FYI, see [first answer](https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript) for how compiler could catch.

garyluu: ok, i interpreted your comment differently I guess.  Thought you meant just removing default when in fact you need to pull it to a different function and have strictnullcheck ensure a non-null response from it.

PR #997, opened 2020-06-04
broken loading gif

NatalieEO:
dockstore/dockstore#3446
https://github.com/dockstore/dockstore-ui2/blob/develop/.circleci/config.yml#L178
The src for this image only had a `/` in front so it was not being picked up by the regex above so the image wasn't being served by gui.dockstore.org

### DIFF
--- a/src/app/loginComponents/auth/auth.component.html
+++ b/src/app/loginComponents/auth/auth.component.html
@@ -1,5 +1,5 @@
 <div class="container">
   <div class="absCenter">
-    <img src="/assets/images/dockstore/loading.gif" />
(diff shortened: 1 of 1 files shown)


garyluu:
> I don't think it's obvious why this fixes it to a new engineer or somebody who hasn't worked in this area; I feel your commit message should have an explanation.

Do we know what changed to cause it to be broken?

NatalieEO:
> I don't think it's obvious why this fixes it to a new engineer or somebody who hasn't worked in this area; I feel your commit message should have an explanation.

updated

PR #998, opened 2020-06-05
Feature/3458/remove ngx bs popover and dropdown

ByteMap:
Removed ngx-popover, ngx-dropdown and ngx-bootstrap

dockstore/dockstore#3458

Included material-extended/mde into the package.json in order to
implement a replacement for ngx-bootstrap popover

Replaced ngx-bootstrap popover with material-extended popover

Replaced ngx-bootstrap dropdown with Material Select

Removed the rest of the ngx-bootstrap dependencies from search.module

Removed ngx-bootstrap from all of the modules and package.json

Adjusted the spec files to have the material counterpart of
ngx-bootstrap import

Removed getTooltipConfig() from tooltip.ts and removed that method
from all modules importing it

Replaced TabsetComponent in entry.ts with MatTabGroup for the
entryTabs ViewChild

Screen Shot 2020-05-29 at 12 45 32 PM
Screen Shot 2020-05-29 at 12 45 56 PM
Screen Shot 2020-05-29 at 12 46 00 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11976,11 +11976,6 @@
         "tslib": "^1.9.0"
       }
     },
-    "ngx-bootstrap": {
-      "version": "3.0.1",
-      "resolved": "https://registry.npmjs.org/ngx-bootstrap/-/ngx-bootstrap-3.0.1.tgz",
-      "integrity": "sha512-ni91yYtn8ldgf/pxrlwl9lkVcLURGzopSpJnEbbgG1v1EZWTobI8y7J3mx4Kxptkn0EeiQwnLel67G7XJSox4A=="
-    },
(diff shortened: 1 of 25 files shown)


agduncan94:
I agree with Charles, I would try to fix the clipboard icon.

coverbeck:
> Looks like we're trading ngx-bootstrap for another library. what's the size of this one and what's the size of the previous one?

It's just a one component library and will still allow us to remove bootstrap proper, as it is an extension of Angular Material.

The npm pages says its 661Kb vs 17.1Mb for ngx-bootstrap, although once tree shaking kicks in, it's probably harder to tell.

Sorry to jump in, but I'd suggested this one to Andy.

PR #1003, opened 2020-06-09
Add preview to friendly name

garyluu:
Simple change that appends `(preview)` to almost every place that mentions Galaxy.  This includes every dropdown in the register modal, the info tab (both stub dropdown and current descriptor type), also the search page galaxy bucket filter

### DIFF
--- a/src/app/entry/extendedDescriptorLanguage.ts
+++ b/src/app/entry/extendedDescriptorLanguage.ts
@@ -101,7 +101,7 @@ const extendedService: ExtendedDescriptorLanguageBean = {
 
 const extendedGalaxy: ExtendedDescriptorLanguageBean = {
   value: 'gxformat2',
-  shortFriendlyName: 'Galaxy',
+  shortFriendlyName: 'Galaxy (preview)',
   friendlyName: 'Galaxy Workflow Format',
   defaultDescriptorPath: '/Dockstore.yml',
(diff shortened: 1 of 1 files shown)


garyluu:
> To clarify, this is literally changing test to say "Galaxy (preview)" as opposed to the preview badge, correct?
> 
> Screenshot from 2020-06-09 17-04-14

Ya, trying to get the preview badge to be placed in the dropdown is...uncharted territory (never seen an icon like that before in Material or most dropdowns).  This is just a simple text append.

denis-yuen:
> Ya, trying to get the preview badge to be placed in the dropdown is...uncharted territory (never seen an icon like that before in Material or most dropdowns). This is just a simple text append.

Understood, I was more thinking when viewing a Galaxy workflow (pretty much the same place) or in the whitelist wizard. Can be separate ticket though
Screenshot from 2020-06-09 17-26-21

garyluu:
> > Ya, trying to get the preview badge to be placed in the dropdown is...uncharted territory (never seen an icon like that before in Material or most dropdowns). This is just a simple text append.
> 
> Understood, I was more thinking when viewing a Galaxy workflow (pretty much the same place) or in the whitelist wizard. Can be separate ticket though
> Screenshot from 2020-06-09 17-26-21

So it's there for radio buttons too.  I wasn't sure where to put preview at the top because it might imply the My Workflows in general is also in preview.

PR #1004, opened 2020-06-09
new message

NatalieEO:
dockstore/dockstore#3381

### DIFF
--- a/src/app/container/view/view.component.ts
+++ b/src/app/container/view/view.component.ts
@@ -83,7 +83,18 @@ export class ViewContainerComponent extends View implements OnInit {
   }
 
   deleteHostedTag(): void {
-    const deleteMessage = 'Are you sure you want to delete tag ' + this.version.name + ' for tool ' + this.tool.tool_path + '?';
+    let deleteMessage;
+    if (this.defaultVersion === this.version.name) {
+      deleteMessage =
+        'You are trying to delete the default version ' +
+        this.defaultVersion +
+        ' for the tool ' +
+        this.tool.tool_path +
+        '. Deleting this will set the default version to be the latest version. Are you sure you want to continue?';
+    } else {
+      deleteMessage = 'Are you sure you want to delete tag ' + this.version.name + ' for tool ' + this.tool.tool_path + '?';
+    }
+
     const confirmDelete = confirm(deleteMessage);
     if (confirmDelete) {
       this.hostedService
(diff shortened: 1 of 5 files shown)


src/app/container/view/view.component.ts
```diff
@@ -83,7 +83,18 @@ export class ViewContainerComponent extends View implements OnInit {
...
+    let deleteMessage;
+    if (this.defaultVersion === this.version.name) {
+      deleteMessage =
+        'You are trying to delete the default version ' +
+        this.defaultVersion +
+        ' for the tool ' +
+        this.tool.tool_path +
+        '. Deleting this will set the default version to be the latest version. Are you sure you want to continue?';
+    } else {
+      deleteMessage = 'Are you sure you want to delete tag ' + this.version.name + ' for tool ' + this.tool.tool_path + '?';
```
coverbeck, line 95: This is fine,  was like this, and is a matter of preference, but FYI, in TypeScript you can use string interpolation, so you don't have to have the `+` signs (surround with backticks instead of single/double quotes)

```
deleteMessage = `Are you sure you want to delete tag ${this.version.name} for tool ${this.tool.tool_path}?'
```

src/app/workflow/view/view.component.ts
```diff
@@ -139,8 +140,19 @@ export class ViewWorkflowComponent extends View implements OnInit {
   }
 
   deleteHostedVersion() {
-    const deleteMessage =
-      'Are you sure you want to delete version ' + this.version.name + ' for workflow ' + this.workflow.full_workflow_path + '?';
+    let deleteMessage;
+    console.log(this.defaultVersion + ' ' + this.version.name);
```
coverbeck, line 144: I don't think you should leave this in.

src/app/workflow/view/view.component.ts
```diff
@@ -139,8 +140,18 @@ export class ViewWorkflowComponent extends View implements OnInit {
...
+      deleteMessage =
+        'You are trying to delete the default version ' +
+        this.defaultVersion +
+        ' for the workflow ' +
+        this.workflow.full_workflow_path +
+        '. Deleting this will set the default version to be the latest version. Are you sure you want to continue?';
+    } else {
+      deleteMessage =
+        'Are you sure you want to delete version ' + this.version.name + ' for workflow ' + this.workflow.full_workflow_path + '?';
+    }
```
garyluu, line 154: Too similar to previous, combine somehow

PR #1007, opened 2020-06-10
Add subresource integrity to embedded content

chmreid:
This pull request adds subresource integrity attributes to embedded content. This consists of a checksum of the content that will be embedded in the page. If the content has been modified by an attacker somewhere between the server and the visitor, the content checksum will change and the content will not be run by the visitor's browser. This helps protect against XSS and other content-hijacking attacks.

Adding subresource integrity is accomplished by adding the `--subresource-integrity` flag to the `ng build` command. This PR updates the `ng build` scripts defined in `package.json` to include the `--subresource-integrity` flag. This should be picked up by the CircleCI, which uses the `package.json` scripts.

Related Jira ticket: [SEAB-1203](https://ucsc-cgl.atlassian.net/browse/SEAB-1203)


chmreid:
circleci upload to s3 is failing b/c this PR is coming from a fork

garyluu:
Is the `WARNING in webpack-subresource-integrity: Cannot determine asset for chunk scripts, computed="undefined", available=scripts.eb693ef3606739758eeb.js. Please report this full error message along with your Webpack configuration at https://github.com/waysact/webpack-subresource-integrity/issues/new` something we should be concerned about?

chmreid:
> something we should be concerned about?

Short answer: no.

Long answer: Per [this thread](https://github.com/waysact/webpack-subresource-integrity/issues/104), that warning is caused by conflicting versions of `webpack-resource-integrity` package in npm. Angular requires an older version (1.1.x) that causes this error, while the latest is 1.3.x that resolves the error. There was a [proposed fix](https://github.com/angular/angular-cli/issues/9338) for angular-cli, but the PR was closed and not merged. No other PRs are mentioned in the thread, so I don't know what happened there.

I tried to get rid of this warning by pinning the version of `webpack-resource-integrity` in package.json, but when I did that it made quite a mess and raised a lot of new errors, so I was not able to figure out how to properly do it using package.json/package-lock.json. Another way of resolving this would be switching to yarn, which allows you to specify "resolutions" that tell it what to do in such cases.

The last observation in favor of not worrying about this error is the fact that, when the `ng build --subresource-integrity` command is run, the generated `index.html` has `integrity=...` attributes on all the `<script>` tags, including the one mentioned in the warning.

chmreid:
It looks like the webpack-subresource-integrity package version was bumped past 1.3.x when Angular cli version 9 was released: https://github.com/angular/angular-cli/blob/8.3.x/packages/angular_devkit/build_angular/package.json#L64 versus https://github.com/angular/angular-cli/blob/9.0.x/packages/angular_devkit/build_angular/package.json#L69

So the warning should go away if we use angular cli >= 9.x

chmreid:
Following up for the benefit of reviewing this PR... 

When this change was deployed to staging, we got errors from the same chunked script that caused the error mentioned above:

image

PR #1012, opened 2020-06-16
Feature/3549/workflow badge

garyluu:
For dockstore/dockstore#3549

Putting the preview badge next the radio button (can't put it below because it looks weird when there's additional text once selected)

image

image

image

Put the preview badge next to the title (kind of weird, might not be entirely obvious why the workflow is in preview mode)


Ldcabansay:
> I also think the preview badge next to the title is odd, but I'm not sure what would help. Can you put a tool tip? I feel like users, other than the developer, wouldn't know it's in preview mode just because of the descriptor type.

Also (or alternatively) maybe put the badge in the info tab where it says descriptor type? If it's possible.

garyluu:
> > I also think the preview badge next to the title is odd, but I'm not sure what would help. Can you put a tool tip? I feel like users, other than the developer, wouldn't know it's in preview mode just because of the descriptor type.
> 
> Also (or alternatively) maybe put the badge in the info tab where it says descriptor type? If it's possible.

It already has "preview" in brackets there. I considered putting the badge there instead, but thought it might be weird because the stub mode dropdown that is in the same place can't deal with badges

PR #1014, opened 2020-06-17
Feature/seab 1367/assist lambda display

ByteMap:
Assist Gary with the lambda display

SEAB-1367

Imported DatePipe in order to format the eventDate into a string,
which enables the filter string to match with the eventDate

Added a new animation state and transition in order to fix
material sort triggering expandable rows

Added a new map friendly value to replace the 'true' and 'false'
values to 'Success' and 'Failed' for the 'success' column

Added a new map friendly value to replace 'PUSH', 'DELETE', and 'INSTALL' to
'Git Push', 'Git Delete', and 'Install' for the 'type' column

Image of how the log would look with new friendly values:
Screen Shot 2020-06-17 at 11 19 55 AM

### DIFF
--- a/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
+++ b/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
@@ -22,9 +22,10 @@ import { finalize } from 'rxjs/operators';
   styleUrls: ['./github-apps-logs.component.scss'],
   animations: [
     trigger('detailExpand', [
-      state('collapsed', style({ height: '0px', minHeight: '0' })),
+      state('collapsed, void', style({ height: '0px', minHeight: '0' })),
       state('expanded', style({ height: '*' })),
-      transition('expanded <=> collapsed', animate('225ms cubic-bezier(0.4, 0.0, 0.2, 1)'))
+      transition('expanded <=> collapsed', animate('225ms cubic-bezier(0.4, 0.0, 0.2, 1)')),
+      transition('expanded <=> void', animate('225ms cubic-bezier(0.4, 0.0, 0.2, 1)'))
     ])
   ]
 })
--- a/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
+++ b/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
@@ -1,4 +1,5 @@
 import { animate, state, style, transition, trigger } from '@angular/animations';
+import { DatePipe } from '@angular/common';
 import { Component, Inject, OnInit, ViewChild } from '@angular/core';
 import { MAT_DIALOG_DATA, MatPaginator, MatSnackBar, MatSort, MatTableDataSource } from '@angular/material';
 import { AlertService } from 'app/shared/alert/state/alert.service';
@@ -28,11 +29,7 @@ import { finalize } from 'rxjs/operators';
   ]
 })
 export class GithubAppsLogsComponent implements OnInit {
-  constructor(
-    @Inject(MAT_DIALOG_DATA) public data: string,
-    private lambdaEventsService: LambdaEventsService,
-    private matSnackBar: MatSnackBar
-  ) {}
+  pipe: DatePipe;
   columnsToDisplay: string[] = ['repository', 'reference', 'success', 'type'];
   displayedColumns: string[] = ['eventDate', 'githubUsername', ...this.columnsToDisplay];
   lambdaEvents: LambdaEvent[] | null;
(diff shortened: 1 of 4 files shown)


src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -45,6 +40,20 @@ export class GithubAppsLogsComponent implements OnInit {
...
   isExpansionDetailRow = (i: number, row: Object) => row.hasOwnProperty('detailRow');
 
+  constructor(
+    @Inject(MAT_DIALOG_DATA) public data: string,
+    private lambdaEventsService: LambdaEventsService,
+    private matSnackBar: MatSnackBar
+  ) {
+    this.pipe = new DatePipe('en');
+    const defaultPredicate = this.dataSource.filterPredicate;
+    // tslint:disable-next-line:no-shadowed-variable
```
coverbeck, line 50: it would be preferable to not have the shadowed variable. Actually, is still one there? I can't tell what it would be. In any case, there's a reason the linter is (was?) giving a warning, so, with exceptions, it's generally best not to suppress these warnings and instead deal with them.

src/app/search/map-friendly-values.pipe.ts
```diff
@@ -103,7 +103,9 @@ export class MapFriendlyValuesPipe implements PipeTransform {
         [SourceFile.TypeEnum.DOCKSTOREYML, 'Configuration'],
         [SourceFile.TypeEnum.DOCKSTORESERVICEOTHER, 'Service Files']
       ])
-    ]
+    ],
+    ['success', new Map([['true', 'Success'], ['false', 'Failed']])],
+    ['type', new Map([['PUSH', 'Git Push'], ['DELETE', 'Git Delete'], ['INSTALL', 'Install']])]
```
coverbeck, line 108: Discussion on Slack may override this comment, but I would just say Push and Delete instead of Git Push and Git Delete.

You want the most significant word to be on the left (in a left-to-right language, which we are). If I'm scanning the column, I have to read past the "Git" to get to the interesting part of the data, the push or the delete.

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
ByteMap: Fixed

src/app/search/map-friendly-values.pipe.ts
agduncan94: I agree, I like simply Push and Delete and Install.

src/app/search/map-friendly-values.pipe.ts
```diff
@@ -103,7 +103,9 @@ export class MapFriendlyValuesPipe implements PipeTransform {
         [SourceFile.TypeEnum.DOCKSTOREYML, 'Configuration'],
         [SourceFile.TypeEnum.DOCKSTORESERVICEOTHER, 'Service Files']
       ])
-    ]
+    ],
+    ['success', new Map([['true', 'Success'], ['false', 'Failed']])],
```
garyluu, line 107: though I guess it doesn't matter too much, does the filtering still work for these columns intuitively?

ByteMap: Filtering doesn't work on the new map-friendly values, but currently working on a method to fix that issue.

ByteMap: Fixed. Though the filter still works with the original values, along with the new mapped values.

src/app/search/map-friendly-values.pipe.ts
ByteMap: Done.

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -45,12 +42,30 @@ export class GithubAppsLogsComponent implements OnInit {
...
+    private matSnackBar: MatSnackBar
+  ) {
+    this.datePipe = new DatePipe('en');
+    this.mapPipe = new MapFriendlyValuesPipe();
+    const defaultPredicate = this.dataSource.filterPredicate;
+    this.dataSource.filterPredicate = (data, filter) => {
+      const formatted = this.datePipe.transform(data.eventDate, 'medium').toLowerCase();
+      return formatted.indexOf(filter) >= 0 || defaultPredicate(data, filter);
+    };
+    this.dataSource.filterPredicate = (data, filter) => {
```
coverbeck, line 57: I don't understand; isn't this going to replace the filterPredicate from line 53?

ByteMap: Right, didnt realized having multiple filterPredicates would result in the prior one being overridden. Just fixed it.

PR #1017, opened 2020-06-18
Use new restful starring endpoint

garyluu:
For dockstore/dockstore#2395 part 3

Use the new restful starring endpoint (need to do this before previous one is removed).


garyluu:
> Do you want to change the target branch from develop?

We're not removing this endpoint for 1.9.0 so I don't think this PR matters much for the release.

PR #1019, opened 2020-06-22
Feature/reinstate tool tests

emlys:
- IMO the cy.wait(1000) is justified here. Since the delete function is reused in multiple contexts, there's no specific XHR or html content to wait for. But this was a common point of failure/flakiness and the wait seems to resolve it.

- Uncommented tests that were commented out due to the tool deleting issue


coverbeck:
Please target release/2.6.0

PR #1021, opened 2020-06-22
applied stargazers CSS to organizations

emlys:
dockstore/dockstore#3553 organization-stargazers component had the stargazers HTML but not the corresponding CSS, causing the sizing to be wrong.

### DIFF
--- a/src/app/organizations/organization/organization-starring/organization-starring.service.ts
+++ b/src/app/organizations/organization/organization-starring/organization-starring.service.ts
@@ -27,7 +27,10 @@ export class OrganizationStarringService {
   constructor(private usersService: UsersService, private organizationsService: OrganizationsService) {}
 
   setUnstar(organizationID: number): Observable<any> {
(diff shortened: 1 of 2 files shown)


src/app/organizations/organization/organization-starring/organization-starring.service.ts
```diff
@@ -27,7 +27,10 @@ export class OrganizationStarringService {
   constructor(private usersService: UsersService, private organizationsService: OrganizationsService) {}
 
   setUnstar(organizationID: number): Observable<any> {
-    return this.organizationsService.unstarOrganization(organizationID);
```
coverbeck, line 30: This should not be here. I'm guessing you branched off develop and then retargeted to release branch. We don't want to bring this in from develop

emlys: Oops yeah, that was not in the PR originally. Just fixed it

PR #1023, opened 2020-06-24
Feature/3590/update angular

garyluu:
For dockstore/dockstore#3590

Some changes because of Angular 9
- bundle size decreased by 15%
- build time seems faster (differential loading not enabled yet though)
- build error messages are more detailed
- automatic budget setting to warn if components are too large (wdl-viewer)
- new Typescript!

Main downside is that two of our dependencies are not exactly compatible with Angular 9 even though it appears to work just fine
  

- @ngx-lite/json-ld has a Angular 9 branch but no work done on it since October 2019 (using -next unstable Angular 9)
- ng2-ui-auth has no branch, no work done anywhere since December 2019

Creating issues for the below if this PR gets merged.

TODO: 
- Fix and then re-enable the disabled tsconfig rule (pretty sure all these automatic `directive` additions are wrong and it should be component)
- Test everything in more detail
- Find replacement for those two dependencies

List of things I've personally tested:
- ngx-markdown
- sharebuttons (Cypress also tests this)
- login (Cypress might test this?)
- json-ld


garyluu:
Angular 10 is out already 😆

.circleci/config.yml
```diff
@@ -75,9 +75,9 @@ jobs:
           key: cypress-cache-{{ checksum "./package-lock.json" }}
           paths:
             - ~/.cache
-      - run:
-          name: Check dependencies
-          command: bash -i -c 'npm ls'
+              #      - run:
```
denis-yuen, line 78: code comment when this will be uncommented (i.e. when we upgrade/remove transitive dependencies on Angular 8)

PR #1024, opened 2020-06-24
Pointer on hover

garyluu:
For dockstore/dockstore#3606 and dockstore/dockstore#3610

Originally based on "Table with expandable rows" of https://material.angular.io/components/table/examplesAdds , now adds a pointer on hover based on this random guy https://stackoverflow.com/a/46931989

~And no, there's no vertical room to display the message by default or horizontal room to show https://mdbootstrap.com/snippets/jquery/cam/979615~

Added it to the same column as date and then shortened the date to have the expand icons

~The console error is a legit console error and the message displayed is intuitive for me and it's hard for me to "imagine if it was not intuitive"~

Added a category of keys in which we allow null/undefined in the friendly map

### DIFF
--- a/cypress/integration/group2/myworkflows.ts
+++ b/cypress/integration/group2/myworkflows.ts
@@ -108,8 +108,8 @@ describe('Dockstore my workflows', () => {
         response: realResponse
       }).as('refreshWorkflow');
       cy.contains('See GitHub Apps Logs').click();
-      cy.contains('Feb 20, 2020, 2:20:20 AM');
-      cy.contains('Jun 5, 2020, 2:40:41 PM');
+      cy.contains('2020-02-20T02:20');
+      cy.contains('2020-06-05T14:40');
       cy.contains('1 - 2 of 2');
       cy.contains('Close').click();
     });
--- a/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.html
+++ b/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.html
@@ -17,8 +17,9 @@
         <ng-container matColumnDef="eventDate">
           <th mat-header-cell *matHeaderCellDef mat-sort-header>Date</th>
           <td mat-cell *matCellDef="let element">
+            <mat-icon *ngIf="!element.message" style="color: grey;">{{ 'expand_more' }}</mat-icon>
             <mat-icon *ngIf="element.message">{{ element == expandedElement ? 'expand_less' : 'expand_more' }}</mat-icon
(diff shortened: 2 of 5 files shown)


denis-yuen:
Idea 3: clickable arrow on left-hand side https://github.com/dockstore/dockstore-ui2/pull/1000#issue-429140146

garyluu:
I think most of those suggestions would be too repetitive for tables.  At that point, you might as well just write at the top that each row is expandable.

denis-yuen:
> At that point, you might as well just write at the top that each row is expandable 

Wouldn't rule this out actually.
But I think we can be a little more conservative. 
If I understand correctly, only the rows with errors can be expanded (and thus could be matched with down arrows/clickable arrows).

garyluu:
Number 2 and 3 basically has the same issue as what I originally wrote, it will take up horizontal space which we don't have https://mdbootstrap.com/snippets/jquery/cam/979615

garyluu:
> > At that point, you might as well just write at the top that each row is expandable
> 
> Wouldn't rule this out actually.
> But I think we can be a little more conservative.
> If I understand correctly, only the rows with errors can be expanded (and thus could be matched with down arrows/clickable arrows).

The expandable row is for the detailed message (not just error), so all of them should be expandable.

denis-yuen:
Maybe we can get more space. Switching to numeric dates would be a start 

> The expandable row is for the detailed message (not just error), so all of them should be expandable. 

Oh, I think I see the miscommunication. 
Currently in Chrome(?), all the successful events jiggle a little when I click them but there's no extra info being displayed. Only the unsuccessful messages seem to display additional text. Was assuming that only the ones with additional text would get the icon/annotation/handle.

garyluu:
> Maybe we can get more space. Switching to numeric dates would be a start
> 
> > The expandable row is for the detailed message (not just error), so all of them should be expandable.
> 
> Oh, I think I see the miscommunication.
> Currently in Chrome(?), all the successful events jiggle a little when I click them but there's no extra info being displayed. Only the unsuccessful messages seem to display additional text. Was assuming that only the ones with additional text would get the icon/annotation/handle.

It's because the messages happen to be empty right now since GitHub App logs aren't fully fleshed out right now.

garyluu:
Updated, see video in slack

Combination of everything

agduncan94:
Is the video in slack still the final implementation that you did that we should review?

garyluu:
> Is the video in slack still the final implementation that you did that we should review?

Just the last two commits (the ones after my message).  Though slack discussion shows what was changed for those two commits.

PR #1026, opened 2020-06-26
timestamps on news and updates UI

GFJHogue:
As described in https://github.com/dockstore/dockstore/issues/3307

The changes add timestamps to `News and Updates` entries as seen in the following screenshot:
Screenshot_2020-06-25 Dockstore(2)

### DIFF
--- a/src/app/home-page/widget/news-updates/news-updates.component.html
+++ b/src/app/home-page/widget/news-updates/news-updates.component.html
@@ -2,6 +2,9 @@
   <div class="news-card" *ngIf="notification.type === notificationTypeEnum.NEWSBODY">
     <markdown class="markdown-remove-pmargin">
       {{ notification.message }}
+      <span class="subtitle">
+        {{ notification.dbCreateDate | date: 'mediumDate' }}
+      </span>
     </markdown>
   </div>
 </div>
--- a/src/app/home-page/widget/news-updates/news-updates.component.scss
+++ b/src/app/home-page/widget/news-updates/news-updates.component.scss
@@ -2,8 +2,3 @@
   padding: 1rem;
   border-bottom: 0.1rem solid #e0e1e2;
 }
-
-.subtitle {
-  font-size: smaller;
-  color: rgba(0, 0, 0, 0.54);
-}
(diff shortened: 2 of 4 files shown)


src/app/home-page/widget/news-updates/news-updates.component.html
```diff
@@ -2,6 +2,9 @@
   <div class="news-card" *ngIf="notification.type === notificationTypeEnum.NEWSBODY">
     <markdown class="markdown-remove-pmargin">
       {{ notification.message }}
+      <span class="subtitle">
+        {{ notification.dbCreateDate | date: 'mediumDate' }}
```
garyluu, line 6: 👍 would've activated my trap card if you used date.service.ts

src/app/home-page/widget/news-updates/news-updates.component.scss
```diff
@@ -2,3 +2,8 @@
   padding: 1rem;
   border-bottom: 0.1rem solid #e0e1e2;
 }
+
+.subtitle {
```
garyluu, line 6: Duplicate of the one in src/app/home-page/recent-events/recent-events.component.scss  .  Move them to styles.scss, might have to rename the class to something more specific just in case it conflicts with other components on site.

GFJHogue: Since they are in different usage contexts, I probably should just give it a unique name here so it can be independent of the `recent-events` one.

garyluu: No, we do that way too often and then end up with conflicting styles and then the site looks inconsistent.  Since the two classes are exactly identical anyways, have them reference the same class.

denis-yuen:
> @denis-yuen can we get this into 1.9? Or at least first hotfix

Yeah, it doesn't seem disruptive. @GFJHogue after squashing, could you create a new PR to the release branch and cherry-pick this?

GFJHogue:
> > @denis-yuen can we get this into 1.9? Or at least first hotfix
> 
> Yeah, it doesn't seem disruptive. @GFJHogue after squashing, could you create a new PR to the release branch and cherry-pick this?

Will do

PR #1034, opened 2020-07-07
Use bound href attribute (lgtm fixes)

david4096:
This PR accepts two suggested changes from [LGTM](https://lgtm.com/projects/g/dockstore/dockstore-ui2/alerts/?mode=list).

1. Use bound attributes in links instead of double braces. Using double-braces in a link can cause rendering of a broken link in some circumstances. [more info](https://lgtm.com/rules/1945880031/)
1. remove unused imports

This issue was also observed in reviewing the LGTM alerts. https://github.com/dockstore/dockstore/issues/3645

### DIFF
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -49,7 +49,7 @@
     <div class="col-md-7">
       <h3 id="tool-path">
         <span id="verifiedIcon" *ngIf="(extendedTool$ | async)?.versionVerified">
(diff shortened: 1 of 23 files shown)


david4096:
@garyluu Excellent point. I took a look at the LGTM rule's documentation, which suggests ng-href from AngularJS. However, in this "[AngularJS to Angular](https://angular.io/guide/ajs-quick-reference)" upgrade guide it said to use the bounded attribute.

PR #1035, opened 2020-07-10
fill input fields to available width when editing in info tab

GFJHogue:
Main issue: https://github.com/dockstore/dockstore/issues/1450

Using `flex` for the inline forms in a very similar way as Bootstrap 4.1+ https://getbootstrap.com/docs/4.1/utilities/flex/

Screenshots in next comment

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -499,6 +499,15 @@ textarea:focus {
   right: 20px;
 }
 
+.inline-flex-form {
+  display: flex !important;
+  align-items: center;
+}
+
+.flex-grow-1 {
+  flex-grow: 1;
+}
+
 /* BUTTONS - - - - - - - - - - - - - */
 
 div.button {
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -499,15 +499,6 @@ textarea:focus {
   right: 20px;
 }
 
-.inline-flex-form {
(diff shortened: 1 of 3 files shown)


GFJHogue:
Old, Not Editing:
Screenshot_2020-07-10 Dockstore My Tools

Old, Editing:
Screenshot_2020-07-10 Dockstore My Tools(1)

New, Not Editing:
Screenshot_2020-07-10 Dockstore My Tools(2)

New, Editing:
Screenshot_2020-07-10 Dockstore My Tools(3)

src/styles.scss
```diff
@@ -499,6 +499,15 @@ textarea:focus {
   right: 20px;
 }
 
+.inline-flex-form {
+  display: flex !important;
```
coverbeck, line 503: I always shudder when I see `!important`, especially in a global stylesheet.

But my CSS is pretty rusty, and I'm not as up-to-date with the new way of doing things. Will defer to Gary's judgement.

GFJHogue: Yep I'm usually iffy about `!important` too, but my excuse is it's exactly how Bootstrap 4.1 does it: https://getbootstrap.com/docs/4.1/utilities/flex/#enable-flex-behaviors

PR #1037, opened 2020-07-16
Don't send back users or starred users

garyluu:
Temporary fix for dockstore/dockstore#3662 .  Additional webservice fix is preferable


garyluu:
> Being super-paranoid here, double-check that this doesn't wipe out workflow versions, starred users, or users? (ideally, there are tests already?)
> 
> Link to additional ticket for webservice follow-up

We actually wiped out the workflowversions before (so I didn't change anything there).  As for the starred users and users, tools already wipes that out, i guess we forgot to do the same for workflows.  The Cypress tests would definitely error out of the users were actually wiped out.  Looking at the webservice code, the only thing it updates is default paths.

PR #1042, opened 2020-07-29
Feature/3455/remove source file refs

NatalieEO:
This https://github.com/dockstore/dockstore/pull/3699 needs to be merged first.
There are still .sourcefile references related to the query builder for elastic search. I tested it briefly locally, and the search still seemed fine. Not sure how it works.
Opening this as a draft PR for now.

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -97,6 +97,6 @@ export function goToUnexpandedSidebarEntry(organization: string, repo: RegExp | 
 
 export function approvePotatoMembership() {
   cy.exec(
-    'PGPASSWORD=dockstore psql -h localhost -c \'update organization_user set accepted=true where userid=2 and organizationid=1\' webservice_test -U dockstore'
+    "PGPASSWORD=dockstore psql -h localhost -c 'update organization_user set accepted=true where userid=2 and organizationid=1' webservice_test -U dockstore"
   );
 }
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -97,6 +97,6 @@ export function goToUnexpandedSidebarEntry(organization: string, repo: RegExp | 
 
 export function approvePotatoMembership() {
   cy.exec(
-    "PGPASSWORD=dockstore psql -h localhost -c 'update organization_user set accepted=true where userid=2 and organizationid=1' webservice_test -U dockstore"
+    'PGPASSWORD=dockstore psql -h localhost -c \'update organization_user set accepted=true where userid=2 and organizationid=1\' webservice_test -U dockstore'
   );
 }
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -196,6 +196,7 @@
                   id="versionsTab"
                   [verifiedSource]="(extendedTool$ | async)?.verifiedSources"
                   [selectedVersion]="selectedVersion"
+                  [verifiedVersionPlatforms]="versionsWithVerirfiedPlatforms"
                 >
                 </app-versions-container>
               </div>
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -180,6 +180,7 @@
               [path]="tool?.tool_path"
               [toolname]="tool?.toolname"
               [mode]="tool?.mode"
+              [versionsFileTypes]="versionsFileTypes"
             ></app-launch>
             <ng-template #noTags>
               <div class="p-3">
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -225,6 +225,7 @@
                     [default]="defaultVersion"
                     [selectedVersion]="selectedVersion"
                     [publicPage]="publicPage"
+                    [versionsFileTypes]="versionsFileTypes"
                   >
                   </app-files-container>
                 </div>
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -342,7 +342,11 @@
         </div>
       </div>
       <current-collections [id]="tool.id"></current-collections>
-      <app-verified-by *ngIf="selectedVersion" [version]="selectedVersion"></app-verified-by>
+      <app-verified-by
+        *ngIf="selectedVersion"
+        [version]="selectedVersion"
+        [verifiedByPlatform]="versionsWithVerirfiedPlatforms"
+      ></app-verified-by>
       <div class="panel panel-default mb-3" *ngIf="tool">
         <div class="panel-heading">
           <h3>Sharing</h3>
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -197,7 +197,7 @@
                   id="versionsTab"
                   [verifiedSource]="(extendedTool$ | async)?.verifiedSources"
                   [selectedVersion]="selectedVersion"
-                  [verifiedVersionPlatforms]="versionsWithVerirfiedPlatforms"
+                  [verifiedVersionPlatforms]="versionsWithVerifiedPlatforms"
                 >
                 </app-versions-container>
               </div>
@@ -346,7 +346,7 @@
       <app-verified-by
         *ngIf="selectedVersion"
         [version]="selectedVersion"
-        [verifiedByPlatform]="versionsWithVerirfiedPlatforms"
+        [verifiedByPlatform]="versionsWithVerifiedPlatforms"
       ></app-verified-by>
       <div class="panel panel-default mb-3" *ngIf="tool">
         <div class="panel-heading">
--- a/src/app/container/container.component.ts
+++ b/src/app/container/container.component.ts
@@ -33,10 +33,11 @@ import { Entry } from '../shared/entry';
 import { ExtendedDockstoreToolQuery } from '../shared/extended-dockstoreTool/extended-dockstoreTool.query';
 import { GA4GHFilesService } from '../shared/ga4gh-files/ga4gh-files.service';
 import { ImageProviderService } from '../shared/image-provider.service';
-import { EntriesService, VersionVerifiedPlatform } from '../shared/openapi';
+import { ContainertagsService, EntriesService, VersionVerifiedPlatform } from '../shared/openapi';
 import { ProviderService } from '../shared/provider.service';
 import { SessionQuery } from '../shared/session/session.query';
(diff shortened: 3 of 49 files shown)


src/app/container/container.component.html
```diff
@@ -341,7 +343,11 @@ <h3>Source Repositories</h3>
         </div>
       </div>
       <current-collections [id]="tool.id"></current-collections>
-      <app-verified-by *ngIf="selectedVersion" [version]="selectedVersion"></app-verified-by>
+      <app-verified-by
+        *ngIf="selectedVersion"
+        [version]="selectedVersion"
+        [verifiedByPlatform]="versionsWithVerirfiedPlatforms"
```
coverbeck, line 349: Typo in `versionsWithVerirfiedPlatforms`

src/app/container/container.component.ts
```diff
@@ -192,6 +198,14 @@ export class ContainerComponent extends Entry implements AfterViewInit {
       this.contactAuthorHREF = this.emailService.composeContactAuthorEmail(this.tool);
       this.requestAccessHREF = this.emailService.composeRequestAccessEmail(this.tool);
       this.sortedVersions = this.getSortedTags(this.tool.workflowVersions, this.defaultVersion);
+      this.entryService.getVerifiedPlatforms(tool.id).subscribe((verifiedVersions: Array<VersionVerifiedPlatform>) => {
+        this.versionsWithVerirfiedPlatforms = verifiedVersions.map(value => Object.assign({}, value));
+      });
+      if (this.selectedVersion) {
+        this.entryService.getVersionsFileTypes(tool.id, this.selectedVersion.id).subscribe((fileTypes: Array<SourceFile.TypeEnum>) => {
```
coverbeck, line 205: I think you should put this after line 177/183, because it's closer to where selectedVersion changes. And also add another line after 175/181 setting versionsFileTypes to null or empty array.

src/app/container/paramfiles/paramfiles.service.ts
```diff
@@ -126,23 +137,24 @@ export class ParamfilesService {
...
-    if (versions) {
-      for (const version of versions) {
-        if (this.getDescriptors(version).length) {
-          versionsWithParamfiles.push(version);
-        }
-      }
-    }
-    return versionsWithParamfiles;
-  }
+  // getVersions(versions: Array<Tag | WorkflowVersion>) {
```
coverbeck, line 149: Go ahead and delete if it's not being used.

src/app/container/tool-file-editor/tool-file-editor.component.ts
```diff
@@ -37,19 +38,24 @@ export class ToolFileEditorComponent extends FileEditing {
...
   ToolDescriptor = ToolDescriptor;
   @Input() entrypath: string;
+  @Input() id: number;
   @Input() set selectedVersion(value: Tag) {
     this.currentVersion = value;
     this.isNewestVersion = this.checkIfNewestVersion();
     this.editing = false;
     this.clearSourceFiles();
     if (value != null) {
       // Fix the JSON.parse later.  Currently used to deep copy values but not keep the read-only attribute of state management.
```
coverbeck, line 48: You can probably get rid of this comment since you deleted the next line.

src/app/container/tool-file-editor/tool-file-editor.component.ts
```diff
@@ -66,17 +72,20 @@ export class ToolFileEditorComponent extends FileEditing {
    * Splits up the sourcefiles for the version into descriptor files and test parameter files
    */
   loadVersionSourcefiles(): void {
-    this.descriptorFiles = JSON.parse(JSON.stringify(this.getDescriptorFiles(this.currentVersion.sourceFiles)));
-    this.testParameterFiles = JSON.parse(JSON.stringify(this.getTestFiles(this.currentVersion.sourceFiles)));
-    this.dockerFile = JSON.parse(JSON.stringify(this.getDockerFile(this.currentVersion.sourceFiles)));
+    this.containerTagsService.getTagsSourcefiles(this.id, this.currentVersion.id).subscribe((sourcefiles: Array<SourceFile>) => {
+      this.originalSourceFiles = sourcefiles;
+      this.descriptorFiles = JSON.parse(JSON.stringify(this.getDescriptorFiles(this.originalSourceFiles)));
```
coverbeck, line 77: May be missing something, it seems like you have JSON, which you stringify, then convert back to JSON. Can you just use the result directly, with stringifying it and parsing it?

garyluu: Not sure about this case specifically. But for other areas where you see this stringify + parse, it's because it's a leftover issue with Akita integration.  Akita has readonly objects but we sometimes use it as-is with something like ngModel which will not work.

coverbeck: Ah, so I was missing something. A comment explaining why we're doing this might be useful. Or extract it into a method `cloneJSONForAkita` or something like that... But perhaps beyond the scope of this PR.

src/app/container/container.component.ts
```diff
@@ -173,8 +179,14 @@ export class ContainerComponent extends Entry implements AfterViewInit {
         this.published = this.tool.is_published;
         if (this.tool.workflowVersions.length === 0) {
           this.selectedVersion = null;
+          this.versionsFileTypes = [];
         } else {
           this.selectedVersion = this.selectTag(this.tool.workflowVersions, this.urlVersion, this.tool.defaultVersion);
+          if (this.selectedVersion) {
+            this.entryService.getVersionsFileTypes(tool.id, this.selectedVersion.id).subscribe((fileTypes: Array<SourceFile.TypeEnum>) => {
```
coverbeck, line 186: At a brief glance, it seems like we're not very good/consistent on this, but ideally you would handle the error case too. This comment applies to several places.

src/app/container/tool-file-editor/tool-file-editor.component.ts
coverbeck: Still feel this deserves a comment.

src/app/container/tool-file-editor/tool-file-editor.component.ts
```diff
@@ -66,17 +71,20 @@ export class ToolFileEditorComponent extends FileEditing {
...
+    });
   }
 
   /**
    * Combines sourcefiles into one array
    * @return {Array<SourceFile>} Array of sourcefiles
    */
-  getCombinedSourceFiles(): Array<SourceFile> {
-    let baseFiles: Array<SourceFile> = [];
+  getCombinedSourceFiles(): Array<any> {
```
coverbeck, line 86: We're trying to move away from `any` - seems like this could still be `Array<SourceFile>` (looks like this.descriptorFiles already has that type).

src/app/shared/available-logs/available-logs.component.ts
```diff
@@ -26,10 +28,11 @@ export class AvailableLogsComponent extends Base implements OnInit {
     private availableLogsQuery: AvailableLogsQuery,
     private checkerWorkflowQuery: CheckerWorkflowQuery,
     private availableLogsService: AvailableLogsService,
-    @Inject(MAT_DIALOG_DATA) public data: Tag | WorkflowVersion | null
+    @Inject(MAT_DIALOG_DATA) public data: any
```
coverbeck, line 31: Is this necessary?

cypress/support/commands.ts
```diff
@@ -97,6 +97,6 @@ export function goToUnexpandedSidebarEntry(organization: string, repo: RegExp |
 
 export function approvePotatoMembership() {
   cy.exec(
-    'PGPASSWORD=dockstore psql -h localhost -c \'update organization_user set accepted=true where userid=2 and organizationid=1\' webservice_test -U dockstore'
+    "PGPASSWORD=dockstore psql -h localhost -c 'update organization_user set accepted=true where userid=2 and organizationid=1' webservice_test -U dockstore"
```
denis-yuen, line 100: Out of curiosity, why is this changing?
LGTM recommendation?

src/app/shared/available-logs/available-logs.component.ts
NatalieEO: Had to remind myself why I did that. So before the only data being sent was a version or null, but now the data sent is the version and verified platforms shown here at the bottom.
https://github.com/dockstore/dockstore-ui2/blob/ffcb28d850b515c3b011cb19069e2424dd81e803/src/app/shared/entry/verified-by/verified-by.component.ts
Wasn't sure how to express that.

coverbeck: You can define an interface, e.g., syntax not verified, not sure if either is optional or types are right, may not be best name, etc., but something like:

```
interface VersionDialogData {
  version: string;
  verifiedByPlatform: string;
}
```

cypress/support/commands.ts
NatalieEO: Think this was an accident. Changed it back.

src/app/container/tool-file-editor/tool-file-editor.component.ts
NatalieEO: changed to `Array<SourceFile>`

src/app/container/container.component.ts
NatalieEO: Handled error cases by adding alerts and setting variable to empty array here and in other locations.

src/app/workflow/workflow.component.ts
```diff
@@ -247,6 +250,16 @@ export class WorkflowComponent extends Entry implements AfterViewInit {
...
+      this.alertService.start('Getting the workflow\'s verified platforms');
+      this.entryService.getVerifiedPlatforms(workflow.id).subscribe(
+        (verifiedVersion: Array<VersionVerifiedPlatform>) => {
+          this.versionsWithVerirfiedPlatforms = verifiedVersion.map(value => Object.assign({}, value));
+        },
+        error => {
+          this.alertService.detailedError(error);
+          this.versionsWithVerirfiedPlatforms = [];
+        }
+      );
```
garyluu, line 262: Too similar to tool counterpart, make common function

garyluu:
Changing tabs in the files tab seems to call the fileType endpoint every time even though the endpoint parameters are not related.

It should be calling the endpoint every time the version is changed and that's it

PR #1049, opened 2020-08-11
use id to track workflow path

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-1447
Add this id so I can grab just the workflow path. Can't use the id above since it could contain `done` or `preview` in the text google tag manager will grab using the id.


NatalieEO:
> Seems fine, but couldn't you also grab it from the url?

D'oh, you're right.

PR #1054, opened 2020-08-19
Feature/2915/throw friendly error for large entity

ByteMap:
Friendly error message for large workflows

dockstore/dockstore#2915

Added a new optional parameter for detailedError() in alert.service.ts
which overrides the default message with a custom message

workflow-file-editor.component.ts now catches a 413 error and outputs
a custom message detailing the 100K byte limit

### DIFF
--- a/src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
+++ b/src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
@@ -142,7 +142,7 @@ export class WorkflowFileEditorComponent extends FileEditing {
       },
       (error: HttpErrorResponse) => {
         if (error.status === 413) {
-          this.alertService.detailedError(error, 'Cannot upload new version: new version would exceed the 60 kilobyte limit');
+          this.alertService.detailedError(error, 'Cannot save new version: versions have a 60 kilobyte limit');
         } else if (error) {
           this.alertService.detailedError(error);
         }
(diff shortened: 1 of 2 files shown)


src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
```diff
@@ -139,8 +140,10 @@ export class WorkflowFileEditorComponent extends FileEditing {
           this.handleNoContentResponse();
         }
       },
-      error => {
-        if (error) {
+      (error: HttpErrorResponse) => {
+        if (error.status === 413) {
+          this.alertService.detailedError(error, 'Cannot upload new version: new version would exceed the 60 kilobyte limit');
```
coverbeck, line 145: This is going to be ridiculously nit-picky, but hey, that's what I do.

1. The user will have clicked "Save as New Version", but the error message is about an upload. I'm probably being over-coddling, but my theory is that the people using this feature (it's intended for those where using GitHub is considered too onerous), won't even necessarily know what upload is, or how it relates to saving a new version.
2. This message is worded assuming that users (should) know that there is a 60K limit, but nobody knows that limit.

Suggest something like: Cannot save new version; versions have a 60K limit.

ByteMap: Done. Changed the error message to be more friendly for users.

PR #1055, opened 2020-08-20
3701/alternative dockstore yml path

agduncan94:
https://github.com/dockstore/dockstore/issues/3701

Support .github/.dockstore.yml as a path. Depends on webservice changes.


agduncan94:
I tested this locally and it works. For whoever takes this over, the version of the webservice needs to be updated to the hotfix/1.9.2

PR #1056, opened 2020-08-24
test against 1.9.1

NatalieEO:
Part of https://ucsc-cgl.atlassian.net/browse/SEAB-1614

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "https://app.circleci.com/pipelines/github/dockstore/dockstore/4129/workflows/83a4e516-e6dc-4641-aa95-a16544fcae0f/jobs/4694"
+    "webservice_version": "1.9.1"
   },
   "scripts": {
     "ng": "npx ng",
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Run `npm run build` to build the project. The build artifacts will be stored in 
 ### Angular Production Build
 
 For an Angular Production Build, you will need to install Nginx. 
-Replace your nginx.conf file (location depends on your installation) with this template, filling in the two paths:
+Replace your nginx.conf file (location depends on your installation) with this template, filling in the two paths: 
 ```
 events {
 }


NatalieEO:
> You're targeting master... Not sure if the intent is to target develop or the hotfix?
> 
> Also, run-webservice-script.sh doesn't even use this setting; not clear how to resolve that.

Oh I think I misunderstood the second bullet point in the ticket:`update to at least 1.9.1 prod for the master branch tests`
Would that be the hotfix then?

coverbeck:
> Oh I think I misunderstood the second bullet point in the ticket:`update to at least 1.9.1 prod for the master branch tests`
> Would that be the hotfix then?

Hmm, hadn't read that and maybe I misunderstood. But I haven't seen PRs directly against master like this. We already released what's on master and don't run tests against it, so I would think it goes into 1.9.2 hotfix branch, but @denis-yuen should confirm.

denis-yuen:
> Hmm, hadn't read that and maybe I misunderstood. But I haven't seen PRs directly against master like this. We already released what's on master and don't run tests against it, so I would think it goes into 1.9.2 hotfix branch, but @denis-yuen should confirm.

Yeah, I was probably inaccurate. I
I think to be kosher, targeting this at the hot-fix branch makes sense.

garyluu:
Interesting...are the tests not re-triggered when the base is changed?

NatalieEO:
> Interesting...are the tests not re-triggered when the base is changed?
Ah yeah... doesn't look like it. Maybe I should make a new PR with a branch created off of the hotfix?

denis-yuen:
You can also trigger a build by updating the readme.md or something

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "1.9.0-beta.0"
+    "webservice_version": "1.9.1"
```
coverbeck, line 6: Now that this is on 1.9.2 hotfix branch, shouldn't this be 1.9.2? Which isn't out yet? This is a broader question, but how are we suppose to handle this? As soon as we do the 1.9.2 web service release, update this to point to 1.9.2?

Also I had a comment about CircleCI artifacts, which I thought take precedence over this, but Gary approved, so I'm not sure.

NatalieEO: I thought the run-webservice-script.sh would use this webservice version as long as it was commented correctly? Unless you meant something else.

And yeah, I don't see a tag for 1.9.2. I guess bare minimum we make sure it was done before release to Dockstore? I don't know where in the release process Gary has those instructions.

coverbeck: > I thought the run-webservice-script.sh would use this webservice version as long as it was commented correctly? Unless you meant something else.

You're right; I was looking at a branch that was pointing at CircleCI, but this branch is using the artifactory one.

PR #1057, opened 2020-08-24
Feature/3714/delete orgs

ByteMap:
Allow users to delete their pending/rejected orgs

dockstore/dockstore#3714

Added a delete button to every mat-card for curator's
allpendingorganizations, and user's allrejectedorganizations and
allpendingorganizations

Created a new dialog template 'organization-delete-confirm-dialog.html'
to correspond with the deletion of an organization

requests.module.ts now imports the new OrganizationDeleteConfirmDialog
module

Created a new method for opening the delete dialog when the delete
button is clicked, and calling the delete method from the Akita service
when the confirmation button is pressed

Implemented a new method 'deleteOrganization()' in requests.service.ts
that calls on the endpoint to delete user's rejected/pending orgs, then
updates the state of the Akita store to remove the org

Included an integration test to check that the organization delete
functionality works

Included ids to every delete button corresponding to the pending
or rejected organizations


ByteMap:
Tests are currently blocked since the script isn't generating the new API call that was recently merged to dev. To test the new feature manually, the generate-openapi-script.sh must be edited so it would generate the files based off of the your local webservice.

ByteMap:
Here is how the new feature would look like.
Screen Shot 2020-08-24 at 1 48 01 PM
Screen Shot 2020-08-24 at 1 48 19 PM
Screen Shot 2020-08-24 at 1 48 26 PM

src/app/loginComponents/requests/organization-delete-confirm-dialog.html
```diff
@@ -0,0 +1,10 @@
+<h1 mat-dialog-title>Delete Organization</h1>
```
coverbeck, line 1: Why can't you use the existing confirmation dialog component?  If it's missing something, and not too much work, I think it would be better to update that component rather than create a special one-off version.

src/app/loginComponents/requests/requests.component.ts
```diff
@@ -115,6 +127,19 @@ export class RequestsComponent extends Base implements OnInit {
     });
   }
 
+  openDeleteDialog(name: string, id: number): void {
+    const dialogRef = this.dialog.open(OrganizationDeleteConfirmDialogComponent, {
+      width: '400px',
```
coverbeck, line 132: No pixels :)

src/app/loginComponents/requests/requests.component.html
```diff
@@ -93,6 +110,15 @@ <h3>My Requests</h3>
             }}</a>
           </mat-card-title>
           <mat-card-subtitle> Requested on {{ membership.dbCreateDate | date }} </mat-card-subtitle>
+          <button
+            mat-flat-button
+            [id]="'delete-my-rejected-org-' + i"
+            class="ml-auto h-50"
+            (click)="openDeleteDialog(membership.organization.displayName, membership.organization.id)"
+            matTooltip="Delete this organization request"
```
coverbeck, line 118: For here and the other buttons above, this is minor but there's no reason to withhold information:

1. I would mention the status, i.e., Delete this rejected organization
2. Or I would mention the org, e.g., Delete "Acme Explosives" organization

Combining 1 and 2 seems tricky for some reason.

src/app/loginComponents/requests/requests.component.html
```diff
@@ -26,11 +26,19 @@ <h3>Curator Requests</h3>
               on {{ org.dbCreateDate | date }}
             </p>
           </mat-card-subtitle>
+          <button
+            mat-flat-button
+            [id]="'delete-pending-org' + i"
```
coverbeck, line 31: I see this pattern was already there, but it looks like we have this ids only for use by Cypress (?). If so, we should probably be using data-cy instead. That said, don't bother changing this, just keep it in mind for the future.

cypress/integration/immutableDatabaseTests/dropdown.ts
```diff
@@ -252,6 +252,74 @@ describe('Dropdown test', () => {
...
       cy.get('#my-pending-org-card-0').should('be.visible');
     });
+
+    it('Should be able to delete pending/reject org request', () => {
+      // Both pending and rejected orgs should be visible
+      cy.get('#my-pending-org-card-0').should('be.visible');
+      cy.get('#my-rejected-org-card-0').should('be.visible');
+
+      // New mocked memberships after deleting the rejected organization
+      let memberships = [
```
coverbeck, line 262: It would be better if this were const instead of let. It's harder to follow code where variables are changing values.

Either create a different variable for the post-DELETE response, or filter this variable.

src/app/loginComponents/requests/requests.component.ts
ByteMap: Fixed, openDeletetDialog() is now deleted as the organization deletion function now uses deregister-modal.

src/app/loginComponents/requests/organization-delete-confirm-dialog.html
ByteMap: Done. Modified the existing confirmation dialog component and is now implemented with the organization deletion function.

src/app/loginComponents/requests/requests.component.html
ByteMap: Combined 1 and 2.

cypress/integration/immutableDatabaseTests/dropdown.ts
ByteMap: Done, created a separate variable for the post-DELETE response.

ByteMap:
New screenshots of how deleting a non-approved organization would look like.
Screen Shot 2020-08-26 at 5 12 08 PM
Screen Shot 2020-08-26 at 5 12 13 PM

cypress/integration/immutableDatabaseTests/dropdown.ts
```diff
@@ -252,6 +252,74 @@ describe('Dropdown test', () => {
...
+        .should('be.visible')
+        .click();
+      cy.contains('div', 'Are you sure you wish to delete this rejected organization: orgThree?').within(() => {
+        cy.contains('button', 'Delete').click();
+      });
+      cy.get('#my-pending-org-card-0').should('be.visible');
+      cy.get('#my-rejected-org-card-0').should('not.exist');
+
+      // New membership JSON object after deleting the pending organization
+      const memberships2 = [{ id: 1, role: 'MAINTAINER', accepted: false, organization: { id: 1000, status: 'PENDING', name: 'orgOne' } }];
```
coverbeck, line 293: You don't have to change this, but if you made `{ id: 1, ...` a const, then you could use it here and online 263 without repeating it.

Also I would give it a more meaningful name, like membershipsAfterDeletion.

src/app/container/deregister-modal/deregister-modal.component.ts
```diff
@@ -27,28 +28,62 @@ import { RegisterToolService } from './../register-tool/register-tool.service';
...
   openDeleteDialog() {
-    const confirmationDialogData: ConfirmationDialogData = {
+    const toolDialogData: ConfirmationDialogData = {
       title: 'Are you sure you wish to delete this tool?',
       message: `All information associated with this tool will be deleted.`,
       cancelButtonText: 'Cancel',
       confirmationButtonText: 'Delete'
     };
+
+    const organizationDialogData: ConfirmationDialogData = {
```
coverbeck, line 62: It doesn't feel right to have this here. This component is for something else, not deleting organizations.

src/app/loginComponents/requests/organization-delete-confirm-dialog.html
garyluu: i was here, this doesn't look addressed

coverbeck, line 1: Is this file still needed?

cypress/integration/immutableDatabaseTests/dropdown.ts
ByteMap: Done. Made the first membership a constant, and renamed the memberships to have more meaningful names.

src/app/container/deregister-modal/deregister-modal.component.ts
ByteMap: Implemented an additional input that takes in the dialog data and moved the two existing dialogDatas to their respectful component files.

src/app/loginComponents/requests/organization-delete-confirm-dialog.html
ByteMap: File deleted. Forgot to clean that up.

src/app/loginComponents/state/requests.service.ts
```diff
@@ -186,4 +189,22 @@ export class RequestsService {
         }
       );
   }
+
+  deleteOrganization(id: number): void {
+    this.openApiOrgService
+      .deleteRejectedOrPendingOrganization(id)
+      .pipe(finalize(() => this.requestsStore.setLoading(false)))
```
garyluu, line 196: when was this set to loading true?

src/app/container/deregister-modal/deregister-modal.component.ts
```diff
@@ -27,28 +28,45 @@ import { RegisterToolService } from './../register-tool/register-tool.service';
...
-      message: `All information associated with this tool will be deleted.`,
-      cancelButtonText: 'Cancel',
-      confirmationButtonText: 'Delete'
-    };
     this.confirmationDialogService
-      .openDialog(confirmationDialogData, bootstrap4mediumModalSize)
+      .openDialog(this.confirmationDialogData, bootstrap4mediumModalSize)
       .pipe(first())
       .subscribe(result => {
+        console.log(this.organizationID);
```
garyluu, line 60: Remove when done

src/app/loginComponents/requests/requests.component.html
```diff
@@ -26,11 +26,22 @@ <h3>Curator Requests</h3>
...
           </mat-card-subtitle>
+          <app-deregister-modal
+            [id]="'delete-pending-org' + i"
+            class="ml-auto h-50"
+            [toolTipMessage]="'Delete this pending organization: ' + org.displayName"
+            [option]="'organization'"
+            [organizationName]="org.displayName"
+            [organizationID]="org.id"
+            [organizationStatus]="'pending'"
+            [confirmationDialogData]="getDeregisterDialogData(org.displayName, 'pending')"
```
garyluu, line 37: this gets executed a bajillion times

src/app/container/deregister-modal/deregister-modal.component.ts
```diff
@@ -27,28 +28,45 @@ import { RegisterToolService } from './../register-tool/register-tool.service';
   styleUrls: ['./deregister-modal.component.css']
 })
 export class ModalComponent implements OnInit {
+  @Input() confirmationDialogData: ConfirmationDialogData;
+  @Input() toolTipMessage: string;
   @Input() refreshMessage: boolean;
+  @Input() option: string;
+  @Input() organizationID: number;
+  @Input() organizationStatus: string;
+  @Input() organizationName: string;
```
garyluu, line 37: where are these used?

src/app/container/deregister-modal/deregister-modal.component.ts
```diff
@@ -27,28 +28,45 @@ import { RegisterToolService } from './../register-tool/register-tool.service';
   styleUrls: ['./deregister-modal.component.css']
 })
 export class ModalComponent implements OnInit {
+  @Input() confirmationDialogData: ConfirmationDialogData;
+  @Input() toolTipMessage: string;
   @Input() refreshMessage: boolean;
+  @Input() option: string;
```
garyluu, line 34: make enum

src/app/loginComponents/state/requests.service.ts
ByteMap: loading is set to true by default when Akita initializes the store.

src/app/container/deregister-modal/deregister-modal.component.ts
ByteMap: No longer using deregister-modal. Created a new method to display a new delete dialog for requests.component.

src/app/loginComponents/requests/requests.component.html
ByteMap: No longer using deregister-modal, so the properties are deleted.

src/app/container/deregister-modal/deregister-modal.component.ts
ByteMap: Deleted.

src/app/container/deregister-modal/deregister-modal.component.ts
ByteMap: No longer using deregister-modal.

src/app/loginComponents/requests/requests.component.html
```diff
@@ -72,6 +83,18 @@ <h3>My Requests</h3>
             }}</a>
           </mat-card-title>
           <mat-card-subtitle> Requested on {{ membership.dbCreateDate | date }} </mat-card-subtitle>
+          <app-deregister-modal
```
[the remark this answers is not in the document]
ByteMap: Done. Created a new dialog specifically for deleting organizations.

ByteMap:
New screenshots of how the delete buttons and delete dialogs look.
Screen Shot 2020-08-28 at 3 20 38 PM
Screen Shot 2020-08-28 at 3 20 45 PM
Screen Shot 2020-08-28 at 3 21 02 PM
Screen Shot 2020-08-28 at 3 21 08 PM

coverbeck:
I would lean towards making the Delete a link like "Reject" -- it's not an operation one should take lightly, even though in this case it's not a huge deal, so you want to make it slightly harder to find/click.

src/app/loginComponents/state/requests.service.ts
coverbeck: Hmm... makes sense that Akita has it set to true when Akita initializes, but something is going to set it to false, even this code.... Even this code sets it to false, what happens if you delete a second org?

src/app/shared/entry-actions/tool-actions.component.html
```diff
@@ -28,7 +28,13 @@
     </button>
   </ng-template>
   <ng-template #enabledDeleteButton>
-    <app-deregister-modal *ngIf="!tool?.is_published" [refreshMessage]="isRefreshing$ | async"></app-deregister-modal>
+    <app-deregister-modal
```
coverbeck, line 31: Should this still be here?

src/app/shared/entry-actions/tool-actions.component.ts
```diff
@@ -23,6 +24,13 @@ export class ToolActionsComponent extends EntryActionsComponent implements OnIni
     super(alertQuery, entryActionsService);
   }
 
+  toolDialogData: ConfirmationDialogData = {
```
coverbeck, line 27: Should this still be here?

src/app/shared/entry-actions/tool-actions.component.html
ByteMap: Completely reverted the file to the version from develop

src/app/shared/entry-actions/tool-actions.component.ts
ByteMap: dito

src/app/loginComponents/state/requests.service.ts
ByteMap: I'm able to delete multiple orgs successfully. Currently investigating where else setLoading() is set to true aside from the store initialization.

ByteMap: Had the wrong impression that the setLoading() flag blocks the store from updating with a new state. Though setLoading() is not set to true anywhere, so I've set the flag to true at the start of the method in the case that we may utilize the flag for future features related to deleting an organization.

ByteMap:
Requiring some inputs on whether the delete button should be a mat-button or if it should be a link button. The link button would just make it harder for the user to delete the organization. I've attached the screenshots of how the current button would look compared to the link button.
Screen Shot 2020-08-31 at 1 40 09 PM
Screen Shot 2020-08-31 at 1 42 26 PM

coverbeck:
I prefer the first one; but I lean towards it not being red -- to instead be the same color as Reject.

src/app/loginComponents/requests/requests.component.html
```diff
@@ -30,8 +30,16 @@ <h3>Curator Requests</h3>
         <mat-card-content class="px-4">
           {{ org.topic }}
         </mat-card-content>
-
         <mat-card-actions fxLayout="row" fxLayoutAlign="end center" fxLayoutGap="10px">
+          <button
+            [id]="'delete-pending-org' + i"
+            mat-raised-button
+            color="warn"
+            [matTooltip]="'Delete this pending organization: ' + org.displayName"
```
coverbeck, line 38: Not wild about "this" nor the colon. Would just say "Delete pending organization <display name>." Optional.

src/app/loginComponents/requests/requests.component.html
```diff
@@ -76,6 +84,19 @@ <h3>My Requests</h3>
         <mat-card-content class="px-4">
           {{ membership.organization.topic }}
         </mat-card-content>
+        <mat-card-actions fxLayout="row" fxLayoutAlign="end center" fxLayoutGap="10px">
+          <button
+            [id]="'delete-my-pending-org-' + i"
+            mat-raised-button
+            color="warn"
+            [matTooltip]="'Delete this pending organization: ' + membership.organization.displayName"
```
coverbeck, line 92: Same as previous comment

src/app/loginComponents/requests/requests.component.ts
```diff
@@ -118,4 +122,31 @@ export class RequestsComponent extends Base implements OnInit {
   rerequestReview(membership: OrganizationUser) {
     this.requestsService.requestRereview(membership.organization.id);
   }
+
+  removeOrganizationDialog(organizationName: string, organizationStatus: string, organizationID: number) {
+    const confirmationDialogData: ConfirmationDialogData = {
+      title: `Are you sure you wish to delete this ${organizationStatus} organization: ${organizationName}?`,
```
coverbeck, line 128: Sort of a long title for a dialog. I would just say "Delete <org>". Then add it to the message: "Are you sure you want to the delete <org>? All information associated with <org> will be deleted."

src/app/loginComponents/requests/requests.component.ts
```diff
@@ -118,4 +122,31 @@ export class RequestsComponent extends Base implements OnInit {
...
+      .subscribe(result => {
+        if (result) {
+          this.deleteOrganization(organizationID);
+        }
+      });
+  }
+
+  deleteOrganization(organizationID: number) {
+    this.userQuery.isAdminOrCurator$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(isAdminOrCurator => {
+      if (isAdminOrCurator) {
```
coverbeck, line 145: You could change this whole if/else block into one line, `this.requestsService.deleteOrganization(organizationID, isAdminOrCurator)`.

src/app/loginComponents/state/requests.service.ts
```diff
@@ -186,4 +189,26 @@ export class RequestsService {
...
+      .subscribe(
+        (organization: Organization) => {
+          this.alertService.simpleSuccess();
+          if (isAdminOrCurator) {
+            this.updateCuratorOrganizations();
+          }
+          this.updateMyMemberships();
+        },
+        () => {
+          this.updateMyMembershipState(null, null, null, null);
```
coverbeck, line 208: What is this doing? Presumably the state hasn't changed since the delete failed. Is there any state that was changed that you need to rollback here?

ByteMap: Just chatted about this. That line is unnecessary since there's really no point in resetting the state on a fail case if the state has not been changed. However, leaving this line in as it fits in with the patterns on lines 121, 167, and 186.

src/app/loginComponents/requests/requests.component.ts
ByteMap: Done.

src/app/loginComponents/requests/requests.component.ts
ByteMap: Title is now 'Delete Organization' and added more info the the message.

src/app/loginComponents/requests/requests.component.html
ByteMap: Done.

src/app/loginComponents/requests/requests.component.html
ByteMap: Done.

ByteMap:
New screenshots for the minor style changes.
Screen Shot 2020-09-01 at 10 50 48 AM
Screen Shot 2020-09-01 at 10 51 58 AM

src/app/loginComponents/requests/requests.component.ts
```diff
@@ -118,4 +122,28 @@ export class RequestsComponent extends Base implements OnInit {
   rerequestReview(membership: OrganizationUser) {
     this.requestsService.requestRereview(membership.organization.id);
   }
+
+  removeOrganizationDialog(organizationName: string, organizationStatus: string, organizationID: number) {
+    const confirmationDialogData: ConfirmationDialogData = {
+      title: 'Delete Organization',
+      message: `Are you sure you wish to delete this ${organizationStatus} organization?
```
coverbeck, line 129: In your screenshot the status is all caps, would be nicer if it were lowercase.

ByteMap: Done. Changed the status to lowercase in the message.

src/app/loginComponents/requests/requests.component.scss
```diff
@@ -0,0 +1,4 @@
+.removeOrg {
```
garyluu, line 1: What does it look like before this class, what does it looks like after?

src/app/loginComponents/requests/requests.component.ts
```diff
@@ -118,4 +122,28 @@ export class RequestsComponent extends Base implements OnInit {
...
+      .pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe(result => {
+        if (result) {
+          this.deleteOrganization(organizationID);
+        }
+      });
+  }
+
+  deleteOrganization(organizationID: number) {
+    this.userQuery.isAdminOrCurator$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(isAdminOrCurator => {
```
garyluu, line 145: - Don't subscribe here, causes issue if state is refreshed.
- adminOrCurator check before the confirmation dialog pops up.  Another possibility, disable the button if they're not an admin/curator

src/app/loginComponents/requests/requests.component.scss
ByteMap: Before:
Screen Shot 2020-09-01 at 2 00 57 PM
After:
Screen Shot 2020-09-01 at 2 00 25 PM

The biggest difference would be how the cursor looks when hovering over the link. The before image would have the cursor as a normal pointer, but the after image would have the cursor hand pointer which is exactly the same pointer as hovering over the "Accept" or "Reject" buttons.

src/app/loginComponents/requests/requests.component.ts
ByteMap: Removed the method entirely and created a boolean field to reflect the admin/curator status. The new field is assigned a value at initialization when the code decides to update the curator orgs or not.

Checking for the admin/curator status is to prevent updating the curators' orgs else an error would be thrown if the user is not an admin/curator. Re-requesting organizations seems to be running into this issue as well; will be creating a ticket for that.

src/app/loginComponents/requests/requests.component.scss
garyluu: Just use a button instead of anchor, automatically black and hover

garyluu (approved):

PR #1059, opened 2020-08-26
update organization user roles

NatalieEO:
dockstore/dockstore#3742
Relates to this web service PR https://github.com/dockstore/dockstore/pull/3762#pullrequestreview-475849753

### DIFF
--- a/scripts/generate-openapi-script.sh
+++ b/scripts/generate-openapi-script.sh
@@ -11,7 +11,7 @@ GENERATOR_VERSION="3.3.4"
 # Uncomment this to use the actual Dockstore webservice release from the package.json
 #BASE_PATH="https://raw.githubusercontent.com/dockstore/dockstore/$npm_package_config_webservice_version"
 # Uncomment this to use the CircleCI swagger/openapi instead
-CIRCLE_CI_PATH="https://4694-33383826-gh.circle-artifacts.com/0/tmp/artifacts" 
+CIRCLE_CI_PATH="https://4872-33383826-gh.circle-artifacts.com/0/tmp/artifacts"
 # DOCKSTORE-2428 - demo how to add new workflow language, generate from local copy of swagger
 # Uncomment this to use your local copy of swagger instead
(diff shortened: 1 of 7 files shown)


src/app/organizations/state/organization-members.service.ts
```diff
@@ -93,9 +93,11 @@ export class OrganizationMembersService {
           // appears as if he still has permissions to edit.
           if (this.userQuery.getValue().user) {
             const currentUserId = this.userQuery.getValue().user.id;
-            const canEdit = organizationUsers.some(user => user.id.userId === currentUserId && user.accepted);
+            const canEdit = organizationUsers.some(
+              user => user.id.userId === currentUserId && user.accepted && user.role !== OrganizationUser.RoleEnum.MEMBER
```
coverbeck, line 97: Optional, because this works, but I feel it's generally better to check for an explicit role, rather than infer its capabilities from what the role is not.

If we add a fourth role some day and miss updating this code, then the user will be able to edit. 

If we change this code now, add a fourth role some day and forget to update this code, the user will not be able to edit. Better to have too few permissions than too many.

Also, I know this would be caught by the API in any case, so there wouldn't really be a security hole, but still.

scripts/generate-openapi-script.sh
```diff
@@ -11,7 +11,7 @@ GENERATOR_VERSION="3.3.4"
 # Uncomment this to use the actual Dockstore webservice release from the package.json
 #BASE_PATH="https://raw.githubusercontent.com/dockstore/dockstore/$npm_package_config_webservice_version"
 # Uncomment this to use the CircleCI swagger/openapi instead
-CIRCLE_CI_PATH="https://4694-33383826-gh.circle-artifacts.com/0/tmp/artifacts" 
+CIRCLE_CI_PATH="https://4872-33383826-gh.circle-artifacts.com/0/tmp/artifacts"
```
garyluu, line 14: change the package.json too

PR #1063, opened 2020-09-03
Addressing the issue where NGINX returns HTML

ByteMap:
Fix for dockstore/dockstore#2915

Currently displaying an error page for HostedWorkflowEntity too large
returns with HTML tags being displayed from NGINX

Created a new method 'customDetailedError()' that displays a custom
error message instead of passing the NGINX error HTML into the
original method and displaying HTML tags

How the error display
Screen Shot 2020-09-03 at 5 02 04 PM
 would look:


src/app/shared/alert/state/alert.service.ts
```diff
@@ -67,21 +67,27 @@ export class AlertService {
...
     let details: string;
     if (error.status === 0) {
       // Error code of 0 means the webservice is not responding, likely down
       message = 'The webservice is currently down, possibly due to load. Please wait and try again later.';
     } else {
-      message = customMessage ? customMessage : 'The webservice encountered an error.';
+      message = 'The webservice encountered an error.';
       details = AlertService.getDetailedErrorMessage(error);
     }
     const previousMessage = this.alertQuery.getValue().message;
```
coverbeck, line 93: I think this block is identical to the body of your method starting on line 70. Just call the method instead.

coverbeck (changes requested):

src/app/shared/alert/state/alert.service.ts
ByteMap: Done. Removed the three lines and replaced them by calling the new method.

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
```diff
@@ -142,7 +142,10 @@ export class WorkflowFileEditorComponent extends FileEditing {
       },
       (error: HttpErrorResponse) => {
         if (error.status === 413) {
-          this.alertService.detailedError(error, 'Cannot save new version: versions have a 60 kilobyte limit');
+          this.alertService.customDetailedError(
```
coverbeck, line 145: A comment here would be nice, mentioning that nginx returns HTML for this error. I know it's in  your commit message (good!), but if somebody is looking at this one year from now, they'll wonder why this error is handled differently than all the others in the codebase, and it would be nice if they could do so without having go through the Git history.

coverbeck (approved):

garyluu (approved):

denis-yuen (approved):

NatalieEO (approved):

src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts
ByteMap: Done.

PR #1066, opened 2020-09-08
info-tab DOI links

GFJHogue:
Main issue: https://github.com/dockstore/dockstore/issues/3673

Use Zenodo-badge DOI links on Info tab:

Screenshot_2020-09-08 Dockstore My Workflows(1)

Copying the usage in the Versions tab:

Screenshot_2020-09-08 Dockstore My Workflows

### DIFF
--- a/src/app/workflow/info-tab/info-tab.component.ts
+++ b/src/app/workflow/info-tab/info-tab.component.ts
@@ -48,6 +48,7 @@ export class InfoTabComponent extends EntryTab implements OnInit, OnChanges {
   currentVersion: WorkflowVersion;
   downloadZipLink: string;
   isValidVersion = false;
+  zenodoUrl: string;
   @Input() selectedVersion: WorkflowVersion;
 
   public validationPatterns = validationDescriptorPatterns;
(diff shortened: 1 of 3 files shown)


GFJHogue:
Not sure about this LGTM alert... complaining that I should use `href` instead of `ng-href`, but doing so breaks the link

src/app/workflow/info-tab/info-tab.component.ts
```diff
@@ -105,6 +106,7 @@ export class InfoTabComponent extends EntryTab implements OnInit, OnChanges {
   }
 
   ngOnInit() {
+    this.zenodoUrl = Dockstore.ZENODO_AUTH_URL ? Dockstore.ZENODO_AUTH_URL.replace('oauth/authorize', '') : '';
```
garyluu, line 109: `this.zenodoUrl = Dockstore.ZENODO_AUTH_URL?.replace('oauth/authorize', '') || '';` _might_ work if you want to be fancy.  same deal with the `selectedVersion` in the html

garyluu (approved):
fine either way

NatalieEO (approved):

david4096:
The LGTM thing is solved by using bound attributes, which is apparently the best way to do it now. It avoids the risk of a link being clickable with templating information in there. Gary mentioned that the *ngIf might put in enough of a guard to avoid this from happening. In either case, you can take a look at this PR that attempted to normalize our usage of templating in href links.

https://github.com/dockstore/dockstore-ui2/commit/4a30a57686c95406bb4153ce02bf09dcf5042bd3

https://github.com/dockstore/dockstore-ui2/pull/1034

src/app/workflow/info-tab/info-tab.component.ts
GFJHogue: The version of prettier/TS we're using doesn't like the `?.` syntax. Could only apply to the HTML.

denis-yuen (approved):
On build success

PR #1067, opened 2020-09-09
Feature/update prettier

garyluu:
Update prettier to latest mainly because of https://github.com/prettier/prettier/issues/7428 , then manually ran prettier on all relevenant files.

The new trailing comma looks weird to me but...it is what it is.

### DIFF
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -1,8 +1,8 @@
 const wp = require('@cypress/webpack-preprocessor');
 
-module.exports = on => {
+module.exports = (on) => {
   const options = {
-    webpackOptions: require('../../webpack.config')
+    webpackOptions: require('../../webpack.config'),
   };
   on('file:preprocessor', wp(options));
 };
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -38,10 +38,10 @@ describe('AppComponent', () => {
         RouterLinkStubDirective,
         RouterOutletStubComponent,
         TosBannerStubComponent,
-        NotificationStubComponent
+        NotificationStubComponent,
       ],
       imports: [RouterTestingModule, MatSnackBarModule],
-      providers: [{ provide: TrackLoginService, useClass: TrackLoginStubService }]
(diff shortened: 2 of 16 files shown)


cypress/plugins/index.js
```diff
@@ -1,8 +1,8 @@
 const wp = require('@cypress/webpack-preprocessor');
 
-module.exports = on => {
+module.exports = (on) => {
   const options = {
-    webpackOptions: require('../../webpack.config')
+    webpackOptions: require('../../webpack.config'),
```
denis-yuen, line 5: You're right, it does look odd

src/app/app.component.spec.ts
```diff
@@ -38,10 +38,10 @@ describe('AppComponent', () => {
         RouterLinkStubDirective,
         RouterOutletStubComponent,
         TosBannerStubComponent,
-        NotificationStubComponent
+        NotificationStubComponent,
```
denis-yuen, line 41: Yep

denis-yuen (commented):
Does `--none` https://prettier.io/docs/en/options.html#trailing-commas 
fix the comma?

garyluu:
Adding `--trailing-comma none` to the command does work but...

https://prettier.io/blog/2020/03/21/2.0.0.html first paragraph says they deliberately chose to do that.  As weird as it looks to me, the point of prettier is that it's opinionated so we don't debate on which style to use https://prettier.io/docs/en/option-philosophy.html

denis-yuen (approved):
ok

coverbeck:
> The new trailing comma looks weird to me but...it is what it is.

Fun fact: IE6 did not allow trailing commas.

cypress/plugins/index.js
coverbeck: The plus is that it makes it easier to add another line, saving you the massive effort of typing in a comma on the previous line. And it does make the diffs a little cleaner. And easier to sort.

But yeah, it still looks weird.

coverbeck (approved):

NatalieEO (approved):

Ldcabansay (approved):

PR #1068, opened 2020-09-09
update to build off develop, remove sourcefile line

NatalieEO:
Looks like I accidentally left a line in 
dockstore/dockstore#3644 
 dockstore/dockstore#3712

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "1.10.0-alpha.0"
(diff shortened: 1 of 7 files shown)


garyluu:
Kind of seems like an important line to remove

NatalieEO:
> Kind of seems like an important line to remove.
> 
> Can you check that editing a hosted file, and then cancelling it, reverts it to the original state?

I had already done the changes needed on the file; I just didn't delete that line for some reason. But I double checked if the file gets reverted after editing and cancelling it. It does, but only once which is the same behavior on develop. I'll write a bug ticket. Looks like there's some unit tests for me to fix now as well.

garyluu (approved):

coverbeck (approved):

denis-yuen (approved):
Link back to the ticket this is supposed to be amending

Ldcabansay (approved):

PR #1071, opened 2020-09-16
feature/3649/refresh-tools-button-tooltip

latfariha:
https://github.com/dockstore/dockstore/issues/3649

Update tool tip message for the 'Refresh Namespace' button

Add cypress test to check tool-tip message

### DIFF
--- a/src/app/shared/refresh-organization/refresh-organization.component.html
+++ b/src/app/shared/refresh-organization/refresh-organization.component.html
@@ -17,12 +17,12 @@
 <button
   mat-button
   color="accent"
-  matTooltip="Refresh all tools in the namespace"
   type="button"
   class="mr-1"
   data-cy="refreshOrganization"
   (click)="refreshOrganization()"
   [disabled]="isRefreshing$ | async"
+  matTooltip="{{ tooltipText }}"
 >
   {{ buttonText }}
 </button>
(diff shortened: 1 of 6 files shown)


src/app/shared/refresh-organization/refresh-organization.component.html
```diff
@@ -17,7 +17,7 @@
 <button
   mat-button
   color="accent"
-  matTooltip="Refresh all entries in the organization"
+  matTooltip="Refresh all tools in the namespace"
```
coverbeck, line 20: Based on the directory name `shared`, I think this is a shared component for tools and workflows. If so, you will need to have different messages for workflows and tools. And possibly services. Things aren't always as straightforward as they seem. :(

coverbeck (changes requested):

src/app/shared/refresh-organization/refresh-organization.component.html
Ldcabansay: We completely removed the refresh organization button from workflows (and services) so this might not actually be an issue. AFAIK only the tools ui is using this logic.

Ldcabansay: Ah nvm I thought it was refresh all

NatalieEO (approved):

coverbeck (changes requested):

Ldcabansay (approved):

garyluu (commented):
Lots of formatting changes, need to merge/rebase with develop

coverbeck (approved):

garyluu (approved):
Well, it certainly does what the issue says

denis-yuen (approved):

PR #1073, opened 2020-09-17
Delete unneeded debounce line in organizations widget

nrzhao4:
- was working on [3182](https://github.com/dockstore/dockstore/issues/3182), which appears to be already fixed
- organizations.component.ts had a `debounceTime(formInputDebounceTime)` which is not needed
- removed the line, searching in the Organizations widget on the logged-in homepage should continue to wait for typing to stop before making any API calls

### DIFF
--- a/src/app/home-page/widget/organizations/organizations.component.ts
+++ b/src/app/home-page/widget/organizations/organizations.component.ts
@@ -1,8 +1,7 @@
 import { Component } from '@angular/core';
-import { formInputDebounceTime } from 'app/shared/constants';
 import { OrganizationUpdateTime } from 'app/shared/openapi/model/organizationUpdateTime';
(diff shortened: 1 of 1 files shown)


Ldcabansay (approved):

garyluu (approved):

coverbeck (approved):
Would be nice if your commit message explained more clearly explained why the debounce line isn't needed. I first thought you meant we don't need to debounce at all. I had to go look at the code to figure it out -- we do need debounce, but it is already implemented in the base class.

BTW, in your PR comment you referenced the GitHub issue with a fully qualified link. Which is fine, but if you want to save yourself some typing next time :), you can reference it with [GitHub flavored markdown](https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown).

PR #1076, opened 2020-09-21
Added histogram to search facets

nrzhao4:
For [#2965](https://github.com/dockstore/dockstore/issues/2965)
Added histogram to the facets on search page. The histogram width is calculated by a pipe that returns a percentage. The percentage is number of items out of total items in the facet.
Currently looks like this: 
Screen Shot 2020-09-24 at 10 25 09 AM
Screen Shot 2020-09-24 at 10 26 00 AM


denis-yuen (approved):

src/app/search/search.component.html
```diff
@@ -82,7 +82,7 @@
           </div>
           <div *ngFor="let subBucket of getKeys(orderedBuckets.get(key).Items); let i = index">
             <div class="panel-container-label">
-              <div *ngIf="i < 5" fxLayout="row" fxLayoutAlign="space-between center">
+              <div *ngIf="i < 5" fxLayout="row" fxLayoutAlign="space-between center" id="facet-row">
```
coverbeck, line 85: An id [should be unique](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) in the whole document. If I'm reading this right, because this is inside an `ngFor`, you will end up with lots of duplicate ids.

src/app/search/search.component.scss
```diff
@@ -152,3 +152,17 @@
   flex-wrap: wrap;
   justify-content: space-between;
 }
+
+.number-of-items {
+  position: absolute;
+  right: 30px;
```
coverbeck, line 158: We try to avoid using pixels in styling. https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide. Use rems instead

src/app/search/search.component.ts
```diff
@@ -261,6 +265,12 @@ export class SearchComponent implements OnInit, OnDestroy {
     });
   }
 
+  getHistogramWidth(items: number) {
+    const rowWidth = document.getElementById('facet-row').offsetWidth;
+    const histogramWidth = (items / 1000) * rowWidth;
```
coverbeck, line 270: Where does 1000 come from?

NatalieEO (approved):

src/app/search/search.component.html
```diff
@@ -95,7 +95,8 @@
                     </mat-checkbox>
                   </span>
                 </div>
-                <span>{{ orderedBuckets.get(key).Items.get(subBucket) }}</span>
+                <span class="number-of-items">{{ orderedBuckets.get(key).Items.get(subBucket) }}</span>
+                <div class="histogram" [ngStyle]="{ 'width.px': getHistogramWidth(orderedBuckets.get(key).Items.get(subBucket)) }"></div>
```
coverbeck, line 99: Can you please make sure that `getHistogramWidth` is not being invoked "zillions" of times? In general, you run that risk when you're referencing a method from the HTML template.

src/app/search/search.component.ts
```diff
@@ -261,6 +265,12 @@ export class SearchComponent implements OnInit, OnDestroy {
     });
   }
 
+  getHistogramWidth(items: number) {
+    const rowWidth = document.getElementById('facet-row').offsetWidth;
```
coverbeck, line 269: 1. Pretty sure this is just using the first element it finds (see my other comment on ids). Which may be OK, but just want to make sure.
2. In general, try to avoid directly accessing the DOM like this. Instead use [ElementRef](https://angular.io/api/core/ElementRef), or according to that page, even better [Renderer2](https://angular.io/api/core/Renderer2).

See https://ucsc-cgl.atlassian.net/browse/SEAB-1766

@garyluu Will probably have better guidance on this, and you may need to get to the DOM directly for this kind of control, so this a general point where there might be an exception.

src/app/search/search.component.html
nrzhao4: Removed this id and instead tagged the #resetButton as there is only one. The button is pretty much the same width as the facet row so using that to determine the histogram width works the same

src/app/search/search.component.ts
nrzhao4: This gets the number of items to become a decimal to calculate the width of the histogram. It probably shouldn't be hard coded as 1000, though I'm not sure how else we can evaluate the number of items as a percentage of something

src/app/search/search.component.html
nrzhao4: Checked it, the function runs for every facet that gets rendered. For example upon the initial page load, only runs for the facets that appear in each accordion, and when you expand individual accordions the new rows that appear will then call the function

src/app/search/search.component.ts
nrzhao4: Changed it to use ElementRef, and since it is only accessing a property and not being used to modify the DOM, this shouldn't be a security concern.

coverbeck (commented):
I feel strongly about the parameter type mismatch; the other stuff, I'm not sure; best to see what Gary says.

garyluu (commented):
crud, I screwed up my prettier PR (both of them)

src/app/search/get-histogram-width.pipe.ts
```diff
@@ -0,0 +1,29 @@
...
+   * @param {Map<string, number>} bucket The items in one facet
+   * @param {string} subBucket The sub-bucket value (e.g. http://edamontology.org/data_9090)
+   * @returns {number} The width of the histogram for that row
+   * @memberof GetHistogramWidthPipe
+   */
+  transform(bucket: Map<string, number>, subBucket: string): number {
+    // Get number of items in the subBucket
+    const items = bucket.get(subBucket);
+
+    // Sum up number of items in the bucket
```
garyluu, line 19: Use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

src/app/search/search.component.scss
```diff
@@ -152,3 +152,17 @@
   flex-wrap: wrap;
   justify-content: space-between;
 }
+
+.number-of-items {
```
garyluu, line 156: Put a comment in here about the first and last class being related to histogram

garyluu (approved):

src/app/search/get-histogram-width.pipe.ts
```diff
@@ -0,0 +1,27 @@
...
+export class GetHistogramWidthPipe implements PipeTransform {
+  /**
+   * This pipe sums number of items in a facet and returns the width of the histogram as a proportion
+   *
+   * @param {Map<string, number>} bucket The items in one facet
+   * @param {string} subBucket The sub-bucket value (e.g. http://edamontology.org/data_9090)
+   * @returns {number} The width of the histogram for that row
+   * @memberof GetHistogramWidthPipe
+   */
+  transform(bucket: Map<string, number>, subBucket: string): number {
```
coverbeck, line 15: This was already merged, but pretty sure the type should be Map<string, string>.  That's probably why you need the `Number` on line 24.

https://github.com/dockstore/dockstore-ui2/blob/b59e590eeea9da1c8ca38fdc3bc50d4a0b400d0a/src/app/shared/models/SubBucket.ts#L19

PR #1077, opened 2020-09-21
npm audit fixes

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-1402 
Removed pipeline-builder because of https://github.com/epam/pipeline-builder/issues/57 and it doesn't handle WDL 1.0. Commented out wdl-viewer.
Used Gary's suggestion in his comment for rc because it needs to update minimist https://github.com/dominictarr/rc/issues/113

### DIFF
--- a/src/app/workflow/dag/dag.component.html
+++ b/src/app/workflow/dag/dag.component.html
@@ -48,18 +48,20 @@
                 (descriptorType$ | async) === ToolDescriptor.TypeEnum.CWL &&
                 (workflow$ | async)?.source_control_provider === 'GITHUB'
               "
-              style="line-height: 20px;"
+              style="line-height: 20px"
             >
               <label class="radio-inline"> <input type="radio" name="dagType" [(ngModel)]="dagType" value="classic" />Classic </label>
               <label class="radio-inline">
                 <input type="radio" name="dagType" [(ngModel)]="dagType" value="cwlviewer" />Common Workflow Language Viewer
               </label>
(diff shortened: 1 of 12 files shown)


src/app/workflow/dag/dag.component.html
```diff
@@ -48,18 +48,20 @@
                 (descriptorType$ | async) === ToolDescriptor.TypeEnum.CWL &&
                 (workflow$ | async)?.source_control_provider === 'GITHUB'
               "
-              style="line-height: 20px;"
+              style="line-height: 20px"
```
coverbeck, line 51: Did prettifier do this? I always thought the semi-colon was proper, although browsers would typically let  you get away without it. [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/style) example has semi-colon.

coverbeck (approved):

denis-yuen (approved):

src/app/workflow/dag/dag.component.html
NatalieEO: yeah, it's getting changed when I make a commit.

coverbeck: Looked deeper and I was wrong. [CSS spec](https://drafts.csswg.org/css-style-attr/#style-attribute) doesn't have a trailing semi-colon.

garyluu (approved):

PR #1078, opened 2020-09-23
add git secrets checks

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-1517

### DIFF
--- a/.gitallowed
+++ b/.gitallowed
@@ -1,0 +1,36 @@
+# Each line is a regular expression against the output of git-secrets
+# travisci/dbdump.sql:.*
+# Can match on the content of the string
+
+# travisci/dbdump.sql
+travisci/db_dump.sql:.*
+
+# src/app/organizations/events/events.component.html
+src/app/organizations/events/events.component.html:.*EventType.REMOVEFROMCOLLECTION
+
+# src/app/organizations/state/events.service.ts
+src/app/organizations/state/events.service.ts:.*EventType.REMOVEFROMCOLLECTION
+
+# cypress/fixtures/searchTableResponse.json
+cypress/fixtures/searchTableResponse.json:.*image_id":*
+
+# cypress/integration/immutableDatabaseTests/searchTable.ts
+cypress/integration/immutableDatabaseTests/searchTable.ts:.*image_id:*
+
+# src/app/container/paramfiles/paramfiles.service.spec.ts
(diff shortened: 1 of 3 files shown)


wshands:
David, @david4096 I know you said we don't need to automatically instal git secrets but maybe it is a good idea to do that so the user doesn't have to do it? If so I suppose a shell script could be added and called from package.json scripts prebuild?

david4096:
@wshands I like this idea. It's probably not too hard to write a shell script that detects the two environments we run under, checks for `git secrets` on the path, and installs either via homebrew or directly via make if its not detected. That same script would be present at other repos to check for whether `git secrets` is on the path so it's probably reusable. The objection in general is people avoid mucking with system requirements and leave that up to the developer since it can cause other problems. git secrets is simple and stable enough I'm not too worried about it.

At dockstore-java git secrets is installed in the CI, so you can take the changes made there to perform the install part of the script for linux, at least. I think mac people prefer to use homebrew to manage source installs and I think this script would also have to test for the presence of `brew`.

wshands:
Added git secrets install script

garyluu:
> Added git secrets install script but it should be run from "prebuild"?

No, I would not want to install gitsecrets every time i build

denis-yuen (approved):

garyluu (approved):

david4096 (approved):
Great, I tested the install script and it worked as expected on linux/ubuntu.

I also tested that a credentials file could not be added to a commit.

It should be pretty easy to add this to CI so that builds that includes secrets will fail. @Ldcabansay is this install script useful to you?

Ldcabansay:
> Great, I tested the install script and it worked as expected on linux/ubuntu.
> 
> I also tested that a credentials file could not be added to a commit.
> 
> It should be pretty easy to add this to CI so that builds that includes secrets will fail. @Ldcabansay is this install script useful to you?

Yeah I'm using the same, working on the CI part

coverbeck (approved):
The minimal is to run git-secrets in the pre-commit. And just fail if the user didn't install git-secrets. I would have been ok with that.

But since you wrote the script to actually install, you could go ahead with the CI job as David suggests. I think it's somewhat useful in that if we had to remove a credential from a branch, at least it wouldn't be in the main branch, but even so, it would be a little too late.

PR #1082, opened 2020-09-30
Can add and edit forumurl to workflow

nrzhao4:
- frontend changes for [3739](https://github.com/dockstore/dockstore/issues/3739)
- discussion link appears in workflow info-tab
Screen Shot 2020-09-30 at 4 20 05 PM
- the link can be edited by owner
Screen Shot 2020-09-30 at 4 16 58 PM

### DIFF
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -88,7 +88,12 @@
                   <button
                     *ngIf="!isPublic"
                     type="button"
-                    [disabled]="(workflowPathEditing && !editWorkflowPathForm.valid) || (isRefreshing$ | async)"
+                    [disabled]="
+                      defaultTestFilePathEditing ||
+                      forumUrlEditing ||
+                      (workflowPathEditing && !editWorkflowPathForm.valid) ||
+                      (isRefreshing$ | async)
+                    "
                     class="btn btn-link"
                     (click)="toggleEditWorkflowPath()"
                   >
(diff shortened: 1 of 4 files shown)


nrzhao4:
Also, not sure about the design for the case where there is no forumUrl and an Add button is displayed. Currently it falls in line with the other Edit buttons.
Screen Shot 2020-09-30 at 2 26 38 PM
Don't know if it should lined with the checker flow Add button, or if the add checker flow button should be moved in line with the rest of the buttons

denis-yuen (commented):
Watching discussion, I have an opinion but was curious what others thought

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -135,6 +148,53 @@
...
+                *ngIf="!(isNFL$ | async) && workflow?.mode !== WorkflowType.ModeEnum.DOCKSTOREYML"
+              >
+                <div class="form-group" fxFlex="noshrink" fxLayout fxLayoutAlign=" center">
+                  <strong matTooltip="Link to a discussion forum for this workflow">Discussion Forum</strong>:
+                  <a *ngIf="!forumUrlEditing" [href]="workflow?.forumUrl" fxFlexOffset="4px"> {{ workflow?.forumUrl }} </a>
+                  <input
+                    *ngIf="forumUrlEditing"
+                    minlength="3"
+                    maxlength="256"
+                    type="text"
```
coverbeck, line 164: The type should ideally be `url`. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url

But we don't use this anywhere else, so it's worth checking how it behaves. If it works, I don't think you need minLength. If you do keep minLength, it should be at least 7 (http://).

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -135,6 +148,53 @@
...
+                    class="input-default form-control"
+                    name="forumUrl"
+                    [(ngModel)]="workflow.forumUrl"
+                    placeholder="Link to discussion forum"
+                    fxFlex="noshrink"
+                    fxFlexOffset="4px"
```
coverbeck, line 170: Ideally no pixels, but the pattern already exists, so I guess you'd have to change it everywhere on the form. :( Don't bother, unless Gary pushes for it.

NatalieEO:
> Watching discussion, I have an opinion but was curious what others thought

I think I would prefer that all the buttons be where the +Add checker workflow button is. When the whole box is too big, the other ones are awkwardly far away (imo)

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -189,7 +249,7 @@
       </ul>
     </mat-card-content>
   </mat-card>
-  <mat-card class="mt-3" *ngIf="workflow && workflow?.mode !== WorkflowType.ModeEnum.STUB && selectedVersion">
```
coverbeck, line 192: Did you remove this on purpose?

coverbeck:
> > Watching discussion, I have an opinion but was curious what others thought
> 
> I think I would prefer that all the buttons be where the +Add checker workflow button is. When the whole box is too big, the other ones are awkwardly far away (imo)

I think the buttons should all be aligned.

On whether right-most aligned or left-aligned, I could go either way. Making it left-aligned would be trickier, although not insurmountable, because the length of the text will vary.

nrzhao4:
> > > Watching discussion, I have an opinion but was curious what others thought
> > 
> > 
> > I think I would prefer that all the buttons be where the +Add checker workflow button is. When the whole box is too big, the other ones are awkwardly far away (imo)
> 
> I think the buttons should all be aligned.
> 
> On whether right-most aligned or left-aligned, I could go either way. Making it left-aligned would be trickier, although not insurmountable, because the length of the text will vary.

For now I've made it right aligned
image

Ldcabansay:
On the alignment thing, I don't have a strong opinion. I feel like for the fields with values (like paths), it makes sense to right align them. It does look a bit odd to me when the blank fields (checker and discussion forum) are right aligned though, a bit too much dead space, but yeah no strong opinion.

denis-yuen (approved):

wshands (approved):

NatalieEO (approved):

Ldcabansay (approved):

src/app/workflow/info-tab/info-tab.service.ts
```diff
@@ -90,6 +90,7 @@ export class InfoTabService {
           this.workflowService.upsertWorkflowToWorkflow(refreshResponse);
           this.workflowService.setWorkflow(refreshResponse);
           this.alertService.detailedSuccess();
+          console.log(refreshResponse);
```
coverbeck, line 93: Please remove this line.

coverbeck (changes requested):

coverbeck (approved):

PR #1085, opened 2020-10-13
Feature/2062/search in authors

nrzhao4:
For [#2062](https://github.com/dockstore/dockstore/issues/2062)
Added search bar to search within Authors, with autocomplete terms.
Screen Shot 2020-10-13 at 4 41 27 PM
Queries for the search term within the field 'author' and sets the results into the appropriate place in the checkBoxMap and entryOrder map, so the search updates the Author facet in the sidebar.

### DIFF
--- a/src/app/search/basic-search/facet-search.component.ts
+++ b/src/app/search/basic-search/facet-search.component.ts
@@ -9,35 +9,34 @@ import { AdvancedSearchComponent } from '../advancedsearch/advancedsearch.compon
 import { SearchQuery } from '../state/search.query';
 import { SearchService } from '../state/search.service';
 
-
 @Component({
   selector: 'app-facet-search',
-  templateUrl: './facet-search.component.html'
+  templateUrl: './facet-search.component.html',
 })
 export class FacetSearchComponent extends Base implements OnInit {
-    constructor(private searchService: SearchService, private searchQuery: SearchQuery, private matDialog: MatDialog) {
-      super();
-    }
-    public searchFormControl = new FormControl();
-    public autocompleteTerms$: Observable<Array<string>>;
-    public hasAutoCompleteTerms$: Observable<boolean>;
+  constructor(private searchService: SearchService, private searchQuery: SearchQuery, private matDialog: MatDialog) {
+    super();
+  }
+  public searchFormControl = new FormControl();
+  public autocompleteTerms$: Observable<Array<string>>;
+  public hasAutoCompleteTerms$: Observable<boolean>;
 
-    ngOnInit() {
-        this.searchQuery.searchText$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((searchText) => {
-            // This keeps the state and the view in sync but this is slightly awkward
-            // Changes to the searchText$ state will change searchFormControl
-            // However, changes to searchFormControl will change searchText$
-            // The ONLY reason why this doesn't go infinite loop is because Akita doesn't emit when it's the same value
-            // Ideally, we should probably be using AkitaFormManager because then there would only be one variable
-            this.searchFormControl.setValue(searchText);
-          });
-          this.autocompleteTerms$ = this.searchQuery.autoCompleteTerms$;
-          this.hasAutoCompleteTerms$ = this.searchQuery.hasAutoCompleteTerms$;
+  ngOnInit() {
+    this.searchQuery.facetSearchText$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((searchText) => {
+      // This keeps the state and the view in sync but this is slightly awkward
+      // Changes to the searchText$ state will change searchFormControl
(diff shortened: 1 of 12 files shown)


nrzhao4:
Currently, if there are no results a message is shown but the box continues to display the top authors. Not sure if this design is better or if it should show only the message and no authors instead.
Screen Shot 2020-10-13 at 4 46 24 PM

garyluu (commented):
Not sure about others, but I was expecting a simple filter for the keys in bucketmap

src/app/search/basic-search/facet-search.component.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Base } from '../../shared/base';
+import { SearchQuery } from '../state/search.query';
+import { SearchService } from '../state/search.service';
+
+@Component({
+  selector: 'app-facet-search',
+  templateUrl: './facet-search.component.html',
+  styleUrls: ['./facet-search.component.scss'],
+})
+export class FacetSearchComponent extends Base implements OnInit {
```
coverbeck, line 15: Haven't actually run this, but this looks really cool...

Just skimming the code, I feel like with a little more work it could be more generic, and not tied directly to author, like on one 20 -- we would eventually want to use this component for other facets as well.

But I don't know this code very well, so it might be trickier to do than say. Also, you could genericize it in the next iteration, when you use it for a different component.

src/app/search/basic-search/facet-search.component.ts
```diff
@@ -0,0 +1,42 @@
...
+    super();
+  }
+  public searchFormControl = new FormControl();
+  public authorAutocompleteTerms$: Observable<Array<string>>;
+  public hasAutoCompleteTerms$: Observable<boolean>;
+  public inputDebounceTime = 500;
+
+  ngOnInit() {
+    this.searchQuery.facetSearchText$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((searchText) => {
+      // This keeps the state and the view in sync but this is slightly awkward
```
denis-yuen, line 26: Thanks for the comment!

src/app/search/search.component.ts
```diff
@@ -366,8 +415,74 @@ export class SearchComponent implements OnInit, OnDestroy {
       this.filters
     );
     this.resetEntryOrder();
-    this.updateSideBar(sideBarQuery);
     this.updateResultsTable(tableQuery);
+    this.updateSideBar(sideBarQuery);
+  }
+
+  // Searches for an author
+  // TODO: right now only returns exact matches, is case sensitive. Can be improved with new ES features when updated
```
denis-yuen, line 423: Thanks for the heads-up, I agree that a more sophisticated approach can wait for the ES update

garyluu (changes requested):
I spoke with Denis and Charles and I think what we want is a similar implementation to https://dcc.icgc.org/search  . 

Namely, no network request is performed when the user types in the facet search bar (client side filtering of the existing bucket keys).
Additionally, the buckets shown in the facet is filtered out as the user types in more characters into the search bar.

coverbeck (changes requested):
I tried running this locally, and it's making API calls to the server when I type something in the author search.

Per Gary's comment, it should be able to do the filtering client-side, i.e., just by filtering all the authors that you already have in the browser directly with TypeScript/JavaScript, without having to make a request to ElasticSearch.

nrzhao4:
> I tried running this locally, and it's making API calls to the server when I type something in the author search.
> 
> Per Gary's comment, it should be able to do the filtering client-side, i.e., just by filtering all the authors that you already have in the browser directly with TypeScript/JavaScript, without having to make a request to ElasticSearch.

The actual filtering is done client side but it makes one call to the server for the autocomplete terms, the same way as in Basic Search

coverbeck:
> The actual filtering is done client side but it makes one call to the server for the autocomplete terms, the same way as in Basic Search

Thanks for clarifying. I don't know this code very well. I'm not convinced we need to do that call, but it might be hard to take out, the way everything is intertwined.

I played with it some more though, and I'm seeing some odd behavior in a local build:

1. Go to search page
2. Click on Search or Author
3. Type `Brian`
4. Select `Brian O'Connor` from autocomplete
5. It seems like you're already in a weird state -- on my local build, he has 13 items, but you can see the other facets are still at the full numbers. (screenshot 1)
6. Select `Brian O'Conner` checkbox that is being displayed
7. Now the author facets seem to show all entries, Brian O'Connor appears twice. But the other facets are correct. (screenshot 2)

### Screenshot 1
Screen Shot 2020-10-26 at 9 31 07 PM

### Screenshot 2
Screen Shot 2020-10-26 at 9 34 24 PM

nrzhao4:
Closing this. New approach in https://github.com/dockstore/dockstore-ui2/pull/1107

PR #1086, opened 2020-10-14
Add some WAF tests

david4096:
* Adds tests that shows the blocking behavior of most rules to match the current state of the WAF in dev.
* Adds tests that show the allow behavior of rules that would break legitimate traffic.
* Adds hooks to the package.json for running against dev/staging/prod
* Left in a couple of tests, commented out, that may be used if we end up turning on more rules.
 
This should probably be added to the release checklist so that we can verify the WAF is working as expected before promoting from dev -> staging and staging -> prod.

When the WAF is turned on for blocking, these tests should pass. Blocked requests will return a 403.

Goes with this PR to turn on some WAF rules. https://github.com/dockstore/dockstore-deploy/pull/130

Also have a PR to add a blurb about maintaining this with WAF changes to -deploy. https://github.com/dockstore/dockstore-deploy/pull/243


denis-yuen:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- cypress/integration/manualTests/waf.ts  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dockstore/dockstore-ui2/pullRequest?prid=6341077&bid=20597413)

coverbeck (approved):
These sort of seem like they don't belong here, not really UI, but then again why not...

Since they are here, it makes sense to add them to the nightly build. It would be good to now if we accidentally turned off the WAF.

david4096:
> These sort of seem like they don't belong here, not really UI, but then again why not...

I agree it does seem a little odd here but we have all of the scaffolding. The alternative would be adding it to -deploy I suppose?



> Since they are here, it makes sense to add them to the nightly build. It would be good to now if we accidentally turned off the WAF.

I added them to the nightly build (against dev), makes me think it would be nice to have for staging/prod as well?

.circleci/config.yml
```diff
@@ -60,6 +60,25 @@ jobs:
           fail_only: true
           failure_message: Nightly auth tests failed!
           webhook: $SLACK_WEBHOOK
+  uptime_monitor_waf:
+    working_directory: ~/repo
+    docker:
+      - image: circleci/buildpack-deps:18.04-browsers
+        auth:
+          username: dockstoretestuser
```
denis-yuen, line 68: Thought, is this a good time to start experimenting with contexts?

coverbeck: I think this using a context.

.circleci/config.yml
```diff
@@ -60,6 +60,25 @@ jobs:
...
+          password: $DOCKERHUB_PASSWORD
+    steps:
```
coverbeck, line 70: As an alternative, you could just make the WAF tests part of the other nightly tests, i.e., don't change this file at all, and just add the WAF to the list of tests run as part of `test-dev-no-auth` in the package.json.

Probably faster instead of setting up a new container and what not. Plus it makes invocation manually for staging/prod more straightforward (less likely to overlook).

On the other hand, UI smoke tests have been a little fragile, and you might want to separate these out as more serious.

coverbeck (approved):

coverbeck:
> I added them to the nightly build (against dev), makes me think it would be nice to have for staging/prod as well?

Agreed, but unfortunately we don't have those automated yet.

david4096:
I accepted Charles' suggestion and moved the WAF tests to the same stage as the smoke tests.

garyluu (approved):

PR #1089, opened 2020-10-16
Fix/3862/public dockerfile download

garyluu:
Basically...use the custom download path no matter what for CONTAINERFILES because there's surprisingly no endpoint that gets the raw text for it

WDL => PLAIN_WDL
CWL => PLAIN_CWL
CONTAINERFILE => ???

For dockstore/dockstore#3862

### DIFF
--- a/cypress/integration/immutableDatabaseTests/toolDetails.ts
+++ b/cypress/integration/immutableDatabaseTests/toolDetails.ts
@@ -82,6 +82,9 @@ describe('Dockstore Tool Details of quay.io/A2/a', () => {
 
       it('Should have content in file viewer', () => {
         cy.get('.ace_content').should('be.visible');
+        cy.contains('a')
+          .should('have.attr', 'href')
+          .and('include', 'data:text/plain');
       });
     });
 
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "https://app.circleci.com/pipelines/github/dockstore/dockstore/4333/workflows/7d647f3a-5ca5-481d-93c1-9361afb142f1/jobs/4949/artifacts"
+    "webservice_version": "1.9.2"
   },
   "scripts": {
     "ng": "npx ng",
--- a/scripts/generate-openapi-script.sh
+++ b/scripts/generate-openapi-script.sh
@@ -9,9 +9,9 @@ GENERATOR_VERSION="3.3.4"
 
 
 # Uncomment this to use the actual Dockstore webservice release from the package.json
-#  BASE_PATH="https://raw.githubusercontent.com/dockstore/dockstore/$npm_package_config_webservice_version"
+BASE_PATH="https://raw.githubusercontent.com/dockstore/dockstore/$npm_package_config_webservice_version"
 # Uncomment this to use the CircleCI swagger/openapi instead
-CIRCLE_CI_PATH="https://4949-33383826-gh.circle-artifacts.com/0/tmp/artifacts"
+# CIRCLE_CI_PATH="https://4949-33383826-gh.circle-artifacts.com/0/tmp/artifacts"
 # DOCKSTORE-2428 - demo how to add new workflow language, generate from local copy of swagger
 # Uncomment this to use your local copy of swagger instead
 # BASE_PATH="../dockstore"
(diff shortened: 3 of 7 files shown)


denis-yuen:
> Basically...use the custom download path no matter what for CONTAINERFILES because there's surprisingly no endpoint that gets the raw text for it 

:cry:

denis-yuen (commented):
Can we determine the path and then grab it from relative files?

garyluu:
> Can we determine the path and then grab it from relative files?


The relative files endpoint requires a type as a path parameter

It's possible, but kind of hacky though because that endpoint requires a type (PLAIN_CWL, PLAIN_WDL, etc)

https://dockstore.org/api/api/ga4gh/v2/tools/quay.io%2Fpancancer%2Fpcawg-sanger-cgp-workflow/versions/2.1.0/PLAIN-WDL/descriptor/Dockerfile

I don't know what will happen if there's no valid CWL or WDL descriptor

denis-yuen:
> I don't know what will happen if there's no valid CWL or WDL descriptor

Long term, we should create a PR to TRS to add raw text containerfiles. 

In the meantime, without a valid CWL or WDL descriptor (if you cannot determine a type), I'd be fine with disabling the link (i.e. keep the current file display, but don't bother with a TRS link)

garyluu:
> > I don't know what will happen if there's no valid CWL or WDL descriptor
> 
> Long term, we should create a PR to TRS to add raw text containerfiles.
> 
> In the meantime, without a valid CWL or WDL descriptor (if you cannot determine a type), I'd be fine with disabling the link (i.e. keep the current file display, but don't bother with a TRS link)

Seems like using the custom download link is a better alternative than disabling the link. But I guess others should weigh in

denis-yuen:
> Seems like using the custom download link is a better alternative than disabling the link. But I guess others should weigh in

Think there is a misunderstanding, I thought you were saying this would be a pain due to 

> I don't know what will happen if there's no valid CWL or WDL descriptor 

But if you're saying that we can use relative TRS when valid descriptors are available and it is simply to use a custom (proprietary) link otherwise than sounds fine to me.

garyluu:
> > Seems like using the custom download link is a better alternative than disabling the link. But I guess others should weigh in
> 
> Think there is a misunderstanding, I thought you were saying this would be a pain due to
> 
> > I don't know what will happen if there's no valid CWL or WDL descriptor
> 
> But if you're saying that we can use relative TRS when valid descriptors are available and it is simply to use a custom (proprietary) link otherwise than sounds fine to me.



> > Seems like using the custom download link is a better alternative than disabling the link. But I guess others should weigh in
> 
> Think there is a misunderstanding, I thought you were saying this would be a pain due to
> 
> > I don't know what will happen if there's no valid CWL or WDL descriptor
> 
> But if you're saying that we can use relative TRS when valid descriptors are available and it is simply to use a custom (proprietary) link otherwise than sounds fine to me.

This is how the conversation/PR went from my POV:
1. I created a PR that uses a custom download link based on contents of the file
2. You asked if relative paths endpoint was possible (won't require custom download link because it's a direct link to the raw text file)
3. I said it's hacky because of "I don't know what will happen if there's no valid CWL or WDL descriptor"
4. You said abandon both solutions and disable the link?

denis-yuen:
Yeah, I think 3 is where the miscommunication starts and 4 is wrong as a result.

coverbeck (commented):
Do I have this right? The options are:

1. Always Dockstore API (the PR as is)
2. Use TRS API when possible; use Dockstore API if not possible

What are the

garyluu:
> Do I have this right? The options are:
> 
> 1. Always Dockstore API (the PR as is)
> 2. Use TRS API when possible; use Dockstore API if not possible
> 
> What are the

Options are:
1. Use the TRS /containers endpoint to get the JSON which has the contents of file and then create a custom download link with that content
2. Hackily guess the descriptor type and use the TRS relative paths endpoint as a direct link to download. Multiple fallbacks with extra calls + complexity.
3. Try and dig up the old PR and issue that made this change (something about why the Dockerfile tab has a descriptor type dropdown)

denis-yuen:
I think we can rule out 4, could use more explanation about 3 (what drop down?)

garyluu:
> > Seems like using the custom download link is a better alternative than disabling the link. But I guess others should weigh in
> 
> Think there is a misunderstanding, I thought you were saying this would be a pain due to
> 
> > I don't know what will happen if there's no valid CWL or WDL descriptor
> 
> But if you're saying that we can use relative TRS when valid descriptors are available and it is simply to use a custom (proprietary) link otherwise than sounds fine to me.

The relative files endpoint requires a type as a path parameter

It's possible, but kind of hacky though because that endpoint requires a type (PLAIN_CWL, PLAIN_WDL, etc)

> I think we can rule out 4, could use more explanation about 3 (what drop down?)

You know how in the Descriptor Files tab, there's a Descriptor Type dropdown where you can choose between CWL and WDL.  We used to have that in the Dockerfile tab, but I think someone on our team said it didn't make sense so it was removed.

denis-yuen:
Oh I see, as in we had a drop-down at some point but we do not now. Thanks

denis-yuen (approved):
To clarify current situation:

1. Current deploy has a broken link (or rather a link to json content) on TRS
2. This PR would create a custom download link, which isn't TRS but is functional 
3. It seems like using TRS would either be difficult or require TRS changes

If this is true, I'm ok with this for now as long as you open a new TRS ticket to make it easier to get TRS dockerfile content

coverbeck (approved):

wshands (approved):

garyluu:
18 days elapsed, merging

PR #1079, opened 2020-09-24
Feature/3758/infinite loading when editing unpub tools

ByteMap:
Editing unpublished hosted tools no longer loads dialog

dockstore/dockstore#3758

Editing unpublished hosted tools no longer tries to load mat-dialog
indefinitely

Asserted that getTestParameterFiles() passed before opening the
mat dialog with VersionModalComponent

Included CypressCI tests and added a few data-cy tags

PR #1091, opened 2020-10-19
Fix double test parameter call

garyluu:
For dockstore/dockstore#3871

Kind of reverting https://github.com/dockstore/dockstore-ui2/pull/1079 but not quite because of the tests


denis-yuen (approved):

NatalieEO (approved):

coverbeck:
> Kind of reverting #1079 but not quite because of the tests

If you're kind of reverting #1079, does that mean that bug #3758 is coming back?

garyluu:
> > Kind of reverting #1079 but not quite because of the tests
> 
> If you're kind of reverting #1079, does that mean that bug #3758 is coming back?

It's possible. But I handled the error in the original calls, so it shouldn't. The test from before remains. Though tester should test both issues.

coverbeck (approved):

PR #1095, opened 2020-10-21
Feature/1651/material icons sri

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-1651

Copied the styles into the top level CSS, also downloaded the WOFF file and added it into assets. This is due to Google fonts not supporting SRI as they serve different fonts by user-agent, [see here for github issue](https://github.com/google/fonts/issues/473). 

We can see the effect by OS (using a helpful code segment provided in the above-linked issue):
```
curl -s -A "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36" https://fonts.googleapis.com/icon\?family\=Material+Icons\&display\=swap | openssl dgst -sha384 -binary | openssl base64 -A | xargs -I {} echo 'sha384-'{}
sha384-9xTNEWf/x+eniVYEmg5lrp9W4Jn36lm+nJ8YwNU6f+NvJEi6qZOFKx4xAKXY+/xi
```
```
curl -s -A "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.113 Safari/537.36" https://fonts.googleapis.com/icon\?family\=Material+Icons\&display\=swap | openssl dgst -sha384 -binary | openssl base64 -A | xargs -I {} echo 'sha384-'{}
sha384-cDG6RnnZ7vc/IbSU/nl2Ex4sgFcU531zMa6l9ouWUp7kHPJOGof99+vZWjoXj/9o
```

An alternate option would be to provide a list of checksums for this resource, but this could be quite large/difficult to maintain.


garyluu (commented):
What happens when you don't include the script here, but instead add to angular.json?

Richard-Hansen:
> What happens when you don't include the script here, but instead add to angular.json?

I can create a separate CSS file and include it within angular.json styles for the same results, would this be the preferred approach?

garyluu:
> > What happens when you don't include the script here, but instead add to angular.json?
> 
> I can create a separate CSS file and include it within angular.json styles for the same results, would this be the preferred approach?

That would be better. But was more wondering what would happen with something like: `@import url("https://example.com/themes/default/style.css");`  or angular.json referencing that url

src/index.html
```diff
@@ -16,6 +15,34 @@
...
 
+    <!-- Style sheet from https://fonts.googleapis.com/icon?family=Material+Icons&display=swap  -->
+    <!-- Google fonts does not support SRI (https://github.com/google/fonts/issues/473) -->
+    <style>
+      @font-face {
+        font-family: 'Material Icons';
+        font-style: normal;
+        font-weight: 400;
+        font-display: swap;
+        src: url('assets/fonts/MaterialIcons.woff2') format('woff2');
```
coverbeck, line 26: I am concerned this may not work in a deployed environment. You may need to precede it with `..`. See other examples of `src: url`. The reason is [here](https://github.com/dockstore/dockstore-ui2/blob/0f3a414bd285649c9bbbc9211b2b948a8bac2d84/.circleci/config.yml#L174) -- when we build the UI for deployment, we replace `../assets` with `https://gui.dockstore.org/....`. This is for the CloudFront deploy that I touched upon the other day in storytime.

coverbeck (commented):
I'm wondering if this is worth it. Lots of sites use Google Fonts. Reading up on it in reviewing this ticket, one of the advantages that is offered is that because the fonts are used by many sites, the caching persists across sites.

Now we are introducing our own copy, so

1. It won't benefit from being used on other sites
2. We have a snapshot of the fonts; we won't automatically get the latest ones. Although that could be a feature.

How confident are you that the fonts you are committing will handle all OS? BTW, that's another downside, because it now sounds like in addition to not getting the benefit of caching, the fonts themselves will be larger, as the ones we are serving up are not user-agent aware.

Would using the `src: url(`, but pointing to an external server, address the immediate NetSparker complaint? It sounds like it from the ticket, although that doesn't make sense to me.

I'm just brainstorming at this point

src/index.html
Richard-Hansen: I've been looking into other approaches, per Gary's comments above. The idea was to see if importing from the url was handled at build or runtime, and if we could potentially leverage this to avoid SRI issues. From what I've been able to determine, importing from a remote url is resolved at runtime.

It looks like Google has a node module of the material icons, https://www.npmjs.com/package/material-design-icons , which we could potentially switch to?

david4096: It looks like the WOFF file they generate gets some unpredictable URL https://fonts.gstatic.com/s/materialicons/v55/flUhRq6tzZclQEJ-Vdg-IuiaDsNc.woff2 . It doesn't seem like a good idea to include that URL instead in our assets, but worth considering.

Fonts and icons are a big download and having them downloaded from CDN can have a noticeable impact on experience.

Regarding the problem of not knowing the base URL, maybe we could use a directory called "fonts" that is managed separately from releases? This would allow us to get caching behavior across releases and have good guarantees for the base URL. Or, we could just put the fonts in develop and link to that build, specifically, here.

david4096:
I'm pretty sure including the CSS via `src: url(` would get around this too.

src/index.html
Richard-Hansen: After reading through, https://www.tunetheweb.com/blog/should-you-self-host-google-fonts/ (interesting read!) there doesn't seem to be a clear winner between self-host and using the google web-fonts, even though the author concludes with recommending self-hosting.

If we were to follow the self-hosting approach, I tested the node-module release of the material icons and it seems well-rounded (better than the current implementation in the PR). It supplies multiple file types (eot, ttf, woff, woff2) bundled in the build for browser compatibility, of which the browser only needs to load 1.

Referencing David's suggestion above, we could include the multiple file types seen in the node module in a separate "fonts" directory, to get a little more breadth to the caching potential?

denis-yuen:
> I'm wondering if this is worth it. Lots of sites use Google Fonts. Reading up on it in reviewing this ticket, one of the advantages that is offered is that because the fonts are used by many sites, the caching persists across sites.

I didn't look into the details, but is this still a thing post this change? 
https://www.zdnet.com/article/chrome-changes-how-its-cache-system-works-to-improve-privacy/

denis-yuen (commented):
commented

Richard-Hansen:
> > I'm wondering if this is worth it. Lots of sites use Google Fonts. Reading up on it in reviewing this ticket, one of the advantages that is offered is that because the fonts are used by many sites, the caching persists across sites.
> 
> I didn't look into the details, but is this still a thing post this change?
> https://www.zdnet.com/article/chrome-changes-how-its-cache-system-works-to-improve-privacy/

I think using google's node module for material icons is a good balance between our options, I've updated the PR. We miss out on CDNs and simultaneously SRI issues, while still having browser compatibility through a few different font files available. 

If we think the performance impact of this isn't worth the gains, I believe the next best approach would be to keep as-is for now. Directly linking the remote font file (which changes url and filetype depending on user-agent) would be prone to breakage.

david4096 (approved):

angular.json
```diff
@@ -28,7 +28,8 @@
               "node_modules/bootstrap-social/bootstrap-social.css",
               "src/styles.css",
               "src/styles.scss",
-              "src/material.scss"
+              "src/material.scss",
+              "node_modules/material-design-icons/iconfont/material-icons.css"
```
coverbeck, line 32: What are the contents of the CSS doing? Is it fetching the fonts from a 3rd party site, or from the node_modules? Too lazy to install to find out. :)

Richard-Hansen: The CSS references local copies of the fonts, that are added to the project at build-time. Similar to this: 
```
@font-face {
  font-family: 'Material Icons';
  font-style: normal;
  font-weight: 400;
  src: url(https://example.com/MaterialIcons-Regular.eot); /* For IE6-8 */
  src: local('Material Icons'),
    local('MaterialIcons-Regular'),
    url(MaterialIcons-Regular.woff2) format('woff2'),
    url(MaterialIcons-Regular.woff) format('woff'),
    url(MaterialIcons-Regular.ttf) format('truetype');
}
```

coverbeck (approved):
Not 100% convinced that this is the best approach, but I'm willing to go with it. The doc seems to be split on this.

garyluu (approved):

denis-yuen (approved):

Richard-Hansen:
To handle the CircleCI integration test issues, added a versioning tag to the start of the cached data's checksum. This environment variable is defined within the CircleCI environment variables, changing this will cause the cache to be reloaded on the next branch that is tested. (Thanks for the [suggestion](https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-dependency-cache) Charles)

Nothing has changed with the material icons portion of the PR, but wanted to run the config.yml (and a few test file changes) by everyone before merging.

garyluu (approved):

coverbeck (approved):
Just curious, although it probably exists not going forward, does it still work if CACHE_VERSION is not set? Presumably it would.

denis-yuen (approved):

david4096 (approved):

PR #1104, opened 2020-11-05
Switched material icons module to a maintained version

Richard-Hansen:
Original ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-1651

Using a forked and separately maintained [module](https://www.npmjs.com/package/material-design-icons-iconfont) of Google's material icons. This is due to Google's material icons module being out-of-date, see [issue](https://github.com/google/material-design-icons/issues/798).

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -28,8 +28,7 @@
               "node_modules/bootstrap-social/bootstrap-social.css",
               "src/styles.css",
               "src/styles.scss",
-              "src/material.scss",
-              "node_modules/material-design-icons/iconfont/material-icons.css"
+              "src/material.scss"
             ],
             "scripts": [
               "node_modules/jquery/dist/jquery.js",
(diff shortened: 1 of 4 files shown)


coverbeck (commented):
How hard would it be to have a test to catch the error that was happening? Smoke or integration... It would be useful if we ever decide to go back to the Google Material module, switch versions, etc.

garyluu:
> How hard would it be to have a test to catch the error that was happening? Smoke or integration... It would be useful if we ever decide to go back to the Google Material module, switch versions, etc.

Could do unauthenticated footer check, seems easier than the other page

Richard-Hansen:
> How hard would it be to have a test to catch the error that was happening? Smoke or integration... It would be useful if we ever decide to go back to the Google Material module, switch versions, etc.

Holding off on tests for now, as it would require visual testing rather than the functional testing provided by Cypress.

We could return to this if there is a need for visual testing other parts of the UI in the future.

denis-yuen (approved):

coverbeck (approved):

garyluu (approved):

PR #1107, opened 2020-11-10
Feature/2062/facet search

nrzhao4:
- For [2062](https://github.com/dockstore/dockstore/issues/2062l)
- Originally made [this PR](https://github.com/dockstore/dockstore-ui2/pull/1085) but changed to a completely different approach
- filtering is done via pipes, rather than making actual changes to the entryOrder map

### DIFF
--- a/src/app/search/facet-search/facet-search-update.pipe.ts
+++ b/src/app/search/facet-search/facet-search-update.pipe.ts
@@ -4,7 +4,7 @@ import { Pipe, PipeTransform } from '@angular/core';
   name: 'getFacetSearchUpdate',
 })
 /**
- * This pipe filters results in the facet according to the facet search text
+ * This pipe updates the number of hidden items in the facet when a search is completed
  *
  * @param {Array<any>} items The items in the facet
  * @param {string} searchText The search text entered
--- a/src/app/search/facet-search/facet-search-update.pipe.ts
+++ b/src/app/search/facet-search/facet-search-update.pipe.ts
@@ -1,0 +1,28 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'getFacetSearchUpdate',
+})
+/**
+ * This pipe filters results in the facet according to the facet search text
+ *
+ * @param {Array<any>} items The items in the facet
+ * @param {string} searchText The search text entered
+ * @returns {string} The numbers of hidden items in the facet
+ * @memberof GetFacetSearchResultsPipe
+ */
+export class GetFacetSearchUpdatePipe implements PipeTransform {
+  transform(unfilteredItems: Array<any>, searchText: string): string {
+    let hiddenFilteredSize;
+    if (!searchText && unfilteredItems.length > 5) {
+      hiddenFilteredSize = unfilteredItems.length - 5;
+      return hiddenFilteredSize.toString() + ' more';
+    }
+    const filteredSize = unfilteredItems.filter((item) => item.toLowerCase().includes(searchText)).length;
+    if (filteredSize < 5) {
+      return null;
+    }
+    hiddenFilteredSize = filteredSize - 5;
+    return hiddenFilteredSize.toString() + ' more';
+  }
+}
--- a/src/app/search/facet-search/facet-search-update.pipe.ts
+++ b/src/app/search/facet-search/facet-search-update.pipe.ts
@@ -12,17 +12,19 @@ import { Pipe, PipeTransform } from '@angular/core';
  * @memberof GetFacetSearchResultsPipe
  */
 export class GetFacetSearchUpdatePipe implements PipeTransform {
-  transform(unfilteredItems: Array<any>, searchText: string): string {
+  transform(unfilteredItems: Array<any>, searchText: string): string | null {
     let hiddenFilteredSize;
-    if (!searchText && unfilteredItems.length > 5) {
-      hiddenFilteredSize = unfilteredItems.length - 5;
+    const value = searchText.toLowerCase();
+    const itemsDisplayed = 5;
+    if (!searchText && unfilteredItems.length > itemsDisplayed) {
+      hiddenFilteredSize = unfilteredItems.length - itemsDisplayed;
       return hiddenFilteredSize.toString() + ' more';
     }
-    const filteredSize = unfilteredItems.filter((item) => item.toLowerCase().includes(searchText)).length;
+    const filteredSize = unfilteredItems.filter((item) => item.toLowerCase().includes(value)).length;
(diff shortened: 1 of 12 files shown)


src/app/search/state/search.service.ts
```diff
@@ -285,6 +285,21 @@ export class SearchService {
     });
   }
 
+  setFacetAutocompleteTerms(hits: any) {
+    const autocompleteTerms = [];
+    for (let i = 0; i <= 3; i++) {
```
coverbeck, line 290: Where does 3 come from? Suggest a const with a meaningful name

src/app/search/search.component.ts
```diff
@@ -516,6 +526,16 @@ export class SearchComponent implements OnInit, OnDestroy {
     this.sortModeMap.get(category).SortBy = sortMode;
   }
 
+  // Get autocomplete terms
+  onFacetSearchKey(key) {
+    const values = this.facetSearchText.toLowerCase();
+    const unfilteredItems = Array.from(this.orderedBuckets.get(key).Items.entries());
+    const filteredItems = unfilteredItems.filter((item) => item[0].toLowerCase().includes(values)).map((item) => item[0]);
+    if (filteredItems) {
```
coverbeck, line 534: This will always be true. If you want the logic this seems to imply, you probably need to test that it's empty.

That said, what happens if it is empty? Do you not want to clear the existing terms (I don't know)?

src/app/search/state/search.service.ts
nrzhao4: Following the other autocomplete implementation where at most there are 4 terms

src/app/search/search.component.ts
nrzhao4: Oops, my logic went awry here. If its empty, the function setFacetAutocompleteTerms can still be called but will leave autocomplete array empty

src/app/search/facet-search/facet-search-update.pipe.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @param {Array<any>} items The items in the facet
+ * @param {string} searchText The search text entered
+ * @returns {string} The numbers of hidden items in the facet
+ * @memberof GetFacetSearchResultsPipe
+ */
+export class GetFacetSearchUpdatePipe implements PipeTransform {
+  transform(unfilteredItems: Array<any>, searchText: string): string {
+    let hiddenFilteredSize;
+    if (!searchText && unfilteredItems.length > 5) {
```
coverbeck, line 17: Magic number 5 used 3 times in this method; should extract to a const. I thought our linter caught stuff like this.

src/app/search/facet-search/facet-search-update.pipe.ts
```diff
@@ -0,0 +1,28 @@
...
+      hiddenFilteredSize = unfilteredItems.length - 5;
+      return hiddenFilteredSize.toString() + ' more';
+    }
+    const filteredSize = unfilteredItems.filter((item) => item.toLowerCase().includes(searchText)).length;
```
coverbeck, line 21: Shouldn't you make a lower-case copy of searchText at the beginning of this method? Otherwise you're relying on the caller always passing in lower-case.

src/app/search/facet-search/facet-search-update.pipe.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * This pipe updates the number of hidden items in the facet when a search is completed
+ *
+ * @param {Array<any>} items The items in the facet
+ * @param {string} searchText The search text entered
+ * @returns {string} The numbers of hidden items in the facet
+ * @memberof GetFacetSearchResultsPipe
+ */
+export class GetFacetSearchUpdatePipe implements PipeTransform {
+  transform(unfilteredItems: Array<any>, searchText: string): string {
```
coverbeck, line 15: Return type should be `string | null`, unless the interface doesn't allow it.

src/app/search/facet-search/facet-search.pipe.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @param {string} facet The facet the search applies to
+ * @returns {Array<any>} The results that match the search
+ * @memberof GetFacetSearchResultsPipe
+ */
+export class GetFacetSearchResultsPipe implements PipeTransform {
+  transform(items: Array<any>, searchText: string, facet: string): Array<any> {
+    if (!items || !searchText || facet !== 'author') {
+      return items;
+    }
+    const filteredItems = items.filter((item) => item.toLowerCase().indexOf(searchText.toLowerCase()) !== -1);
```
coverbeck, line 20: 1. Micro-optimization that maybe the compiler already handles, but you would ideally create a variable out of `searchText.toLowerCase()` -- otherwise you are executing it again and again for every item in the array.
2. You could use `includes` instead of `indexOf` like you do in facet-search-update-pipe

coverbeck (commented):
I ran this locally and it looks good to me!

Some nit-pick comments..

src/app/search/facet-search/facet-search-update.pipe.ts
garyluu: Only our Java checkstyle (not tests) seems to complain about magic numbers.  Javascript/Typescript linters/projects typically don't have that as a default unless otherwise specified

denis-yuen: Can confirm I configured it for Java non-tests. 
Would be nice for the UI (non-tests) as well

src/app/search/facet-search/facet-search-update.pipe.ts
```diff
@@ -12,17 +12,19 @@ import { Pipe, PipeTransform } from '@angular/core';
  * @memberof GetFacetSearchResultsPipe
  */
 export class GetFacetSearchUpdatePipe implements PipeTransform {
-  transform(unfilteredItems: Array<any>, searchText: string): string {
+  transform(unfilteredItems: Array<any>, searchText: string): string | null {
     let hiddenFilteredSize;
-    if (!searchText && unfilteredItems.length > 5) {
-      hiddenFilteredSize = unfilteredItems.length - 5;
+    const value = searchText.toLowerCase();
```
coverbeck, line 17: Can searchText be null? This will fail if so.

coverbeck (approved):

denis-yuen (approved):

src/app/search/facet-search/facet-search-update.pipe.ts
nrzhao4: searchText can't be null, if empty it will be an empty string

PR #1109, opened 2020-11-16
Changed class name for star icon styling to prevent duplicate icons

Richard-Hansen:
Issue: dockstore/dockstore#3925

Changed class name for mat-icon stars to prevent icon duplication.

### DIFF
--- a/cypress/integration/immutableDatabaseTests/app-spec.ts
+++ b/cypress/integration/immutableDatabaseTests/app-spec.ts
@@ -47,14 +47,14 @@ describe('Logged in Dockstore Home', () => {
     if (type === 'workflow') {
       cy.get('[data-cy=workflows-tab]').click();
     }
-    cy.get('.mat-icon.star').should('not.exist');
+    cy.get('.mat-icon.star-icon').should('not.exist');
     cy.visit(url);
     cy.get('#starringButton').click();
     cy.visit('');
     if (type === 'workflow') {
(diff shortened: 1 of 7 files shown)


coverbeck (approved):
Looks good, but can you explain why this broke? Do the icons we are now using have different names?

Richard-Hansen:
> Looks good, but can you explain why this broke? Do the icons we are now using have different names?

We still have the same classes, the 2 icons are slightly different ("star" vs "star_rate")
Screen Shot 2020-11-16 at 12 51 11 PM

The issue, if I'm understanding this correctly, is due to this block of code within the new material-icons package: 
```
@if $material-design-icons-class-icons-generate-codepoints {
    @each $name, $codepoint in $material-icons-codepoints {
      $codepoint: map-get($material-icons-codepoints, $name);

      $class-name: $material-design-icons-class-icons-prefix + $name;
      $class-name: normalize-class-name($class-name);

      &.#{$class-name} {
        &:before {
          content: material-icons-content($codepoint);
        }
      }
    }
  }
```
Which creates CSS class attributes such as
 `
.material-icons.star:before {
    content: "\e838"; }
` 
for the mat-icons component. So if we have a custom class with the same name as a material icon, and we use that class in a `<mat-icon>` component, this bug will pop up.

denis-yuen (approved):

NatalieEO (approved):

garyluu (approved):

PR #1110, opened 2020-11-17
feature/1781/upgrade to elasticsearch 7

nrzhao4:
- corresponds with [this PR](https://github.com/dockstore/dockstore/pull/3931)
- change ES requests to match new endpoints
- if there is no search term, query for tools and workflows separately or else ES will return all 200 entries as tools

### DIFF
--- a/src/app/search/search-results/search-results.component.ts
+++ b/src/app/search/search-results/search-results.component.ts
@@ -73,8 +73,8 @@ export class SearchResultsComponent extends Base implements OnInit {
 
   createToolTagCloud(toolQuery, type) {
     ELASTIC_SEARCH_CLIENT.search({
-      index: 'entries',
-      type: 'all',
+      index: 'tools',
+      type: 'entry',
       body: toolQuery,
     })
       .then((hits) => {
--- a/src/app/search/search-results/search-results.component.ts
+++ b/src/app/search/search-results/search-results.component.ts
@@ -72,7 +72,6 @@ export class SearchResultsComponent extends Base implements OnInit {
   }
 
   createToolTagCloud(toolQuery, type) {
-    console.log(toolQuery);
     ELASTIC_SEARCH_CLIENT.search({
       index: 'entries',
       type: 'all',
--- a/src/app/search/search-results/search-results.component.ts
+++ b/src/app/search/search-results/search-results.component.ts
@@ -72,9 +72,10 @@ export class SearchResultsComponent extends Base implements OnInit {
   }
 
   createToolTagCloud(toolQuery, type) {
+    console.log(toolQuery);
     ELASTIC_SEARCH_CLIENT.search({
-      index: 'tools',
-      type: 'entry',
+      index: 'entries',
+      type: 'all',
       body: toolQuery,
     })
       .then((hits) => {
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -70,7 +70,8 @@ export class SearchComponent implements OnInit, OnDestroy {
   public hits: Hit[];
 
   // Possibly 100 workflows and 100 tools (extra +1 is used to see if there are > 200 results)
-  public readonly query_size = 201;
+  // Set to 201 if searching both queries
+  public query_size;
   searchTerm = false;
 
   /** a map from a field (like _type or author) in elastic search to specific values for that field (tool, workflow) and how many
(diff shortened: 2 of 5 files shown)


src/app/search/search-results/search-results.component.ts
```diff
@@ -72,9 +72,10 @@ export class SearchResultsComponent extends Base implements OnInit {
   }
 
   createToolTagCloud(toolQuery, type) {
+    console.log(toolQuery);
     ELASTIC_SEARCH_CLIENT.search({
-      index: 'tools',
-      type: 'entry',
+      index: 'entries',
```
garyluu, line 77: I think It's gotten to the point where 'entries' should be a constant somewhere

src/app/search/search-results/search-results.component.ts
```diff
@@ -72,9 +72,10 @@ export class SearchResultsComponent extends Base implements OnInit {
   }
 
   createToolTagCloud(toolQuery, type) {
+    console.log(toolQuery);
```
garyluu, line 75: extra console.log

src/app/search/search.component.ts
```diff
@@ -410,6 +421,41 @@ export class SearchComponent implements OnInit, OnDestroy {
...
+      })
+      .catch((error) => console.log(error));
+    ELASTIC_SEARCH_CLIENT.search({
+      index: 'entries',
+      type: 'workflows',
+      body: value,
+    })
+      .then((hits) => {
+        workflowHits = hits.hits.hits;
+        this.hits = toolHits.concat(workflowHits);
```
garyluu, line 452: there is no guarantee that toolHits will return before workflowHits, need to join them somehow

src/app/search/search.component.ts
```diff
@@ -410,6 +421,41 @@ export class SearchComponent implements OnInit, OnDestroy {
...
+   * Updates the results table when there is no search term
+   * Queries for 100 tools and 100 workflows
+   *
+   * @param {string} value the elastic search query
+   * @memberof SearchComponent
+   */
+  updateResultsTableSeparately(value: string) {
+    let toolHits;
+    let workflowHits;
+    const numOfResults = 201;
```
garyluu, line 434: move the constant so that https://github.com/dockstore/dockstore-ui2/pull/1110/files#diff-f92f57d57ef2daf0a8070240bc9d165e4a2a9a51285247c4b1941633bdeac556R356 can use it too

garyluu (changes requested):

src/app/search/search.component.ts
```diff
@@ -349,6 +350,11 @@ export class SearchComponent implements OnInit, OnDestroy {
     // The second query updates the result table
     const advancedSearchObject = this.advancedSearchQuery.getValue().advancedSearch;
     const values = this.advancedSearchQuery.getValue().searchText;
+    if (!values) {
+      this.query_size = 101;
```
coverbeck, line 354: Magic numbers (I see Gary commented from the other side further below).

coverbeck (commented):
Forgot to add in your web service PR that you'll also need a compose_setup PR.

NatalieEO (approved):

src/app/search/search.component.ts
```diff
@@ -410,6 +420,40 @@ export class SearchComponent implements OnInit, OnDestroy {
...
+      .then((tools) => {
+        toolHits = tools.hits.hits;
+        ELASTIC_SEARCH_CLIENT.search({
+          index: 'tools',
+          type: 'entry',
+          body: workflowsQuery,
+        })
+          .then((workflows) => {
+            workflowHits = workflows.hits.hits;
+            this.hits = toolHits.concat(workflowHits);
```
garyluu, line 448: The two calls are now sequentially ran which should work (but not ideal)

garyluu (approved):
I'm just asssuming everything works (the index,type,body)

denis-yuen (approved):

src/app/search/search.component.ts
```diff
@@ -410,6 +420,40 @@ export class SearchComponent implements OnInit, OnDestroy {
       .catch((error) => console.log(error));
   }
 
+  /**
+   * Updates the results table when there is no search term
+   * Queries for 100 tools and 100 workflows
```
denis-yuen, line 425: tiny comment fix, query size is actually based on the query, not hardcoded anymore (which is good)

src/app/search/search.component.ts
coverbeck: I'd never paid attention to this client before -- is it returning a Promise? If so, I think you can create Observables from them, then use `forkJoin`.

coverbeck (approved):
With the caveat of merging other two ES PRs at the same time.

PR #1112, opened 2020-11-20
Added success message and methods

latfariha:
recording (1)

added two methods: getIdOfNewestVersion, and getNewestVersion. Not sure if the former is necessary, also added detailed success message

### DIFF
--- a/src/app/container/tool-file-editor/tool-file-editor.component.ts
+++ b/src/app/container/tool-file-editor/tool-file-editor.component.ts
@@ -114,9 +114,7 @@ export class ToolFileEditorComponent extends FileEditing {
           this.toggleEdit();
           this.containerService.setTool(editedDockstoreTool);
           const updatedVersion = this.getNewestVersion(editedDockstoreTool.workflowVersions);
-          this.alertService.detailedSuccess(
-            'Saved version ' + updatedVersion.name + ' ' + 'of hosted workflow ' + editedDockstoreTool.namespace
-          );
+          this.alertService.detailedSuccess('Saved version ' + updatedVersion.name + ' of hosted entry ' + editedDockstoreTool.namespace);
         } else {
           // Probably encountered a 204
           this.handleNoContentResponse();
--- a/src/app/container/tool-file-editor/tool-file-editor.component.ts
+++ b/src/app/container/tool-file-editor/tool-file-editor.component.ts
@@ -66,14 +66,6 @@ export class ToolFileEditorComponent extends FileEditing {
     return this.currentVersion.id === mostRecentId;
   }
 
-  getIdOfNewestVersion(versions: Array<Tag>): number {
-    return versions.reduce((max, n) => Math.max(max, n.id), this.versions[0].id);
-  }
-
-  getNewestVersion(versions: Array<Tag>): Tag {
-    return versions.reduce((p, c) => (p.id > c.id ? p : c));
-  }
-
   /**
    * Fix the JSON.parse later.  Currently used to deep copy values but not keep the read-only attribute of state management.
    * Splits up the sourcefiles for the version into descriptor files and test parameter files
--- a/src/app/container/tool-file-editor/tool-file-editor.component.ts
+++ b/src/app/container/tool-file-editor/tool-file-editor.component.ts
@@ -121,7 +121,10 @@ export class ToolFileEditorComponent extends FileEditing {
           // Only stop editing when version change was successful (not 204)
           this.toggleEdit();
           this.containerService.setTool(editedDockstoreTool);
-          this.alertService.detailedSuccess();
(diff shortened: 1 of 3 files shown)


src/app/container/tool-file-editor/tool-file-editor.component.ts
```diff
@@ -66,6 +66,14 @@ export class ToolFileEditorComponent extends FileEditing {
     return this.currentVersion.id === mostRecentId;
   }
 
+  getIdOfNewestVersion(versions: Array<Tag>): number {
+    return versions.reduce((max, n) => Math.max(max, n.id), this.versions[0].id);
+  }
+
+  getNewestVersion(versions: Array<Tag>): Tag {
```
coverbeck, line 73: This is the same as the code in [workflow-file-editor.component.ts](https://github.com/dockstore/dockstore-ui2/blob/23d6e519f214dfaec4b9476a0ece63129a96a04a/src/app/workflow/workflow-file-editor/workflow-file-editor.component.ts#L82). It would be good if you could combine the two and put it in a service, then both this file and that file use the common method.

coverbeck (changes requested):

denis-yuen (commented):
For reviewers, this is https://ucsc-cgl.atlassian.net/browse/DOCK-1289
Approved after addressing feedback

latfariha:
recording (2)

updated recording of success message

src/app/container/tool-file-editor/tool-file-editor.component.ts
```diff
@@ -113,7 +113,10 @@ export class ToolFileEditorComponent extends FileEditing {
           // Only stop editing when version change was successful (not 204)
           this.toggleEdit();
           this.containerService.setTool(editedDockstoreTool);
-          this.alertService.detailedSuccess();
+          const updatedVersion = this.getNewestVersion(editedDockstoreTool.workflowVersions);
+          this.alertService.detailedSuccess(
+            'Saved version ' + updatedVersion.name + ' ' + 'of hosted workflow ' + editedDockstoreTool.namespace
```
coverbeck, line 118: Must fix: This is a message for a tool, not a workflow.

Optional: Trivial, but instead of `...+ ' ' + 'of hosted...`, you could just put the space directly in the second string, e.g., `+ ' of hosted`. Makes is a little easier to read. Or you could use string interpolation, as I believe I suggested in another PR, but that's a matter of preference.

src/app/shared/file-editing.ts
```diff
@@ -17,6 +17,8 @@
 import { Directive } from '@angular/core';
 import { AlertService } from './alert/state/alert.service';
 import { Files } from './files';
+import { WorkflowVersion } from './swagger';
+import { Tag } from './swagger';
```
coverbeck, line 21: Also minor/optional, but you should just have import statement for WorkflowVersion and Tag. IDE should do this for you automatically: `import { Tag, WorkflowVersion }from './swagger';`

coverbeck (changes requested):

garyluu (approved):
Not sure about others, but calling it "hosted entry" instead of "hosted tool" is fine with me.

src/app/container/tool-file-editor/tool-file-editor.component.ts
latfariha: no problem! sorry about not catching those minor errors. fixing right now

src/app/container/tool-file-editor/tool-file-editor.component.ts
```diff
@@ -113,7 +113,8 @@ export class ToolFileEditorComponent extends FileEditing {
           // Only stop editing when version change was successful (not 204)
           this.toggleEdit();
           this.containerService.setTool(editedDockstoreTool);
-          this.alertService.detailedSuccess();
+          const updatedVersion = this.getNewestVersion(editedDockstoreTool.workflowVersions);
+          this.alertService.detailedSuccess('Saved version ' + updatedVersion.name + ' of hosted entry ' + editedDockstoreTool.namespace);
```
coverbeck, line 117: I would have said `tool`, even though Gary suggests `entry`. Why not be specific? You use `workflow` in the corresponding workflow component. But since he and I disagree, it's fine to leave it.

coverbeck (approved):

NatalieEO (approved):

denis-yuen (approved):

src/app/container/tool-file-editor/tool-file-editor.component.ts
garyluu: It's because it adds no value to the user to be specific like that. However, there's a chance for developers to make mistakes like this (saying it's a workflow when it's not) which would confuse the user more. The same mistake can be made for services or any other entry type we may add in the future.

PR #1115, opened 2020-11-25
Feature/1781/elastic search response size

nrzhao4:
- for [SEAB-1323](https://ucsc-cgl.atlassian.net/browse/SEAB-1323)
- biggest response size was from the results-table query, before ES7 it was ~2mb consistently, the other responses average 30-60kb
- after upgrade to ES7, the shards decreased from 5 to 1 for each index and the response size for the results-table query shrank to ~900kb but load time still slow (especially the initial search page load)
- the query was excluding some sourcefiles and workflow versions, but was returning all the info relating to the entries
- changed the query to only include the properties needed to display the results (basically everything needed to populate each column of the table)
- this drastically reduced the response size to less than 100kb and the results load significantly faster


nrzhao4:
This PR also includes changes that were made in [this PR](https://github.com/dockstore/compose_setup/pull/171), since i branched it off the ES7 branch

coverbeck:
>     * biggest response size was from the results-table query, before ES7 it was ~2mb consistently, the other responses average 30-60kb

If I go to staging and [WDL and workflows](https://staging.dockstore.org/search?descriptorType=WDL&_index=workflows&searchMode=files), I get a 4.3 MB response (compresses to 246KB, but still).

What does it return with your changes?

Just starting to review, but these numbers look really good!

src/app/search/query-builder.service.ts
```diff
@@ -103,14 +101,23 @@ export class QueryBuilderService {
...
-    tableBody = this.excludeVerifiedContent(tableBody);
+    tableBody = this.sourceOptions(tableBody);
     tableBody = this.appendQuery(tableBody, values, advancedSearchObject, searchTerm);
     tableBody = this.appendFilter(tableBody, null, filters);
     const builtTableBody = tableBody.build();
     const tableQuery = JSON.stringify(builtTableBody);
     return tableQuery;
   }
 
+  getResultSingleIndexQuery(query_size: number, index: string): string {
```
coverbeck, line 112: Type of index should be `index: 'tools' | 'workflows'`

src/app/search/search.component.ts
```diff
@@ -410,6 +420,39 @@ export class SearchComponent implements OnInit, OnDestroy {
...
+      body: toolsQuery,
+    })
+      .then((tools) => {
+        toolHits = tools.hits.hits;
+        ELASTIC_SEARCH_CLIENT.search({
+          index: 'tools',
+          type: 'entry',
+          body: workflowsQuery,
+        })
+          .then((workflows) => {
```
coverbeck, line 445: I thought I reviewed this already, adding on to Gary's comment that you can use forkJoin. Are there 2 PRs in one here?

src/app/search/query-builder.service.ts
```diff
@@ -65,26 +65,24 @@ export class QueryBuilderService {
...
-        'workflowVersions.valid',
-        'workflowVersions.workflow_path',
-        'workflowVersions.workingDirectory',
-        'workflowVersions.reference',
-      ],
-    });
+  // These are the properties to return in the search to display the results table correctly
+  private sourceOptions(body: any) {
+    return body.rawOption('_source', [
+      'author',
```
coverbeck, line 71: I wish these strings weren't hard-coded like this, but I guess there's no easy solution. Also, was like this already.

nrzhao4:
> > ```
> > * biggest response size was from the results-table query, before ES7 it was ~2mb consistently, the other responses average 30-60kb
> > ```
> 
> If I go to staging and select [WDL and workflows](https://staging.dockstore.org/search?descriptorType=WDL&_index=workflows&searchMode=files), I get a 4.3 MB response (compresses to 246KB, but still).
> 
> What does it return with your changes?
> 
> Just starting to review, but these numbers look really good!

Its now two responses to query the two indices separately, each one returns a 36-38kb response

garyluu (approved):

NatalieEO (approved):

src/app/search/search.component.ts
```diff
@@ -410,6 +420,36 @@ export class SearchComponent implements OnInit, OnDestroy {
...
+   *
+   * @param {string} value the elastic search query
+   * @memberof SearchComponent
+   */
+  updateResults(toolsQuery: string, workflowsQuery: string) {
+    const tools = ELASTIC_SEARCH_CLIENT.search({
+      index: 'tools',
+      type: 'entry',
+      body: toolsQuery,
+    }).catch((error) => console.log(error));
```
coverbeck, line 437: This was like this, but I think here and on line 422 we should be surfacing the error to the user, rather than just silently failing. Everything that can fail will fail at some point. :) Could you please log it as a followup ticket?

coverbeck (approved):

src/app/search/search.component.ts
nrzhao4: Ok, logged here [3952](https://github.com/dockstore/dockstore/issues/3952)

PR #1116, opened 2020-12-02
Feature/2076/new tool tag and badges

nrzhao4:
- started implement [new search page UI ](https://projects.invisionapp.com/share/VNZ92Z0T2B3#/screens/436191418)
- changed Tag Cloud to Popular Keywords and moved button
- descriptorType column has badge for each language
Screen Shot 2020-12-02 at 1 19 43 PM
Screen Shot 2020-12-02 at 1 19 52 PM
Screen Shot 2020-12-02 at 1 20 01 PM


src/app/search/search-workflow-table/search-workflow-table.component.scss
```diff
@@ -10,3 +10,12 @@
 .mat-column-projectLinks {
   max-width: 105px;
 }
+
+.language-badge {
```
garyluu, line 14: Looks like a duplicate.  Could also use a comment on where the background color and color was gotten from and whether it's something we should be using for all "badges" (not to be confused with Material badges).

garyluu (commented):
For everyone: Just to confirm, this new uninteractable badge style is similar to the interactable Material Chips (https://material.io/components/chips) and we're all ok with that right?

coverbeck:
> For everyone: Just to confirm, this new uninteractable badge style is similar to the interactable Material Chips (https://material.io/components/chips) and we're all ok with that right?

Have you mentioned this to Kim? Also, is she on GitHub? Maybe we could add her to these reviews if so.

garyluu:
> > For everyone: Just to confirm, this new uninteractable badge style is similar to the interactable Material Chips (https://material.io/components/chips) and we're all ok with that right?
> 
> Have you mentioned this to Kim? Also, is she on GitHub? Maybe we could add her to these reviews if so.

Nope, when I first saw it, I thought it was interactable in some way I didn't know about (so just use a chip). Now that I see the code PR, it's not interactable.

kcullion:
@garyluu this sounds good. The "chips" are dismissible in this context. But usually they are just labels/tags.
Screen Shot 2020-12-02 at 3 33 43 PM

coverbeck:
I feel these badges look just different enough from Angular chips that they're OK 

* No `x`
* Rounding of corners is a little different.

But it's borderline; I could be tempted to click on one.

src/app/search/search-results/search-results.component.html
```diff
@@ -56,25 +65,34 @@
...
-            [data]="workflowTagCloudData"
-            [width]="options.width"
-            [height]="options.height"
-            [overflow]="options.overflow"
-            (clicked)="tagClicked($event)"
-          >
-          </angular-tag-cloud>
+        <div fxLayout="row wrap" fxLayoutAlign="space-between center">
+          <span fxFlex="65%">
+            <p>A <b>Workflow</b> can use multiple containers and exectues multiple actions or steps, outlined by a descriptor</p>
```
coverbeck, line 70: 1. A workflow can have multiple descriptors. Maybe "outlined by one or more descriptors"?
2. Typo in "exectues"

src/app/search/search-results/search-results.component.html
```diff
@@ -56,25 +65,34 @@
...
+          </span>
+          <div class="tag-cloud-dropdown">
```
coverbeck, line 72: This block looks the same as the block starting on line 30. Good candidate for a component. Your call; don't want to over-engineer either.

src/app/search/search-tool-table/search-tool-table.component.html
```diff
@@ -49,11 +49,15 @@
...
     </ng-container>
-    <ng-container matColumnDef="descriptorType">
-      <mat-header-cell fxShow fxHide.lt-sm *matHeaderCellDef mat-sort-header>Format</mat-header-cell>
-      <mat-cell data-cy="toolTypes" fxShow fxHide.lt-sm *matCellDef="let tool">{{
-        tool?.descriptorType ? (tool?.descriptorType.toString() | uppercase) : ''
-      }}</mat-cell>
+    <ng-container matColumnDef="descriptorType" style="width: fit-content">
+      <mat-header-cell fxShow fxHide.lt-sm *matHeaderCellDef mat-sort-header fxFlex="15%">Format</mat-header-cell>
+      <mat-cell data-cy="toolTypes" fxShow fxHide.lt-sm *matCellDef="let tool" fxFlex="15%">
+        <div *ngFor="let language of tool?.descriptorType; let i = index">
```
coverbeck, line 55: You're not using `i` -- unless there's some Angular reason you need it.

kcullion:
> I feel these badges look just different enough from Angular chips that they're OK
> 
> * No `x`
> * Rounding of corners is a little different.
> 
> But it's borderline; I could be tempted to click on one.

So is there no way to implement these material design components and access the css style to make a few adjustments (like colour)?

nrzhao4:
> > I feel these badges look just different enough from Angular chips that they're OK
> > 
> > * No `x`
> > * Rounding of corners is a little different.
> > 
> > But it's borderline; I could be tempted to click on one.
> 
> So is there no way to implement these material design components and access the css style to make a few adjustments (like colour)?

Should I be using the material design chips? They look like they're designed to be clickable, unlike the badges here. In which case simple CSS would be easier to implement

kcullion:
> > > I feel these badges look just different enough from Angular chips that they're OK
> > > 
> > > * No `x`
> > > * Rounding of corners is a little different.
> > > 
> > > But it's borderline; I could be tempted to click on one.
> > 
> > 
> > So is there no way to implement these material design components and access the css style to make a few adjustments (like colour)?
> 
> Should I be using the material design chips? They look like they're designed to be clickable, unlike the badges here. In which case simple CSS would be easier to implement

So I think there are three versions of this component. And maybe that could help you decide what material design component to use. Described in the following screenshots...

coverbeck:
> So is there no way to implement these material design components and access the css style to make a few adjustments (like colour)?

I would think so. Your later comment with the screenshot clarifies things for me. At this point I think it's a "detail" as to whether Nina implements all 3 variants with the chip or not, unless Gary has strong opinions about overloading the chip too much.

nrzhao4:
> > So is there no way to implement these material design components and access the css style to make a few adjustments (like colour)?
> 
> I would think so. Your later comment with the screenshot clarifies things for me. At this point I think it's a "detail" as to whether Nina implements all 3 variants with the chip or not, unless Gary has strong opinions about overloading the chip too much.

I played around with using a `mat-chip` element but I think for this case it makes more sense to use CSS, the mat-chip by default behaves as a clickable element so you'd have to change it a lot

garyluu:
> > > So is there no way to implement these material design components and access the css style to make a few adjustments (like colour)?
> > 
> > 
> > I would think so. Your later comment with the screenshot clarifies things for me. At this point I think it's a "detail" as to whether Nina implements all 3 variants with the chip or not, unless Gary has strong opinions about overloading the chip too much.
> 
> I played around with using a `mat-chip` element but I think for this case it makes more sense to use CSS, the mat-chip by default behaves as a clickable element so you'd have to change it a lot

Assuming everyone is fine with them looking similar then it boils down to:
Is it interactable?  Use and modify the mat-chip. Otherwise, use new custom css.

coverbeck:
Are your screenshots up-to-date? Could you please update if not? Thanks.

src/app/search/search-tool-table/search-tool-table.component.html
coverbeck: > You're not using `i` -- unless there's some Angular reason you need it.

Still seeing this.

src/app/search/search-results/search-results.component.html
coverbeck: I'm OK leaving this as is, but you have a thumbs up, but didn't change it. Just want to make sure that's your intent, and that you didn't overlook it.

nrzhao4: Yup, chose not to change it

nrzhao4:
> Are your screenshots up-to-date? Could you please update if not? Thanks.

Yup, they're up to date

coverbeck (approved):

src/app/search/search-workflow-table/search-workflow-table.component.scss
garyluu: also why do the tools and workflows have different padding?

nrzhao4: They look different if you put the exact same padding. Think it might have something to do with the fact that the tool styling is a span inside a div element for the ngFor, but for workflows its in a single span

garyluu (changes requested):
The language-badge style is still mostly duplicated. It seems like it's intended to be reused (not just for languages), needs to be placed in styles.scss

NatalieEO (approved):

garyluu (approved):

PR #1121, opened 2020-12-09
made some minor changes to the detailed success message

latfariha:
switched from "hosted entry" to "hosted tool" as suggested, and replaced the namespace var with name to get the hosted tool's name instead of it's namespace. The namespace and my hosted tool's names were similar, I think, so that was why I didn't catch it.

### DIFF
--- a/src/app/container/tool-file-editor/tool-file-editor.component.ts
+++ b/src/app/container/tool-file-editor/tool-file-editor.component.ts
@@ -119,7 +119,7 @@ export class ToolFileEditorComponent extends FileEditing {
               updatedVersion.name +
               ' of hosted tool ' +
               editedDockstoreTool.name +
-              (editedDockstoreTool.toolname ? '/' + editedDockstoreTool.toolname : null)
+              (editedDockstoreTool.toolname ? '/' + editedDockstoreTool.toolname : '')
           );
         } else {
           // Probably encountered a 204
--- a/src/app/container/tool-file-editor/tool-file-editor.component.ts
+++ b/src/app/container/tool-file-editor/tool-file-editor.component.ts
@@ -114,7 +114,13 @@ export class ToolFileEditorComponent extends FileEditing {
           this.toggleEdit();
           this.containerService.setTool(editedDockstoreTool);
           const updatedVersion = this.getNewestVersion(editedDockstoreTool.workflowVersions);
-          this.alertService.detailedSuccess('Saved version ' + updatedVersion.name + ' of hosted tool ' + editedDockstoreTool.toolname);
+          this.alertService.detailedSuccess(
+            'Saved version ' +
+              updatedVersion.name +
+              ' of hosted tool ' +
+              editedDockstoreTool.name +
+              (editedDockstoreTool.toolname ? '/' + editedDockstoreTool.toolname : null)
+          );
         } else {
           // Probably encountered a 204
           this.handleNoContentResponse();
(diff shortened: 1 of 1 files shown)


chmreid (approved):

src/app/container/tool-file-editor/tool-file-editor.component.ts
```diff
@@ -114,7 +114,7 @@ export class ToolFileEditorComponent extends FileEditing {
           this.toggleEdit();
           this.containerService.setTool(editedDockstoreTool);
           const updatedVersion = this.getNewestVersion(editedDockstoreTool.workflowVersions);
-          this.alertService.detailedSuccess('Saved version ' + updatedVersion.name + ' of hosted entry ' + editedDockstoreTool.namespace);
+          this.alertService.detailedSuccess('Saved version ' + updatedVersion.name + ' of hosted tool ' + editedDockstoreTool.toolname);
```
coverbeck, line 117: I too am confused by all the different parts of a tool :). @Richard-Hansen can you confirm this is what you're expecting?

coverbeck (approved):

src/app/container/tool-file-editor/tool-file-editor.component.ts
Richard-Hansen: I think this will print ```Saved version 1 of hosted tool null``` (or throw an error) for cases where toolname is not defined. 

I believe the complete format of an entry is as follows:
```registry/namespace/name/*toolname*```, where toolname is an optional parameter. See the bottom dialogue box.
Screen Shot 2020-12-10 at 2 42 01 PM

For example, both the following are valid hosted tool names:
```quay.io/richardhansen/myHostedTool```
```quay.io/richardhansen/myHostedTool/myCustomToolName```

I think we would want to print the custom tool name if it is present, but omit it otherwise. So, using a ternary, something along the lines of 
```editedDockstoreTool.toolname ? editedDockstoreTool.name + "/" + editedDockstoreTool.toolname : editedDockstoreTool.name```
 would return only the ```name``` variable if toolname is not truthy, otherwise it would return the string ```name/toolname```.

**Review by coverbeck — CHANGES_REQUESTED** [review]
Please see @Richard-Hansen 's comments.

File: src/app/container/tool-file-editor/tool-file-editor.component.ts [hunk 2]
```diff
@@ -114,7 +114,13 @@ export class ToolFileEditorComponent extends FileEditing {
           this.toggleEdit();
           this.containerService.setTool(editedDockstoreTool);
           const updatedVersion = this.getNewestVersion(editedDockstoreTool.workflowVersions);
-          this.alertService.detailedSuccess('Saved version ' + updatedVersion.name + ' of hosted entry ' + editedDockstoreTool.namespace);
+          this.alertService.detailedSuccess(
+            'Saved version ' +
+              updatedVersion.name +
+              ' of hosted tool ' +
+              editedDockstoreTool.name +
+              (editedDockstoreTool.toolname ? '/' + editedDockstoreTool.toolname : null)
```
Richard-Hansen, line 122: I believe this line will add the string 'null' to the end of the message. Javascript is funky and will convert null and undefined to strings when you try and concatenate them with a string variable, so:
Screen Shot 2020-12-11 at 2 54 22 PM


You could replace ```null``` with an empty string, ```""```, which would work as intended.

Richard-Hansen (changes requested):

Richard-Hansen (approved):

coverbeck (approved):

PR #1126, opened 2020-12-22
Feature/seab 2108/bodybuilder update

garyluu:
For SEAB-2108

Sadly, bodybuilder update didn't do much. On the positive side, elasticsearch-browser can be removed because the frontend doesn't interract with ES directly.  Previously we pretended the Dockstore endpoint behaved the same as the ES endpoint which isn't true anymore with ES7.  ES7 expects the endpoint to accept a 'type' query parameter

This is about how much space ES-browser took up:

image


denis-yuen (approved):
> This is about how much space ES-browser took up:

:fireworks:

src/app/search/search.component.ts
```diff
@@ -389,16 +390,19 @@ export class SearchComponent implements OnInit, OnDestroy {
...
-    })
-      .then((hits) => {
-        this.setupAllBuckets(hits);
-        this.setupOrderBuckets();
-      })
-      .catch((error) => console.log(error));
+    this.extendedGA4GHService.toolsIndexSearch(value).subscribe(
+      (hits: any) => {
+        if (hits) {
+          this.setupAllBuckets(hits);
```
coverbeck, line 396: Without understanding entirely, if there aren't any hits, don't you need to "clear"? Although look at `setupAllBuckets()`, it would have failed before as well.

src/app/search/search.component.ts
```diff
@@ -438,23 +439,18 @@ export class SearchComponent implements OnInit, OnDestroy {
...
+    const workflowsObservable: Observable<any> = this.extendedGA4GHService.toolsIndexSearch(workflowsQuery);
+    forkJoin([toolsObservable, workflowsObservable]).subscribe((results: Array<any>) => {
+      if (results[0] && results[1]) {
+        const toolHits = results[0].hits.hits;
+        const workflowHits = results[1].hits.hits;
+        this.hits = toolHits.concat(workflowHits);
+        const filteredHits: [Array<Hit>, Array<Hit>] = this.searchService.filterEntry(this.hits, this.query_size_full);
+        this.searchService.setHits(filteredHits[0], filteredHits[1]);
+      } else {
+        console.error('No hits returned');
```
coverbeck, line 452: Same question about no hits as my other comment

coverbeck (approved):

src/app/search/search.component.ts
garyluu: ya, that's true. i put this hit check for testing purposes only, it's never falsey. it won't affect anything in practice but i suppose i should find a better solution

NatalieEO (approved):

wshands (approved):

PR #1129, opened 2021-01-12
Have sample code when creating new descriptor files for hosted tool/workflow

GFJHogue:
Main ticket: https://github.com/dockstore/dockstore/issues/3518

When composing a new CWL/WDL/NF descriptor file for a hosted tool/workflow, the editor will initialize with some sample code. The file contents are selected and the editor has focus. This is to make life a little easier for a beginner.

Editor behaviour for existing files and non-descriptors should remain the essentially the same as before (except it focuses now).

I picked the best "hello world" code I could find, but feel free to suggest better alternatives for each language.


coverbeck (commented):
Is this going to load the same files for both tools and workflows? Won't the CWL be invalid for both? Or maybe a CWL tool can have an embedded tool in it...

GFJHogue:
> Is this going to load the same files for both tools and workflows? Won't the CWL be invalid for both? Or maybe a CWL tool can have an embedded tool in it...

@coverbeck Yeah... the CWL file I made is invalid as it has examples for both a Tool & Workflow (see comments in `helloworld.cwl`).

My thought was that the user would be editing off the sample code anyway, so maybe it's alright if the comments are clear enough? (ie. `Use this for a TOOL:`, `Use this for WORKFLOW:`, etc..)

If we want it to have different sample code for each context, then it'll be a little more tricky (need to inform the editor whether the file is for a Tool/Workflow).

src/assets/text/helloworld.cwl
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env cwl-runner
+
+# 1st-tool.cwl example from https://www.commonwl.org/user_guide/02-1st-example/
```
denis-yuen, line 3: Add example file format metadata
https://www.commonwl.org/user_guide/16-file-formats/index.html

src/assets/text/helloworld.cwl
```diff
@@ -0,0 +1,38 @@
...
+inputs:
+  message:
+    type: string
+    inputBinding:
+      position: 1
+outputs: []
+
+# 1st-workflow.cwl example from https://www.commonwl.org/user_guide/21-1st-workflow/index.html
+cwlVersion: v1.0
+class: Workflow
```
denis-yuen, line 16: Add a dummy example author metadata
https://www.commonwl.org/user_guide/17-metadata/index.html

src/assets/text/helloworld.nf
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env nextflow
+
+// tutorial.nf example from https://www.nextflow.io/docs/latest/getstarted.html
+
```
denis-yuen, line 4: Ditto for authorship and description
https://docs.dockstore.org/en/develop/advanced-topics/best-practices/nfl-best-practices.html?highlight=metadata#authorship-metadata

src/assets/text/helloworld.cwl
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env cwl-runner
+
+# 1st-tool.cwl example from https://www.commonwl.org/user_guide/02-1st-example/
+cwlVersion: v1.0
+class: CommandLineTool
+baseCommand: echo
```
denis-yuen, line 6: Should also have label and doc

src/assets/text/helloworld.nf
```diff
@@ -0,0 +1,31 @@
...
+    """
+    printf '${params.str}' | split -b 6 - chunk_
+    """
+}
+
+
+process convertToUpper {
+
+    input:
```
denis-yuen, line 20: Example should use a pinned ubuntu Docker image

src/assets/text/helloworld.wdl
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  command {
+    echo 'hello ${name}!'
+  }
+  output {
+    File response = stdout()
+  }
+  runtime {
+    docker: 'ubuntu:latest'
```
denis-yuen, line 15: Pinned image
and metadata https://docs.dockstore.org/en/develop/advanced-topics/best-practices/wdl-best-practices.html?highlight=metadata#authorship-metadata

denis-yuen (changes requested):

denis-yuen:
> If we want it to have different sample code for each context, then it'll be a little more tricky (need to inform the editor whether the file is for a Tool/Workflow).

 Seems like it should be worth doing but can spin out ticket if the implementation is tricky

src/assets/text/helloworld.cwl
GFJHogue: Might just be easier to just copy & use this example for CWL Tools (maybe just swap out the names & emails).

src/assets/text/helloworld.wdl
GFJHogue: Again, might be better to use a modified version of our bamstats example since it's more complete.

denis-yuen: Whatever works, just want to show a "best practices" example

src/assets/text/helloworld.nf
GFJHogue: Also need to add a sample `nextflow.config`

GFJHogue:
> > If we want it to have different sample code for each context, then it'll be a little more tricky (need to inform the editor whether the file is for a Tool/Workflow).
> 
> Seems like it should be worth doing but can spin out ticket if the implementation is tricky

Turned out to be much easier than I thought - separate CWL Tool & Workflow sample code now!

src/app/shared/code-editor/code-editor.component.ts
```diff
@@ -15,6 +16,7 @@ export class CodeEditorComponent implements AfterViewInit {
   editorFilepath: string;
   aceId: string;
   readOnly = true;
+  @Input() entryType: string;
```
coverbeck, line 19: Type should be `'tool' | 'workflow'`

src/assets/text/helloworld.wdl
```diff
@@ -0,0 +1,35 @@
...
+        File bamstats_report = "bamstats_report.zip"
+    }
+
+    runtime {
+        docker: "quay.io/collaboratory/dockstore-tool-bamstats:1.25-6_1.0"
+        memory: mem_gb + "GB"
+    }
+}
+
+workflow bamstatsWorkflow {
```
coverbeck, line 29: Is this both a valid tool and valid workflow? I suspect yes, since we've said a tool is just a workflow with one task, but just making sure.

src/app/shared/code-editor/code-editor.component.ts
```diff
@@ -57,13 +59,43 @@ export class CodeEditorComponent implements AfterViewInit {
...
+        sampleCodeUrl = 'assets/text/helloworld.wdl';
+      } else if (this.mode === 'nfl') {
+        sampleCodeUrl = 'assets/text/helloworld.nf';
+      } else if (this.editorFilepath === '/nextflow.config') {
+        sampleCodeUrl = 'assets/text/nextflow.config';
+      }
+    }
+    if (sampleCodeUrl) {
+      const httpOptions: Object = { responseType: 'text' };
+      this.httpClient.get(sampleCodeUrl, httpOptions).subscribe(setContent);
```
coverbeck, line 96: It's unfortunate that it's another http request, but there's always a tradeoff. @garyluu any strong opinions on this? These files are so small it probably doesn't matter either way, but wondering if this best practice or not.

garyluu: I don't have a better solution that this right now.

src/assets/text/helloworld.wdl
GFJHogue: It shows as valid when I use it for a Workflow as-is (even though _technically_ a Workflow should have 2 tasks)

denis-yuen (approved):

src/app/shared/code-editor/code-editor.component.ts
GFJHogue: Yeah, in my mind it's either this or add bloat to the bundle.

coverbeck (approved):

src/app/shared/code-editor/code-editor.component.ts
```diff
@@ -57,13 +59,44 @@ export class CodeEditorComponent implements AfterViewInit {
...
+      if (this.editorContent) {
+        this.editor.setValue(this.editorContent, cursorPos);
+      }
+
+      this.editor.focus();
+    };
+
+    let sampleCodeUrl = '';
+
+    // Load helloworld files by default when editing empty CWL/WDL/NF files
```
garyluu, line 79: https://github.com/dockstore/dockstore/blob/develop/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java#L61 are they ways we describe Nextflow

NatalieEO (approved):
Probably want to rename the example files you added since they seem they aren't hello worlds anymore. 

Also, I don't feel as strongly about this since you already redid the files once by request of others, but I'm wondering if it would've been better if we had used and referenced the trainings Louise came up with for the bosc training https://github.com/dockstore/bcc2020-training

garyluu (approved):

GFJHogue:
> Also, I don't feel as strongly about this since you already redid the files once by request of others, but I'm wondering if it would've been better if we had used and referenced the trainings Louise came up with for the bosc training https://github.com/dockstore/bcc2020-training

Quickly looking through the exercises, the basic helloworld exercises are missing some language features we want to show (see @denis-yuen's comments above). I suppose it would be possible to extend these if others think it's a good idea.

It's also easy enough to change the sample code in the future (after we get some feedback).

denis-yuen:
> It's also easy enough to change the sample code in the future (after we get some feedback).

Ditto, I think this works for now

PR #1130, opened 2021-01-18
Feature/3805/angular material clipboard

wendidy:
- [`dockstore/dockstore#3805`](https://github.com/dockstore/dockstore/issues/3805): Removed `ngx-clipboard` dependency, and changed to using Angular Material Clipboard
- [`dockstore/dockstore#3941`](https://github.com/dockstore/dockstore/issues/3941) : Fixed duplicate `matTooltip ` double hover message


package-lock.json
```diff
@@ -156,9 +142,9 @@
       },
       "dependencies": {
         "@angular-devkit/core": {
-          "version": "10.2.0",
-          "resolved": "https://registry.npmjs.org/@angular-devkit/core/-/core-10.2.0.tgz",
-          "integrity": "sha512-XAszFhSF3mZw1VjoOsYGbArr5NJLcStjOvcCGjBPl1UBM2AKpuCQXHxI9XJGYKL3B93Vp5G58d8qkHvamT53OA==",
+          "version": "10.2.1",
```
coverbeck, line 145: Is this and all the Angular changes correct? Doesn't seem like they should have updated for removing one module.

src/app/workflow/dag/state/dag.service.ts
```diff
@@ -29,7 +29,7 @@ import { DagStore } from './dag.store';
 
 @Injectable()
 export class DagService {
-  readonly style = [
+  readonly style: cytoscape.Stylesheet[] = [
```
coverbeck, line 32: Why did this change?

coverbeck (commented):
There is a failing integration test, which might be flaky, but might be genuine.

src/app/workflow/dag/state/dag.service.ts
wendidy: @coverbeck I was running into some run time errors with styling(don't remember the exact message), inquired Gary, that's his fix

garyluu: Didn't look too much into why, but this branch somehow caused a type conflict in the cytoscape style. This specific line didn't need to change, but might as well for easier debugging in the future.

package-lock.json
garyluu: Looks like what would've happened if one ran `npm update`

garyluu (changes requested):
1. Revert all your package-lock.json and package.json changes. `npm uninstall ngx-clipboard` and save those changes. Those changes should be vastly different than your current ones
2. Remove line 113 of my fix (it should have compilation errors) if the package-lock.json and package.json is correctly done. You can keep line 32.

wendidy:
@garyluu two .json files are being reverted, but if I were to revert line 113 of your changes, it throws an error: Type 'string' is not assignable to type 'number'.

garyluu:
> @garyluu two .json files are being reverted, but if I were to revert line 113 of your changes, it throws an error: Type 'string' is not assignable to type 'number'.

reload your vscode

wendidy:
@garyluu Thank you. But same thing, also tried putting quotations around the name. But it seems like only having no quotations at all on this line works
image

garyluu:
> @garyluu Thank you. But same thing, also tried putting quotations around the name. But it seems like only having no quotations at all on this line works
> image

push your changes, i'll check it out

garyluu:
hmm...looks like our current style's type is not compatible with our current @types/cytoscape. leave out the entire fix then (don't change dag.service.ts)

wendidy:
@garyluu So I will leave it as it is right now? Or remove the quotations

garyluu:
> @garyluu So I will leave it as it is right now? Or remove the quotations

When you go to https://github.com/dockstore/dockstore-ui2/pull/1130/files . there should be no changes to dag.service.ts

coverbeck (commented):
I'm still wondering why there are so many package-lock.json changes, even though it's much better than before... I think the only thing that should have changed is ngx-clipboard being removed. I did it this way from the develop branch, without your changes (I had to look it up), and it seems to give the desired result:

```
npm uninstall ngx-clipboard --save
```

wendidy:
@coverbeck Thank you Charles! I will try that right now. Before I ran it without `--save`. It's still interesting how I only did `npm ci` after, it's not supposed to modify package-lock.json file.

garyluu (approved):

coverbeck (approved):
You have a failing check which seems to be blocking. Took a brief look, and it looks OK to me (it's just an HTML snippet that has to exist in many places), but you may need an admin to do the merge.

PR #1134, opened 2021-01-25
Feature/3942/private tool status bar text

wendidy:
[dockstore/dockstore#3942](https://github.com/dockstore/dockstore/issues/3942)
This feature fixes a part of the issue. It takes care of displaying registry name properly when hovered over the "Request Access" button. The rest of the solution to this issue please refer to the [other PR](https://github.com/dockstore/dockstore-ui2/pull/1135)

### DIFF
--- a/src/app/container/container.component.ts
+++ b/src/app/container/container.component.ts
@@ -59,7 +59,7 @@ export class ContainerComponent extends Entry implements AfterViewInit {
   containerEditData: any;
   thisisValid = true;
   ModeEnum = DockstoreTool.ModeEnum;
-  public requestAccessHREF: string;
+  public requestAccessHREF$: Observable<string>;
   public contactAuthorHREF: string;
   public missingWarning: boolean;
   public tool: ExtendedDockstoreTool;
@@ -197,13 +197,14 @@ export class ContainerComponent extends Entry implements AfterViewInit {
     if (tool) {
       this.tool = tool;
       this.initTool();
-      this.extendedTool$.pipe(takeUntil(this.ngUnsubscribe)).subscribe(data => {
+      this.requestAccessHREF$ = this.extendedTool$.pipe(map(data => {
         if (Object.keys(data).length !== 0 ) {
-          this.requestAccessHREF = this.emailService.composeRequestAccessEmail(this.tool, data.imgProvider);
+          console.log(data);
+          return this.emailService.composeRequestAccessEmail(this.tool, data.imgProvider);
         } else {
-          this.requestAccessHREF = this.emailService.composeRequestAccessEmail(this.tool, 'not-valid');
+          return this.emailService.composeRequestAccessEmail(this.tool, 'undefined');
         }
-      });
+      }));
       this.contactAuthorHREF = this.emailService.composeContactAuthorEmail(this.tool);
       this.sortedVersions = this.getSortedTags(this.tool.workflowVersions, this.defaultVersion);
       this.updateVerifiedPlatforms(this.tool.id);
--- a/src/app/container/container.component.ts
+++ b/src/app/container/container.component.ts
@@ -62,7 +62,7 @@ export class ContainerComponent extends Entry implements AfterViewInit {
   public requestAccessHREF$: Observable<string>;
   public contactAuthorHREF: string;
   public missingWarning: boolean;
-  public tool: ExtendedDockstoreTool;
+  public tool: DockstoreTool;
   public toolCopyBtn: string;
   public sortedVersions: Array<Tag | WorkflowVersion> = [];
   public DockstoreToolType = DockstoreTool;
@@ -198,12 +198,8 @@ export class ContainerComponent extends Entry implements AfterViewInit {
       this.tool = tool;
       this.initTool();
       this.requestAccessHREF$ = this.extendedTool$.pipe(
-        map((data) => {
-          if (data.imgProvider) {
-            return this.emailService.composeRequestAccessEmail(this.tool, data.imgProvider);
-          } else {
-            return this.emailService.composeRequestAccessEmail(this.tool, 'undefined');
-          }
+        map((extendedTool) => {
+          return this.emailService.composeRequestAccessEmail(extendedTool);
         })
       );
       this.contactAuthorHREF = this.emailService.composeContactAuthorEmail(this.tool);
--- a/src/app/container/container.component.ts
+++ b/src/app/container/container.component.ts
@@ -197,14 +197,15 @@ export class ContainerComponent extends Entry implements AfterViewInit {
     if (tool) {
       this.tool = tool;
       this.initTool();
-      this.requestAccessHREF$ = this.extendedTool$.pipe(map(data => {
-        if (Object.keys(data).length !== 0 ) {
-          console.log(data);
-          return this.emailService.composeRequestAccessEmail(this.tool, data.imgProvider);
-        } else {
-          return this.emailService.composeRequestAccessEmail(this.tool, 'undefined');
-        }
-      }));
+      this.requestAccessHREF$ = this.extendedTool$.pipe(
+        map((data) => {
+          if (data.imgProvider) {
+            return this.emailService.composeRequestAccessEmail(this.tool, data.imgProvider);
+          } else {
+            return this.emailService.composeRequestAccessEmail(this.tool, 'undefined');
+          }
+        })
+      );
       this.contactAuthorHREF = this.emailService.composeContactAuthorEmail(this.tool);
       this.sortedVersions = this.getSortedTags(this.tool.workflowVersions, this.defaultVersion);
       this.updateVerifiedPlatforms(this.tool.id);
--- a/src/app/container/container.component.ts
+++ b/src/app/container/container.component.ts
@@ -21,7 +21,7 @@ import { MatDialog } from '@angular/material/dialog';
 import { ActivatedRoute, Router } from '@angular/router';
 import { AlertService } from 'app/shared/alert/state/alert.service';
 import { Observable } from 'rxjs';
-import { takeUntil } from 'rxjs/operators';
+import { map, takeUntil } from 'rxjs/operators';
 import { ListContainersService } from '../containers/list/list.service';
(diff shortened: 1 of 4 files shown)


src/app/container/container.component.ts
```diff
@@ -197,8 +197,15 @@ export class ContainerComponent extends Entry implements AfterViewInit {
     if (tool) {
       this.tool = tool;
       this.initTool();
+      this.requestAccessHREF$ = this.extendedTool$.pipe(map(data => {
+        if (Object.keys(data).length !== 0 ) {
+          console.log(data);
```
coverbeck, line 202: Best not to log this.

src/app/container/email.service.ts
```diff
@@ -11,47 +11,15 @@ export class EmailService {
...
-  private getRequestEmailMailTo(maintainerEmail: string, email: string): string {
-    return this.dockstoreService.getRequestAccessEmail(maintainerEmail, email);
-  }
-
   /**
    * This gets the email subject when the Request Access button is clicked
    * @param path The path of the tool whose access is being requested
    */
-  private getRequestEmailSubject(path: string): string {
+  private static getRequestEmailSubject(path: string): string {
```
coverbeck, line 22: Why did you make this and other methods static?

src/app/container/container.component.ts
```diff
@@ -197,8 +197,15 @@ export class ContainerComponent extends Entry implements AfterViewInit {
     if (tool) {
       this.tool = tool;
       this.initTool();
+      this.requestAccessHREF$ = this.extendedTool$.pipe(map(data => {
+        if (Object.keys(data).length !== 0 ) {
```
coverbeck, line 201: Since you only need imgProvider, suggest you just look for that, instead of seeing if it has many properties, one of which may or may not be imgProvider.
```
if (data.imgProvider) {
```

src/app/container/container.component.ts
```diff
@@ -197,8 +197,15 @@ export class ContainerComponent extends Entry implements AfterViewInit {
...
+          console.log(data);
+          return this.emailService.composeRequestAccessEmail(this.tool, data.imgProvider);
+        } else {
+          return this.emailService.composeRequestAccessEmail(this.tool, 'undefined');
```
coverbeck, line 205: I don't think 'undefined' should be part of the message.

src/app/container/email.service.ts
wendidy: Gary suggested to make some functions in this file static so testing is easier. I think this is what he is talking about

src/app/container/container.component.ts
wendidy: I feel so bad, can't believe I overlooked this...thank you Charles!

src/app/container/container.component.ts
wendidy: I was actually not sure what to put in. This line is there so just in case imgProvider is not defined, there will be a string, and we know this case is handled properly. 
Because I believe the message we had before was also "undefined", but we didn't know what was missing

src/app/container/container.component.ts
```diff
@@ -197,8 +197,16 @@ export class ContainerComponent extends Entry implements AfterViewInit {
     if (tool) {
       this.tool = tool;
       this.initTool();
+      this.requestAccessHREF$ = this.extendedTool$.pipe(
+        map((data) => {
```
garyluu, line 201: rename to extendedTool

src/app/container/container.component.ts
```diff
@@ -59,10 +59,10 @@ export class ContainerComponent extends Entry implements AfterViewInit {
   containerEditData: any;
   thisisValid = true;
   ModeEnum = DockstoreTool.ModeEnum;
-  public requestAccessHREF: string;
+  public requestAccessHREF$: Observable<string>;
   public contactAuthorHREF: string;
   public missingWarning: boolean;
-  public tool: DockstoreTool;
+  public tool: ExtendedDockstoreTool;
```
garyluu, line 65: This is not an ExtendedDockstoreTool

src/app/container/email.service.ts
```diff
@@ -70,23 +38,55 @@ export class EmailService {
...
-  private getInquiryEmailBody(): string {
+  private static getInquiryEmailBody(): string {
     return '';
   }
+
+  /**
+   * Composes the href for the Request Access email
+   * @param tool The tool to request access for
+   */
+  public composeRequestAccessEmail(tool: ExtendedDockstoreTool, imgProvider: string): string {
```
garyluu, line 65: signature should've remained the same

src/app/container/container.component.ts
```diff
@@ -197,8 +197,16 @@ export class ContainerComponent extends Entry implements AfterViewInit {
     if (tool) {
       this.tool = tool;
       this.initTool();
+      this.requestAccessHREF$ = this.extendedTool$.pipe(
+        map((data) => {
+          if (data.imgProvider) {
+            return this.emailService.composeRequestAccessEmail(this.tool, data.imgProvider);
```
garyluu, line 203: the extendedTool should've been passed in here, not `this.tool`

src/app/container/container.component.ts
```diff
@@ -197,8 +197,16 @@ export class ContainerComponent extends Entry implements AfterViewInit {
...
+          } else {
+            return this.emailService.composeRequestAccessEmail(this.tool, 'undefined');
```
garyluu, line 205: this entire section should just be one return statement. the logic for checking imgProvider should be done in getRequestEmailBody

garyluu (changes requested):

src/app/container/container.component.ts
wendidy: you are referring to data?

src/app/container/container.component.ts
wendidy: I'm still a bit concerned about the problem with consistency, because the parameter names and their types are all the same for a couple functions in `email.service.ts`. Developers might assume they are all the same but in the end they are not. Or I could change `public composeContactAuthorEmail(tool: ExtendedDockstoreTool): string` in `email.service.ts` - their parameter to be type `DockstoreTool`. That might be clearer? @garyluu

wendidy: It's probably more clear that I just push my code for you to see

src/app/container/container.component.ts
garyluu: yes

src/app/container/container.component.ts
garyluu: current changes look fine

garyluu (approved):
I'm ok with this assuming Charle's `getRequestEmailBody()` changes are addressed.

src/app/container/email.service.ts
```diff
@@ -60,8 +29,12 @@ export class EmailService {
    * @param path The path of the tool whose access is being requested
    * @param toolRegistry The tool registry (Quay.io, GitLab, etc) of the tool whose access is being requested
    */
-  private getRequestEmailBody(path: string, toolRegistry: string): string {
-    return encodeURI(`I would like to request access to your Docker image ${path}. ` + `My user name on ${toolRegistry} is <username>`);
+  private static getRequestEmailBody(path: string, toolRegistry: string): string {
+    const imgProvider: string = toolRegistry ? toolRegistry : 'undefined';
+    return encodeURI(
+      `I would like to request access to your Docker image ${path}. ` +
+        `My user name on ${imgProvider} is <put your Docker Hub/quay.io username here>`
```
coverbeck, line 36: 1. `username` instead of `user name` (you have have it as one word later).
2. For `<put your Docker Hub/quay.io username here>`, my suggestion was to actually specify the registry, since we know it, `put your ${imgProvider} username here`.
3. But in discussing it with Gary, he suggested "My username on ${imgProvider} is <username>", so I lean towards that approach.

src/app/container/email.service.ts
wendidy: @coverbeck To be honest I'm not sure myself, but my understanding is that since these functions are static, we don't need to instantiate them over and over again along side with the instantiation of the class object. Plus, since all of these class objects have access to the functions, and they should store the same info across, so there's no need to pass in new parameters. The above is my understanding, it probably doesn't have anything to do with testing in specific. But I guess Gary could explain better? @garyluu

garyluu: @coverbeck Email.service.ts is essential a file full of helper functions. The functions should not have any side effects (like subscriptions).  It's made to static in the event others are tempted to do something like subscribe to tokens/users within the file. Can you explain why you don't want it static?

coverbeck: I'm wasn't necessarily against it. I just saw the changes to add static, but didn't see any code that relied on the change (or am I wrong?), and was wondering why it was changed.

Your explanation " It's made to static in the event others are tempted..." works for me.

wendidy: @garyluu The thing is I did not make all private functions static. There's one that needs to access this.something, so I left it as it is

garyluu: @wendidy yes, i'm well aware not all of them were changed (even though the could've). not really a high priority, so I didn't request changes for that

NatalieEO (approved):
Once tests pass

src/app/container/email.service.ts
```diff
@@ -60,8 +29,11 @@ export class EmailService {
    * @param path The path of the tool whose access is being requested
    * @param toolRegistry The tool registry (Quay.io, GitLab, etc) of the tool whose access is being requested
    */
-  private getRequestEmailBody(path: string, toolRegistry: string): string {
-    return encodeURI(`I would like to request access to your Docker image ${path}. ` + `My user name on ${toolRegistry} is <username>`);
+  private static getRequestEmailBody(path: string, toolRegistry: string): string {
+    const imgProvider: string = toolRegistry ? toolRegistry : 'undefined';
+    return encodeURI(
+      `I would like to request access to your Docker image ${path}. ` + `My username on ${imgProvider} is <input username here>`
```
coverbeck, line 35: You don't need to change this, but you could have gotten rid of the plus/concatenation operator, and just make it one string between backticks.

coverbeck (approved):

PR #1135, opened 2021-01-25
 feature/3942/DisplayUsername

wendidy:
[dockstore/dockstore#3942](https://github.com/dockstore/dockstore/issues/3942)
This feature takes care of displaying username properly when hovered over the "Request Access" button. The rest of the solution to this issue please refer to the other [PR](https://github.com/dockstore/dockstore-ui2/pull/1134)

### DIFF
--- a/src/app/container/container.component.ts
+++ b/src/app/container/container.component.ts
@@ -42,6 +42,7 @@ import { WorkflowVersion } from '../shared/swagger/model/workflowVersion';
 import { ToolQuery } from '../shared/tool/tool.query';
 import { ToolService } from '../shared/tool/tool.service';
 import { TrackLoginService } from '../shared/track-login.service';
+import { UserQuery } from '../shared/user/user.query';
 import { ExtendedDockstoreTool } from './../shared/models/ExtendedDockstoreTool';
 import { ContainersService } from './../shared/swagger/api/containers.service';
 import { DockstoreTool } from './../shared/swagger/model/dockstoreTool';
@@ -72,6 +73,7 @@ export class ContainerComponent extends Entry implements AfterViewInit {
   public schema: BioschemaTool;
   public extendedTool$: Observable<ExtendedDockstoreTool>;
   public isRefreshing$: Observable<boolean>;
+  public username: string;
   constructor(
     private dockstoreService: DockstoreService,
     dateService: DateService,
@@ -97,7 +99,8 @@ export class ContainerComponent extends Entry implements AfterViewInit {
     public dialog: MatDialog,
     private toolService: ToolService,
     alertService: AlertService,
-    entryService: EntriesService
+    entryService: EntriesService,
+    private userQuery: UserQuery
   ) {
     super(
       trackLoginService,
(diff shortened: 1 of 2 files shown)


denis-yuen:
FYI, should probably look into lint issue

wendidy:
@denis-yuen Thanks Denis! I was just looking into that, it seems like the spec file has some problem with testing a function with the wrong number of parameters. Should I manually fix it?

src/app/container/container.component.ts
```diff
@@ -197,8 +200,11 @@ export class ContainerComponent extends Entry implements AfterViewInit {
     if (tool) {
       this.tool = tool;
       this.initTool();
+      this.userQuery.user$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((userData) => {
+        this.username = userData.username ? userData.username : 'not-logged-in';
```
coverbeck, line 204: 1. Doesn't this mean that if you're not logged in, an email will be sent saying "My user name... is not-logged-in"? I guess you have a chance to edit the email before sending it though, don't you?
2. The message in email.service.ts says "My user name on ${toolRegistry} is ${username}", but you're passing the Dockstore username, not the toolRegistry username.

wendidy: 1. Yes you will be able to edit the email before sent!
2. Oh no sorry I had a misunderstanding...have no idea why I thought username on dockstore is the username used there. Now I have been trying to retrieve user information, but it seems like the tool registry username is not stored anywhere in `ExtendedDockstoreTool.ts` or `dockstoreTool.ts`. I'm wondering if you, or any others have any insights on this?(Should I ask this question in the channel?)

denis-yuen:
> @denis-yuen Thanks Denis! I was just looking into that, it seems like the spec file has some problem with testing a function with the wrong number of parameters. Should I manually fix it?

Yeah, this isn't an automatically generated file https://github.com/dockstore/dockstore-ui2/blame/develop/src/app/container/email.service.spec.ts

src/app/container/container.component.ts
garyluu: A few things:
- The username should be the user's username from the Docker registry
- Dockstore may or may not have that username (if the account's not linked in Dockstore)
- If the account is not linked, there's not much we can do. User will have to fill it out.
- If we want to display the username for the accounts that are linked, it would likely involve getting the user's current tokens, hopefully finding the registry there, getting the username from the matching token.

IMO, it's not worth the effort.

What does everyone else think?

NatalieEO: I would think not worth it. At least for now. We only have 2 private tools.

coverbeck: There's also the case of a user not even being logged in to Dockstore at all...

I don't think it's worth a lot of effort, and I think we could have a the message saying something like "<put your Docker Hub/quay.io username here>" (or put the actual name if it's easily accessible -- how does the accounts page do it?).

I don't think we should have a message that says "not-loggged-in" or has the wrong username.

wendidy: So I assume we can just display the message like <put your Docker Hub/quay.io username here> so the user can edit it themselves while sending the email request? If this is the case I will just modify the other PR since it's a very small fix. Then we can disregard this PR

coverbeck: I'm OK with some version of that, although to be clear, you'll know if it's Docker Hub or quay.io, so the message should be specific. But let's see what Gary and Denis say.

denis-yuen: Yeah, Charles's suggestion works for me

PR #1144, opened 2021-02-02
fix license checker script

NatalieEO:
A million commits later and I think this is working. I added a new npm command just for circle-ci to use. For some reason `license-checker --excludePackages dockstore-ui2@2.3.0 --csv` can't run from within the bash script on circleci.


scripts/detect-package-json-changes.sh
```diff
@@ -1,14 +1,10 @@
 #!/bin/bash
 
-packageJsonStatus="$(git diff --name-status develop -- package.json)"
-packageLockJsonStatus="$(git diff --name-status develop -- package-lock.json)"
+diff CIRCLE-THIRD-PARTY-LICENSES.csv THIRD-PARTY-LICENSES.csv
```
coverbeck, line 3: You don't need this line

.circleci/config.yml
```diff
@@ -128,7 +129,21 @@ jobs:
           password: $DOCKERHUB_PASSWORD
     steps:
       - checkout
-      - run: bash scripts/detect-package-json-changes.sh
+      - install_container_dependencies
+      - run:
```
coverbeck, line 133: I think further down, around line 245/260, you can make licenses_test require lint_unit_test_coverage. Then it's already been checked out for you and you don't have to do npm ci.

Or maybe it's easier to just get rid of licenses_test as a job, and just move lines 141 /146 into lint_unit_test_coverage.

Sort of nit-picky, but it would be nice to avoid the duplication of running npm ci, both in terms of code duplication and it executing twice.

NatalieEO: I'll try requiring lint_unit_test_coverage. I tried just putting it in the lint_unit_test_coverage job, but I got errors that the files were different. I couldn't reproduce them being different locally. I'm not sure what step in lint_unit_test_coverage makes that happen.

package.json
```diff
@@ -13,7 +13,8 @@
     "prebuild.prod": "npm run prebuild",
     "build.prod": "npx ng build --progress false --prod --subresource-integrity",
     "build": "npx ng build --progress false --subresource-integrity",
-    "license": "license-checker  --excludePackages dockstore-ui2@2.3.0 --csv > THIRD-PARTY-LICENSES.csv",
+    "license": "license-checker --excludePackages dockstore-ui2@2.3.0 --csv > THIRD-PARTY-LICENSES.csv",
```
denis-yuen, line 16: 2.3.0 is in the file three times now. 
Maybe experiment with variables? https://frontend.irish/npm-config-variables

scripts/detect-package-json-changes.sh
```diff
@@ -1,14 +1,9 @@
 #!/bin/bash
 
-packageJsonStatus="$(git diff --name-status develop -- package.json)"
-packageLockJsonStatus="$(git diff --name-status develop -- package-lock.json)"
+licenseFileStatus="$(diff CIRCLE-THIRD-PARTY-LICENSES.csv THIRD-PARTY-LICENSES.csv)"
```
denis-yuen, line 3: thought: might be able to do a git diff to avoid the extra file too, but might not be worth the effort to change it

package.json
NatalieEO: Hm the version doesn't seem to like the variable
```
npm WARN prepare removing existing node_modules/ before installation
npm ERR! code EINVALIDTAGNAME
npm ERR! Invalid tag name "$npm_package_config_ui_version": Tags may not have any characters that encodeURIComponent encodes.
```
but works okay in the scripts section.

.circleci/config.yml
NatalieEO: Seems like you have to do the checkout, but got rid of the npm ci

denis-yuen (approved):

coverbeck (approved):

garyluu (approved):

PR #1145, opened 2021-02-03
Feature/3968/hosted workflow file naming

Richard-Hansen:
issue: dockstore/dockstore#3968

This PR fixes two minor bugs:
1. Can now multiple files to a hosted tool or workflow, they just need a unique name
2. When hitting the 'download' button for an individual file in a tool/workflow, the file will correctly download based on directory structure. (i.e. /A/B/C.cwl will download as C.cwl and not _A_B_C.cwl, when downloading as a zip file, C.cwl will be in the expected directory structure specified by /A/B/C.cwl)


coverbeck (approved):
In your commit message, when you merge, can you add a comment explaining why you are switching from `path` to `absolutePath`? I know it was discussed in Slack, but it would be good to capture here.

cypress/integration/group1/hostedWorkflows.ts
```diff
@@ -146,5 +148,55 @@ describe('Dockstore hosted workflows', () => {
...
+      // Add a new version with 3 descriptors
+      goToTab('Files');
+      cy.get('#editFilesButton').click();
+
+      // there should be no descriptors files
+      cy.get('app-code-editor').should('have.length', 0);
+
+      // add first file. This will be the primary descriptor, so we wont need to give it a custom name
+      cy.contains('Add File').click();
+      cy.wait(100);
```
garyluu, line 178: Quitea few waits added. We generally add them when the test randomly fails and there's no better solution.

Instead of waiting after a file is added, just check that there's an additional editor or similar.

Richard-Hansen: Removed a few excess `cy.wait()`s, there are a few that prevent the tests from failing:
1. After each `cy.contains('Add File').click();`, if the test types too quickly, it may type into the content box of the file, rather than the name input field. This is due to loading a default descriptor from a file into the content box and then switching focus to it, which pulls focus from the name input field. Looked into different approaches to get around this, but haven't had any luck in waiting until the editor is populated with content, thoughts?
2. After saving as a new version, if the test checks the length of the list of files before the webservice has responded, it'll count all the local file changes in the edit window, rather than what is saved in the database.

cypress/integration/group1/hostedWorkflows.ts
```diff
@@ -36,6 +36,8 @@ describe('Dockstore hosted workflows', () => {
     goToUnexpandedSidebarEntry('dockstore.org/A', /hosted/);
   }
 
+  const NEW_WORKFLOW_NAME = 'newHostedWorkflow';
```
denis-yuen, line 39: A very developer friendly name. Not that related to the bug, but while adding this could boundary push a bit and kill two birds with one stone. No spaces allowed, but could do "new_Hosted-Workflow8" for example

cypress/integration/group1/hostedWorkflows.ts
```diff
@@ -146,5 +148,53 @@ describe('Dockstore hosted workflows', () => {
...
+      // there should be no descriptors files
+      cy.get('app-code-editor').should('have.length', 0);
+
+      // add first file. This will be the primary descriptor, so we wont need to give it a custom name
+      cy.contains('Add File').click();
+
+      // add second file
+      cy.contains('Add File').click();
+      cy.wait(100);
+      cy.get('.editor-file-name').last().type('{backspace}{backspace}{backspace}{backspace}{backspace}/AAA.cwl{enter}');
```
denis-yuen, line 180: Assuming these backspaces are due to the default names like `/.cwl` this becomes `/AAA.cwl`
Is it possible to make this more readable by looping through the length of a string or even better the actual constant in that we use to populate new files?

cypress/integration/group1/hostedWorkflows.ts
garyluu: I would have to play around with it to see what works and what doesn't. If you think you've removed as many waits as possible, that's fine.

cypress/integration/group1/hostedWorkflows.ts
Richard-Hansen: Changed to a very not developer friendly name.

cypress/integration/group1/hostedWorkflows.ts
Richard-Hansen: Cheated a bit, Cypress provides `{selectall}` to select all content in the current input bar and paired that with a single `{backspace}`. Also moved the file creations into a loop, no reason to have those lines duplicated.

denis-yuen (approved):

NatalieEO (approved):

NatalieEO:
you'll have to get the latest from develop so your merge isn't blocked

garyluu (approved):

PR #1149, opened 2021-02-08
Footer Reimplementation

wendidy:
[SEAB-1951](https://ucsc-cgl.atlassian.net/browse/SEAB-1951)
Refer to mock up: https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60074bdcbb1d4b5c92a996d3

Screenshots:
image
image
image
image
(When the width is less than 425px, the page has a white border on the right, I noticed this is also the case before this footer reimplementation, have talked with Denis, and decided to leave it for now. Example below:)
image

- Deleted the "In Affiliation" section on pages' main bodies, since we moved it to the footer.
- "The Team", "About", and "Contact Us" do not have `href` right now, since the pages are a part of the redesign that need to be implemented.
- The reason why this section in the mock up and implementation are different is because in this case, resizing the screen with Angular `fxFlex` will look nicer in any screen sizes, if it needs to be exactly the same as the mock-up, I could look more into it.
image

- All the logos are in `.svg` format except UCSC, which is in `.png`
- The content in folder `sponsors` will be modified further when we get to implement then "About" page. Right now the analysis is yelling at having a duplicate of `sponsor.model.ts`, I was thinking about leaving the duplicate there for now until the new "About" page is created, but if there's a better way then that will be wonderful
- All tests are successful when ran locally, not sure if I should fix the 
  `Chrome Headless 88.0.4324.150 (Linux x86_64) ERROR
    An error was thrown in afterAll...` as you can see


src/app/footer/footer.component.css
```diff
@@ -17,18 +17,70 @@
...
+  line-height: 26px;
+}
+
+/* the title of each column */
+.title {
+  /* used for making the text semi-bold */
+  text-shadow: 0 0 0 #c5cae9;
+  color: #c5cae9;
+  text-transform: uppercase;
+  padding: 10px 0;
```
coverbeck, line 41: I will beat Gary to his comment and say we should avoid the use of pixels in css.

src/app/footer/footer.component.ts
```diff
@@ -49,10 +59,17 @@ export class FooterComponent extends Base implements OnInit {
    */
   private readonly WEBSERVICE_DOWN_STATUS_CODES = [0, 404, 502, 504];
 
-  constructor(private metadataService: MetadataService, private footerService: FooterService) {
+  constructor(private metadataService: MetadataService, private footerService: FooterService, private _snackBar: MatSnackBar) {
```
coverbeck, line 62: It should just be `snackbar` -- we don't use the convention of fields starting with underscores.

src/app/footer/sponsor.model.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+export class Sponsor {
+  private static colouredPath = '../assets/images/sponsors/non-coloured/';
```
coverbeck, line 18: This is confusing :) The name of the variable is 'colouredPath', but the path is `non-coloured`.

src/styles.scss
```diff
@@ -1287,3 +1279,8 @@ mat-tab-group.no-pagination {
     flex-direction: column;
   }
 }
+
+.custom_copy_snack_bar {
+  min-width: 0 !important;
```
coverbeck, line 1284: Old-school CSS rule-of-thumb is to avoid `!important`, if possible. May not apply as much any more, but do you need it?

wendidy: Unfortunately I believe so, since snack bar already has their own styling for `min-width`, in order to override it, I had to apply `!important`

src/app/footer/footer.component.css
wendidy: @coverbeck I see, as I noticed many other stylings are in `px` as well. Are you suggesting we covert them to `rem`?

coverbeck: > @coverbeck I see, as I noticed many other stylings are in `px` as well. Are you suggesting we covert them to `rem`?

Yes :) .  Rems work better with different screen sizes/resolutions.

See https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide , about the 10th item.

And yes, we have a ton of styles using px, but we're trying to avoid adding additional ones going forward.

denis-yuen (commented):
Thanks for the screenshots!
will defer to others on the PR

src/ds-style-fix.scss
```diff
@@ -686,8 +686,8 @@ div.modal form {
   text-decoration: none;
 }
 
-.footer-list > li {
-  margin: 0px !important;
+.footer-list li {
+  margin: 0 0 10px 0 !important;
```
coverbeck, line 690: Can you use rem?

coverbeck (approved):
I think there are px usages you can still tackle, and as discussed in Slack, a more faded border for Copy Build Info button would be nice.

NatalieEO (approved):
Once you get the tests passing.

denis-yuen (approved):
Ditto with tests

src/app/footer/footer.component.css
garyluu: Ah, I should've probably mentioned this months ago. I think Emily had a PR that fixes px for different screen sizes. though it wouldn't really hurt to just keep using rem.

src/app/footer/footer.component.ts
```diff
@@ -15,19 +15,22 @@
...
 import { Metadata } from './../shared/swagger/model/metadata';
 import { FooterService } from './footer.service';
+import { Sponsor } from './sponsor.model';
 import { versions } from './versions';
 
 @Component({
   selector: 'app-footer',
   templateUrl: './footer.component.html',
   styleUrls: ['./footer.component.css'],
+  providers: [FooterService],
```
garyluu, line 33: What happens without this addtion?

src/app/footer/sponsor.model.ts
```diff
@@ -0,0 +1,28 @@
+/*
+ *    Copyright 2017 OICR
```
garyluu, line 2: This file looks duplicated, simplify somehow.

src/ds-style-fix.scss
```diff
@@ -686,8 +686,8 @@ div.modal form {
   text-decoration: none;
 }
 
-.footer-list > li {
-  margin: 0px !important;
+.footer-list li {
```
garyluu, line 689: Might as well move this to the footer scss

src/app/footer/footer.component.css
```diff
@@ -17,18 +17,69 @@
...
+.title {
+  /* used for making the text semi-bold */
+  text-shadow: 0 0 0 #c5cae9;
+  color: #c5cae9;
+  text-transform: uppercase;
+  padding: 1rem 0;
+}
+
+.sponsor-img {
+  height: 60px;
```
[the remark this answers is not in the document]
wendidy: @coverbeck We don't want the image to resize according to the screen size so that's why I left it as `60px`, then the reason why I left a bunch of `1px`s is because we don't need to worry about resizability at all, but I guess I could change them to rem it shouldn't matter that much because it's so small

src/app/footer/footer.component.ts
wendidy: @garyluu nothing I think, will get rid of this

src/app/home-page/home-logged-out/home.component.html
```diff
@@ -211,11 +211,6 @@ <h3 class="text-center">Join the discussion</h3>
         Discourse <mat-icon>open_in_new</mat-icon>
       </a>
     </div>
-
-    <p class="text-center">And</p>
```
denis-yuen, line 215: Is this deletion intentional?

src/app/footer/footer.component.css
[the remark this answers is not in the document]
coverbeck: Fair... I just picked this one at random because I just saw the letters "px"... If it makes sense in this case, that's fine, and as Gary pointed, maybe I was taking the guidance too literally.

src/app/app.module.ts
```diff
@@ -209,6 +210,7 @@ export function configurationServiceFactory(configurationService: ConfigurationS
     ViewService,
     TosBannerService,
     ConfigurationService,
+    FooterService,
```
coverbeck, line 213: Is this needed (fine if it is)?

coverbeck (approved):
I'm wondering if the button is a faded as the mocks and as we had discussed in Slack, but looks good.

src/ds-style-fix.scss
wendidy: @garyluu Tried to do that but unfortunately the check yells at me for using `!important `in component style sheet

PR #1150, opened 2021-02-08
Seab 1632/tool tester logs smoke test

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-1632

### DIFF
--- a/cypress/integration/manualTests/basic-enduser-prod-only.ts
+++ b/cypress/integration/manualTests/basic-enduser-prod-only.ts
@@ -1,0 +1,23 @@
+describe('Test existence of Logs', () => {
+  it('Find Logs in Workflows', () => {
+    cy.visit('/workflows/github.com/DataBiosphere/topmed-workflows/UM_variant_caller_wdl:1.32.0?tab=info');
+    cy.get('[data-cy=verificationLogsDialog]').click();
+
+    cy.contains('.mat-card-title', 'Verification Information');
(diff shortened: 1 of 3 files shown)


coverbeck (commented):
Does this only work on prod? I'm pretty sure it doesn't work on dev, and may not have been configured for staging....

If so, it needs to come out of basic-enduser.ts, and the smoke tests in package.json for prod (and staging?) should include an additional file.

NatalieEO:
> Does this only work on prod? I'm pretty sure it doesn't work on dev, and may not have been configured for staging....
> 
> If so, it needs to come out of basic-enduser.ts, and the smoke tests in package.json for prod (and staging?) should include an additional file.

checked and the logs table only exists in production right now

denis-yuen:
> checked and the logs table only exists in production right now 

The webservice reads content from a S3 bucket for that I believe. 
Maybe we should have another S3 bucket for staging or point staging at the prod bucket (there are no secrets there)

denis-yuen:
> Does this only work on prod? I'm pretty sure it doesn't work on dev, and may not have been configured for staging....
> 
> If so, it needs to come out of basic-enduser.ts, and the smoke tests in package.json for prod (and staging?) should include an additional file.

Didn't see this, but this probably covers it

coverbeck (approved):

garyluu (approved):

PR #1151, opened 2021-02-08
Fix for text search

garyluu:
Part 2 of 2 for dockstore/dockstore#3985

The code comment explains why this fix is needed and why it works. Seems the current implementation gave the illusion that search works because it's searching through a lot of other fields.  

New behaviour:
Matches when tool_path or full_workflow_path contains the search text even if the path has a prefix or suffix

Note: the results are still unordered. Sourcefile content matches have the same priority as full_workflow_path matches.

Reviewers should test it locally to be thorough.

### DIFF
--- a/src/app/search/query-builder.service.ts
+++ b/src/app/search/query-builder.service.ts
@@ -179,7 +179,7 @@ export class QueryBuilderService {
   appendQuery(body: any, values: string, oldAdvancedSearchObject: AdvancedSearchObject, searchTerm: boolean): any {
     const advancedSearchObject = { ...oldAdvancedSearchObject };
     if (values.toString().length > 0) {
-      this.searchEverything(body, values);
+      body = this.searchEverything(body, values);
     } else {
       body = body.query('match_all', {});
     }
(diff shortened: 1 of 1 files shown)


coverbeck:
Running web service locally with feature/3985/searchtext web service branch and this UI branch, the numbers next to the facets does not correspond to the number of entries displayed.

If you just type "aligner", the facet numbers and the number of results match.

If you then click "WDL", with "aligner" still ther, the number of display results shoots up

Screenshot from 2021-02-10 10-20-42

garyluu:
> Running web service locally with feature/3985/searchtext web service branch and this UI branch, the numbers next to the facets does not correspond to the number of entries displayed.
> 
> If you just type "aligner", the facet numbers and the number of results match.
> 
> If you then click "WDL", with "aligner" still ther, the number of display results shoots up
> 
> Screenshot from 2021-02-10 10-20-42

Strange, i see:
image

garyluu:
Ok, definitely not working without multiple facets. back to the drawing board

garyluu:
Re-opening, seems better now?

garyluu:
image
image
image
image

NatalieEO (approved):

denis-yuen (approved):

src/app/search/query-builder.service.ts
```diff
@@ -198,22 +198,25 @@ export class QueryBuilderService {
   /**
    * Appends search-everything filter to the query
    * Currently searches sourcefiles, description, labels, author, and path
-   * @param {*} body The original body
-   * @param {string} searchString the search string
-   * @memberof QueryBuilderService
+   * Some requiremnts:
```
denis-yuen, line 201: Will probably end up trying this on dev.
Also typo here

denis-yuen (approved):

coverbeck (approved):
I am seeing slightly different results, but as discussed, I am using a different snapshot of data, so I think it's good.

PR #1154, opened 2021-02-10
Feature/broken styles

wendidy:
[SEAB-2456](https://ucsc-cgl.atlassian.net/browse/SEAB-2456?filter=-1)
Each of the numbers below corresponds to the numbers in the ticket. 5 in total

1. Default label
image

2. Size of the icon
image

3. I removed the "Verified" and "Stars" columns when the screen size is small. I imagined we don't really care about small screens so I just temporarily decided to do this so the styling is not broken, if in the future, we can think of better alternatives, it's really easy to change it back.
image

4. Now they are always displayed in rows
image
image

5.
image

### DIFF
--- a/src/app/container/version-modal/version-modal.component.css
+++ b/src/app/container/version-modal/version-modal.component.css
@@ -13,3 +13,9 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
+
+#file_copy {
+  font-size: initial;
+  transform: scale(1.3);
+  height: fit-content;
+}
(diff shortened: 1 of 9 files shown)


denis-yuen:
> I removed the "Verified" and "Stars" columns when the screen size is small. I imagined we don't really care about small screens so I just temporarily decided to do this so the styling is not broken, if in the future, we can think of better alternatives, it's really easy to change it back.

Sounds good

coverbeck (approved):

Richard-Hansen (approved):

NatalieEO (approved):

PR #1158, opened 2021-02-18
Feature/4009/copy snackbar

wendidy:
[DOCK-1692 dockstore#4009](https://github.com/dockstore/dockstore/issues/4009)
image
Created a directive and a module for this custom snackbar.
The reason for the presence of the new module is due to the fact that snackbar directive is needed to be included in the declarations of multiple modules, but we are unable to directly do so with the directive. The way to go about it is to create a module dedicated for snackbar. (Although it seems redundant)
Both of them are in the shared folder.

I believe the duplicate lines were there before, not sure if I could do anything about it?

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -81,6 +81,7 @@ import { ListContainersModule } from './shared/modules/list-containers.module';
 import { ListWorkflowsModule } from './shared/modules/list-workflows.module';
 import { CustomMaterialModule } from './shared/modules/material.module';
 import { OrderByModule } from './shared/modules/orderby.module';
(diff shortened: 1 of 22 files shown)


garyluu (approved):

coverbeck (approved):
Instead of `appSnackbar`/`SnackbarDirective`, I would have called it something like `appCopiedMessage`/`CopiedMessageDirective`.

1. Maybe someday the implementation is not a snackbar
2. It's not a generic Snackbar directive, it specifically displays a Copied message.

But don't change it! It can evolve to take a message parameter for step 2 (if parameters can be passed to directives), in which case the name would be perfect.

PR #1161, opened 2021-02-23
Feature/collection reimplementation

wendidy:
[SEAB-2285](https://ucsc-cgl.atlassian.net/browse/SEAB-2285)
Refer to mock up: [page 1](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60073f22c824b7b8dfde055c), and [page 2](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60073cb04ab9d01346312637)

Screenshots:
image
image

Screenshots maintainers/admins see:
image

Some things to be noted:
1. The colored chips beside the updated time and the star on the right are going to be implemented in another ticket I have created.
2. The 2 pages of mock up have an inconsistency on whether or not the organization name is displayed or a placeholder "Collections" is displayed. I went with organization name but we can change that if needed.
3. The reason why I merged `develop` into this branch is because I want to see what the styling looks here before I merge this branch in. And also to make sure the there will be no conflicts by that time.
4. *A question: the description of collections is not from the UI I assume? There's a lot of inconsistency in the title of sections, for example: 
 and 
I'm just wondering if I should look into this with an additional ticket?


src/app/organizations/collection/collection.component.html
```diff
@@ -37,25 +37,19 @@
...
-              </mat-card-title>
-              <mat-card-subtitle>
-                {{ collection.topic }}
-              </mat-card-subtitle>
-            </mat-card-header>
+            <h2>
+              <a [routerLink]="['/organizations/', organization.name]">
+                <span [matTooltip]="organization?.name" class="org-name">{{ organization.displayName }}</span></a
+              >
+              : <span [matTooltip]="collection?.name" class="colleciton-name">{{ collection.displayName }}</span>
```
coverbeck, line 50: typo in `colleciton`

src/app/organizations/collection/collection.component.html
```diff
@@ -37,25 +37,19 @@
...
-                  / <span [matTooltip]="collection?.name">{{ collection.displayName }}</span>
-                </h1>
-              </mat-card-title>
-              <mat-card-subtitle>
-                {{ collection.topic }}
-              </mat-card-subtitle>
-            </mat-card-header>
+            <h2>
+              <a [routerLink]="['/organizations/', organization.name]">
+                <span [matTooltip]="organization?.name" class="org-name">{{ organization.displayName }}</span></a
```
coverbeck, line 48: Was like this and I'm not sure what the alternative is, but it's weird to me to just have the name as the tooltip for a display name. I guess just leave it. :)

src/app/organizations/collection/collection.component.html
```diff
@@ -72,68 +66,93 @@ <h1>
...
+                <a
+                  [routerLink]="[
+                    '/' + (entry.entryType === 'tool' ? 'container' : 'workflow') + 's/',
+                    entry.entryPath + (entry.versionName ? ':' + entry.versionName : '')
+                  ]"
+                >
+                  <div fxLayout="row" fxLayoutAlign="space-between">
+                    <div fxLayout="row" fxLayoutAlign="flex-start" fxLayoutGap="1rem">
+                      <img *ngIf="entry.entryType === 'workflow'" src="../../../assets/svg/workflow.svg" alt="workflow icon" />
+                      <img *ngIf="entry.entryType === 'tool'" src="../../../assets/svg/tool.svg" alt="tools icon" />
```
coverbeck, line 89: Real nitpicky, but you have "workflow icon", singular, and "tools icon", plural.

src/app/organizations/collection/collection.component.html
```diff
@@ -72,68 +66,93 @@ <h1>
...
+                        >
+                          <mat-icon>clear</mat-icon> Remove
+                        </button>
+                      </span>
+                    </div>
+                  </div>
+                </a>
+                <hr />
+                <div fxLayout="row" fxLayoutAlign="space-between">
+                  <span class="update-time"> Last updated {{ entry.dbUpdateDate | date }} </span>
```
coverbeck, line 119: The code on line 95 is checking if entry exists, but this code assumes it always does.

wendidy: @coverbeck I think I will just remove line 95, since no where else inside this file checks whether entry exists.  In addition to this, entry is an element inside the array, we could likely assume entry is not empty. 
Then take the anchor link somewhere else in this project for example, if entry does not exist then the redirection will fail. We could very well check if it exists and then determine whether anchor will be disabled accordingly, but like mentioned above, I don't think entry should be empty...

denis-yuen:
> *A question: the description of collections is not from the UI I assume? There's a lot of inconsistency in the title of sections, for example. I'm just wondering if I should look into this with an additional ticket?

This might be out of scope and would require a curator I'd suspect since this is user generated

coverbeck (approved):

src/app/organizations/collection/collection.component.scss
```diff
@@ -1,4 +1,40 @@
...
+  padding: 0.5rem 1rem;
+}
+
+hr {
+  margin: 0.5rem 0;
+  border: 1px solid #dfe0e2;
+}
+
+h5 {
+  color: #151c3d;
```
garyluu, line 19: should make a style in the main scss. i feel like this color is going to be repeated a lot

garyluu (approved):

PR #1162, opened 2021-02-23
Feature/4051/UI bugs

wendidy:
[dockstore/dockstore/issues/4051](https://github.com/dockstore/dockstore/issues/4051)
Screenshots:
1.
image
2.
image
3. 
image
4. When screen is medium sized
image

### DIFF
--- a/src/app/search/search-workflow-table/search-workflow-table.component.html
+++ b/src/app/search/search-workflow-table/search-workflow-table.component.html
@@ -33,8 +33,10 @@
       <mat-cell fxShow fxHide.lt-sm class="duration-cell" *matCellDef="let workflow">{{ workflow?.author || 'n/a' }}</mat-cell>
     </ng-container>
     <ng-container matColumnDef="descriptorType">
-      <mat-header-cell fxShow fxHide.lt-sm data-cy="descriptorTypeHeader" *matHeaderCellDef mat-sort-header>Format</mat-header-cell>
-      <mat-cell fxShow fxHide.lt-sm data-cy="descriptorType" *matCellDef="let workflow">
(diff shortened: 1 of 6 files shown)


garyluu (changes requested):
1. In my opinion, the version should:
  - avoid breaking a word that can fit on its own line
  - if the word is too long, only then should it break
  
Looks like on prod, it does the first point but not the 2nd. The versions table I believe does both. Your PR looks like it does only the 2nd point but not the first.

2. This is fine i think

3. Seems worse:

image

4. No comment

src/app/search/search-workflow-table/search-workflow-table.component.html
```diff
@@ -33,16 +33,18 @@
...
+      >
+      <mat-cell fxShow fxHide.lt-sm data-cy="descriptorType" *matCellDef="let workflow" fxFlex="20%">
         <div>
           <span class="nonclick-badge">{{ workflow?.descriptorType | uppercase }}</span>
         </div>
       </mat-cell>
     </ng-container>
     <ng-container matColumnDef="projectLinks">
       <mat-header-cell fxShow fxHide.lt-md *matHeaderCellDef>Project Links</mat-header-cell>
-      <mat-cell fxShow fxHide.lt-md *matCellDef="let entry" class="align-languages">
```
garyluu, line 45: Is this the only use of the class? If so, remove the class.

garyluu (approved):
Adding Charles because I have no opinion on number 4

NatalieEO (approved):

coverbeck (approved):

src/app/search/search-workflow-table/search-workflow-table.component.html
wendidy: @garyluu I need to use media query on this, fxLayout.lt doesn't work in this case for some reason...

PR #1163, opened 2021-02-25
Feature/3729/multiple account error messages

Richard-Hansen:
When a 409 is received while attempting to link a new account, the snackbar message that pops up will have a link to the docs. Currently, the if there is an error linking your Google account, that uses a banner error instead of the snackbar component (apparently there is some special setup that needs to be done? See [here](https://github.com/dockstore/dockstore-ui2/blob/11ca91a3d79a0ff5395ed744f2db7bc60b335023/src/app/loginComponents/accounts/external/accounts.service.ts#L54))

GH issue: dockstore/dockstore#3729
Pairs with: https://github.com/dockstore/dockstore/pull/4063


NatalieEO (approved):

denis-yuen (approved):

denis-yuen:
(TIL, note readthedocs latest is not like Docker Hub latest. It is just the default branch for the repo)

src/app/shared/alert/state/alert.service.ts
```diff
@@ -111,6 +111,24 @@ export class AlertService {
...
+   *
+   * @param error {HttpErrorResponse} error  The HttpErrorResponse return by the failed Http call
+   * @param actionButtonText {string} The text of the action button
+   * @param link {link} The link that will open in a new tab
+   */
+  public detailedSnackBarErrorWithLink(error: HttpErrorResponse, actionButtonText: string, link: string) {
+    this.clearEverything();
+    const detailedError: string = AlertService.getDetailedErrorMessage(error);
+    this.matSnackBar
+      .open(detailedError, actionButtonText)
```
coverbeck, line 127: You could increase the duration of the snackbar for this case, I think the default is set to 5 seconds; it might be hard to locate and click on the link before it disappears. But I don't feel that strongly about it.

coverbeck (approved):
Doesn't the snackbar di

garyluu (approved):

PR #1165, opened 2021-03-01
Feature/footer modifications

wendidy:
very small detail modifications: 
image

Side note: right now I'm trying to use svg format for all images, there's a svg folder and an image folder, which one should I use?


coverbeck:
> very small detail modifications:

Can you explain what you changed? Was this the (lack of a )gap between the icon and the text?

denis-yuen:
Think the underlining included the leading space before
> Side note: right now I'm trying to use svg format for all images, there's a svg folder and an image folder, which one should I use?
SVG sounds good (assuming folder is full of PNG and the like)

coverbeck (approved):

wendidy:
@coverbeck @denis-yuen Yeah for sure, so the spacing is enlarged, underline in the gap is removed, and icons are smaller now

src/app/footer/footer.component.html
```diff
@@ -149,28 +149,29 @@
           <ul class="footer-list m-0 affiliation-list">
             <li class="m-0 title">Find us on</li>
             <li class="m-0">
-              <a [href]="Dockstore.DOCKSTORE_REPO" target="_blank" rel="noopener noreferrer" class="footer-link">
-                <img
-                  class="site-icons-tab hidden-sm"
-                  src="../assets/images/dockstore/github-mark-light.png"
```
garyluu, line 155: This appears to be the only place that uses the png and now it's removed

wendidy: @garyluu Yeah so I deleted the png

garyluu (commented):
Can you explain the pros and cons of switching from the material icon to svg for RSS?

wendidy:
> Can you explain the pros and cons of switching from the material icon to svg for RSS?

@garyluu Simply because styling is easier across all icons..Plus the quality of svg is guaranteed so I thought there shouldn't be a problem, correct me if I'm wrong

garyluu:
> > Can you explain the pros and cons of switching from the material icon to svg for RSS?
> 
> @garyluu Simply because styling is easier across all icons..Plus the quality of svg is guaranteed so I thought there shouldn't be a problem, correct me if I'm wrong

mat-icons are SVG icons under the hood so image quality so the mat-icons shouldn't be any worse. it's clear where the SVG of the mat-icon came from but not from the changes here. I'd like to retain the material icons unless there's a clear advantage. You can read more about mat-icons here: https://material.angular.io/components/icon/overview

garyluu (approved):

PR #1166, opened 2021-03-02
Feature/update schema and icons

wendidy:
[SEAB-2481](https://ucsc-cgl.atlassian.net/browse/SEAB-2481) and [SEAB-2480](https://ucsc-cgl.atlassian.net/browse/SEAB-2480)
[Search page mock up](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/5fc690e874e0c807afeb299b)(can also navigate to other pages)

Screenshots:
Search page tab bar:

Tool:
image

Workflow:
image

Nav bar:
image

Updates:
image
image
image
image


src/app/workflow/workflow.component.html
```diff
@@ -42,21 +42,27 @@ <h2 class="text-center">The workflow linked to could not be found!</h2>
...
-            <mat-icon matTooltip="Verified">done</mat-icon>
-          </a>
+        <span>
+          <img src="../../assets/svg/workflow-circle.svg" alt="workflow icon" />
         </span>
         <span>
           {{ title
           }}<span *ngIf="workflow?.workflowVersions.length > 0"
-            >:<span class="ds-green">{{ selectedVersion?.name }}</span></span
+            >:<span class="ds-steelblue">{{ selectedVersion?.name }}</span></span
```
wendidy, line 55: if I make each span on their own lines it seems like the test doesn't detect the `title+":"+selectedVersion.name `correctly?

src/app/container/container.component.html
```diff
@@ -42,20 +42,26 @@
...
+      <h3 id="tool-path" fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="1rem">
+        <span>
+          <img src="../../assets/svg/tool-circle.svg" alt="tool icon" />
         </span>
         <span *ngIf="tool?.private_access" class="private-lock">
           <app-private-icon></app-private-icon>
         </span>
         {{ tool?.tool_path }}<span *ngIf="tool?.workflowVersions.length > 0">:</span
-        ><span class="ds-green">{{ selectedVersion?.name }}</span>
+        ><span class="ds-steelblue">{{ selectedVersion?.name }}</span>
```
wendidy, line 56: same problem as below

denis-yuen:
> I think we don't have the service icon for headers, correct me if I'm wrong, so I substituted the workflows icon instead:

this is probably an oversight, there should be an icon for example on https://projects.invisionapp.com/share/VNZ92Z0T2B3#/screens/434537756 the left, albeit not coloured

denis-yuen:
>  If we care about the mobile version then I will also fix the word break of the other links on that page 

We care insofar as it is easy (i.e. it is not a focus and should not take up much time, but it is a nice to have)

coverbeck:
Default badge formatting is wrong for me on versions table when there is a DOI, at least I think that's the cause. github.com/gatk-workflows/gatk4-data-processing/processing-for-variant-discovery-gatk4:2.1.0?tab=versions

Screen Shot 2021-03-03 at 9 51 21 AM

garyluu:
> Default badge formatting is wrong for me on versions table when there is a DOI, at least I think that's the cause. github.com/gatk-workflows/gatk4-data-processing/processing-for-variant-discovery-gatk4:2.1.0?tab=versions
> 
> Screen Shot 2021-03-03 at 9 51 21 AM

We've always had issues with not having enough space to work with

coverbeck:
In the My account dropdown, the new icons look disabled/gray -- is there any reason they're not in color?

And I think you have that same underline problem that was on the footer.


Screen Shot 2021-03-03 at 3 09 50 PM

src/app/container/container.component.html
```diff
@@ -42,20 +42,26 @@
...
+        <span id="verifiedIcon" *ngIf="(extendedTool$ | async)?.versionVerified" fxHide.lt-sm>
+          <mat-chip matTooltip="Verified" class="container-label-chip">
+            <a [href]="getVerifiedLink()" class="verified-check" fxLayoutGap="0.5rem" fxLayoutAlign=" center">
```
coverbeck, line 59: How often is this invoked? It's generally not good to invoke functions like this from HTML to set an attribute value, as they get invoked a lot.

coverbeck (commented):
I like the green underline for selected tabs. It's a small thing, but it really is more obvious what's selected.

src/app/container/container.component.html
garyluu: was like that before. the function does get called a lot and should definitely be changed

wendidy:
> In the My account dropdown, the new icons look disabled/gray -- is there any reason they're not in color?
> 
> And I think you have that same underline problem that was on the footer.
> 
> Screen Shot 2021-03-03 at 3 09 50 PM

Yeah before the icons were a blue-ish gray color. So I went along with the gray. There are colored versions though, we can for sure go with that

wendidy:
> I like the green underline for selected tabs. It's a small thing, but it really is more obvious what's selected.

What do you say about this:
When hovered: 
image

After clicked:
image

coverbeck:
> What do you say about this:

It works for me. Does it match Kim's mocks?

wendidy:
> > What do you say about this:
> 
> It works for me. Does it match Kim's mocks?

Kim's mock-up is completely different 
image
We can only do this when the My Account pages are implemented
So I guess we will go with what we have right now?

coverbeck:
> We can only do this when the My Account pages are implemented
> So I guess we will go with what we have right now?

Ah, right. If by right now, you mean color, yes I vote for that. Or at least a less grayish color that doesn't look disabled.

I may be pessimistic, but I don't know if we'll be able to get whole rework into 1.11, so this might be there for a while.

garyluu:
ugh, the files changed shows my other PR.  Anyways...there's a material.scss if you want to blanket change colors

wendidy:
> ugh, the files changed shows my other PR. Anyways...there's a material.scss if you want to blanket change colors

Since we are talking about it, I don't know why I decided to rebase and not merge :) And what do you mean? What's up with material.scss?

coverbeck (approved):

garyluu (approved):

PR #1167, opened 2021-03-02
add nextflow install instructions

NatalieEO:
dockstore/dockstore#3679

### DIFF
--- a/src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
+++ b/src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
@@ -168,7 +168,7 @@ $ cwltool --version
 \`\`\`
 
 #### Part 5 - Install Nextflow (Optional)
-Nextflow workflows cannot run on the Dockstore CLI. Users must run them directly by using the Nextflow command line tool.
+The Dockstore CLI does not run Nextflow workflows. Users must run them directly by using the Nextflow command line tool.
 1. You can install the Nextflow CLI with the following command:
 \`\`\`
 // Creates a file nextflow in the current directory
--- a/src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
+++ b/src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
@@ -166,6 +166,15 @@ pip3 install -r requirements.txt
 $ cwltool --version
 /usr/local/bin/cwltool ${this.cwltoolVersion}
 \`\`\`
+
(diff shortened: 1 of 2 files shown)


src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -166,6 +166,15 @@ pip3 install -r requirements.txt
 $ cwltool --version
 /usr/local/bin/cwltool ${this.cwltoolVersion}
 \`\`\`
+
+#### Part 5 - Install Nextflow (Optional)
+Nextflow workflows cannot run on the Dockstore CLI. Users must run them directly by using the Nextflow command line tool.
```
coverbeck, line 171: I would say "The Dockstore CLI does not run Nextflow workflows", but I don't feel super strongly about it. I would argue that the CLI is not a platform that you run things on.

coverbeck (approved):
Can you also please update [this](https://github.com/dockstore/dockstore-ui2/blob/920a8164d41313379c3ae47aa19f06eabb9f7b17/src/app/shared/launch.service.ts#L76) to be https? I was actually expecting your changes to be on that panel (now I realized they shouldn't be), so I was looking there and happened to notice.

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -166,6 +166,15 @@ pip3 install -r requirements.txt
 $ cwltool --version
 /usr/local/bin/cwltool ${this.cwltoolVersion}
 \`\`\`
+
+#### Part 5 - Install Nextflow (Optional)
+The Dockstore CLI does not run Nextflow workflows. Users must run them directly by using the Nextflow command line tool.
```
denis-yuen, line 171: must -> can 
(minor, just opening the door that there are other options like Nextflow Tower as well)

denis-yuen (approved):

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -166,6 +166,15 @@ pip3 install -r requirements.txt
...
+1. You can install the Nextflow CLI with the following command:
```
garyluu, line 172: Just my opinion of course, but I rather link out to the actual nextflow documentation because I wouldn't want to maintain this (in case they somehow change the installation instructions)

https://github.com/nextflow-io/nextflow#download-the-package

Approved assuming I don't need to maintain this.

garyluu (approved):

PR #1168, opened 2021-03-04
added sort by feature in organizations

wendidy:
[SEAB-2494](https://ucsc-cgl.atlassian.net/browse/SEAB-2494)
[mock](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60072419e08ae47561013a80)

Screenshots:
image
image
image
image

Changing sort by:
image

Sort by name:
image

Side note:
1. Ran into some issues with pagination, it was only half working. Since right now we only have 20+ organizations, so I'm thinking of leaving this for now. Will leave a note on Jira to explain. Still have the code for it, if anyone has any idea I would love to discuss about it!


src/app/organizations/organizations/organizations-paginator.directive.ts
```diff
@@ -0,0 +1,38 @@
+import { Directive, OnInit, ViewChild } from '@angular/core';
```
wendidy, line 1: This file is used to create something similar to the functionalities in `search-entry-table.ts` I tried to have `organizations.component.ts` to extend from this, but then there's this "No providers for OrganizationsPaginatorDirective
". 
**Right now this file is not being used**

src/app/organizations/organizations/organizations.component.html
```diff
@@ -20,59 +20,90 @@
...
-  >
-    Create Organization Request
-  </button>
-  <div *ngIf="filteredOrganizations$ | async as organizations; else noOrganizationsFound">
-    <form [formGroup]="organizationSearchForm">
-      <mat-form-field>
-        <input matInput placeholder="Search organizations" formControlName="name" />
-      </mat-form-field>
-    </form>
+  <div *ngIf="obs | async as organizations; else noOrganizationsFound">
```
wendidy, line 23: `obs` is the observable I messed around with paginator in the `.ts` file, right now it doesn't work. `obs` is dead

src/app/organizations/organizations/organizations.component.ts
```diff
@@ -13,9 +13,12 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-import { Component, OnInit } from '@angular/core';
+import { AfterViewInit, ChangeDetectorRef, Component, OnInit, ViewChild } from '@angular/core';
```
wendidy, line 16: So I'm referring to this link: [paginator but not with table](https://stackblitz.com/edit/mat-paginator-with-cards?file=app%2Fapp.component.ts) and also the `search-entry-table.ts`

src/app/organizations/organizations/organizations.component.ts
```diff
@@ -76,4 +99,8 @@ export class OrganizationsComponent extends Base implements OnInit {
   createOrganization(): void {
     this.matDialog.open(RegisterOrganizationComponent, { data: { organization: null, mode: TagEditorMode.Add }, width: '600px' });
   }
+
+  // privateNgOnInit(): Observable<Array<Organization>>{
```
wendidy, line 103: This is when this class extends `OrganizationsPaginatorDirective`

src/app/organizations/state/organizations.service.ts
```diff
@@ -52,4 +52,23 @@ export class OrganizationsStateService {
...
+  updateSort(sortBy: string): void {
+    this.organizationsStore.update((state: OrganizationsState) => {
+      return {
+        ...state,
+        sortBy: sortBy,
+      };
+    });
+  }
+
+  setPageSizeAndIndex(pageSize: number, pageIndex: number) {
```
wendidy, line 65: Yeah tried to copy `search.service.ts`

src/app/organizations/organizations/organizations.component.scss
```diff
@@ -4,14 +4,56 @@
...
+  padding: 0 !important;
+}
+
+.info {
+  font-size: 12px;
+  line-height: 2rem;
+}
+
+hr {
+  margin-bottom: 2rem;
```
garyluu, line 46: we already have classes for this. check the bootstrap4.scss

src/app/organizations/organizations/organizations.component.ts
```diff
@@ -53,20 +54,30 @@ export class OrganizationsComponent extends Base implements OnInit {
...
     this.isLoggedIn$ = this.trackLoginService.isLoggedIn$;
     this.organizationsStateService.updateSearchNameState('');
-    this.organizationSearchForm = this.formBuilder.group({ name: '' });
+    this.organizationSearchForm = this.formBuilder.group({ name: '', sort: '' });
     this.loading$ = this.alertQuery.showInfo$;
     // The real loading$ is currently not being used because the alertQuery global loading is used instead
     // this.loading$ = this.organizationsQuery.loading$;
     this.organizationsStateService.updateOrganizations();
     this.filteredOrganizations$ = this.organizationsQuery.filteredOrganizations$;
+    this.filteredOrganizations$.pipe(filter(orgs => orgs != null), take(1))
```
garyluu, line 63: don't subscribe, just map to another observable and move this to the query.ts

src/app/organizations/state/organizations.query.ts
```diff
@@ -22,13 +23,16 @@ export class OrganizationsQuery extends Query<OrganizationsState> {
...
+   * Filters the organizations based on a string
    * Case insensitive
    * Partial match
    * Searches the name, description, displayName, location, and topic of each organization (does not search its collections)
    *
+   * Also able to sort the organizations by name or number of stars
+   *
    * @param {Array<Organization>} organizations  List of all approved organizations
    * @param {string} searchName                  Search string
+   * @param {string} sortBy                      Search string
```
garyluu, line 35: should be of type `'starred' | 'name'`

src/app/organizations/organizations/organizations.component.scss
wendidy: Thanks I forgot about it, but now the funny thing is bootstrap4 doesn't include anything `2rem`

src/app/organizations/organizations/organizations.component.ts
wendidy: @garyluu This length can't be an observable, I only need this once at the beginning

src/app/organizations/state/organizations.query.ts
```diff
@@ -50,4 +54,20 @@ export class OrganizationsQuery extends Query<OrganizationsState> {
     }
     return null;
   }
+
+  filterAndSortOrganizations(organizations: Array<Organization>, searchName: string, sortBy: string): Array<Organization> | null {
+    if (organizations) {
+      organizations = this.filterOrganizations(organizations, searchName);
```
coverbeck, line 60: This is just me, minor and won't affect anything in this case should you change it, but my preference is not to reassign parameter values.

coverbeck (approved):

src/app/organizations/state/organizations.query.ts
```diff
@@ -50,4 +54,20 @@ export class OrganizationsQuery extends Query<OrganizationsState> {
     }
     return null;
   }
+
+  filterAndSortOrganizations(organizations: Array<Organization>, searchName: string, sortBy: string): Array<Organization> | null {
+    if (organizations) {
```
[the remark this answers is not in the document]
wendidy: @coverbeck  Yeah for sure, I definitely agree on changing the parameter type to `Array | something`, the thing is this function I wrote is very similar to one that's already there in this file, and the latter has `| null` instead of `[]`. I'm wondering if we should keep the `null`

src/app/organizations/organizations/organizations.component.scss
garyluu: find one that's similar

src/app/organizations/organizations/organizations.component.ts
garyluu: the view doesn't do a re-draw if the value is the same value.  if your code only works if on the first observable emit then something is very wrong

denis-yuen (approved):
Looking good in general

src/app/organizations/organizations/organizations.component.html
```diff
@@ -20,57 +20,93 @@
...
   <div *ngIf="filteredOrganizations$ | async as organizations; else noOrganizationsFound">
-    <form [formGroup]="organizationSearchForm">
-      <mat-form-field>
-        <input matInput placeholder="Search organizations" formControlName="name" />
-      </mat-form-field>
-    </form>
+    <div
+      fxLayout="row"
+      fxLayoutAlign="space-between center"
+      class="mt-3 mb-3"
```
garyluu, line 27: my-3

src/app/organizations/organizations/organizations.component.html
```diff
@@ -20,57 +20,93 @@
...
+      <div fxLayoutAlign="start stretch" fxLayout="row wrap" fxLayoutGap="10px">
+        <mat-card
+          style="height: 340px"
+          *ngFor="let org of organizations"
+          class="my-2 mat-elevation-z3"
+          fxFlex="32"
+          fxFlex.lt-lg="48"
+          fxFlex.lt-md="100"
+        >
+          <div style="height: 100%" fxLayout="column" fxLayoutAlign="space-between stretch">
```
garyluu, line 67: h-100

garyluu (approved):

PR #1127, opened 2020-12-23
Feature/2073/seperate tools and workflows tab

nrzhao4:
- for part of [SEAB-2073](https://ucsc-cgl.atlassian.net/browse/SEAB-2073)
- implemented some of the [new UI](https://projects.invisionapp.com/share/VNZ92Z0T2B3#/screens/436191418)
- sidebar facets are now independent for workflows and tabs, only applies to current tab
- searching the in the search bar returns results only from the current tab
- clicking a different tab resets all search terms and filters
- to make tabs above sidebar and results table, moved the mat-tab from `search-results.component` to `search.component`
- updated icons and text colour in tab labels
Screen Shot 2020-12-23 at 3 00 48 PM
Screen Shot 2020-12-23 at 3 00 35 PM

PR #1169, opened 2021-03-04
Fix tests + tab change filter reset

garyluu:
Fix for tests and also the tab change not triggering facet reset

Edit to add: follows #1127

### DIFF
--- a/cypress/integration/manualTests/basic-enduser.ts
+++ b/cypress/integration/manualTests/basic-enduser.ts
@@ -100,14 +100,14 @@ describe('Test logged out home page', () => {
   it('home page search bar works', () => {
     cy.visit('/');
     cy.get('input[id=searchBar]').type('test{enter}');
-    cy.url().should('include', '/search?search=test');
+    cy.url().should('include', '/search?entryType=workflows&search=test');
   });
 });
 
 describe('Test search page functionality', () => {
   it('displays tools', () => {
     cy.visit('/search');
-    cy.get('[data-cy=toolNames]').should('have.length.of.at.least', 1);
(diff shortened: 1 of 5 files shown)


denis-yuen:
Assuming this is #2073

cypress/integration/manualTests/basic-enduser.ts
```diff
@@ -100,14 +100,14 @@ describe('Test logged out home page', () => {
   it('home page search bar works', () => {
     cy.visit('/');
     cy.get('input[id=searchBar]').type('test{enter}');
-    cy.url().should('include', '/search?search=test');
+    cy.url().should('include', '/search?entryType=workflows&search=test');
```
denis-yuen, line 103: Thought, this is a test so not strictly responsible but got me thinking,
what happens with an old search that someone had bookmarked that doesn't specify an entry type?

garyluu: defaults to workflow with possibly weird facets selected

denis-yuen: I'm fine with defaults to workflows. Why weird facets selected though?

coverbeck (approved):

cypress/integration/manualTests/basic-enduser.ts
garyluu: if you previously selected tool-specific facets and have a link for that, it'll now default to looking for a workflow with that facet selected...which won't yield any results anymore

denis-yuen: Oh ok, I think I can live with that

denis-yuen (approved):

NatalieEO (approved):

PR #1177, opened 2021-03-11
fix circle ci build

wshands:


Richard-Hansen (commented):
This may fix the current build, but it would end up breaking builds down that line when the webservice gets updated.

It looks like the webservice develop branch hasn't built in a few days (https://app.circleci.com/pipelines/github/dockstore/dockstore?branch=develop), which would be causing this. Without the updated webservice JARs, the ui will be running on old generated files. Once the webservice completes a build this should resolve itself.

wshands:
Closing based on Richard's comment

PR #1178, opened 2021-03-16
styling bugs

wendidy:
1. [dockstore/issues/4106](https://github.com/dockstore/dockstore/issues/4106)
image

2. [SEAB-2283 comment about card titles](https://ucsc-cgl.atlassian.net/browse/SEAB-2283)
The screenshot of the messed up tools and workflows line height is fixed, it seems like that only happened on Firefox. At the same time I realized on Chrome the line does not break when screen size is small. They are all fixed now.

3. [SEAB-2283 comment about organization img](https://ucsc-cgl.atlassian.net/browse/SEAB-2283)
Gave it an empty `div` when `src` is unknown, so the problem is right now some `src` are there but they are no longer valid. Wondering if anyone could give me some input on this
image

### DIFF
--- a/src/app/search/search-workflow-table/search-workflow-table.component.html
+++ b/src/app/search/search-workflow-table/search-workflow-table.component.html
@@ -39,7 +39,7 @@
       <mat-cell fxShow fxHide.lt-sm data-cy="descriptorType" *matCellDef="let workflow" fxFlex="20%">
         <div>
           <span class="nonclick-badge">{{
-            workflow?.descriptorType === 'gxformat2' ? 'Galaxy' : (workflow?.descriptorType | uppercase)
+            workflow?.descriptorType === 'gxformat2' ? galaxyShortfriendlyName : (workflow?.descriptorType | uppercase)
           }}</span>
         </div>
       </mat-cell>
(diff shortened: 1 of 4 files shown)


src/app/search/search-workflow-table/search-workflow-table.component.html
```diff
@@ -38,7 +38,9 @@
       >
       <mat-cell fxShow fxHide.lt-sm data-cy="descriptorType" *matCellDef="let workflow" fxFlex="20%">
         <div>
-          <span class="nonclick-badge">{{ workflow?.descriptorType | uppercase }}</span>
+          <span class="nonclick-badge">{{
+            workflow?.descriptorType === 'gxformat2' ? galaxyShortfriendlyName : (workflow?.descriptorType | uppercase)
```
coverbeck, line 42: I think you could change `'gxformat2'` to `DescriptorLanguageService.knownGalaxyValue`.

coverbeck (approved):

denis-yuen (approved):

src/app/search/search-workflow-table/search-workflow-table.component.html
wendidy: @coverbeck I think then I will need to make `knownGalaxyValue` `static`, but it will break other code. 
And I feel like it's unnecessary to initialize a private `descriptorLanguageService` and then create `ngOnInit()`, and also make `knownGalaxyValue` `public`?

coverbeck: Good point, I hadn't noticed that when I grepped.

It is exposed statically as an enum elsewhere, in ExtendedDescriptorLanguageBean, but you'd still have to do a little extra work on it, so just leave it as is.

PR #1190, opened 2021-04-02
Upgrade akita to latest

david4096:
https://ucsc-cgl.atlassian.net/browse/SEAB-2524


coverbeck (commented):
I am wary of such a big Akita version change, since it is a key component almost everywhere. Even though all the tests passed....

Going to wait until for Gary to approve first.

david4096:
> I am wary of such a big Akita version change, since it is a key component almost everywhere. Even though all the tests passed....
> 
> Going to wait until for Gary to approve first.

I had the same thought, definitely don't want to merge without Gary. In addition, I wonder what kind of tests would give you more confidence? The fact that nothing broke could speak highly to the akita developers for keeping API parity?

coverbeck:
> In addition, I wonder what kind of tests would give you more confidence?

I think our tests are fairly comprehensive, but I still can't quite get over a 2 major version upgrade for a major component requiring no changes.

> The fact that nothing broke could speak highly to the akita developers for keeping API parity?

Could be... I have vague recollections of Gary mentioning some Akita migration being difficult.

garyluu:
The migration difficulty depends on which major version.  Sometimes there's a lot of breaking changes, sometimes there is not. Since there's no compilation errors, just need to play around with the website. In particular, check that the akita-ng-forms-manager still works.

david4096:
At Gary's suggestion I tested out the parts of the site that rely on akita-ng-forms-manager: Collections and Organizations. I was able to add an organization, edit its members, create a collection, modify the collection, and add entries to it. If anyone has other suggestions of parts to test before this is merge-able feel free to ask. It does seem to me like this is covered by the cypress tests, too https://github.com/dockstore/dockstore-ui2/blob/develop/cypress/integration/group2/organizations.ts#L134

image

image

image

image

coverbeck (approved):

garyluu (approved):

denis-yuen (approved):

NatalieEO (approved):

Richard-Hansen (approved):

PR #1200, opened 2021-04-13
New OAuth window

garyluu:
For https://ucsc-cgl.atlassian.net/browse/SEAB-2782 

This now only applies to ORCID account linking

Before PR:
- Click on link account (could be on several different pages)
- Page gets redirected to OAuth screen
- User accepts OAuth stuff
- Page gets redirected to the previous page that the user was on
- Fail message shows if it failed

This PR:
- Click on link account (could be on several different pages)
- A new tiny window pops up
- User accepts OAuth stuff on this new tiny window
- This new tiny window gets redirected to the AuthComponent of Dockstore and does a bit of stuff
  - if it succeeded then automatically close
  - if it failed then the error message would display in that tiny window
- The original page has the userObject refreshed (if the oauth succeeded)

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -47,6 +47,7 @@ export function assertNoTab(tabName: string): any {
 export function resetDB() {
   before(() => {
     cy.exec('java -jar dockstore-webservice.jar db drop-all --confirm-delete-everything travisci/web.yml');
+    // cy.exec('docker exec -i postgres1 psql webservice_test -U postgres < travisci/db_dump.sql');
     cy.exec('PGPASSWORD=dockstore psql -h localhost -f travisci/db_dump.sql webservice_test -U dockstore');
     cy.exec(
       'java -jar dockstore-webservice.jar db migrate -i 1.5.0,1.6.0,1.7.0,1.8.0,1.9.0,1.10.0,alter_test_user_1.10.2,1.11.0 travisci/web.yml'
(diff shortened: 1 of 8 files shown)


denis-yuen:
> Kind of a big change. Feels like I'm going backwards and reverting something that was intentional. This affects linking ALL accounts.

 Don't know how I feel about this. We need to change it for ORCID since ORCID has branding guidelines, it doesn't seem necessary for the other ones

garyluu:
> > Kind of a big change. Feels like I'm going backwards and reverting something that was intentional. This affects linking ALL accounts.
> 
> Don't know how I feel about this. We need to change it for ORCID since ORCID has branding guidelines, it doesn't seem necessary for the other ones

Little messier and inconsistent, but I just made it only affect ORCID

cypress/support/commands.ts
```diff
@@ -47,6 +47,7 @@ export function assertNoTab(tabName: string): any {
 export function resetDB() {
   before(() => {
     cy.exec('java -jar dockstore-webservice.jar db drop-all --confirm-delete-everything travisci/web.yml');
+    // cy.exec('docker exec -i postgres1 psql webservice_test -U postgres < travisci/db_dump.sql');
```
denis-yuen, line 50: Intentional?

denis-yuen (approved):
> Little messier and inconsistent, but I just made it only affect ORCID

Will take a look when it makes its way out to dev

cypress/support/commands.ts
garyluu: yes

coverbeck (approved):

wshands (approved):

NatalieEO (approved):

PR #1202, opened 2021-04-14
search.ts reset filters

wendidy:
https://github.com/dockstore/dockstore/issues/4008

Seems like I'm unable to reproduce this problem, added wait(1000) and modified assertion.
Note, right now there's a `michaelgatzen/dockstore_demo/HelloWorldWorkflow` when searching for `dockstore_d`, so I changed that to `dockstore_i`


garyluu (approved):
If the wait isn't necessary after the dockstore_i change, remove it

coverbeck (approved):

denis-yuen (approved):

PR #1203, opened 2021-04-14
passing in the correct parameter value of descriptor type before displaying code block

wendidy:
https://github.com/dockstore/dockstore/issues/3676
The problem before was when displaying the code block inside a container, _sometimes_ the correct descriptor type(for exmaple WDL) was not passed into some `launchService` functions. Thus this resulted in a missing `--descriptor wdl`

Here is after the fix:
image

Optimization ticket: https://github.com/dockstore/dockstore/issues/4194

### DIFF
--- a/src/app/container/launch/launch.component.ts
+++ b/src/app/container/launch/launch.component.ts
@@ -72,7 +72,6 @@ export class LaunchComponent extends Base implements OnInit, OnChanges {
       .subscribe((descriptors: Array<Workflow.DescriptorTypeEnum>) => {
         this.descriptors = descriptors;
         this.filteredDescriptors = this.filterDescriptors(this.descriptors, this._selectedVersion, this.versionsFileTypes);
-        this.reactToDescriptor();
       });
     this.published$ = this.toolQuery.toolIsPublished$;
   }
@@ -81,6 +80,7 @@ export class LaunchComponent extends Base implements OnInit, OnChanges {
     this._selectedVersion = this.selectedVersion;
     this.filteredDescriptors = this.filterDescriptors(this.descriptors, this._selectedVersion, this.versionsFileTypes);
     this.reactToDescriptor();
+    this.changeDescriptor();
   }
 
   // Returns an array of descriptors that are valid for the given tool version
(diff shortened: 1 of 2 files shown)


src/app/container/launch/launch.component.ts
```diff
@@ -72,14 +72,15 @@ export class LaunchComponent extends Base implements OnInit, OnChanges {
       .subscribe((descriptors: Array<Workflow.DescriptorTypeEnum>) => {
         this.descriptors = descriptors;
         this.filteredDescriptors = this.filterDescriptors(this.descriptors, this._selectedVersion, this.versionsFileTypes);
```
coverbeck, line 74: After a quick skim of the code, I would be inclined to put this line inside `reactToDescriptor`, just to avoid "everytime you call `reactToDescriptor`, make sure to first call `this.filterDescriptors`". 

But "everywhere" is only twice, and there may be a subtlety I'm missing, so your call.

coverbeck (approved):

src/app/container/launch/launch.component.ts
wendidy: Thanks Charles for the suggestion. I tried with it, but interestingly enough the page did not get updated correctly. I'm actually very curious as to why, if anyone knows please let me know. But I think we will leave the code like that for now

garyluu (changes requested):
logging the amount of 'reactToDescriptor' calls, I originally see 2 calls (not great). Now i I see 4 calls. 

Try to fix this, otherwise. Spin off to another ticket and someone else can take care of it.

wendidy:
> logging the amount of 'reactToDescriptor' calls, I originally see 2 calls (not great). Now i I see 4 calls.
> 
> Try to fix this, otherwise. Spin off to another ticket and someone else can take care of it.

As it turns out it could have been kept at 2 calls.
Because the html is calling the `reactToDescriptor` function when the `currentDescriptor` changes; I made a new function that holds all the things that will be changed, and left out the unchanged one behind in `reactToDescriptor`.
Now that I try to log, each function is called twice at the beginning

wendidy:
> logging the amount of 'reactToDescriptor' calls, I originally see 2 calls (not great). Now i I see 4 calls.
> 
> Try to fix this, otherwise. Spin off to another ticket and someone else can take care of it.

Added a new `changeDescriptor` for the purpose of separating functions that need to be called when descriptor is changing and those that don't. In this case `reactToDescriptor` is called 2 times, and `changeDescriptor` 4 times. There really is no improvement but I think so far that's all I can think of. I think I might need to get help from someone on this

garyluu (approved):
😢

denis-yuen (approved):

PR #1206, opened 2021-04-16
Hide display of non-Dockstore tokens

kathy-t:
First part of [SEAB-1713](https://ucsc-cgl.atlassian.net/browse/SEAB-1713)

Hid the show token button and copy token button for non-Dockstore tokens so users can't access it. Looks like the following:

Hide tokens

### DIFF
--- a/src/app/loginComponents/accounts/external/accounts.component.html
+++ b/src/app/loginComponents/accounts/external/accounts.component.html
@@ -74,21 +74,6 @@
       </div>
 
       <span fxLayout="row" fxLayoutAlign="start center" *ngIf="row.isLinked">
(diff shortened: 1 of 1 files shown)


denis-yuen (approved):

coverbeck (approved):
Looks good, but note there will have to be another iteration where the Link/unlink buttons' display can't be determined by the existence of a token. And looking at this code, we may need to send the username in the response as well.

garyluu (commented):
row.isLinked is checked twice. The fxlayout and fxlayoutalign is probably not needed anymore

NatalieEO (approved):

garyluu (approved):

PR #1208, opened 2021-04-21
Use webservice endpoint to get user's GitHub organizations

kathy-t:
Part 2 of [SEAB-1713](https://ucsc-cgl.atlassian.net/browse/SEAB-1713).

The UI was getting the user's GitHub organizations itself, so I removed it and used the webservice endpoint instead. From what I could find, the function was only being called in my-workflow.

### DIFF
--- a/src/app/myworkflows/my-workflow/my-workflow.component.ts
+++ b/src/app/myworkflows/my-workflow/my-workflow.component.ts
@@ -127,9 +127,8 @@ export class MyWorkflowComponent extends MyEntry implements OnInit {
   ngOnInit() {
     this.myWorkflowsService.clearPartialState();
     this.gitHubAppInstallationLink$ = this.sessionQuery.gitHubAppInstallationLink$;
-    this.tokenQuery.gitHubToken$
-      .pipe(takeUntil(this.ngUnsubscribe))
-      .subscribe((token: string) => this.tokenService.getGitHubOrganizations(token));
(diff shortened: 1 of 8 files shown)


src/app/myworkflows/my-workflow/my-workflow.component.ts
```diff
@@ -127,9 +127,8 @@ export class MyWorkflowComponent extends MyEntry implements OnInit {
   ngOnInit() {
     this.myWorkflowsService.clearPartialState();
     this.gitHubAppInstallationLink$ = this.sessionQuery.gitHubAppInstallationLink$;
-    this.tokenQuery.gitHubToken$
-      .pipe(takeUntil(this.ngUnsubscribe))
-      .subscribe((token: string) => this.tokenService.getGitHubOrganizations(token));
+    this.tokenQuery.gitHubToken$.pipe(takeUntil(this.ngUnsubscribe)).subscribe();
```
coverbeck, line 130: I don't think you need this line at all, do you? You're subscribing to it, but not observing anything. Unless there is some side effect I'm missing.

kathy-t: You're right, it's not needed. I was planning on removing token-related code in my next PR, but I'll remove this line now since we're here

garyluu (approved):
Once test passes

denis-yuen (approved):
on passing build, github will force that anyway

NatalieEO (approved):

coverbeck (approved):

PR #1210, opened 2021-04-22
minor adjustments to multiple pages

wendidy:
Please refer to the points that are crossed off in this document: https://docs.google.com/document/d/1GLshBv7Ue09ceCFDSLmFvrgbEofGkY6R0jXlKxVjqHc/edit?ts=607ded9d

Screenshots(only some minor changes):
1. Added a side scrollbar for cards that have too much content(only visible when scrolling):
image

other things are pretty straight forward, if screenshots are needed then I will insert them here

### DIFF
--- a/src/app/about/about.component.html
+++ b/src/app/about/about.component.html
@@ -1,13 +1,13 @@
-<div>
-  <div fxLayout="column" fxLayoutAlign="space-around" fxLayoutGap="4rem" class="mb-5">
+<div class="hidden-overflow">
+  <div fxLayout="column" fxLayoutAlign="space-around" fxLayoutGap="4rem" class="mb-5 w-100">
     <div class="about-dockstore">
       <div fxLayout="row" fxLayout.lt-md="column" LayoutAlign="space-between center" fxLayoutGap="1rem" class="container">
         <div fxLayout="column" fxLayoutAlign="space-between start" fxLayoutGap="2rem">
           <h1>About Dockstore</h1>
-          <h5>
+          <h3>
             Dockstore’s mission is to enable researchers and developers to share and reuse analytical workflows and tools in a way that
             makes them machine readable and runnable in a variety of environments.
-          </h5>
+          </h3>
           <p>
             Established in 2015, Dockstore has been an evolving proof of concept designed as a starting point for community standards within
             the <a target="_blank" rel="noopener noreferrer" href="https://www.ga4gh.org/">GA4GH</a>.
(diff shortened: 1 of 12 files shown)


src/styles.scss
```diff
@@ -1341,3 +1341,12 @@ hr {
   margin: 0.5rem 0;
   border: 1px solid #dfe0e2;
 }
+
+.orgLogo {
+  height: 50px;
+  object-fit: scale-down;
+}
+
+.hidden-overflow {
```
garyluu, line 1350: try:

```
@each $value in $overflows {
  .overflow-#{$value} { overflow: $value !important; }
}
```
and using class overflow-hidden and see what happens. if it works, then ur .content style can also be replaced

assuming everything works, put that in  the boostrap4 scss file

wendidy: It does, thank you!

src/app/about/about.component.html
```diff
@@ -36,29 +36,29 @@ <h5>
...
 
     <div fxLayout="column" class="container" fxLayoutAlign="space-between center" fxLayoutGap="1rem">
-      <h3>Dockstore follows the FAIR Guiding Principles</h3>
+      <h2>Dockstore follows the FAIR Guiding Principles</h2>
       <span
         >Our
         <a
           target="_blank"
           rel="noopener noreferrer"
-          [href]="Dockstore.DOCUMENTATION_URL + 'advanced-topics/best-practices/best-practices-secure-fair-workflows.html'"
```
garyluu, line 45: Assuming this is intended

garyluu (approved):
Overall assuming everything looks right

src/app/about/about.component.html
denis-yuen: Hard-coding this does seem a bit odd

denis-yuen (approved):
Sonarcloud smells are intentional, removal of DOCUMENTATION_URL seems odd

src/app/about/about.component.html
denis-yuen: In discussion, oddly https://docs.dockstore.org/en/develop/advanced-topics/best-practices/best-practices-secure-fair-workflows.html exists but https://docs.dockstore.org/en/latest/advanced-topics/best-practices/best-practices-secure-fair-workflows.html does not

denis-yuen: Probably due to https://ucsc-cgl.atlassian.net/browse/SEAB-2737

PR #1211, opened 2021-04-23
orcid branding

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-2781
Added orcid branded button and changed wording of linking account buttons.
Add an explanation of what we do with ORCID credentials.
After reading the documentation, I don't think the person's name is necessary. The name is an option if you aren't using the full or compact id display

### DIFF
--- a/src/app/loginComponents/accounts/external/accounts.component.html
+++ b/src/app/loginComponents/accounts/external/accounts.component.html
@@ -89,8 +89,9 @@
           (click)="link(row.source)"
           matTooltip="Link the account"
         >
-          <mat-icon>link</mat-icon>
-          <span>Link Account</span>
+          <mat-icon *ngIf="row.name !== 'ORCID'">link</mat-icon>
+          <img *ngIf="row.name === 'ORCID'" src="../../assets/images/account-logos/orcid.svg" alt="ORCID iD logo" class="orcid-id-logo" />
+          <span> Create or connect your {{ row.name }} Account</span>
         </button>
       </span>
       <ng-template #linked>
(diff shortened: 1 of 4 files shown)


src/app/loginComponents/accounts/external/accounts.component.html
```diff
@@ -89,8 +89,9 @@ <h3>
           (click)="link(row.source)"
           matTooltip="Link the account"
         >
-          <mat-icon>link</mat-icon>
-          <span>Link Account</span>
+          <mat-icon *ngIf="row.name !== 'ORCID'">link</mat-icon>
+          <img *ngIf="row.name === 'ORCID'" src="../../assets/images/account-logos/orcid.svg" alt="ORCID iD logo" class="orcid-id-logo" />
```
coverbeck, line 93: You could alternatively do this more generically by adding a property to row in accounts.component.ts. At this point it's a wash, but if we ever have to do this for a second button, then we probably should move it into Typescript.

garyluu (approved):
the two lines are repeated 3 times. you can consider using ng-template. approved either way

coverbeck (approved):

denis-yuen (approved):
>  After reading the documentation, I don't think the person's name is necessary. The name is an option if you aren't using the full or compact id display 

Ah, that's why.

src/app/loginComponents/accounts/external/accounts.component.ts
```diff
@@ -188,6 +193,9 @@ export class AccountsExternalComponent implements OnInit, OnDestroy {
     this.tokenQuery.tokens$.subscribe((tokens: Token[]) => {
       this.setTokens(tokens);
     });
+    this.userQuery.user$.subscribe((user: User) => {
```
garyluu, line 196: don't subscribe, pipe map. this way, if user.orcid isn't needed, it won't be subscribed to

src/app/loginComponents/accounts/external/accounts.component.html
```diff
@@ -76,6 +76,10 @@ <h3>
       <span *ngIf="row.isLinked">
         <strong>Username: </strong>
         {{ row.source | getTokenUsername: tokens }}
+        <span *ngIf="row.name === 'ORCID'">
+          <img src="../../assets/images/account-logos/{{ row.logo }}" alt="ORCID iD logo" class="orcid-id-logo" />
+          <a href="https://orcid.org/{{ user.orcid }}" style="font-size: 0.8em"> {{ user.orcid }}</a>
```
garyluu, line 81: avoid inline style. this is to prevent our components from having a variety of different font-sizes

PR #1217, opened 2021-04-30
Use Kim's color scheme for material

garyluu:
The idea is that all primary, secondary, and error colors of Kim's new color scheme is defined here and only here. This is to maintain a consistent UI.

The grayscale, borders, and others will need another solution (different file) since material doesn't seem to care about those.

Before and after.

image

The primary material color has not appeared to be changed. The lighter and darker variants (used in progress bar for example). was changed but probably doesn't matter too much to find a screenshot

### DIFF
--- a/src/material.scss
+++ b/src/material.scss
@@ -21,80 +21,65 @@
 // Be sure that you only ever include this mixin once!
 @include mat-core();
 
-$my-blue: (
-  50: #e8ecf1,
-  100: #c4cedf,
-  200: #9fafc9,
-  300: #7b8fb2,
-  400: #5d77a3,
-  500: #3f5f95,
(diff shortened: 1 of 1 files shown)


denis-yuen (approved):

denis-yuen:
> If PR is approved, I'm banning all mention of any colors specified in this file elsewhere. For example, we have "397a84" mentioned 6 times in other files 

Automatically or by team convention?

NatalieEO (approved):

garyluu:
> > If PR is approved, I'm banning all mention of any colors specified in this file elsewhere. For example, we have "397a84" mentioned 6 times in other files
> 
> Automatically or by team convention?

Team convention

PR #1218, opened 2021-04-30
Remove duplicate secondary color

garyluu:
PR just to remove duplicate 397a84

Demonstrates how not to repeat colors

Split material.scss to a different file because:

https://material.angular.io/guide/theming#defining-a-custom-theme
"Your custom theme file should not be imported into other SCSS files. This will duplicate styles in your CSS output. If you want to consume your theme definition object (e.g., $candy-app-theme) in other SCSS files, then the definition of the theme object should be broken into its own file, separate from the inclusion of the mat-core and angular-material-theme mixins."

### DIFF
--- a/src/app/about/about.component.scss
+++ b/src/app/about/about.component.scss
@@ -1,3 +1,5 @@
+@import '/src/materialColorScheme.scss';
+
 .about-dockstore {
   margin-top: 6rem;
   background-image: url(../../assets/images/about-page/city.png);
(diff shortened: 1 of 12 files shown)


coverbeck (approved):

garyluu:
Also removed duplicate 21335b

garyluu:
Removed the rest too. Very weird `color: mat-color($dockstore-app-warn, darker);` works but `color: mat-color($dockstore-app-warn, 1);` doesn't

denis-yuen (approved):
FYI @e34lin

coverbeck (approved):

wshands (approved):

NatalieEO (approved):

PR #1219, opened 2021-05-03
seab-2789/redirect after linking orcid account

NatalieEO:
Following guidelines https://info.orcid.org/documentation/integration-guide/user-experience-display-guidelines/#ORCID_OAuth_sign-in_screens

https://user-images.githubusercontent.com/16004905/116945202-f5107600-ac2b-11eb-9dc9-51de26b80319.mp4

### DIFF
--- a/src/app/orcid/orcid.component.html
+++ b/src/app/orcid/orcid.component.html
@@ -1,0 +1,17 @@
+<app-header>
+  <div fxLayout="row" fxLayoutGap="1rem">
+    <img src="../../../assets/svg/sub-nav/organization.svg" alt="organization icon" />
+    <span>ORCID</span>
+  </div>
+</app-header>
+
+<div fxLayout="column" fxLayoutAlign="start center">
+  <div fxLayout="row">
+    <mat-card>
+      Thank you for connecting your ORCID iD
+      <img src="../../assets/images/account-logos/orcid.svg" alt="ORCID iD logo" class="orcid-id-logo" />
+      <br />
+      You will now be automatically directed back to your Accounts page.
+    </mat-card>
+  </div>
+</div>
--- a/src/app/shared/user/user.service.ts
+++ b/src/app/shared/user/user.service.ts
@@ -1,8 +1,11 @@
 import { Injectable } from '@angular/core';
+import { Router } from '@angular/router';
 import { transaction } from '@datorama/akita';
 import { AuthService } from 'ng2-ui-auth';
+import { Observable } from 'rxjs';
 import { Md5 } from 'ts-md5/dist/md5';
 import { AlertService } from '../alert/state/alert.service';
+import { TokenQuery } from '../state/token.query';
 import { TokenService } from '../state/token.service';
 import { WorkflowService } from '../state/workflow.service';
 import { Configuration, ExtendedUserData, User, UsersService, Workflow } from '../swagger';
@@ -10,6 +13,7 @@ import { UserStore } from './user.store';
 
 @Injectable({ providedIn: 'root' })
 export class UserService {
+  orcidAccountIsLinked$: Observable<boolean>;
   constructor(
     private userStore: UserStore,
     private authService: AuthService,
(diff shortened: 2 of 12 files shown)


src/app/orcid/orcid.component.html
```diff
@@ -0,0 +1,17 @@
...
+  <div fxLayout="row" fxLayoutGap="1rem">
+    <img src="../../../assets/svg/sub-nav/organization.svg" alt="organization icon" />
+    <span>ORCID</span>
+  </div>
+</app-header>
+
+<div fxLayout="column" fxLayoutAlign="start center">
+  <div fxLayout="row">
+    <mat-card>
+      Thank you for connecting your ORCID iD
```
coverbeck, line 11: Period at end, unless guidelines are somehow specific about not having it.

src/app/orcid/orcid.component.html
```diff
@@ -0,0 +1,17 @@
...
+      <img src="../../assets/images/account-logos/orcid.svg" alt="ORCID iD logo" class="orcid-id-logo" />
+      <br />
+      You will now be automatically directed back to your Accounts page.
```
coverbeck, line 14: Totally optional, but you could make the accounts page a link, in case the user doesn't feel like waiting; I've seen sites like that. Unless the guidelines don't allow it.

src/app/shared/user/user.service.ts
```diff
@@ -74,14 +80,28 @@ export class UserService {
     }
   }
 
-  getSingleUser(): void {
+  getSingleUser(redirectAfterSuccess: boolean, url?: string): void {
```
coverbeck, line 83: I suggest just having one parameter, the redirectUrl. Instead of having a parameter that says to honor the second parameter, just have/don't have the url param. I don't think you need it unless there's ever a case where the url is set but you don't want to redirect.

src/app/shared/user/user.service.ts
```diff
@@ -74,14 +80,28 @@ export class UserService {
...
     if (this.configuration.apiKeys['Authorization']) {
       this.usersService.getUser().subscribe(
         (user: User) => {
           this.updateUser(user);
           if (user) {
             this.tokenService.get(user.id);
+            if (redirectAfterSuccess) {
+              switch (url) {
+                case '/orcid':
+                  this.orcidAccountIsLinked$ = this.tokenQuery.hasOrcidToken$;
```
coverbeck, line 94: Rusty on Akita, not sure if these make sense:

1. Does this need to be a class property? Can't `this.orcidAccountIsLinked$` be a local variable?
2. Do you need to unsubscribe?
3. Could you just get the value instead of using a subscription? Or do you expect the subscription to become true at some point after this?

garyluu (changes requested):
Well this is awkward. This issue should've been rejected as duplicate and closed. Back when i was working on this, the idea of having a new page for the redirect was explored and deemed to be not worth the effort. Some reasons were:

- Additional code and component that's too specific to ORCID and will not be consistent with other account linkng behaviour
- The new component will need logic that remembers which page the user was previously on (which may not be /account)
- Just a whole lot of work

Instead "Provide feedback that their record has been connected" and such was taken more liberally. The accounts page with buttons and icons being changed + the (now seemingly missing) matsnackbar success alert should've been enough.

In any case, a new page for the redirect is welcome but should behave this way I think:

- the redirect URL (the new component) should be displayed inside the small popup window as mentioned by "Remember that your [redirect URI call](Link to documentation) will also be displayed inside this popup window. Make sure your content and formatting are appropriately styled and sized."
- once closed automatically or by user, a variety of things can happen like the previous accounts page being updated (like it already is)

So for this PR (assuming you're proceeding with new component)...after the oauth process, I should not see accounts page redirecting to new component then redirecting back to accounts page. Even creating an autoclosing material dialog or snackbar would've been better.

src/app/shared/user/user.service.ts
garyluu: 1. no, yes
2. yes
3. should be yes

PR #1220, opened 2021-05-04
search term limit

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-2515
Needed to add front end portion

### DIFF
--- a/src/app/search/state/search.service.ts
+++ b/src/app/search/state/search.service.ts
@@ -21,6 +21,8 @@ import { transaction } from '@datorama/akita';
 import { AdvancedSearchObject, initialAdvancedSearchObject } from 'app/shared/models/AdvancedSearchObject';
 import { Explanation } from 'elasticsearch';
 import { BehaviorSubject } from 'rxjs';
+import { AlertService } from '../../shared/alert/state/alert.service';
+import { searchTermLengthLimit } from '../../shared/constants';
 import { Dockstore } from '../../shared/dockstore.model';
 import { ImageProviderService } from '../../shared/image-provider.service';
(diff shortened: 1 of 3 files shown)


garyluu (approved):

src/app/search/state/search.service.ts
```diff
@@ -166,6 +169,13 @@ export class SearchService {
 
   @transaction()
   setSearchText(text: string) {
+    if (text.length > searchTermLengthLimit) {
+      text = '';
+      this.alertService.customDetailedError(
```
coverbeck, line 174: Same question as other comment -- are you displaying this before the request is even sent?

Also:

1. It would be good to avoid the duplication, but it might be hard to find a good place to put it.
2. I'm wary of a function behaving unexpectedly like this without any indication that it ignored the input (the error message is for the end user, not the code that invoked this method). But this is an extreme edge case presumably, most likely only run into somebody trying to break the code.

coverbeck (approved):

denis-yuen (approved):

src/app/search/state/search.service.ts
NatalieEO: Yeah, I'm displaying before ever sending a request so I'll remove the status code. 

And I'm going to remove one of these alerts because it wasn't actually doing anything... If a user performs a search using the search box, they'll get the red alert box. If they do a search by using a url, then the search term gets replaced with the empty string, which causes the permalink to be updated. There won't be an error displayed to the user, the page just loads like navigating to /search. But yeah I can't imagine a user using such a long term for a genuine search so I think this is okay.

PR #1222, opened 2021-05-05
Add hover and active states

kathy-t:
For [SEAB-2845](https://ucsc-cgl.atlassian.net/browse/SEAB-2845)
[Zeplin link](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/607da9695f1d1e9c06b6e74e)

**Navigation bar**
This picture shows the active state (user is on the Search page) and the hover state for the "About" button:
active and hover navigation bar
Notes:
- There's no hover or active state for the Dockstore home button
- There's no active state for Docs or Forum because they open a new window. The active state is what ever page they were on previously
- Some of the buttons on the Home page, About page, or Organizations page take you to other sections of the navigation bar, so the active state is applied accordingly, depending on what page you're currently on. It does not depend on what button you click on the navigation bar. 
- Example 1: if you're on the search page (active state is Search), and you click on a workflow, the active state on Search is gone and there is no active state on the navigation bar
- Example 2: if you're on the home page, and you click the Browse Organizations button, the active state will be on the Organizations button in the navigation bar

**White links on dark background**
hover white links on dark background

**Green accent buttons**
This shows the hover state when the mouse is hovering over the "About Dockstore" button:
Hover for green buttons
Other green buttons on the home page also have this hover effect, including the login button. 

I made some changes so other green mat-raised-buttons on other pages, like Organizations, also have this effect. For example, the View buttons:
Hover state for view buttons

Another example is on the Search page:
hover search green buttons

**Tabs**
This is how the tabs look like with no hover:
no hover tabs

This is how they look like with hover:
hover tabs
I did it by changing the opacity of the tab on hover, but in my opinion it's still a little hard to tell since the font is thin. I could change the opacity to 1 on hover, but then it may look too much like an active tab. Any thoughts about this would be appreciated.

This hover effect on tabs is also applied to any other tabs on the site, like the My Workflows page.
hover tabs my workflows

**Misc**

Fixed the How to Star link because I noticed the page was highlighting the word "star" which made it hard to read.

Refactored the code that highlighted the active page in this menu so it was similar to how I did it for the navigation bar:
Active state Menu
The appearance is still the same, but the active state is remembered even if the page is refreshed. Previously, it was lost when refreshed.


src/app/about/about.component.scss
```diff
@@ -18,10 +18,7 @@
 }
 
 button {
-  background-color: mat-color($dockstore-app-accent);
-  color: white;
-  padding: 1rem;
-  text-decoration: none;
```
garyluu, line 24: 👍

src/app/navbar/navbar.component.ts
```diff
@@ -69,6 +71,7 @@ export class NavbarComponent extends Logout implements OnInit {
       }
     });
     this.userQuery.extendedUserData$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((extendedUser) => (this.extendedUser = extendedUser));
+    this.navbarService.activePage$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((activePage) => (this.activePage = activePage));
```
garyluu, line 74: shouldn't need to manually set and activePage. The activePage can probably be figured out using some field within activatedRoute. If activatedRoute isn't possible. this should not be a subscribe. an async pipe in the html should do

src/styles.scss
```diff
@@ -551,6 +555,14 @@ div.button a {
   margin-top: 8px;
 }
 
+.mat-raised-button:hover {
+  text-decoration: none;
```
garyluu, line 559: it should already be none. Is there something else causing it to not be? Find and remove that instead

denis-yuen (commented):
That's a pretty thorough write-up, thanks!

> I did it by changing the opacity of the tab on hover, but in my opinion it's still a little hard to tell since the font is thin. I could change the opacity to 1 on hover, but then it may look too much like an active tab. Any thoughts about this would be appreciated.

I did do a double-take before noticing the change. That said, it might be more obvious "in motion" when the cursor is moving, I think we can run with this for now.

coverbeck (approved):
I find the distinction between active and hover in the navigation bar pretty subtle in your screenshot. I think it's partially due to the fact that the image is cropped in the PR -- in the Zepplin link the difference is more noticeable. But even there, I can't really tell which is which.

It might be obvious though when you see it in context of the mouse moving, similar to Denis' other comment.

src/styles.scss
kathy-t: My mistake, that shouldn't be there. Forgot to remove it after testing. Removed now

src/app/navbar/navbar.component.ts
kathy-t: Removed that in my latest PR. Setting active state for the navbar and menu items using `routerLinkActive` in the html now

src/app/navbar/navbar.component.scss
```diff
@@ -14,18 +14,34 @@
...
   font-weight: bold;
   text-decoration: none;
   color: white;
   font-size: 14px;
+  border-radius: 0;
+  height: 64px;
+}
+
+.icon:hover {
+  background-color: #112246;
```
garyluu, line 36: what is this color? only used for this component and not elsewhere or in the color palette?

src/app/navbar/navbar.component.scss
```diff
@@ -14,18 +14,34 @@
...
+}
+
+.icon-active {
+  background-color: mat-color($dockstore-app-primary, darker);
+}
+
+.menu-item-active {
+  background-color: #dfe0e2;
```
garyluu, line 44: ditto

src/app/home-page/home-logged-out/home.component.scss
```diff
@@ -28,10 +28,7 @@
 }
 
 button {
-  background-color: mat-color($dockstore-app-accent);
-  color: white;
-  padding: 1rem;
-  text-decoration: none;
+  padding: 0.3rem 1rem;
```
garyluu, line 31: what's so special about the buttons on the logged out homepage that it has different padding than all the buttons in the entire site?

src/app/navbar/navbar.component.scss
kathy-t: It's the color for the navigation bar's hover state, according to the [zeplin](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/607da9695f1d1e9c06b6e74e) and it's only used in this component. The actual color isn't specified in the zeplin, but using the svg image given, I was able to identify the color as `#112246`. I couldn't find that color in [kim's color palette](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/5fc699d068c46f17551f745b)

garyluu: put a code comment and say whether it's darker or lighter than the ones in kim's color palette

src/app/navbar/navbar.component.scss
kathy-t: This color appears to be in the gray scale of kim's color palette, but the gray scale isn't in `materialColorScheme.scss`. The color is used in 3 styles in `styles.scss` and in the about component as well

src/app/home-page/home-logged-out/home.component.scss
garyluu: what about other pages? what padding are tehy using?

src/app/navbar/navbar.component.scss
garyluu: See if you can somehow specify it in the material color scheme and then reference it. if not possible via that, do something from https://sass-lang.com/documentation/variables

src/app/home-page/home-logged-out/home.component.scss
kathy-t: Seems like mat-raised-buttons on other pages don't have padding (just the default padding from mat-raised-buttons)

garyluu: looks like the default is 0px and 16px. if you're going to add padding for buttons, add it globally (unless there's a good reason)

denis-yuen (approved):
With Gary's feedback

src/app/home-page/home-logged-out/home.component.scss
kathy-t: Edit: removed the additional padding for buttons on the home page and about page because it didn't make that much of a difference (pics above for comparison). It still has the default 0px and 16px padding

kathy-t:
Also added the grayscale from https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/5fc699d068c46f17551f745b to the material color scheme because several of the grayscale colors are used in the UI. Removed the duplicate grayscale colors

garyluu (approved):

NatalieEO (approved):

PR #1228, opened 2021-05-07
Feature/3588/remove bootstrap social

garyluu:
Remove bootstrap social, not replacing it with another library, just angular fontawesome.

Currently pretty basic replacement, others can prettify it

image


NatalieEO (approved):

coverbeck (approved):
For some reason I thought the social network buttons we have would have been affected by this. But presumably that would have caused a compiler error, so I guess it's good.

denis-yuen:
FYI @e34lin , think you need to accept an invite to the dockstore organization to be tagged in reviews

denis-yuen (approved):

kathy-t (approved):

garyluu:
> For some reason I thought the social network buttons we have would have been affected by this. But presumably that would have caused a compiler error, so I guess it's good.

That's the ngx-sharebuttons

PR #1232, opened 2021-05-12
Feature/3764/galaxy multi cloud launch with

NatalieEO:
Since i don't do much in the UI I wanted to get some initial feedback that this is generally okay and if there is other functionality I need to consider.

I have a webservice PR here: https://github.com/dockstore/dockstore/pull/4241

https://user-images.githubusercontent.com/16004905/118013381-e8181480-b306-11eb-93a4-4aaa037ea87f.mp4

### DIFF
--- a/src/app/shared/dockstore.model.ts
+++ b/src/app/shared/dockstore.model.ts
@@ -38,6 +38,7 @@ export class Dockstore {
   static BD_CATALYST_TERRA_IMPORT_URL = 'https://terra.biodatacatalyst.nhlbi.nih.gov/#import-tool/dockstore';
   static CAVATICA_IMPORT_URL = 'https://cavatica.sbgenomics.com/integration/trs/import';
   static NEXTFLOW_TOWER_IMPORT_URL = 'https://tower.nf/launch';
+  static GALAXY_INSTANCE_IMPORT_URL = 'https://usegalaxy.org/workflows/trs_import?trs_server=dockstore.org&trs_id=';
 
   static GITHUB_CLIENT_ID = 'will be filled in by configuration.service';
   static GITHUB_AUTH_URL = 'https://github.com/login/oauth/authorize';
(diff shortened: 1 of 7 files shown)


coverbeck (commented):
Just going off video (plus one other comment), looks like it's going down the right path.

I feel it does reveal the need to be able to remove entries, which I don't think the mocks have.

NatalieEO:
> I feel it does reveal the need to be able to remove entries, which I don't think the mocks have.

Yeah I didn't notice removal in the mocks. Should I try adding that now or wait till there is a mock?

coverbeck:
> > I feel it does reveal the need to be able to remove entries, which I don't think the mocks have.
> 
> Yeah I didn't notice removal in the mocks. Should I try adding that now or wait till there is a mock?

Probably wait... In practice I don't think that many people will run into yet. But I'd give @kcullion a heads up.

denis-yuen (commented):
Might be obvious, but deletion would probably only be for those custom URLs.

kcullion:
I didn't know that "Specify url" text field would add to the url radiobutton list. I thought it would just be saved locally for them in that text field for when they come back to launch this workflow. Was there a use case for adding/removing custom urls from the list?

NatalieEO:
> I didn't know that "Specify url" text field would add to the url radiobutton list. I thought it would just be saved locally for them in that text field for when they come back to launch this workflow. Was there a use case for adding/removing custom urls from the list?

@kcullion I think the intention is that there are sooo many galaxy instances, it's easier to let the users specify where they want to launch instead of us needing to know which ones to add. But maybe most users would only ever want to add just one custom url? @denis-yuen @coverbeck

coverbeck:
> @kcullion I think the intention is that there are sooo many galaxy instances, it's easier to let the users specify where they want to launch instead of us needing to know which ones to add. But maybe most users would only ever want to add just one custom url? @denis-yuen @coverbeck

Yes, I think typically users would want to have at most one custom url.

OTOH, although we don't know yet if this will work for the Launch with Button, Terra makes it super-easy to spin up and down Galaxy instances, so I could see that one url changing.

denis-yuen:
I'm ok with just having the "one" custom choice as @kcullion describes for this pass and worrying about extending it later when we get requests. It doesn't sound like we're cutting ourselves off from that possibility in the future anyway

PR #1233, opened 2021-05-14
Clean up collections and stargazers page

kathy-t:
For [SEAB-2849](https://ucsc-cgl.atlassian.net/browse/SEAB-2849)

Did 17, 19 - 22 from the [Dockstore UI review doc](https://docs.google.com/document/d/1GLshBv7Ue09ceCFDSLmFvrgbEofGkY6R0jXlKxVjqHc/edit)

- **Collection page breadcrumb**
Collection breadcrumb

- **Consistent logo location**
    - Stargazer page has the same header as the organization page so the logo and org title location is the same
Organization header
Also note the stargazers breadcrumb
Stargazers header
    - For the most part, the logo location from the organization page to the collection page is the same
Screenshot_2021-05-14 Dockstore Organization
Screenshot_2021-05-14 Dockstore Collection
The logo location may be a little different if the description of the organization is longer than the collection (i.e. 2 line org description vs 1 line collection description). I couldn't really find a way to make it consistent with different length descriptions unless I put the logo at the top of the container, but that makes the logo look unbalanced with the title.

- **Stargazer member cards**
I did a grid of 5 boxes, instead of 6 like the UI review doc suggests so that I could get the Orcid ID on one line

    - Normal view:
Stargazers normal size
    - Medium view:
Stargazers medium
    - Small view:
stargazers small

- **Tooltip for disabled starring button**
Tooltip to login and star

- **Hide overflow in organization cards**
Removed the overflow scrolling bar and just hid the overflow with "..."
Organization card overflow
If there's enough room (ex: small screen so the cards are big), then the full content is shown
Organization card no overflow

### DIFF
--- a/src/app/organizations/organization/organization.component.html
+++ b/src/app/organizations/organization/organization.component.html
@@ -62,7 +62,7 @@
           </div>
           <div>
             <h2 class="name">{{ org.displayName }}</h2>
-            <span>{{ org.topic }}</span>
+            <span style="word-break: break-word">{{ org.topic }}</span>
           </div>
         </div>
         <div fxFlex>
@@ -152,7 +152,7 @@
               </span>
               <span fxLayout="row" fxLayoutGap="10px" *ngIf="org?.location">
                 <mat-icon class="side-icons">location_on</mat-icon>
-                <span>{{ org.location }}</span>
+                <span style="word-break: break-word">{{ org.location }}</span>
               </span>
             </div>
             <markdown-wrapper *ngIf="org.description" fxFlex [data]="org.description"></markdown-wrapper>
--- a/src/app/organizations/organization/organization.component.html
+++ b/src/app/organizations/organization/organization.component.html
@@ -62,7 +62,7 @@
           </div>
           <div>
             <h2 class="name">{{ org.displayName }}</h2>
-            <span style="word-break: break-word">{{ org.topic }}</span>
+            <span class="text-break">{{ org.topic }}</span>
           </div>
         </div>
         <div fxFlex>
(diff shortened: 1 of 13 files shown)


src/app/organizations/organization/organization.component.html
```diff
@@ -19,37 +19,24 @@
       <img src="../../../assets/svg/sub-nav/organization.svg" alt="organization icon" />
       <a class="category-text" [routerLink]="['/organizations/']">Organizations</a>
       <span fxHide.lt-sm>/</span>
-      <span fxHide.lt-sm>{{ (organization$ | async)?.displayName }}</span>
+      <div fxHide.lt-sm *ngIf="organizationStarGazersClicked; else organizationName" fxLayoutGap="1rem">
+        <a class="category-text" [routerLink]="['./']" (click)="organizationStarGazersChange()">{{
```
coverbeck, line 23: Cool, I did not know that routerLink can be relative.

src/app/organizations/organization/organization.component.html
```diff
@@ -87,82 +74,91 @@ <h2 class="name">{{ org.displayName }}</h2>
...
-            </span>
-            <span fxLayout="row" fxLayoutGap="10px" *ngIf="org?.location">
-              <mat-icon class="side-icons">location_on</mat-icon>
-              <span class="ellipsis-lines">{{ org.location }}</span>
-            </span>
+                <mat-icon>edit</mat-icon>
+              </button>
+            </div>
+            <div *ngIf="!org.description" fxFlex>
+              <div *ngIf="canEdit$ | async" class="text-muted">Select edit to add a markdown description</div>
```
coverbeck, line 143: It was like this, but I think "Click" is better than "Select" in this context.

coverbeck (approved):

garyluu (changes requested):
Remove inline style, see if:

```
.text-break {
  word-break: break-word !important; // Deprecated, but avoids issues with flex containers
  word-wrap: break-word !important; // Used instead of `overflow-wrap` for IE & Edge Legacy
}
```
is already in bootstrap4.scss. if not, add it and use that instead

garyluu (approved):

denis-yuen (approved):

NatalieEO (approved):

PR #1234, opened 2021-05-14
Feature/3764/galaxy multi cloud launch with

NatalieEO:
dockstore/dockstore#3764
web service PR: https://github.com/dockstore/dockstore/pull/4241
Trying to make the styling match Kim's mocks is taking me longer than expected, so I decided break out that work to another PR since Kathy needs this work to do her DNAstack ticket this sprint.

https://user-images.githubusercontent.com/16004905/118317952-91007400-b4ad-11eb-80b6-6e99732e2d53.mp4

### DIFF
--- a/src/app/shared/dockstore.model.ts
+++ b/src/app/shared/dockstore.model.ts
@@ -38,7 +38,6 @@ export class Dockstore {
   static BD_CATALYST_TERRA_IMPORT_URL = 'https://terra.biodatacatalyst.nhlbi.nih.gov/#import-tool/dockstore';
   static CAVATICA_IMPORT_URL = 'https://cavatica.sbgenomics.com/integration/trs/import';
   static NEXTFLOW_TOWER_IMPORT_URL = 'https://tower.nf/launch';
-  static GALAXY_INSTANCE_IMPORT_URL = 'https://usegalaxy.org/workflows/trs_import?trs_server=dockstore.org&trs_id=';
 
   static GITHUB_CLIENT_ID = 'will be filled in by configuration.service';
   static GITHUB_AUTH_URL = 'https://github.com/login/oauth/authorize';
@@ -89,5 +88,6 @@ export class Dockstore {
 
   static FEATURES = {
     enableCwlViewer: true,
+    enableLaunchWithGalaxy: true,
   };
 }
--- a/src/app/shared/dockstore.model.ts
+++ b/src/app/shared/dockstore.model.ts
@@ -38,6 +38,7 @@ export class Dockstore {
   static BD_CATALYST_TERRA_IMPORT_URL = 'https://terra.biodatacatalyst.nhlbi.nih.gov/#import-tool/dockstore';
   static CAVATICA_IMPORT_URL = 'https://cavatica.sbgenomics.com/integration/trs/import';
   static NEXTFLOW_TOWER_IMPORT_URL = 'https://tower.nf/launch';
+  static GALAXY_INSTANCE_IMPORT_URL = 'https://usegalaxy.org/workflows/trs_import?trs_server=dockstore.org&trs_id=';
 
   static GITHUB_CLIENT_ID = 'will be filled in by configuration.service';
   static GITHUB_AUTH_URL = 'https://github.com/login/oauth/authorize';
--- a/src/app/workflow/launch-third-party/filterCloudInstances.pipe.ts
+++ b/src/app/workflow/launch-third-party/filterCloudInstances.pipe.ts
@@ -5,7 +5,7 @@ import { CloudInstance } from '../../shared/openapi';
   name: 'filterCloudInstances',
 })
 export class FilterCloudInstancesPipe implements PipeTransform {
-  transform(cloudInstances: Array<CloudInstance>, filterFor: string): Array<CloudInstance> {
+  transform(cloudInstances: Array<CloudInstance>, filterFor: string): Array<CloudInstance> | null {
     if (cloudInstances && filterFor) {
       return cloudInstances.filter((ci: CloudInstance) => ci.partner === filterFor);
     }
--- a/src/app/workflow/launch-third-party/filterCloudInstances.pipe.ts
+++ b/src/app/workflow/launch-third-party/filterCloudInstances.pipe.ts
@@ -1,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { CloudInstance } from '../../shared/openapi';
+
+@Pipe({
+  name: 'filterCloudInstances',
+})
+export class FilterCloudInstancesPipe implements PipeTransform {
+  transform(cloudInstances: Array<CloudInstance>, filterFor: string): Array<CloudInstance> {
+    if (cloudInstances && filterFor) {
+      return cloudInstances.filter((ci: CloudInstance) => ci.partner === filterFor);
+    }
+    return null;
+  }
+}
--- a/src/app/workflow/launch-third-party/filterCloudInstances.pipe.ts
+++ b/src/app/workflow/launch-third-party/filterCloudInstances.pipe.ts
@@ -6,9 +6,10 @@ import { CloudInstance } from '../../shared/openapi';
 })
 export class FilterCloudInstancesPipe implements PipeTransform {
   transform(cloudInstances: Array<CloudInstance>, filterFor: string): Array<CloudInstance> | null {
-    if (cloudInstances && filterFor) {
-      return cloudInstances.filter((ci: CloudInstance) => ci.partner === filterFor);
+    if (!cloudInstances || !filterFor) {
+      return cloudInstances;
     }
-    return null;
+
+    return cloudInstances.filter((ci: CloudInstance) => ci.partner === filterFor);
   }
 }
--- a/src/app/workflow/launch-third-party/launch-third-party.component.html
+++ b/src/app/workflow/launch-third-party/launch-third-party.component.html
@@ -108,13 +108,16 @@
             </a>
           </div>
 
-          <div fxFlex="100%" *ngIf="workflow?.descriptorType === 'gxformat2'">
+          <div
+            fxFlex="100%"
+            *ngIf="config.FEATURES.enableLaunchWithGalaxy && workflow?.descriptorType === WorkflowModel.DescriptorTypeEnum.Gxformat2"
+          >
             <app-multi-cloud-launch
               [cloudInstances]="cloudInstances"
               [usersCloudInstances]="usersCloudInstances"
               [languagePartner]="partner.GALAXY"
               [appendToLaunchUrl]="'%23workflow/' + workflow?.full_workflow_path"
-              [user]="user"
+              [user]="user$ | async"
             ></app-multi-cloud-launch>
           </div>
 
--- a/src/app/workflow/launch-third-party/launch-third-party.component.html
+++ b/src/app/workflow/launch-third-party/launch-third-party.component.html
@@ -13,7 +13,7 @@
             </div>
             <a
               mat-raised-button
-              [matTooltip]="!(hasContent$ | async) ? 'The WDL has no content.' : ''"
+              [matTooltip]="!(hasContent$ | async) ? 'The WDL workflow has no content.' : ''"
               target="_blank"
               rel="noopener"
               [attr.href]="config.DNASTACK_IMPORT_URL + '?descriptorType=wdl&path=' + workflowPathAsQueryValue"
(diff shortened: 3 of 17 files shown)


src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.scss
```diff
@@ -0,0 +1,22 @@
+.radio-options {
```
garyluu, line 1: short explanation for each of these and why its style is different than the rest of the site

src/app/workflow/launch-third-party/launch-third-party.component.html
```diff
@@ -110,10 +115,21 @@ <h3>Launch with</h3>
               ><img src="../assets/images/thirdparty/cavatica.png" /> Cavatica
             </a>
           </div>
+
+          <div fxFlex="100%" *ngIf="workflow?.descriptorType === 'gxformat2'">
+            <app-multi-cloud-launch
+              [cloudInstances]="cloudInstances"
+              [usersCloudInstances]="usersCloudInstances"
+              [languagePartner]="partner.GALAXY"
+              [launchUrlAppend]="'%23workflow/' + workflow?.full_workflow_path"
```
garyluu, line 124: this looks like the TRS ID. calling it TRS ID may be more suitable

src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  @Input()
+  languagePartner: CloudInstance.PartnerEnum;
+
+  @Input()
+  launchUrlAppend: string;
+
+  @Input()
+  user: User;
+  _customLaunchWithOption: string;
```
garyluu, line 28: nothing particularly wrong with using underscore prefix for private variables but we haven't really be using that convention (same with most code on the Angular site)

src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (localStorage.getItem('useCustomLaunch') === 'true') {
+      this._launchWithOption = 'other';
+    }
+
+    switch (this.languagePartner) {
+      case 'GALAXY':
+        this.partner = 'Galaxy';
+        break;
+      case 'DNA_STACK':
+        this.partner = 'DNAstack';
```
garyluu, line 64: is there an enum for these?

garyluu (commented):
Interesting hacky solution with mat-menu. I would've assumed a dialog with parent element's position passed in would be easier.

src/app/workflow/launch-third-party/launch-third-party.component.ts
```diff
@@ -199,6 +209,17 @@ export class LaunchThirdPartyComponent extends Base implements OnChanges, OnInit
   }
 
   ngOnInit(): void {
+    this.userQuery.user$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((user) => (this.user = user));
+    this.cloudInstanceService.getCloudInstances().subscribe((cloudInstances: Array<CloudInstance>) => {
+      this.cloudInstances = cloudInstances;
+    });
+    // Uncomment when we want to get the users custom cloud instances
+    // if (this.user) {
+    //   this.usersService.getUserCloudInstances(this.user.id).subscribe((cloudInstances: Array<CloudInstance>) => {
```
garyluu, line 218: this is unlikely to work if the user object is slow to get

src/app/workflow/launch-third-party/filterCloudInstances.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { CloudInstance } from '../../shared/openapi';
+
+@Pipe({
+  name: 'filterCloudInstances',
+})
+export class FilterCloudInstancesPipe implements PipeTransform {
+  transform(cloudInstances: Array<CloudInstance>, filterFor: string): Array<CloudInstance> | null {
+    if (cloudInstances && filterFor) {
```
coverbeck, line 9: Is it possible for cloudInstances to exist and filterFor to not be set? In that case, wouldn't you want to return `cloudInstances`? Right now it would return null.

src/app/workflow/launch-third-party/launch-third-party.component.html
```diff
@@ -110,10 +115,21 @@ <h3>Launch with</h3>
...
+              [user]="user"
+            ></app-multi-cloud-launch>
+          </div>
+
           <div fxFlex="100%" *ngIf="workflow?.descriptorType === 'NFL'">
             <a
               mat-raised-button
-              [matTooltip]="!(hasContent$ | async) ? 'The WDL has no content.' : 'Export this workflow to Nextflow Tower'"
+              [matTooltip]="!(hasContent$ | async) ? 'The Nextflow has no content.' : 'Export this workflow to Nextflow Tower'"
```
coverbeck, line 132: `The Nextflow` sounds odd to me. I'd say "The Nextflow workflow" or "The workflow".

src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  set customLaunchWithOption(value: string) {
+    this._customLaunchWithOption = value;
+    this.updateLaunchWithUrl();
+  }
+
+  selectDefault(): void {
+    localStorage.setItem('defaultLaunchWith', this.launchWith.url);
+    this.defaultLaunchWith = localStorage.getItem('defaultLaunchWith');
+    console.log('Custom Launch Option: ' + this._customLaunchWithOption);
```
coverbeck, line 91: I recommend removing console.logs

src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.html
```diff
@@ -0,0 +1,59 @@
...
+  <div class="mat-menu-content-formatting" fxLayout="row" fxLayoutAlign="end end">
+    <button mat-button class="launch-buttons" color="accent" (click)="closeMatMenu()">Cancel</button>
+    <a
+      mat-raised-button
+      class="launch-buttons"
+      target="_blank"
+      rel="noopener"
+      [attr.href]="launchWith.url + launchUrlAppend"
+      color="accent"
+      [disabled]="this._launchWithOption === 'other' && (_customLaunchWithOption == null || _customLaunchWithOption === '')"
```
coverbeck, line 53: In JavaScript, you can just do `!_customLanuchWithOption` -- catches null, undefined, and empty string. Not sure that it's a best practice though in TypeScript.

src/app/workflow/launch-third-party/launch-third-party.component.html
NatalieEO: I figured it wouldn't always look like the TRS id so I gave it a more generic name. Renamed it a bit though

src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.scss
NatalieEO: removed a lot these instead

src/app/workflow/launch-third-party/launch-third-party.component.html
```diff
@@ -110,10 +115,21 @@ <h3>Launch with</h3>
               ><img src="../assets/images/thirdparty/cavatica.png" /> Cavatica
             </a>
           </div>
+
+          <div fxFlex="100%" *ngIf="workflow?.descriptorType === 'gxformat2'">
```
garyluu, line 119: here too

src/app/workflow/launch-third-party/launch-third-party.component.ts
```diff
@@ -199,6 +210,17 @@ export class LaunchThirdPartyComponent extends Base implements OnChanges, OnInit
   }
 
   ngOnInit(): void {
+    this.userQuery.user$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((user) => (this.user = user));
```
garyluu, line 213: just async pipe this instead

src/styles.scss
```diff
@@ -1466,3 +1466,8 @@ hr {
   height: 50px;
   object-fit: scale-down;
 }
+
+// material sometimes sets the font-weight to be bold (radio buttons for example) when we want it to be normal.
```
garyluu, line 1470: if we want it to be normal, then it can be applied globally to all radio button groups or w/e instead of manually adding the class to each material component.

src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.scss
```diff
@@ -0,0 +1,4 @@
+// Since this appears a small dropdown, font-size is decreased. Better matches Kim's mock
```
garyluu, line 1: I think you a word

src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.scss
```diff
@@ -0,0 +1,4 @@
...
+.mat-form-field {
+  font-size: 12px;
```
garyluu, line 3: unless you're absolutely sure we won't have a small dropdown, put it in the global styles.scss and add an additional class identifier for it (so that all mat-form-fields with the "small" class for example would have its font size changed). hopefully we would remember to use it

denis-yuen (approved):

src/app/shared/dockstore.model.ts
```diff
@@ -38,6 +38,7 @@ export class Dockstore {
   static BD_CATALYST_TERRA_IMPORT_URL = 'https://terra.biodatacatalyst.nhlbi.nih.gov/#import-tool/dockstore';
   static CAVATICA_IMPORT_URL = 'https://cavatica.sbgenomics.com/integration/trs/import';
   static NEXTFLOW_TOWER_IMPORT_URL = 'https://tower.nf/launch';
+  static GALAXY_INSTANCE_IMPORT_URL = 'https://usegalaxy.org/workflows/trs_import?trs_server=dockstore.org&trs_id=';
```
denis-yuen, line 41: Not sure why comment was dropped. 
We may need a on/off switch, particularly for DNAstack but also for this depending on https://github.com/dockstore/dockstore/issues/3764#issuecomment-833821786 (is it not working for just versions or overall?)

garyluu (approved):

src/app/workflow/launch-third-party/launch-third-party.component.html
```diff
@@ -107,10 +114,24 @@
...
+
+          <div
+            fxFlex="100%"
+            *ngIf="config.FEATURES.enableLaunchWithGalaxy && workflow?.descriptorType === WorkflowModel.DescriptorTypeEnum.Gxformat2"
+          >
+            <app-multi-cloud-launch
+              [cloudInstances]="cloudInstances"
+              [usersCloudInstances]="usersCloudInstances"
+              [languagePartner]="partner.GALAXY"
+              [appendToLaunchUrl]="'%23workflow/' + workflow?.full_workflow_path"
```
coverbeck, line 126: Doesn't the version need to be set too? I know it's not working on their side yet, but we should have it working correctly on our side.

coverbeck (commented):
How does the trs_server get set? Is that by invoking the cloudinstances API?

src/app/workflow/launch-third-party/launch-third-party.component.html
NatalieEO: I didn't include the version because it's not working on their end, but I thought it'd be okay for us to not use it for now. The user can select the version once they are on the galaxy side of things.

coverbeck: I feel like it's worth doing. It seems like it might not be trivial (you can't just append) and worth figuring out while you're here rather than a hotfix. Presumably DNAstack will have a similar pattern in any case?

NatalieEO: trs_server is set by putting it as the url field in the cloud instance endpoint. Mine is set to `https://usegalaxy.org/workflows/trs_import?trs_server=dockstore.org&trs_id=`

The Galaxy people gave this as an example of launch with working, although it doesn't https://workflowhub.eu/workflows/110, but it shows the version as just getting appended. I didn't append the version here because if you launch with that, a little error message pops up on the Galaxy side that the version doesn't exist, but you can just select the version after that. I'm not sure where they're at with this. I asked on Gitter today, but haven't received a response yet.

I thought we could have this enabled without the version/error message. But if we only want it enabled with the version working, then I can add it. Or if the small error message the user would see is fine, then I can add it anyway.

coverbeck: I thought when I was playing with it there was no error, but yes, I see it, so best to not have it.

coverbeck (approved):

PR #1236, opened 2021-05-18
Add frontend pagination

garyluu:
For SEAB-2494

Frontend pagination only. Kind of weird numbers I've chosen (9, 18, 36).

https://user-images.githubusercontent.com/24548904/118706504-6bde6f00-b7e7-11eb-95fc-88294fb864b3.mp4

Also feels like I'm missing something

Unrelated: at the end of the video, i show myself clicking in the text field. Does it look weird like that for anyone?

### DIFF
--- a/src/app/organizations/organizations/organizations.component.html
+++ b/src/app/organizations/organizations/organizations.component.html
@@ -20,7 +20,7 @@
   </div>
 </app-header>
 <div *ngIf="!(loading$ | async)" class="container mb-5">
-  <div *ngIf="filteredOrganizations$ | async as organizations; else noOrganizationsFound">
+  <div *ngIf="pagedOrganizations as organizations; else noOrganizationsFound">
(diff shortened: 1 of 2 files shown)


denis-yuen:
> Does it look weird like that for anyone? 

I think if its the quick highlight/not highllight it doesn't occur for me. But not sure if that's what you meant

coverbeck (approved):
Those numbers are indeed strange. I do suggest the 10, 25, 50 that we use on search.

garyluu:
> Those numbers are indeed strange. I do suggest the 10, 25, 50 that we use on search.

It's because we mostly display the organizations in columns of three whereas search is a table with 1 result per row

garyluu:
> > Does it look weird like that for anyone?
> 
> I think if its the quick highlight/not highllight it doesn't occur for me. But not sure if that's what you meant

The left side of the highlight is momentarily thinner than the rest

denis-yuen:
Divisible by three makes sense to me. No opinion on exact numbers.

> The left side of the highlight is momentarily thinner than the rest

Oh, then yes. It does.

denis-yuen (approved):

NatalieEO (approved):

PR #1238, opened 2021-05-19
Adding register button 

e34lin:
The register button was implemented as per mock up but realized that the register button on the nav bar is on the left whereas on the page it is on the right, should the register button be on the right instead for consistency?

Probably: https://ucsc-cgl.atlassian.net/browse/SEAB-2844

### DIFF
--- a/src/app/login/login.component.ts
+++ b/src/app/login/login.component.ts
@@ -32,6 +32,7 @@ import { LoginService } from './login.service';
 export class LoginComponent {
   faGithub = faGithub;
   faGoogle = faGoogle;
+  public tabindex: number;
   constructor(
     private trackLoginService: TrackLoginService,
     private loginService: LoginService,
@@ -41,6 +42,11 @@ export class LoginComponent {
     iconRegistry: MatIconRegistry,
     sanitizer: DomSanitizer
   ) {
+    if (this.router.url.includes('register')) {
+      this.tabindex = 1;
+    } else {
+      this.tabindex = 0;
+    }
     iconRegistry.addSvgIcon('google', sanitizer.bypassSecurityTrustResourceUrl('../assets/svg/btn_google_light_normal_ios.svg'));
   }
 
--- a/src/app/navbar/navbar.component.html
+++ b/src/app/navbar/navbar.component.html
@@ -41,9 +41,14 @@
           <img class="site-icons-small hidden-sm" src="../assets/svg/main-nav/icons-navigation-forum.svg" alt="stacked pages" /> Forum
         </a>
         <span class="spacer"></span>
-        <a *ngIf="!isLoggedIn" routerLink="/login">
-          <button mat-raised-button color="accent">Login</button>
-        </a>
+        <div fxLayoutGap="1rem" *ngIf="!isLoggedIn">
+          <a routerLink="/login">
+            <button mat-raised-button color="accent">Login</button>
+          </a>
+          <a routerLink="/register">
+            <button mat-raised-button color="accent">Register</button>
+          </a>
+        </div>
         <template [ngTemplateOutlet]="loginButton"></template>
         <mat-menu #menuLarge="matMenu">
           <button
--- a/src/app/app.routing.ts
+++ b/src/app/app.routing.ts
@@ -106,6 +106,7 @@ const APP_ROUTES: Routes = [
     loadChildren: () => import('app/search/search.module').then((m) => m.SearchModule),
     data: { title: 'Dockstore | Search' },
   },
+  { path: 'register', component: LoginComponent, data: { title: 'Dockstore | Register' } },
   { path: 'login', component: LoginComponent, data: { title: 'Dockstore | Login' } },
   { path: 'logout', component: LogoutComponent, data: { title: 'Dockstore | Logout' } },
   { path: 'session-expired', component: SessionExpiredComponent, data: { title: 'Dockstore | Session Expired' } },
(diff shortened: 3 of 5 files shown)


denis-yuen:
FYI @kcullion for the left-right question

denis-yuen:
@e34lin actually, the build test failure seems legitimate and we could probably use a screenshot to prompt the discussion

e34lin:
Screen Shot 2021-05-19 at 4 27 37 PM

src/app/login/login.component.ts
```diff
@@ -32,6 +32,7 @@ import { LoginService } from './login.service';
 export class LoginComponent {
   faGithub = faGithub;
   faGoogle = faGoogle;
+  public state: any;
```
coverbeck, line 35: Avoid `any` if possible.

src/app/login/login.component.ts
```diff
@@ -41,6 +42,13 @@ export class LoginComponent {
     iconRegistry: MatIconRegistry,
     sanitizer: DomSanitizer
   ) {
+    if (this.router.url.includes('register')) {
+      this.state = 1;
+    } else {
+      this.state = 0;
+    }
+
+    this.state = window.history.state.data;
```
coverbeck, line 51: Won't this clobber the value you set in line 46 or 48? It seems like you don't need lines 45-49 if you have this line. Or you don't need this and you should keep lines 45-49.

src/app/navbar/navbar.component.html
```diff
@@ -41,10 +41,16 @@
           <img class="site-icons-small hidden-sm" src="../assets/svg/main-nav/icons-navigation-forum.svg" alt="stacked pages" /> Forum
         </a>
         <span class="spacer"></span>
-        <a *ngIf="!isLoggedIn" routerLink="/login">
-          <button mat-raised-button color="accent">Login</button>
-        </a>
-        <template [ngTemplateOutlet]="loginButton"></template>
+        <div fxLayoutGap="1rem">
+          <a *ngIf="!isLoggedIn" routerLink="/register" [state]="{ data: 1 }">
```
coverbeck, line 45: Why do you need to pass `state` here and on line 49? Can't you just go off like the url, like you are already doing, in login.component.ts?

src/app/navbar/navbar.component.html
```diff
@@ -41,10 +41,16 @@
...
+            <button mat-raised-button color="accent">Register</button>
+          </a>
+          <template [ngTemplateOutlet]="loginButton"></template>
+          <a *ngIf="!isLoggedIn" routerLink="/login" [state]="{ data: 0 }">
+            <button mat-raised-button color="accent">Login</button>
+          </a>
+          <template [ngTemplateOutlet]="loginButton"></template>
```
coverbeck, line 52: Do you need this twice? It's also on line 48.

coverbeck (commented):
I did not try running this, so my comments may have been different if I had, but I don't think you need to pass state around.

src/app/login/login.component.ts
e34lin: Yes, thank you for catching that. I was testing line 51 beforehand and forgot to delete it.

src/app/navbar/navbar.component.html
e34lin: Thank you, just tried it and I don't need to pass state. I think I initially had this because I wasn't going to make a new URL for register.

src/app/navbar/navbar.component.html
e34lin: Thank you for catching this as well, I don't need it twice.

src/app/login/login.component.ts
```diff
@@ -32,6 +33,7 @@ import { LoginService } from './login.service';
 export class LoginComponent {
   faGithub = faGithub;
   faGoogle = faGoogle;
+  public state: NumberInput;
```
garyluu, line 36: https://material.angular.io/components/tabs/api selectedInex says the type is `number`, just use that. Not familiar with cdk/coercion.

src/app/navbar/navbar.component.html
```diff
@@ -41,10 +41,15 @@
           <img class="site-icons-small hidden-sm" src="../assets/svg/main-nav/icons-navigation-forum.svg" alt="stacked pages" /> Forum
         </a>
         <span class="spacer"></span>
-        <a *ngIf="!isLoggedIn" routerLink="/login">
-          <button mat-raised-button color="accent">Login</button>
-        </a>
-        <template [ngTemplateOutlet]="loginButton"></template>
+        <div fxLayoutGap="1rem">
```
garyluu, line 44: probably just put the ngIf here instead

kcullion:
> FYI @kcullion for the left-right question

So I typically see login/register like this with the form and then the other option for new users to follow:

image

I couldn't find an example that was separated by tabs. But I would say, if you keep the tabs then lead with the priority task on the website. I would assume that's "Login" right @denis-yuen ?

In that case you can reorder the buttons in the header to Login | Register if you would like to keep this order consistent.

denis-yuen:
Thanks @kcullion 

@e34lin let's re-order the buttons in the header for now to lead with login to wrap this up for 1.11, please also create a ticket to remove the tabs for 1.12

garyluu (approved):

e34lin:
> Thanks @kcullion
> 
> @e34lin let's re-order the buttons in the header for now to lead with login to wrap this up for 1.11, please also create a ticket to remove the tabs for 1.12

Just made the change for the navbar

NatalieEO (approved):

coverbeck (approved):

PR #1239, opened 2021-05-19
feature/3764/UI fixes for Galaxy launch with

NatalieEO:
This PR:
fix bug that saves the default
doesn't require users know the pattern and adds it for them. Which also means we should not save the pattern to the db too. The db and users just need `https://usegalaxy.org`
Aligns the UI more with Kim's mocks, but the color of the label also just matches the theme instead of whatever color is in her mock.
mock: https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60219b781aeb7a60c283c2d1
mine:

https://user-images.githubusercontent.com/16004905/119037298-a4f11d80-b966-11eb-9921-8bd741d5ddfa.mp4

Hopefully close enough


garyluu (approved):

kathy-t (approved):

denis-yuen (commented):
> Aligns the UI more with Kim's mocks, but I couldn't figure out a way to get the form field to be outlined and display label text above. The color of the label also just matches the theme instead of whatever color is in her mock.

Think I can live with this being spun out into a different ticket 

The icon for the Galaxy button itself is missing too

src/styles.scss
```diff
@@ -1474,3 +1474,31 @@ hr {
 .mat-form-field.small-text {
   font-size: 12px;
 }
+
+// Style the float label of the mat-form-field
+.mat-form-field-appearance-legacy.mat-form-field-can-float.mat-form-field-should-float
```
coverbeck, line 1479: Are these all going to change any other forms elsewhere? Maybe that's the intent if so, but I just want to make sure that other UI isn't being inadvertently changed.

coverbeck (approved):

NatalieEO:
> > Aligns the UI more with Kim's mocks, but I couldn't figure out a way to get the form field to be outlined and display label text above. The color of the label also just matches the theme instead of whatever color is in her mock.
> 
> Think I can live with this being spun out into a different ticket
> 
> The icon for the Galaxy button itself is missing too

Weird I totally thought I updated my description and video before I requested review from people again. I must have forgot to hit Update. But the icon for galaxy is still missing yeah

src/styles.scss
NatalieEO: Yeah the intention was to change the color of all the float labels to keep it consistent.

src/app/shared/dockstore.model.ts
```diff
@@ -21,7 +21,9 @@ import { Provider } from './enum/provider.enum';
 
 export class Dockstore {
   static readonly HOSTNAME = window.location.protocol + '//' + window.location.host;
+  static readonly HOSTNAME_NO_PROTOCOL = window.location.hostname;
   static readonly API_URI = Dockstore.HOSTNAME + '/api';
+  static readonly GALAXY_IMPORT_URL = '/workflows/trs_import?trs_server=' + Dockstore.HOSTNAME_NO_PROTOCOL + '&trs_id=%23workflow/';
```
coverbeck, line 26: It's technically not a URL, but the path portion of a URL.

coverbeck (approved):

denis-yuen (approved):

garyluu (approved):

PR #1240, opened 2021-05-21
Feature/2847/odd logo behavior

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-2847

Note: Can reject this PR if we don't want to do anything for this at the moment

The odd logo behavior is caused by out-of-date organization data and urls that no longer exist. 

The broken links are broken in both dev and prod:
https://dev.dockstore.net/organizations/DNAstack -> https://dockstore.org/organizations/DNAstack
https://dev.dockstore.net/organizations/MGI -> https://dockstore.org/organizations/MGI

When run locally using a modern database, logos appear as expected:
**Missing in dev**: https://dev.dockstore.net/organizations/OICR
**When run locally**: Screen Shot 2021-05-21 at 12 18 43 PM

To avoid the broken link icon case I have a few suggestions (although we should still reach out to the organizations for a working link): 
1. The current PR hides the image element if there is an error.
2. We could simply set `alt=''`, this will display no corrupt image icon in case of failed image load (at the expense of the alt text)
3. Leave as-is. This is technically 'correct' behavior, so it seems reasonable to not change anything and live with the corrupt image icons. (Might be worth discussing different approaches to storing images rather than using remote urls).

**Options 1 and 2 will leave us with this on a failed image load:**
Screen Shot 2021-05-21 at 11 06 56 AM
**instead of**
Screen Shot 2021-05-21 at 11 06 34 AM

### DIFF
--- a/src/app/loginComponents/auth/auth.component.ts
+++ b/src/app/loginComponents/auth/auth.component.ts
@@ -95,7 +95,6 @@ export class AuthComponent extends Base implements OnInit {
           if (provider !== Provider.ORCID) {
             this.router.navigate([`${prevPage}`]);
           }
-          this.router.navigate([`${prevPage}`]);
           if (error.status === 409) {
             this.alertService.detailedSnackBarErrorWithLink(
               error,
--- a/src/app/organizations/organization/organization.component.html
+++ b/src/app/organizations/organization/organization.component.html
@@ -58,7 +58,7 @@
       <div fxLayout="row" fxLayout.lt-md="column" fxLayoutAlign="space-between">
         <div fxFlex="90" fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="2rem">
           <div *ngIf="gravatarUrl$ | async">
-            <img [src]="gravatarUrl$ | async" class="big-org-logo" alt="avatar" />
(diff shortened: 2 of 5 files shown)


denis-yuen (commented):
> Leave as-is. This is technically 'correct' behavior, so it seems reasonable to not change anything and live with the corrupt image icons. (Might be worth discussing different approaches to storing images rather than using remote urls). 

Think I'm ok with this to make it more obvious to curators that an image was specified and something went wrong rather than hiding the broken image icon. 
A follow-up ticket to store images rather than using remote urls is probably wise.

garyluu (approved):

coverbeck:
> A follow-up ticket to store images rather than using remote urls is probably wise.

Torn on this one; on the one hand it doesn't break, on the other hand it doesn't stay up to date. :)

coverbeck (approved):
I'm good this way or what Denis suggests. FWIW, BD Cat doesn't have a logo that they don't want used from outside BD Cat and had to set to an empty image to avoid the broken image link, but that solution works, so it's acceptable.

Richard-Hansen:
Since there isn't any strong pushes for this change I'm going to close the PR.

Created a follow-up ticket for the image storage suggestion: https://github.com/dockstore/dockstore/issues/4262

PR #1241, opened 2021-05-24
Save user's custom launch to db

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-2971

Attempts to create and save a user's custom launch url and ignore if it fails so the user can continue with launch.

Doesn't save any custom launch urls if the user isn't logged in.

### DIFF
--- a/src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.ts
+++ b/src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.ts
@@ -1,6 +1,7 @@
 import { HttpErrorResponse } from '@angular/common/http';
 import { Component, EventEmitter, Input, OnInit, Output, ViewChild } from '@angular/core';
 import { MatMenuTrigger } from '@angular/material/menu';
+import { AlertService } from '../../../shared/alert/state/alert.service';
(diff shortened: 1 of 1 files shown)


denis-yuen (approved):

src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.ts
```diff
@@ -116,27 +117,29 @@ export class MultiCloudLaunchComponent extends Base implements OnInit {
...
-    //       this.usersCloudInstances = usersCloudInstances;
-    //     },
-    //     (error: HttpErrorResponse) => {
-    //       console.log(error.message);
-    //     }
-    //   );
-    // }
+    // Create and save a user's custom launch entry to the db
+    if (this.presetLaunchWithOption === 'other' && this.user) {
+      const url: URL = new URL(this.launchWith.url);
```
coverbeck, line 122: This will throw an exception if the URL is invalid, e.g., `usegalaxy.org` instead of `https://usegalaxy.org`, but based on your comments below, it doesn't seem like it matters at this point.

coverbeck (approved):

coverbeck (approved):

PR #1243, opened 2021-05-25
Implementing X icon in search components

e34lin:
Added an X icon for the search components.

Reference: https://docs.google.com/document/d/1GLshBv7Ue09ceCFDSLmFvrgbEofGkY6R0jXlKxVjqHc/edit number 13 under organization


coverbeck:
Screenshots would be helpful to get a sense of what it looks like.

coverbeck (commented):
I think the X should be vertically centered. Or maybe bottom-aligned. Right now it's top-aligned and looks off to me (Firefox, Mac OSx -- see screenshot), but I'll defer to others.

It's hard to tell in the doc how it should be.


Screen Shot 2021-05-25 at 4 22 50 PM

garyluu:
> I think the X should be vertically centered. Or maybe bottom-aligned. Right now it's top-aligned and looks off to me (Firefox, Mac OSx -- see screenshot), but I'll defer to others.
> 
> It's hard to tell in the doc how it should be.
> 
> Screen Shot 2021-05-25 at 4 22 50 PM

Should probably look like the first example: https://material.angular.io/components/input/examples

src/app/home-page/home-logged-out/home.component.html
```diff
@@ -54,10 +54,14 @@ <h2>An app store for bioinformatics</h2>
             aria-label="Search"
             #search
             type="text"
+            [(ngModel)]="value"
```
garyluu, line 57: (keyup.enter) uses search.value. this addition uses value. pick one and use it for both features

src/app/search/basic-search/basic-search.component.scss
```diff
@@ -7,3 +7,8 @@ textarea:focus {
 .form-field {
   display: inherit;
 }
+
+.search-position {
+  vertical-align: middle;
```
garyluu, line 12: need explanation on why this search bar has different styling than everything else on the site

e34lin:
> Screenshots would be helpful to get a sense of what it looks like.

Screen Shot 2021-05-25 at 9 16 35 PM
Screen Shot 2021-05-25 at 9 16 44 PM
Screen Shot 2021-05-25 at 9 20 02 PM

e34lin:
> I think the X should be vertically centered. Or maybe bottom-aligned. Right now it's top-aligned and looks off to me (Firefox, Mac OSx -- see screenshot), but I'll defer to others.
> 
> It's hard to tell in the doc how it should be.
> 
> Screen Shot 2021-05-25 at 4 22 50 PM

It looks vertically aligned on my localhost, but let me take another look at it, thank you!

src/app/home-page/home-logged-out/home.component.html
e34lin: okay, didn't catch that (keyup.enter) uses search.value, will make the change thank you!

src/app/search/basic-search/basic-search.component.scss
e34lin: The search bar previously had a class of "w-100 mb-3" so I added the margin bottom in the styling and the alignment for the button.

garyluu: so i think the implementation is going backwards. .form-field should be removed, this .search-position should be removed. the w-100 and mb-3 should be removed. And a w-100 should be added to the form field.  and mat-label should be added.

This should result in a smaller form field which is more inline with the mockups and this also has the advantage of having little to no css.

e34lin:
Screen Shot 2021-05-28 at 3 28 27 PM
Screen Shot 2021-05-28 at 3 30 15 PM

src/app/home-page/home-logged-out/home.component.html
```diff
@@ -54,10 +54,14 @@ <h2>An app store for bioinformatics</h2>
             aria-label="Search"
             #search
             type="text"
+            [(ngModel)]="search.value"
```
garyluu, line 57: not entirely sure this is needed.

src/app/home-page/home-logged-out/home.component.scss
```diff
@@ -111,6 +111,11 @@ h3 {
   border-radius: 1rem;
 }
 
+.close-alignment {
```
garyluu, line 114: try to put the button without the close-alignment classes in any component. https://stackblitz.com/angular/nlkqgebarxk?file=src%2Fapp%2Finput-clearable-example.ts has no styling and looks close to what it should be.

src/app/search/basic-search/basic-search.component.html
```diff
@@ -1,18 +1,21 @@
...
       matInput
       [matAutocomplete]="auto"
       [formControl]="searchFormControl"
       matTooltip="Find tools and workflows"
       [matTooltipPosition]="'after'"
       list="suggestions"
-      class="w-100 mb-3"
       placeholder="Enter search term"
       type="text"
+      [(ngModel)]="value"
```
garyluu, line 13: i prefer the other method without ngModel. ngModel is for two-way binding which isn't needed here

src/styles.scss
```diff
@@ -1475,6 +1476,23 @@ hr {
...
+  max-width: 200px;
+}
+
+.search-form-field input {
+  padding-right: 35px;
+  transform: translateY(-3px);
+}
+
+.search-form-field button {
+  position: absolute;
```
garyluu, line 1490: it's likely the absolute position is making things worse

src/styles.scss
```diff
@@ -1475,6 +1476,23 @@ hr {
   font-size: 12px;
 }
 
+
+.search-form-field {
+  max-width: 200px;
+}
+
+.search-form-field input {
+  padding-right: 35px;
```
garyluu, line 1485: if you're finding that the x is disappearing without the padding-right, it's likely that the width of the search-form-field is wrong

garyluu (commented):
In general, try to use less styling.

src/app/home-page/home-logged-out/home.component.scss
e34lin: This is how it looks like when I don't do any additional styling for the button
Screen Shot 2021-05-28 at 4 52 15 PM

src/styles.scss
e34lin: I added the padding so when a long input is passed it doesn't cover the x icon

Without padding:
Screen Shot 2021-05-28 at 4 55 37 PM

src/styles.scss
e34lin: I added the absolute position because I found on firefox and chrome the x icons were showing up in different places

src/app/home-page/home-logged-out/home.component.html
e34lin: If I delete this line the x icon only shows up the first time I enter something in the input

src/app/search/basic-search/basic-search.component.html
e34lin: Sorry, which other method are you referring to?

src/styles.scss
garyluu: With 0 styling: 
image

src/app/home-page/home-logged-out/home.component.scss
garyluu: figure out what styling is messing with it and remove that style instead of adding more styles

src/app/home-page/home-logged-out/home.component.html
garyluu: odd...but anyways, assuming you always want the x to show up (i prefer not), try removing this and the *ngIf on the button then

e34lin: okay, thank you I'll try that out. The x doesn't always show, just only when you enter something into the search

src/app/home-page/home-logged-out/home.component.scss
e34lin: okay, thank you. will try to figure it out

src/app/home-page/home-logged-out/home.component.scss
```diff
@@ -118,3 +118,10 @@ h3 {
 .warn-1 {
   color: mat-color($dockstore-app-warn, darker);
 }
+
+// Style home search bar
```
garyluu, line 122: if you can't figure out how to not have this still, put a TODO along with your best guess on what the problem is

garyluu (approved):

denis-yuen (changes requested):
Heads-up, https://github.com/dockstore/dockstore-ui2/pull/1243/commits/50ee7159f14872d94f6449df98c59af842a6bf2a looks like it is taking in the angular 11 changes from develop.

Separately, let's try to get this in by end of day

Edit to add: chatting with @garyluu 
looks like the workaround without upgrading angular is `display: inline-block` with a good comment to remove
Follow-up with a PR to develop after this is merged to the release branch

coverbeck (approved):

src/styles.scss
```diff
@@ -1506,3 +1511,13 @@ hr {
...
+
+// Style home search mat-form-field
+// Make mat-form-field match home search bar, could look into more solutions
+.home-field .mat-form-field-wrapper {
+  padding-bottom: 0;
+}
+.home-field .mat-form-field-infix {
+  padding: 0.8em 0 0.8em 0;
+  border-top: 0em;
+}
```
garyluu, line 1523: umm....if it's a style that's specific to a component (because it can't be generalized), then it should be in the component scss. In this case, there's some weird problems going on and it's unlikely to be occur in a future component so we won't ever use the class elsewhere.

denis-yuen (approved):
Angular 11 looks gone, thanks!

PR #1247, opened 2021-05-27
Feature/2937/run smoke tests against staging and prod

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-2937

Runs the staging no-auth smoke tests during the nightly testing. Since the auth testing user is not currently setup on staging and prod, the auth tests will need to wait until: https://ucsc-cgl.atlassian.net/browse/SEAB-1871 is finished.

To get the correct set of tests to run:
1. Scrape the current webpage for the commit id
2. Checkout the branch with the commit id
3. Run smoke tests
(Note: This PR also changes the dev nightly tests to use the above strategy)

There is currently one smoke test that is known to be broken: https://ucsc-cgl.atlassian.net/browse/SEAB-2071. 

I've removed the breaking input so that the tests wont fail in the upcoming release (they will until then, though). 
We can see the smoke test failing (correctly) in staging: https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/5775/workflows/691f3f94-8109-4c8c-af04-ec4709bb5be2/jobs/18891.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ jobs:
       - upload_nightly_artifacts
       - slack/status:
         fail_only: true
-        failure_message: Nightly << parameters.stack >> non-auth tests failed!
+        failure_message: "Nightly << parameters.stack >> non-auth tests failed!"
         webhook: $SLACK_WEBHOOK
 
   # TODO: Add auth tests for staging/prod once https://ucsc-cgl.atlassian.net/browse/SEAB-2071 is done.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,8 +27,7 @@ jobs:
       - run:
           name: Get commit id.
           command: |
-            git checkout
-            $(curl https://dev.dockstore.net | \
+            git checkout $(curl https://dev.dockstore.net | \
             grep -Eoi "https:\/\/gui\.dockstore\.org\/[A-Za-z0-9]+-[A-Za-z0-9]+" | \
             head -1 | \
             grep -Eoi "[A-Za-z0-9]+$")
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -305,6 +305,8 @@ workflows:
                 - develop
     jobs:
       - uptime_monitor_no_auth:
+          stack: "dev"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -21,49 +21,62 @@ executors:
...
   uptime_monitor_no_auth:
+    parameters:
+      stack:
+        type: string
+      webpage_URL:
+        type: string
     working_directory: ~/repo
     docker:
       - image: circleci/buildpack-deps:18.04-browsers
-        auth:
```
Richard-Hansen, line 29: I removed this auth information as I don't think it impacts the no-auth tests. Unless I'm missing something...?

coverbeck: You should keep this. This is using DockerHub credentials to fetch the Docker image. This will decrease the likelihood of us running into a DockerHub rate limiting issue in the future (authorized users have a higher rate limit). I believe CircleCI has some sort of arrangement with DockerHub for now so that pulls from CircleCI don't run into the rate limit, but that may not always be the case.

garyluu (approved):

coverbeck (approved):
There is a filter against the develop branch for the nightly tests, but you then checkout specific commits.

Just to confirm, the purpose of the filter is to ensure that the nightly tests will only be run once, because there is only one branch named `develop`? I don't see another way around it, but just wanted to make sure that's what's going on.

Richard-Hansen:
> There is a filter against the develop branch for the nightly tests, but you then checkout specific commits.
> 
> Just to confirm, the purpose of the filter is to ensure that the nightly tests will only be run once, because there is only one branch named `develop`? I don't see another way around it, but just wanted to make sure that's what's going on.

Correct. Without the filter the nightly cron job would run against every branch ([CircleCI docs](https://circleci.com/docs/2.0/configuration-reference/#branches)). So we specify to only run the `develop` branch nightly, and then pick specific commits for smoke testing.

denis-yuen (approved):

Richard-Hansen:
I'm planning on leaving this PR here until the updated smoke test from https://github.com/dockstore/dockstore-ui2/pull/1251 make it into the next staging/prod release. This will avoid the headache of nightly failures for both stacks...

PR #1258, opened 2021-06-03
Point at stable

denis-yuen:
Follow-up on SEAB-2737
Also probably need to mess with Dockstore.DOCUMENTATION_URL on deploy

### DIFF
--- a/src/app/about/about.component.html
+++ b/src/app/about/about.component.html
@@ -42,7 +42,7 @@
         <a
           target="_blank"
           rel="noopener noreferrer"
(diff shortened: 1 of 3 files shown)


coverbeck (approved):
Ideally these would be using [DOCUMENTATION_URL](https://github.com/dockstore/dockstore-ui2/blob/develop/src/app/shared/dockstore.model.ts#L86), but that's a bit of extra work and it's unlikely we'll change it again.

garyluu (approved):

PR #1255, opened 2021-06-02
Feature/2855/automate smoke tests

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-2855
Smoke test doc: https://docs.google.com/document/d/1CShy6nHne-defoJlA4jXIBJG-TG4g07lZcOZn7t7lU4/edit

I covered a majority of the manual smoke tests, there were a few I'm leaving behind for the moment (services/discourse) as they require a little more tinkering to get right.

There were a lot of minor changes that are hard to keep track of, so here are some of the important ones:
1. Started with some stochastic testing, [here](https://github.com/dockstore/dockstore-ui2/pull/1255/files#diff-6ff190303112ed6cdf5b4b1af7dfb12b7c369ca75e13f23f7512875060024e15R6), to pick a random tool/workflow from the search page and check the info/files/versions tabs. (includes testing the [download as zip](https://github.com/dockstore/dockstore-ui2/pull/1255/files#diff-6ff190303112ed6cdf5b4b1af7dfb12b7c369ca75e13f23f7512875060024e15R34)).
[Sonarcloud](https://sonarcloud.io/project/security_hotspots?id=dockstore_dockstore-ui2&pullRequest=1255&resolved=false&types=SECURITY_HOTSPOT) isn't happy about `Math.random` to pick the index, but it isn't a security risk, so we should be able to ignore.
2. Added a new WDL workflow to test against, [here](https://github.com/dockstore/dockstore-ui2/pull/1255/files#diff-6ff190303112ed6cdf5b4b1af7dfb12b7c369ca75e13f23f7512875060024e15R58). Mainly helps test the appropriate launch-with buttons
3. DAG tests, [here](https://github.com/dockstore/dockstore-ui2/pull/1255/files#diff-6ff190303112ed6cdf5b4b1af7dfb12b7c369ca75e13f23f7512875060024e15R114).
4. Boolean facet filters, [here](https://github.com/dockstore/dockstore-ui2/pull/1255/files#diff-6ff190303112ed6cdf5b4b1af7dfb12b7c369ca75e13f23f7512875060024e15R227).
5. Nextflow Tower button check, [here](https://github.com/dockstore/dockstore-ui2/pull/1255/files#diff-6ff190303112ed6cdf5b4b1af7dfb12b7c369ca75e13f23f7512875060024e15R150).
6. Added a decent number of `data-cy=...` attributes to help with the testing.
[...]

PR #1259, opened 2021-06-03
Feature/2855/automate smoke tests rebased

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-2855
Smoke test doc: https://docs.google.com/document/d/1CShy6nHne-defoJlA4jXIBJG-TG4g07lZcOZn7t7lU4/edit

I covered a majority of the manual smoke tests, there were a few I'm leaving behind for the moment (services/discourse) as they require a little more tinkering to get right.

Passing tests against release/2.8:
Screen Shot 2021-06-03 at 2 39 33 PM

There is a draft PR that is targeted at develop, https://github.com/dockstore/dockstore-ui2/pull/1255, instead of the release/2.8 branch, if we decide that is a better route (it also has some more details on the smoke tests that were changed/added).

### DIFF
--- a/cypress/integration/manualTests/basic-enduser.ts
+++ b/cypress/integration/manualTests/basic-enduser.ts
@@ -63,7 +63,7 @@ const workflowVersionTuples = [
   ],
   ['github.com/nf-core/vipr', 'dev', 'master', '', 'NFL'],
 ];
-describe('Monitor workflows', () => {
+describe.only('Monitor workflows', () => {
   workflowVersionTuples.forEach((t) => testWorkflow(t[0], t[1], t[2], t[3], t[4]));
 });
 function testWorkflow(url: string, version1: string, version2: string, trsUrl: string, type: string) {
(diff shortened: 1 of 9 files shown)


cypress/integration/manualTests/basic-enduser.ts
```diff
@@ -38,58 +111,77 @@ function testWorkflow(url: string, version1: string, version2: string, trsUrl: s
...
-  //
-  // it('DAG tab works', () => {
-  //   goToTab('DAG');
-  //   cy.url().should('contain', '?tab=dag');
-  //   cy.get('[data-cy=dag-holder]');
-  // });
+  it('DAG tab works', () => {
+    cy.contains('.mat-tab-label', 'DAG').click();
+    cy.url().should('contain', '?tab=dag');
+    cy.get('[data-cy=dag-holder]');
```
coverbeck, line 117: What is this testing? Will it verify there actually is a DAG, or only that there is a container for the DAG?

Just asking because of the removed  code above, where it says "This tends to be flaky". I don't think DAG is flaky (but I could be wrong), I think there are versions it can visualize and versions it can't -- I've never seen  it be flaky for a particular version. 

TLDR; there should be a know version that we can always verify has a DAG (and maybe this is already doing this).

cypress/integration/manualTests/basic-enduser.ts
```diff
@@ -38,58 +111,77 @@ function testWorkflow(url: string, version1: string, version2: string, trsUrl: s
...
-      ['Terra', url + ':' + version2],
-      ['AnVIL', url + ':' + version2],
-      ['NHLBI BioData Catalyst', url + ':' + version2],
+      ['dnanexusLaunchWith', '?source=' + trsUrl],
+      ['terraLaunchWith', url + ':' + version2],
+      ['anvilLaunchWith', url + ':' + version2],
+      ['nhlbiLaunchWith', url + ':' + version2],
     ];
+  } else if (type === 'NFL') {
+    launchWithTuples = [['nextflowtowerLaunchWith', url + '&revision=' + version2]];
```
coverbeck, line 150: Would be nice if Galaxy were here too, but I don't want to introduce feature creep. If complicated, could you create a follow up ticket and/or add it the doc for manual testing.

coverbeck (approved):

garyluu (approved):

denis-yuen (approved):

cypress/integration/manualTests/basic-enduser.ts
Richard-Hansen: Currently, this just checks that the container is there (and added a check for atleast 1 child element).

I believe the flakiness came from the use of [goToTab](https://github.com/dockstore/dockstore-ui2/blob/ba23f699615f627c0897e437d6e4f74f2563669d/cypress/support/commands.ts#L17), as it checks for visibility. Depending on the tabs shown, `DAG` may or may not be visible by default, so skipping the visibility check should make it more reliable.

cypress/integration/manualTests/basic-enduser.ts
Richard-Hansen: Created a followup ticket to add smoke testing for both Galaxy and DNA Stack, https://github.com/dockstore/dockstore/issues/4278. They both have a different structure than the other buttons (and I'm not seeing any default options available when running locally... so there may be more setup on my end).

PR #1261, opened 2021-06-04
Checker link: No more social distancing

Richard-Hansen:
Random fix. 

Computed difference is `flex: 1 0 auto` -> `flex: 0 1 auto`. Prevents the container from filling as much space as possible.

**Before:**
Screen Shot 2021-06-04 at 11 41 34 AM

**After:**
Screen Shot 2021-06-04 at 11 41 06 AM

### DIFF
--- a/src/app/shared/entry/info-tab-checker-workflow-path/info-tab-checker-workflow-path.component.html
+++ b/src/app/shared/entry/info-tab-checker-workflow-path/info-tab-checker-workflow-path.component.html
@@ -1,6 +1,6 @@
 <li *ngIf="((checkerId$ | async) || (canAdd$ | async)) && !(isRefreshing$ | async)">
   <form #editCheckerWorkflowPathForm="ngForm" class="form-inline">
-    <div class="form-group" fxFlex="noshrink" fxLayout fxLayoutAlign=" center">
+    <div class="form-group" fxFlex="nogrow" fxLayout fxLayoutAlign=" center">
       <strong matTooltip="Checker workflow of this tool/workflow that tests it">Checker Workflow</strong>:
     </div>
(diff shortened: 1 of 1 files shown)


garyluu (approved):

coverbeck (approved):
Looks good, especially the PR title :). Does this fix the add case too? I couldn't tell for sure by a quick glance at the code.

Screen Shot 2021-06-04 at 9 45 41 AM

garyluu:
> Looks good, especially the PR title :). Does this fix the add case too? I couldn't tell for sure by a quick glance at the code.
> 
> Screen Shot 2021-06-04 at 9 45 41 AM

I think we need to redesign that portion. the other fields are all the way to the right too

denis-yuen (approved):

denis-yuen:
> > Looks good, especially the PR title :). Does this fix the add case too? I couldn't tell for sure by a quick glance at the code.
> > Screen Shot 2021-06-04 at 9 45 41 AM
> 
> I think we need to redesign that portion. the other fields are all the way to the right too

Hmmm, on the other hand. Its on the right in the redesign. 
Doesn't look as far though because its aligned with the footer on the bottom

Richard-Hansen:
Hmm, this may not be as straightforward as I thought.
1. This PR messes up the private page: 
Screen Shot 2021-06-04 at 2 35 43 PM

2. But I think makes the public page look more consistent, as those entries are left-aligned: 
Screen Shot 2021-06-04 at 2 36 53 PM

3. Also the mocks note this other case with 'N/A': 
Screen Shot 2021-06-04 at 2 39 54 PM

I'm thinking leave as-is for now... But we may want to revisit these alignments?

Richard-Hansen:
Closing per the above comment.

PR #1267, opened 2021-06-08
Feature/smoke test suggestions

Richard-Hansen:
In response to nightly dev/staging smoke test failures.

I'm suggesting two changes to improve smoke test reliability:
1. Match the suggested text by regex rather than a specific string, should make the test more reliable regardless of what database a Dockstore stack is using.
2. Added more retries to test runs. I noticed some smoke tests are a little flaky, so this should alleviate some false-positives for test failures.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -6,7 +6,7 @@
   "defaultCommandTimeout": 30000,
   "requestTimeout": 30000,
   "retries": {
-    "runMode": 1,
+    "runMode": 3,
     "openMode": 3
   }
 }
--- a/cypress/integration/manualTests/search.ts
+++ b/cypress/integration/manualTests/search.ts
@@ -19,10 +19,10 @@ describe('Admin UI', () => {
       cy.contains('the Language is WDL').should('not.exist');
     });
     it('should be able to use basic search box and have suggestions', () => {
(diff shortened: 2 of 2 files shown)


cypress.json
```diff
@@ -6,7 +6,7 @@
   "defaultCommandTimeout": 30000,
   "requestTimeout": 30000,
   "retries": {
-    "runMode": 1,
+    "runMode": 3,
```
coverbeck, line 9: This will affect all Cypress tests. Are we OK with that? Not sure what the options are if we don't, unless they have a way to specify the config on the command line.

coverbeck (approved):

cypress.json
Richard-Hansen: I can modify the smoke test commands to specify `retries=3` if we want. i.e. something like 
`npx cypress run -c baseUrl=https://staging.dockstore.org,retries=3 -s my_test.ts`

coverbeck: I'll defer to Gary.

garyluu: I actually thought I set it in runMode instead of open. So this is actually what I originally intended.

garyluu (approved):

denis-yuen (approved):

PR #1268, opened 2021-06-08
Feature/4229/file tree

garyluu:
For dockstore/dockstore#4229

Currently set to only appear for files > 10

Before clicking button
image

Dialog appears once clicked:

image

After file selected:

image

0 css, many TODOs, needs a follow up ticket

Anyone got any low hanging fruit fixes/changes?

### DIFF
--- a/src/app/file-tree/file-tree.component.ts
+++ b/src/app/file-tree/file-tree.component.ts
@@ -1,0 +1,94 @@
+import { Component, Inject } from '@angular/core';
+import { MatTreeFlatDataSource, MatTreeFlattener } from '@angular/material/tree';
+import { FlatTreeControl } from '@angular/cdk/tree';
+import { MAT_DIALOG_DATA } from '@angular/material/dialog';
+import { SourceFile } from 'app/shared/swagger';
+
+/** File node data with possible child nodes. */
+export interface FileNode {
+  name: string;
+  children?: FileNode[];
+}
+
+/**
+ * Flattened tree node that has been created from a FileNode through the flattener. Flattened
+ * nodes include level index and whether they can be expanded or not.
+ */
+export interface FlatTreeNode {
+  name: string;
+  level: number;
+  expandable: boolean;
+}
+
+@Component({
+  selector: 'app-file-tree',
+  templateUrl: './file-tree.component.html',
+})
+export class FileTreeComponent {
+  /** The TreeControl controls the expand/collapse state of tree nodes.  */
+  treeControl: FlatTreeControl<FlatTreeNode>;
+
+  /** The TreeFlattener is used to generate the flat list of items from hierarchical data. */
+  treeFlattener: MatTreeFlattener<FileNode, FlatTreeNode>;
+
+  /** The MatTreeFlatDataSource connects the control and flattener to provide data. */
+  dataSource: MatTreeFlatDataSource<FileNode, FlatTreeNode>;
+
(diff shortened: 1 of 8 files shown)


src/app/file-tree/file-tree.component.ts
```diff
@@ -0,0 +1,105 @@
...
+    this.matDialogRef.close(node.absolutePath);
+  }
+
+  private convertSourceFilesToTree(sourceFiles: SourceFile[]): FileNode[] {
+    // Everything is assumed to be an absolute path starting with "/", removing that slash as it's not needed for path.split
+    const paths = sourceFiles.map((sourceFile) => sourceFile.absolutePath.substring(1));
+    let result: FileNode[] = [];
+    let level = { result };
+    paths.forEach((path) => {
+      path.split('/').reduce((accumulator, filename, i, a) => {
```
coverbeck, line 95: Suggest better names for `i` and `a` -- I hadn't used reduce in a while and had to go look up what they are.

Also confused by `filename` -- it's a path segment.

src/app/file-tree/file-tree.component.ts
```diff
@@ -0,0 +1,105 @@
...
+        if (!accumulator[filename]) {
+          accumulator[filename] = { result: [] };
+          accumulator.result.push({ name: filename, children: accumulator[filename].result, absolutePath: '/' + a.join('/') });
```
coverbeck, line 98: `absolutePath: '/' + a.join('/')` => or `absolutePath: '/' + path`?

src/app/file-tree/file-tree.component.ts
```diff
@@ -0,0 +1,105 @@
...
+  /** Get the children for the node. */
+  getChildren(node: FileNode): FileNode[] | null | undefined {
+    return node.children;
+  }
+
+  selectFile(node: FileNode) {
+    this.matDialogRef.close(node.absolutePath);
+  }
+
+  private convertSourceFilesToTree(sourceFiles: SourceFile[]): FileNode[] {
```
coverbeck, line 89: Seems like a good candidate for a unit test

coverbeck (commented):
It's going to be hard to find a file whose path you don't know. And even if you do know it, may need to expand a lot of nodes.

Didn't try actually running it though -- is there already "free" search functionality in a Angular tree?

garyluu:
> It's going to be hard to find a file whose path you don't know. And even if you do know it, may need to expand a lot of nodes.
> 
> Didn't try actually running it though -- is there already "free" search functionality in a Angular tree?

Ha...no

src/app/file-tree/file-tree.component.ts
garyluu: it's actually a filename or folder name. the i and a are named that way because they weren't really used so it doesn't matter

garyluu:
Just to re-iterate: Anyone got any low hanging fruit fixes/changes? So...changes that don't take more time than the length of this PR

denis-yuen:
>  I think at that point, we're better off embedding stackblitz into dockstore 

Might be an idea for hosted workflows. Wonder how many dependencies it would be

denis-yuen (approved):
Looking good, look forward to messing with this on dev to prioritise TODOs

garyluu:
Got much less feedback on draft PR than I thought...oh well. Closing, cleaning things up before requesting review again.

PR #1270, opened 2021-06-10
Test the 1.11 rc release

denis-yuen:
Probably need to give this a shot before the final release and the swagger/openapi should be frozen now

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
   "name": "dockstore-ui2",
-  "version": "2.3.0",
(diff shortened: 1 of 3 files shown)


denis-yuen:
Don't think this gets merged according to checklist, so just a FYI

garyluu (commented):
There's also another script to modify run-webservice.sh or something like that

coverbeck (approved):

garyluu (approved):

PR #1271, opened 2021-06-10
Feature/fix deps

garyluu:
Random PR `npm ls` reports many errors....now it reports less


denis-yuen:
Curious, is it actually deleting 400 lines of dependencies? 
Or were they just previously superfluous?

garyluu:
> Curious, is it actually deleting 400 lines of dependencies?
> Or were they just previously superfluous?

Hard to say why. Maybe one dependency was using a bunch of other dependencies in an old version and then the new version uses much fewer dependencies

NatalieEO (approved):

PR #1274, opened 2021-06-11
Feature/4229/file tree

garyluu:
For dockstore/dockstore#4229

Added a file tree dialog that's shown when the number of files exceeds 10 in a workflow version
The core issue however, is not quite fixed because it's very much possible for a workflow with under 10 files to have two files with the same relative path. This could be easily resolved by switching to absolute path but that would result in a very long, unreadable file path.

Also removed the light blue border of input text focus. We don't need it if using material design. If we notice there is a place that uses it, we should convert that to material instead.

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -103,6 +103,7 @@ import { TosBannerComponent } from './tosBanner/tos-banner.component';
 import { ExporterStepComponent } from './workflow/snapshot-exporter-modal/exporter-step/exporter-step.component';
 import { SnaphotExporterModalComponent } from './workflow/snapshot-exporter-modal/snaphot-exporter-modal.component';
 import { ViewService } from './workflow/view/view.service';
+import { FileTreeComponent } from './file-tree/file-tree.component';
 
 export const myCustomTooltipDefaults: MatTooltipDefaultOptions = {
   showDelay: 500,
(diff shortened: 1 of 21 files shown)


coverbeck (commented):
Actually tried this locally, unlike the draft PR. I don't like how the dialog keeps resizing as the tree expands/collapses. Could it fixed size, or is the problem you don't know what the fixed size should be?

garyluu:
> Actually tried this locally, unlike the draft PR. I don't like how the dialog keeps resizing as the tree expands/collapses. Could it fixed size, or is the problem you don't know what the fixed size should be?

That's odd. it's supposed to be fixed size. Wonder if the search bar affected things

coverbeck:
> Edit: you were talking about vertical (not horizontal). We use variable size vertical for every modal/dialog so far. If you feel strongly about not using variable size, make an issue, discuss with Kim, and we can convert all modals (organization description, edit version, etc) to non-variable size vertical at once.

This one seems of a different magnitude than the others -- one keypress can radically change the size. But I will log separately.

Jun-15-2021 17-33-45

denis-yuen (approved):

garyluu:
> > Edit: you were talking about vertical (not horizontal). We use variable size vertical for every modal/dialog so far. If you feel strongly about not using variable size, make an issue, discuss with Kim, and we can convert all modals (organization description, edit version, etc) to non-variable size vertical at once.
> 
> This one seems of a different magnitude than the others -- one keypress can radically change the size. But I will log separately.
> 
> Jun-15-2021 17-33-45

Like I said, the modals are all variable in size and this issue would be present in all past and future dialogs that have changing (or rapidly changing content). So no, the dialog behavior is not different (though this may have made it more pronounced). So when you log an issue for this. The expected behaviour needs to work for everything.

NatalieEO (approved):

coverbeck (approved):

PR #1281, opened 2021-06-24
Feature/1829/change staging prod smoke tests

Richard-Hansen:
Changes smoke tests to be pulled from a specific branch, rather than scraped from the webpage. The branches are (tentatively):
1. develop
2. stagingSmokeTests
3. prodSmokeTests

Created both the stagingSmokeTests and prodSmokeTests branches. Turning on prod smoke tests after the release seems easiest.

Also, the smoke test branches should have required PR reviews which I don't have permissions to set. Made a PR to the smoke test branch: https://github.com/dockstore/dockstore-ui2/pull/1282

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -253,7 +253,7 @@ workflows:
             - dockerhub
       - uptime_monitor_no_auth:
           stack: "staging"
-          smoke_testing_branch: "stagingSmokeTests"
+          smoke_testing_branch: "develop"
           context:
             - dockerhub
 #      # Add the tags filter to all jobs so they will run before upload_to_s3
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,15 +247,15 @@ workflows:
   everything:
     jobs:
       - uptime_monitor_no_auth:
-        stack: "dev"
-        smoke_testing_branch: "develop"
-        context:
-          - dockerhub
-    - uptime_monitor_no_auth:
-        stack: "staging"
-        smoke_testing_branch: "stagingSmokeTests"
-        context:
-          - dockerhub
+          stack: "dev"
+          smoke_testing_branch: "develop"
+          context:
+            - dockerhub
+      - uptime_monitor_no_auth:
+          stack: "staging"
+          smoke_testing_branch: "stagingSmokeTests"
+          context:
+            - dockerhub
 #      # Add the tags filter to all jobs so they will run before upload_to_s3
       - lint_license_unit_test_coverage:
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,10 +49,10 @@ jobs:
           environment:
             MOCHA_FILE: nightly-test-results/junit/test-<< parameters.stack >>-waf-[hash].xml
       - upload_nightly_artifacts
-      - slack/status:
-          fail_only: true
-          failure_message: Nightly << parameters.stack >> non-auth tests failed!
-          webhook: $SLACK_WEBHOOK
+#      - slack/status:
+#          fail_only: true
+#          failure_message: Nightly << parameters.stack >> non-auth tests failed!
+#          webhook: $SLACK_WEBHOOK
 
   # Specify the webpage url (https://dev.dockstore.net or https://staging.dockstore.org or https://dockstore.org)
   # and stack (dev or staging or prod) and the no auth smoke tests will be run against the corresponding webpage.
@@ -246,6 +246,16 @@ workflows:
   version: 2
   everything:
     jobs:
+      - uptime_monitor_no_auth:
+        stack: "dev"
+        smoke_testing_branch: "develop"
+        context:
+          - dockerhub
+    - uptime_monitor_no_auth:
+        stack: "staging"
+        smoke_testing_branch: "stagingSmokeTests"
+        context:
+          - dockerhub
 #      # Add the tags filter to all jobs so they will run before upload_to_s3
       - lint_license_unit_test_coverage:
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
     parameters:
       stack:
         type: string
-      webpage_URL:
+      smoke_testing_branch:
         type: string
     working_directory: ~/repo
     docker:
@@ -37,7 +37,7 @@ jobs:
           password: $DOCKERHUB_PASSWORD
     steps:
       - setup_nightly_tests:
-          webpage_URL: << parameters.webpage_URL >>
+          smoke_testing_branch: << parameters.smoke_testing_branch >>
       - run:
           name: Run remote verification test against << parameters.stack >> (no auth)
           command: bash -i -c 'npm run test-<< parameters.stack >>-no-auth'
@@ -60,7 +60,7 @@ jobs:
     parameters:
       stack:
         type: string
-      webpage_URL:
+      smoke_testing_branch:
         type: string
     working_directory: ~/repo
     docker:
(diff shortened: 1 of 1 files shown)


denis-yuen:
super minor, but these don't seem consistent:

> develop
>   stagingSmokeTests
>   prodSmokeTests

One is a branch but the other two are environments. 
Shouldn't it be `dev`, `staging`, `prod` or alternatively the first should be `devSmokeTests`?

denis-yuen (approved):

coverbeck (approved):
New releases will be interesting... You'll need to merge from the release branch or tag into `stageSmokeTests` branch.

I feel like the getting the commit from the running UI might ultimately be the best way to do this, that is was a one-off for this release where the tests don't run against staging. But on the other hand, this gives the ability to add tests after we've released, if we need to do that.

Long-winded way of approving.

Richard-Hansen:
> super minor, but these don't seem consistent:
> 
> > develop
> > stagingSmokeTests
> > prodSmokeTests
> 
> One is a branch but the other two are environments.
> Shouldn't it be `dev`, `staging`, `prod` or alternatively the first should be `devSmokeTests`?

Hmm, I think using the `develop` branch is best for the dev smoke tests as it doesn't create any extra headache. I see the inconsistency over the other 2, but would argue since they are strictly for smoke tests, the naming helps with clarity.

I don't have strong feelings either way, and changing the branch names is easy.

garyluu:
So I want to add a simple test that works for dev, prod, and staging. How would I do this?

Richard-Hansen:
> So I want to add a simple test that works for dev, prod, and staging. How would I do this?

It will require 3 PRs, depending on when you want the tests running against each stage.

If you want the smoke tests running against all 3 stages immediately, that will require 1 PR to each branch (unfortunately).

If you initially make a PR to add it to `develop`, then during a release we merge the release branch to either/both `stagingSmokeTests` or `prodSmokeTests`.

garyluu:
> It will require 3 PRs, depending on when you want the tests running against each stage.
> 
> If you want the smoke tests running against all 3 stages immediately, that will require 1 PR to each branch (unfortunately).
> 
> If you initially make a PR to add it to `develop`, then during a release we merge the release branch to either/both `stagingSmokeTests` or `prodSmokeTests`.

What's the advantages of having two extra branches vs 2 extra directories in dev?

Richard-Hansen:
> What's the advantages/disadvantages of having two extra branches vs 2 extra directories in develop branch?

**3 branch approach:**
1. Easy to fix a test in 1 stage.
2. Annoying to add a new test for all stages immediately -> 3 PRs (Definitely a headache).
3. Merge smoke tests into testing branches during a release.

**3 dir approach:**
1. Easy to fix a test in 1 stage.
2. Annoying to add a new test for all stages immediately -> duplicate the test 3 times (Less PR time, which is nice).
3. Merge one directory into another during a release.
4. Extra clutter from duplicated directories.

**1 shared test dir + 3 individual dir approach** (1 directory contains shared tests for all stages, and then an individual directory for tests that are custom for an individual stage):
1. Annoying to fix a test in 1 stage (You'd have remove it from the shared directory, and copy it into each individual directory).
2. Easy to add a test for all stages.
3. Cleanup could be a hassle during a release. You'd want to move as many tests as you can back into the shared directory, and then merge 1 custom directory into another.
4. Extra clutter from duplicated directories.
5. The better our tests, the more lightweight this approach is.
---
Depending on our use case, I could see each approach being better than the others.
1. If we intend to mainly perform hotfixes on failing smoke tests in staging/prod, and rarely adding new smoke tests for all 3 stages (Same process as before for adding smoke tests to dev), I think the 3 branch approach is effective.
2. If we intend to add new smoke tests for all 3 stages at least somewhat frequently, the 3 directory approach saves a lot of PR time compared to the 3 branch approach.
3. If we are a mix between the above 2, the 1 shared + 3 individual directory approach gives good flexibility, at the cost of test-shimmying (This approach gets better the more reliable our tests are, which is a +)
---
All of the above approaches are simple to implement, so no downsides there. Thoughts?

garyluu:
> > What's the advantages/disadvantages of having two extra branches vs 2 extra directories in develop branch?
> 
> **3 branch approach:**
> 
> 1. Easy to fix a test in 1 stage.
> 2. Annoying to add a new test for all stages immediately -> 3 PRs (Definitely a headache).
> 3. Merge smoke tests into testing branches during a release.
> 
> **3 dir approach:**
> 
> 1. Easy to fix a test in 1 stage.
> 2. Annoying to add a new test for all stages immediately -> duplicate the test 3 times (Less PR time, which is nice).
> 3. Merge one directory into another during a release.
> 4. Extra clutter from duplicated directories.
> 
> **1 shared test dir + 3 individual dir approach** (1 directory contains shared tests for all stages, and then an individual directory for tests that are custom for an individual stage):
> 
> 1. Annoying to fix a test in 1 stage (You'd have remove it from the shared directory, and copy it into each individual directory).
> 2. Easy to add a test for all stages.
> 3. Cleanup could be a hassle during a release. You'd want to move as many tests as you can back into the shared directory, and then merge 1 custom directory into another.
> 4. Extra clutter from duplicated directories.
> 5. The better our tests, the more lightweight this approach is.
> 
> Depending on our use case, I could see each approach being better than the others.
> 
> 1. If we intend to mainly perform hotfixes on failing smoke tests in staging/prod, and rarely adding new smoke tests for all 3 stages (Same process as before for adding smoke tests to dev), I think the 3 branch approach is effective.
> 2. If we intend to add new smoke tests for all 3 stages at least somewhat frequently, the 3 directory approach saves a lot of PR time compared to the 3 branch approach.
> 3. If we are a mix between the above 2, the 1 shared + 3 individual directory approach gives good flexibility, at the cost of test-shimmying (This approach gets better the more reliable our tests are, which is a +)
> 
> All of the above approaches are simple to implement, so no downsides there. Thoughts?

I personally think 3 > 2 > 1. It seems all the downsides of 3 is applicable to 1
For each point of 3:
1. Annoying to fix a test. May have to create 3 different PRs
2. Annoying to add a test. Definitely have to create 3 different PRs.
3. Cleanup is a hassle during release, gotta merge the dev smoke tests to both the staging and prod branches (which probably will have different cypress versions)
4. Extra clutter from having near duplicate branches (cypress, package.json, everything)
5. The better our tests, the more redundent things are in the 3 branches.

garyluu (approved):
Well, since Charles and Denis approved already and I don't plan to maintain synchronization, it doesn't matter too much to me.

coverbeck:
My sense is that we will mainly just be adding new tests to the develop branch. I don't think we'll typically be adding additional tests to what's already in staging/prod, unless there's something in particular we want to start monitoring.

Hotfixes might be an exception.

denis-yuen:
I think these are worthwhile questions actually. The three different PR issue seems like a particular pain given the need for reviews. I haven't worked with the smoke tests but at a superficial level it does seem like 3) would be less work to maintain. I also may have an instinct against long-lived branches that never merge. 

Maybe a question would help clarify, how do we anticipate tests moving when going past hotfixes to actual new releases of Dockstore under these three models?

Richard-Hansen:
> Maybe a question would help clarify, how do we anticipate tests moving when going past hotfixes to actual new releases of Dockstore under these three models?

I think all 3 models would have a similar pattern during a release as, ideally, you would be replacing tests rather than merging. 

For example, when we deploy to staging, we'd replace all the smoke tests currently in the staging branch/directory with the smoke tests in dev (since all the dev smoke tests should be working in dev beforehand). If there is a desync between the smoke tests and the current stage, you'd fix that individually (Like the launchWith buttons were reverted to their previous styling for the release, which broke a smoke test, requiring a PR).

Model 3 does have the following benefit: Once we complete a full deploy (dev -> staging -> prod), smoke tests from dev should be duplicated in the staging and prod directories, at which time we could move those tests back into the 'shared' test directory. Potentially cleaning out some/all of the unique tests.

Richard-Hansen:
Closing in favor of #1287

PR #1283, opened 2021-06-24
Feature/3361/logout incorrect user

garyluu:
For dockstore/dockstore#3361

Simply logout if there's any issues grabbing the user.

Had to move some code around because:

logout.service.ts imports user.service.ts, we grab users in the user.service.ts (which makes sense), but logging out requires logout.service.ts function which results in a circular dependency.

logout.service.ts is now mostly useless and should probably be merged with user.service sometime

So the previous behavior where the top right shows a dropdown with no username should never happen. The user would be instantly redirected to the logout page.

### DIFF
--- a/src/app/myworkflows/my-workflow/my-workflow.component.spec.ts
+++ b/src/app/myworkflows/my-workflow/my-workflow.component.spec.ts
@@ -19,6 +19,7 @@ import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
 import { MatDialogRef } from '@angular/material/dialog';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { RouterTestingModule } from '@angular/router/testing';
+import { TrackLoginService } from 'app/shared/track-login.service';
 import { AuthService } from 'ng2-ui-auth';
 import { AccountsService } from '../../loginComponents/accounts/external/accounts.service';
 import { CustomMaterialModule } from '../../shared/modules/material.module';
@@ -39,6 +40,7 @@ import {
   ConfigurationStub,
   RefreshStubService,
   RegisterWorkflowModalStubService,
+  TrackLoginStubService,
   UrlResolverStubService,
   UsersStubService,
   WorkflowsStubService,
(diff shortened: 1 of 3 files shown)


coverbeck (approved):
Just curious, were you able to reproduce dockstore/dockstore#3361, where it seems a valid token stopped working?

I feel like this is addressing dockstore/dockstore#2105, which is good, but still wondering what was going on in dockstore/dockstore#3361

garyluu:
> Just curious, were you able to reproduce [dockstore/dockstore#3361](https://github.com/dockstore/dockstore/issues/3361), where it seems a valid token stopped working?
> 
> I feel like this is addressing [dockstore/dockstore#2105](https://github.com/dockstore/dockstore/issues/2105), which is good, but still wondering what was going on in [dockstore/dockstore#3361](https://github.com/dockstore/dockstore/issues/3361)

There is no indication that it was a valid token

coverbeck:
> There is no indication that it was a valid token

I thought Denis wrote up the ticket for his experience on prod, where we don't clear/wipe the tokens like we do in other environments.

garyluu:
> > There is no indication that it was a valid token
> 
> I thought Denis wrote up the ticket for his experience on prod, where we don't clear/wipe the tokens like we do in other environments.

Still had every indication that it was the exact same behavior though (failing to get the user)

NatalieEO (approved):

denis-yuen:
> I thought Denis wrote up the ticket for his experience on prod, where we don't clear/wipe the tokens like we do in other environments. 

This is true, there is an external user on that ticket so it should be prod. 

> Still had every indication that it was the exact same behavior though (failing to get the user which likely means invalid token)

What is the theory though for why a redeploy would invalidate a token? 

That said, can't see why this would hurt so approving.

denis-yuen (approved):

garyluu:
> > I thought Denis wrote up the ticket for his experience on prod, where we don't clear/wipe the tokens like we do in other environments.
> 
> This is true, there is an external user on that ticket so it should be prod.
> 
> > Still had every indication that it was the exact same behavior though (failing to get the user which likely means invalid token)
> 
> What is the theory though for why a redeploy would invalidate a token?
> 
> That said, can't see why this would hurt so approving.

If this gets merged, the original issue https://github.com/dockstore/dockstore/issues/3361 would be kind of unreproducible and I doubt users care they have to login (click two buttons) every time we do a release. I personally don't really care exactly why it gets invalidated since it could be due to a myriad of unconfirmable reasons (satellizer implementation, browser security feature, etc)

wshands (approved):

PR #1286, opened 2021-06-29
Add dependabot config file to enable dependabot in dockstore-ui2

chmreid:
Related: [SEAB-3008](https://ucsc-cgl.atlassian.net/browse/SEAB-3008)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  # Enable version updates for npm
+  - package-ecosystem: "npm"
+    # Look for package.json and lock files in the root dir
+    directory: "/"
+    # Check the npm registry for updates every day
+    schedule:
+      interval: "daily"


coverbeck (commented):
It seems we are already getting dependabot PRs in this repo; for example there is an open one, #1284.

I see your comments on the ticket, going to defer to Denis on how these should be set up.

denis-yuen (approved):

denis-yuen:
> It seems we are already getting dependabot PRs in this repo; for example there is an open one, #1284.
> 
> I see your comments on the ticket, going to defer to Denis on how these should be set up.

Those look like dependabot-preview tickets rather than https://github.com/dockstore/dockstore/pull/4268 which is not in preview. Moving toward the new version is fine, with the caveat that both repos have issues with this kind of auto-update due to our generated snyk files (webservice) and generated license files (both).

garyluu (approved):

PR #1282, opened 2021-06-24
Update staging smoke tests

Richard-Hansen:
Targeting stagingSmokeTests branch.

Old launch with buttons use ```<svg>```s insteads of ``````s. 

Tested this alongside https://github.com/dockstore/dockstore-ui2/pull/1281, and reran the smoke tests a few times in circleCI: https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2?branch=feature%2F1829%2FchangeStagingProdSmokeTests

PR #1287, opened 2021-06-30
Feature/1829/smoke test directories

Richard-Hansen:
Issue: https://github.com/dockstore/dockstore/issues/4306
original PR that used branches: https://github.com/dockstore/dockstore-ui2/pull/1281
CircleCI job that ran the smoke tests: https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/5993/workflows/938c8d1b-717d-4b2a-b75f-aaa1cfc295c3

I can also add a short readme to the `smokeTests/` directory, if we think the explanation would help?

Changes how smoke tests are run by using multiple new directories:
- smokeTests/
  - sharedTests/
  - devOnly/
  - stagingOnly/
  - prodOnly/

The npm scripts have been updated to run the set of tests in the shared directory and the directory that matches the stage (dev/staging/prod).

Also split out a failing test regarding launch with buttons into the `devOnly/` and `stagingOnly/` directories. I opted to move over the entire test set (it's part of a larger function), so merging back is simple. But, we can change this if we want.

If we're happy with this approach I'll:
1. Close https://github.com/dockstore/dockstore-ui2/pull/1281
2. Close https://github.com/dockstore/dockstore-ui2/pull/1282
3. Delete the `stagingSmokeTests` and `prodSmokeTests` branches


garyluu (approved):

denis-yuen (approved):
Security warnings seem irrelevant

coverbeck (approved):
Seeing this in action, it seems like there is a fair amount of duplication. But we discussed all the options, so approving and we can always revisit if need be.

garyluu:
> Seeing this in action, it seems like there is a fair amount of duplication. But we discussed all the options, so approving and we can always revisit if need be.

It's like 

> Seeing this in action, it seems like there is a fair amount of duplication. But we discussed all the options, so approving and we can always revisit if need be.

This seems like a lot less duplication than the previous method which was tens of thousands of lines

PR #1298, opened 2021-07-06
Add extra note about what is ORCID

garyluu:
Feedback from ORCID meeting: add an additional note on what is ORCID. Kind of hacky.

image


denis-yuen (commented):
Looks good, but needs to be cherry-picked

PR #1309, opened 2021-07-08
Feature/use new launch buttons

Richard-Hansen:
Two minor changes. Ran smoke tests locally and they passed using the new launch with buttons.
1. Switches back to the new launch with buttons
2. Adds names to the dev/staging smoke test jobs that run


coverbeck (approved):

garyluu (approved):

denis-yuen (approved):

denis-yuen (commented):
Hmmm, with the caveat that we're not sure if DNAstack would make it into 1.12 if we do a fast(?) release. Not sure how likely that is

chmreid (approved):

Richard-Hansen:
> Hmmm, with the caveat that we're not sure if DNAstack would make it into 1.12 if we do a fast(?) release. Not sure how likely that is

Another option is to change the dev smoke tests to match what's in staging, and leave the buttons as-is. Thoughts?

coverbeck:
> Another option is to change the dev smoke tests to match what's in staging, and leave the buttons as-is. Thoughts?

That's probably best. Unfortunately, the property affects all WDL buttons, not just the DNAstack one, I don't think it's worth the effort to change that.

Richard-Hansen:
> > Another option is to change the dev smoke tests to match what's in staging, and leave the buttons as-is. Thoughts?
> 
> That's probably best. Unfortunately, the property affects all WDL buttons, not just the DNAstack one, I don't think it's worth the effort to change that.

Moved the staging and dev tests back into shared, this keeps the old launch with buttons, and the tests run successfully against dev and staging.

coverbeck (approved):

denis-yuen (approved):

PR #1310, opened 2021-07-12
Feature/seab 3119/refactor file tab

garyluu:
For SEAB-3119

Refactor some of the file tab components into one.

The workflow files component's Akita state files still remains because it appears it's used by tools

Currently a few requirements in the doc aren't implemented (but could be in the future):
- No file selection if there’s one file
- Each option in the file selection must be unique (no relative path to parent descriptor)
- put main descriptor before others (even if same type)


denis-yuen:
> Now the tabs are reversed, not sure if that's what we want. 

Think I'd like to keep the current order, unless we make an explicit decision to change it in storytime

garyluu:
> > Now the tabs are reversed, not sure if that's what we want.
> 
> Think I'd like to keep the current order, unless we make an explicit decision to change it in storytime

So in the doc, did you want me to merge the top configuration tab with the bottom configuration tab then?

denis-yuen (approved):

src/app/source-file-tabs/source-file-tabs.service.ts
```diff
@@ -24,6 +32,41 @@ export class SourceFileTabsService {
...
+    return fileTabs;
+  }
+
+  /**
+   * Given the sourcefiles in the current tab, return the validation message object that should be shown
+   */
+  getValidationMessage(sourcefiles: SourceFile[], version: WorkflowVersion): Map<string, string> {
+    let validationMessage = null;
+    version.validations.forEach((validation: Validation) => {
+      sourcefiles.forEach(file => {
```
coverbeck, line 61: Bike-shedding, especially given the typical small number of files in play, but you could call `find()` and avoid unnecessary loops. Unless you're looking to grab the last validation message instead of the first one.

coverbeck (approved):

src/app/source-file-tabs/source-file-tabs.service.ts
garyluu: No idea, copied from previous

NatalieEO (approved):

PR #1312, opened 2021-07-12
Feature/potato

garyluu:
@e34lin to test out

The search image was replaced with mat-icon, not sure of the downsides (if there's something wrong with the material icon for some reason). Change the color of the mat-icon and remove the image if it's not needed.

Few improvements are:
- the vertical size is now closer to https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/602197fe5ca1a163b9ae735b (mock is 36px, this is 37.31px, previously it's 43.81px)
- the outline now appears (though it's still missing the mat-label that goes to the top left after focus)

One main downside is that the input text has shifted slightly down, but considering that most appearances for the default material  form-fields also do that, it's probably not too bad

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -446,7 +446,6 @@ img.contain-img {
   -webkit-border-radius: 8px;
   -moz-border-radius: 8px;
   border-radius: 8px;
-  margin-top: 8px;
 }
 
 .button.login p {
@@ -947,7 +946,7 @@ mat-icon.star-icon {
 }
 
 .landing-search {
-  width: 60%;
+  margin-top: 8px;
   max-width: 600px;
   padding-left: 30px;
   padding-top: 3px;
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -945,20 +945,6 @@ mat-icon.star-icon {
   height: 16px;
 }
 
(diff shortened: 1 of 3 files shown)


src/styles.scss
```diff
@@ -946,20 +945,6 @@ mat-icon.star-icon {
...
-  padding-left: 30px;
-  padding-top: 3px;
-  padding-bottom: 3px;
-  border: none;
-  line-height: 2em;
-  background: white url(../images/search-button.png) left center no-repeat;
-  background-position: 5px;
-  outline: none;
-  color: black;
-}
```
garyluu, line 961: I have no idea what this mess was for since there's no comments

src/styles.scss
```diff
@@ -446,7 +446,6 @@ img.contain-img {
   -webkit-border-radius: 8px;
   -moz-border-radius: 8px;
   border-radius: 8px;
-  margin-top: 8px;
```
garyluu, line 449: also no idea why a class called fully-rounded is adding margin to the top

e34lin (approved):
I tested it out, and it looks good. Didn't experience any problems with it as well. Thank you!

PR #1318, opened 2021-07-23
Policy acknowledgement requires checkbox

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-2345
Force users to click the checkbox to register
Force users who are registered, but don't have the latest policies accepted, to click a checkbox to acknowledge the latest policies on login (they already got automatically logged out before).


cypress/integration/immutableDatabaseTests/tosAndPrivacyPolicy.ts
```diff
@@ -72,4 +73,48 @@ describe('TOS Banner', () => {
...
+      cy.visit('/');
+      cy.wait(500);
+      cy.clearLocalStorage('ng2-ui-auth_token');
+      cy.visit('/login');
+      cy.get('[data-cy=login-with-google]').should('not.be.disabled');
+      cy.get('[data-cy=login-with-github]').should('not.be.disabled');
+    });
+
+    setTokenUserViewPort();
+    it('Auto logg out if policies are not up to date and force a check if accepted policy in local storage is out of date.', () => {
```
denis-yuen, line 109: logg

src/app/login/login.component.html
```diff
@@ -54,13 +56,14 @@ <h3 class="available-containers">Login to Collaboratory Dockstore</h3>
...
+                data-cy="login-with-google"
               >
                 <fa-icon [icon]="faGoogle"></fa-icon> Login with Google
               </button>
-              <p>
-                By clicking Login with GitHub or Login with Google, you agree to Dockstore's
-                <template [ngTemplateOutlet]="PrivacyAndTOSDisclaimer"></template>
-              </p>
+              <mat-checkbox class="text-wrap" *ngIf="mustAcceptCurrentTOS" (change)="loginDisabled = !loginDisabled">
+                You must accept Dockstore's current <template [ngTemplateOutlet]="PrivacyAndTOSDisclaimer"></template
```
denis-yuen, line 65: Very imperative, maybe "Please accept"?

denis-yuen (approved):

src/styles.scss
```diff
@@ -1109,3 +1109,7 @@ hr {
 .mat-radio-button .mat-form-field-can-float.mat-form-field-should-float .mat-form-field-label {
   top: 1.8em;
 }
+
+.mat-checkbox.text-wrap > label.mat-checkbox-layout {
```
garyluu, line 1113: https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide#pr-checklist 

"Global styles that require the developer to actively use it..."

src/app/navbar/navbar.component.ts
```diff
@@ -63,6 +63,10 @@ export class NavbarComponent extends Logout implements OnInit {
   ngOnInit() {
     this.userQuery.user$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((user) => {
       this.user = user;
+      if (this.user) {
+        localStorage.setItem(acceptedTOSVersion, user.tosversion);
```
coverbeck, line 67: I have concerns about using localStorage for this:

1. The same user needs to accept the TOS on every computer the user logs in to Dockstore
2. This could "carry over" between users, although I didn't double-check the flow.
    1. User A accepts TOS, logs out
    2. Use B logs on same computer (unlikely, but shared computer, multiple accounts for same person) and localStorage indicates TOS are accepted, even if B hadn't accepted.

garyluu (approved):

src/app/navbar/navbar.component.ts
NatalieEO: Okay, I don't think 1 is that big a of deal, but yes there would be carry over if people shared the same computer/browser window. Maybe instead of logging them out, I'll just display the banner that follows the user around with somewhat different wording.

david4096 (approved):

src/app/navbar/navbar.component.ts
coverbeck: > Okay, I don't think 1 is that big a of deal, but yes there would be carry over if people shared the same computer/browser window. Maybe instead of logging them out, I'll just display the banner that follows the user around with somewhat different wording.

I recognize this makes it more complicated, alas. You could go with the annoying banner, or you'd have to introduce a new page after they log in if they haven't accepted the latest TOS.

src/app/app.component.ts
```diff
@@ -39,6 +46,16 @@ export class AppComponent implements OnInit, OnDestroy {
     this.isLoggedIn$ = this.trackLoginService.isLoggedIn$;
     this.dismissedLatestTOS$ = this.tosBannerQuery.dismissedLatestTOS$;
     this.dismissedLatestPrivacyPolicy$ = this.tosBannerQuery.dismissedLatestPrivacyPolicy$;
+    this.displayLoggedInTOSBanner$ = this.tosBannerQuery.displayLoggedInTOSBanner$;
+    this.userQuery.user$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((user) => {
```
garyluu, line 50: this ngUnsubscribe doesn't really do anything afaik. then again, app.component should never be destroyed so you kind of never need to unsubscribe

src/app/tosBanner/tos-banner.component.ts
```diff
@@ -1,15 +1,27 @@
...
   templateUrl: './tos-banner.component.html',
   styleUrls: ['./tos-banner.component.css'],
 })
 export class TosBannerComponent {
+  @Input()
+  public bannerText: string;
+  @Input()
+  public loggedIn: boolean;
+  @Input()
+  public user: User;
```
garyluu, line 16: probably best to not pass in the user. it triggers change detection even though it's not needed until acceptTOS() is clicked on. just have the service function grab the user when needed

src/app/tosBanner/state/tos-banner.service.ts
```diff
@@ -33,4 +38,23 @@ export class TosBannerService {
       };
     });
   }
+
+  acceptTOS() {
+    this.userQuery.user$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((user) => {
```
garyluu, line 43: this takeUntil also doesn't do anything. also, this will cause multiple unclosed subscription. need to get the user value directly from the store. search "getValue()" in the code for an example.

garyluu (changes requested):
There's a possible a niche bug where the user object is gotten really slowly, user accepts ToS before that object is gotten.  Probably add that as a comment, not sure what the solution is though.

david4096 (approved):

denis-yuen (approved):

src/app/app.component.ts
```diff
@@ -19,16 +21,20 @@ export class AppComponent implements OnInit, OnDestroy {
   public isLoggedIn$: Observable<boolean>;
   public dismissedLatestTOS$: Observable<User.TosversionEnum>;
   public dismissedLatestPrivacyPolicy$: Observable<User.PrivacyPolicyVersionEnum>;
+  public displayLoggedInTOSBanner$: Observable<boolean>;
   public currentTOSVersion: User.TosversionEnum = currentTOSVersion;
   public currentPrivacyPolicyVersion: User.PrivacyPolicyVersionEnum = currentPrivacyPolicyVersion;
+  protected ngUnsubscribe: Subject<{}> = new Subject();
```
garyluu, line 27: unused

src/app/tosBanner/state/tos-banner.service.ts
```diff
@@ -17,10 +17,14 @@
...
 import { TosBannerStore } from './tos-banner.store';
+import { User, UsersService } from '../../shared/openapi';
+import { UserQuery } from '../../shared/user/user.query';
+import { Subject } from 'rxjs';
 
 @Injectable({ providedIn: 'root' })
 export class TosBannerService {
-  constructor(private tosBannerStore: TosBannerStore) {}
+  constructor(private tosBannerStore: TosBannerStore, private userService: UsersService, private userQuery: UserQuery) {}
+  protected ngUnsubscribe: Subject<{}> = new Subject();
```
garyluu, line 27: unused

garyluu (approved):

src/app/app.component.ts
```diff
@@ -39,6 +44,15 @@ export class AppComponent implements OnInit, OnDestroy {
     this.isLoggedIn$ = this.trackLoginService.isLoggedIn$;
     this.dismissedLatestTOS$ = this.tosBannerQuery.dismissedLatestTOS$;
     this.dismissedLatestPrivacyPolicy$ = this.tosBannerQuery.dismissedLatestPrivacyPolicy$;
+    this.displayLoggedInTOSBanner$ = this.tosBannerQuery.displayLoggedInTOSBanner$;
+    this.userQuery.user$.pipe().subscribe((user) => {
```
coverbeck, line 48: It's the app object, so it may not matter that much, but I think you need to unsubscribe (or `takeUntil(....`).

coverbeck (commented):
If I'm reading this correctly, this doesn't force logged-in users to accept the latest TOS, it just displays a banner when they haven't done so.

If that's OK from a compliance perspective, I'm OK with it, but I just want to make sure that's the intent.

I think my comment about unsubscribe should probably be addressed -- even though it's the app, there is another subscription in the file that unsubscribes, so best to follow the pattern.

src/app/app.component.ts
NatalieEO: I had it before, but Gary explained it was pointless https://github.com/dockstore/dockstore-ui2/pull/1318#discussion_r678389800

But yes, you are reading it correctly. I'm assuming it's okay because Nneka reviewed the documentation that describes this and approved it.

coverbeck (approved):
> I had it before, but Gary explained it was pointless #1318 (comment)

Ah, I'd overlooked that comment. Yeah, I also said it may not matter, because the component will always be there until the app is gone, but there is already another subscription/takeUntil pattern in there, and if it ever gets moved out into a new component...

But I'll go ahead and approve as is.

PR #1320, opened 2021-07-29
Feature/seab 2905/starred redesign

garyluu:
For SEAB-2905

There are arbitrary css classes added where they are different from other similar components because 🤷


e34lin:
Incase the comment
Screen Shot 2021-07-29 at 1 52 56 PM
Screen Shot 2021-07-29 at 1 53 09 PM
 on the other PR were missed: 

I created tickets for both the verified tag, and collection length so they still don't work in my current PR. For the organization card, I just removed the inline style. I was also able to get the colours for each tab to work and there shouldn't be any horizontal or vertical scroll bars anymore.

This is how it currently looks like:
Screen Shot 2021-07-29 at 1 52 50 PM

coverbeck:
Comment based on your screenshots (they look good!) -- if there are none starred for an entity, like you have no starred services, then I think you should either display `0`, or not display the empty gray box like you have. Without actually trying it, I'd lean towards having 0.

This is in reference to where you have the counts on the tabs

src/app/starredentries/starredentries.component.html
```diff
@@ -16,110 +16,248 @@
...
+              </div>
+            </a.no-underline>
+          </mat-card>
+        </div>
       </div>
     </div>
   </mat-tab>
 
-  <mat-tab label="Organizations">
+  <!-- Services not done -->
```
coverbeck, line 171: If they're not done, should you remove the tab until they are?

I see now that my comment about your screenshot may only be applicable to this case, which may be n/a.

src/app/starredentries/starredentries.component.html
```diff
@@ -16,110 +16,248 @@
...
-              style="width: 100px"
-              [workflow]="workflow"
-              (starGazersChange)="starGazersChange()"
-            ></app-starring>
-            <mat-divider></mat-divider>
-          </mat-list-item>
-        </mat-list>
+                    <span id="verifiedIcon" *ngIf="(tool$ | async)?.versionVerified" fxHide.lt-sm>
+                      <mat-chip matTooltip="Verified" class="gray-label-tag">
+                        <a [href]="verifiedLink" class="verified-check" fxLayoutGap="0.5rem" fxLayoutAlign=" center">
```
coverbeck, line 154: I was going to suggest this should be a routerLink instead of an href. But then I went into the editor to see if it was an internal vs. external link (it should be routerLink only if it's internal), and the IDE says both `verifiedLink` and `tool$` on line 152 are undefined. IDE could be funky, but it seems like something is wrong -- can you please check?

src/app/starredentries/starredentries.component.html
```diff
@@ -16,110 +16,248 @@
...
-            <app-organization-starring
-              class="pull-right"
-              style="width: 100px"
-              [organization]="organization"
-              (stargazersChange)="organizationStarGazersChange()"
-            ></app-organization-starring>
-            <mat-divider></mat-divider>
-          </mat-list-item>
-        </mat-list>
+        <div *ngIf="starredOrganizations.length > 0; else noOrganizationsFound">
```
coverbeck, line 200: `noOrganizationsFound` does not exist

coverbeck: More generally, is it possible to use organizations component here? Or a subset of it? It probably won't work out of the box, but if we decide to change the number of orgs per row to 2 or 4, or redo the layout entirely, we'll now have to change it in two places.

I realize this won't be trivial -- the organizations component probably assumes it's displaying all orgs; you may need to extract a new component from the organizations component. But there's a lot of duplication right now.

coverbeck (changes requested):

denis-yuen (commented):
Looking good based on review demo, had similar comment on the consistency of the tabs with zero starred. Think it was easiest to make them show `0` when empty across the board since that was already the case for tools/workflows.

src/app/starredentries/starredentries.component.html
e34lin: Yeah, I mentioned on the old ticket that I was having some trouble getting the verified versions for the workflow and tools, so I must've forgotten to change that. I made a ticket for that, but don't think it's resolved yet. I also just removed the href, since I don't think it is needed. Do you think I should just remove the verified check entirely for now?

e34lin:
At the moment, the services tab is hardcoded to display a 0.

src/app/starredentries/starredentries.component.html
e34lin: I think this was also mentioned in the old PR as well, but I just decided to remove the inline styles for now since I'm not sure if the starred organizations tab would look different than the organizations tab in the future.

coverbeck (commented):
When I run this locally, the org layout seems off, View and collections could is outside of the box.

Screen Shot 2021-08-11 at 10 01 05 AM

@garyluu Can't mark you as a reviewer, since you created this, but can you review without being a formal reviewer.

e34lin:
The spacing for the organization cards should be fixed now
Screen Shot 2021-08-12 at 4 46 40 PM

coverbeck (changes requested):
When I run this locally, I star the UCSC GI organization. On the organizations page, the org has one collection, which is correct. When I look at the same org on My Starred, it says zero collections (might be a web service issue).

Also, when I hover over the starred org, my mouse cursor doesn't turn to a "link cursor" like it does when I hover over an org on the organizations page.

e34lin:
> When I run this locally, I star the UCSC GI organization. On the organizations page, the org has one collection, which is correct. When I look at the same org on My Starred, it says zero collections (might be a web service issue).
> 
> Also, when I hover over the starred org, my mouse cursor doesn't turn to a "link cursor" like it does when I hover over an org on the organizations page.

On the other (closed) PR, I mentioned how I had a problem with getting the correct number of collections on the starred organization page. I think it currently requires me to modify the endpoint so I think Gary told me to create a ticket for it instead, so it still doesn't work on my current PR.

Also made the change for the cursors, so the 'link' cursors should appear when you hover for all the cards. Thanks!

coverbeck (commented):
This looks really nice, but I still noticed a couple of things:

1. When hovering over a my starred organization, a tooltip with the org name appears. This does not happen in prod, and it doesn't happen on the Organizations page. The purpose of a tooltip is to provide more contextual information; in this case it's providing less info than what's already there, so I think it should go.
2. When you click on the number of stars on the my starred page, for both workflows and orgs (didn't try tools), it does 2 things (took me a while to figure out, I was confused at first):
    1. It opens a new tab with the the org/workflow and gives it focus
    2. If you go back to the original tab, it has navigated to all the stargazers. I would expect it to this only (and that's what it does on prod).

e34lin:
> This looks really nice, but I still noticed a couple of things:
> 
> 1. When hovering over a my starred organization, a tooltip with the org name appears. This does not happen in prod, and it doesn't happen on the Organizations page. The purpose of a tooltip is to provide more contextual information; in this case it's providing less info than what's already there, so I think it should go.
> 2. When you click on the number of stars on the my starred page, for both workflows and orgs (didn't try tools), it does 2 things (took me a while to figure out, I was confused at first):
>    
>    1. It opens a new tab with the the org/workflow and gives it focus
>    2. If you go back to the original tab, it has navigated to all the stargazers. I would expect it to this only (and that's what it does on prod).

I removed the tooltip in the PR, and I slightly changed the behaviour of the card which should fix the issue. You can either click on the name of the starred item or 'view' to open a new tab where you can view the starred item for the workflows and tools. So now when you click on the number of stars, it shouldn't open to a new tab and only navigate to the stargazers.

coverbeck (commented):
Looking good! I still see one problem:

1. Go to My Starred
2. Click on the stars of an organization
3. It now correctly takes you to the stargazers without opening a new tab
4. Click on "Back to details" -- you end up on a blankish screen.

Also, following up on the post-mortem best practices, is the ticket you created for the number of collections always showing 0, marked as 1.12 milestone? Do you have the ticket number handy.

e34lin:
> Looking good! I still see one problem:
> 
> 1. Go to My Starred
> 2. Click on the stars of an organization
> 3. It now correctly takes you to the stargazers without opening a new tab
> 4. Click on "Back to details" -- you end up on a blankish screen.
> 
> Also, following up on the post-mortem best practices, is the ticket you created for the number of collections always showing 0, marked as 1.12 milestone? Do you have the ticket number handy.

I'm not too sure why it does that, but I'll take a look at it right now and see if I can figure it out. 

[SEAB-3136](https://ucsc-cgl.atlassian.net/browse/SEAB-3136) is the ticket I created, but I haven't marked the milestone for it.

e34lin:
The page should load properly now when you go back from stargazers.

src/app/starredentries/starredentries.component.html
```diff
@@ -16,110 +16,254 @@
...
+                          <span class="verified-check" fxLayoutGap="0.5rem" fxLayoutAlign=" center">
+                            <mat-icon class="verified-icon">done</mat-icon>
+                            <span>Verified</span>
+                          </span>
+                        </mat-chip>
+                      </mat-chip-list>
+                    </span>
+                  </div>
+                </div>
+                <a routerLink="/workflows/{{ workflow.full_workflow_path }}" target="_blank" rel="noopener" class="no-underline">
```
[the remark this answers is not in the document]
garyluu: https://github.com/dockstore/dockstore-ui2/pull/635#discussion_r282653584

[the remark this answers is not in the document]
coverbeck: That comment is for an external link, outside dockstore.org. This comment is for a link within dockstore.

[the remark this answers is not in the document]
garyluu: Ok, so your process for determining whether a link should be opened in a new tab depends on whether it's external or not? My initial assumption based on that PR is that you wanted every link in the starred page to open in a new tab (since no specific reason was given)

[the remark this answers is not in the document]
coverbeck: Yes, I would go with that. :) And I think we generally behave like that already; just tried a couple of links on the home page for a not vary rigorous confirmation.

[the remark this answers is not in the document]
garyluu: So first of all, that's not part of the frontend style guide yet (I was planning to add it if further thought and research corroborates it). It's not really consistent depending on what's considered internal or not (gui assets like pdfs, dockstore documentation open in new tab).  Looking into it, I think the summary of https://www.nngroup.com/articles/new-browser-windows-and-tabs/ was what I've always thought for links which is more based on the task the user is trying to accomplish and internal vs external is not really a good way of determining it.

So for this specific case (because it used to not open into a new tab, then it was changed to a new tab)... before I change it back to no new tab, I need some kind of good detailed comment to prevent some new team member from stumbling upon it and changing it back to a new tab

coverbeck (approved):

NatalieEO (approved):

denis-yuen (approved):

PR #1325, opened 2021-08-10
Add --prod flag to all npm build steps to fix SRI

chmreid:
This PR modifies package.json to add the `--prod` flag to the build step
for the staging stage (not just for the prod stage).

This is necessary because without the `--prod` flag, the staging site
adds a `<noscript></noscript>` tag block that contains `<link>` tags
in it that do not have an 'integrity' attribute.

In other words, if we do not build our site with the `--prod` flag, it
disables SRI for resources in the `<noscript></noscript>` tags, which
leaves users with Javascript **off** vulnerable to CSS-based cross-site
scripting attacks.

See [SEAB-3225](https://ucsc-cgl.atlassian.net/browse/SEAB-3225).

PR #1327, opened 2021-08-11
Feature/angular cli settings

garyluu:
Should've been done during the Angular 12 CLI.

Essentially copied the default angular.json, package.json, and .browserslistrc

After the first commit (but before actually committing), the hash code for `npm run build` is the same as develop's `npm run build.prod`

I've officially dropped support for IE9 and 10...not like it was working before

More details on what changed:
prod build is now done by default, staging builds are removed

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -5,8 +5,10 @@
 # You can see what browsers were selected by your queries by running:
 #   npx browserslist
 
-> 0.5%
-last 2 versions
+last 1 Chrome version
+last 1 Firefox version
+last 2 Edge major versions
+last 2 Safari major versions
+last 2 iOS major versions
 Firefox ESR
-not dead
-IE 9-11 
+IE 11 # Angular supports IE 11 only as an opt-in. To opt-in, remove the 'not' prefix on this line.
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -15,7 +15,8 @@
     "prebuild": "ts-node -O '{\"module\":\"commonjs\"}' git.version.ts && ./scripts/generate-openapi-script.sh",
     "prebuild.prod": "npm run prebuild",
     "build.prod": "npx ng build --prod --subresource-integrity",
```
coverbeck, line 17: I thought you [said](https://github.com/dockstore/dockstore-ui2/pull/1325#issuecomment-897138979) you were changing this to `-c production` in your PR, or did you mean another PR?

coverbeck (approved):

package.json
garyluu: oh, i only changed the "build"

.browserslistrc
```diff
@@ -5,8 +5,10 @@
 # You can see what browsers were selected by your queries by running:
 #   npx browserslist
 
-> 0.5%
-last 2 versions
+last 1 Chrome version
```
denis-yuen, line 8: Why is this changing to support the last 2 versions of everything except for Chrome which is our most popular browser?

denis-yuen (approved):

.browserslistrc
garyluu: Just copying the default settings from the Angular CLI generated ones. My guess is that Firefox and Chrome auto-updates enough that the last 1 major version is enough

wshands (approved):

PR #1330, opened 2021-08-13
Use develop JAR

garyluu:
Jumped the gun on my previous PR. It was using a branch JAR which i was supposed to update

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "config": {
     "webservice_version": "1.11.3",
     "use_circle": true,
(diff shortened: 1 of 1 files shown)


denis-yuen (approved):

garyluu:
Hmm...looks like a non-flakey test failure. Previously with Amazon ECR, you can type in the docker registry path. but now you can't.

garyluu:
Kathy taking over, closing PR to avoid confusion

PR #1331, opened 2021-08-16
Fix snapshot/export error message

garyluu:
For dockstore/dockstore#4366

Going to use `workflowName || repository` as the workflow name (it's not unique btw, only `workflow.full_workflow_path` is). suggestions welcome.  Another option is to omit it entirely and only mention version

hosted workflows use `repository` only in the error message (because workflowname isn't a thing)


coverbeck (approved):

denis-yuen (commented):
fix version is for 1.11.x (if possible)

garyluu:
Couldn't rebase. Closing in favor of #1332

PR #1333, opened 2021-08-16
Husky 7

garyluu:
For dockstore/dockstore#4281

Important:
After doing this and switching to an older branch, git secrets/linter/prettier no longer works. This is because it previous uses .git/hooks, now it uses .husky/

I think what should happens is that this PR gets reviewed, then merged, then reviewed by security people to ensure git-secrets still works. Then I apply the changes to all the other repos. 

Main process:
Followed instructions for https://github.com/typicode/husky-4-to-7
Then changed the .husky files from `${husky_git_params}` to `$@` because of https://stackoverflow.com/a/57864128


coverbeck (approved):

NatalieEO (approved):

denis-yuen:
(adding security folks)

denis-yuen (approved):
actually, misread

PR #1339, opened 2021-08-24
Allow public and private Amazon ECR tools to be registered

kathy-t:
For https://github.com/dockstore/dockstore/issues/4331
Corresponding webservice PR: https://github.com/dockstore/dockstore/pull/4420

Registering a public Amazon ECR tool: [public.ecr.aws/x5g7o3i3/appropriate/curl](https://gallery.ecr.aws/x5g7o3i3/appropriate/curl)
- Docker registry path is pre-filled and disabled
- `x5g7o3i3` is what Amazon calls a registry alias. For our use, it's a namespace. 
  - Side note: Amazon allows you to request a custom alias that's more readable. For example, for the repository: [public.ecr.aws/ubuntu/ubuntu](https://gallery.ecr.aws/ubuntu/ubuntu), the first `ubuntu` is a custom registry alias
- `appropriate/curl` is the repository name (this one has slashes in it)

Screenshot 2021-08-24 at 08-39-25 Dockstore My Tools

Registering a private Amazon ECR tool: `467982390456.dkr.ecr.us-east-1.amazonaws.com/kathy-test`
- Docker registry path can be edited if the Private Image checkbox is checked
- Private images don't have a registry alias (because it's already uniquely identified by the account id), so in our case, the namespace should always be `_`. 
  - Any thoughts on whether this should this be enforced in the UI? It previously wasn't.
  - EDIT: We specify in the docs that "Amazon ECR images are treated in Dockstore as a custom Docker registry path and an empty namespace": https://docs.dockstore.org/en/develop/advanced-topics/docker-registries.html?highlight=amazon%20ecr#amazon-ecr-tools
- `kathy-test` is the repository name (private repository names can also have slashes)

Screenshot 2021-08-24 at 08-41-27 Dockstore My Tools

Registering a hosted Amazon ECR tool:
Screenshot 2021-08-24 at 08-42-44 Dockstore My Tools

### DIFF
--- a/src/app/container/register-tool/register-tool.component.html
+++ b/src/app/container/register-tool/register-tool.component.html
@@ -204,20 +204,38 @@
           <div class="form-group form-group-sm">
             <label class="col-sm-3 control-label"> Image Registry: </label>
             <div class="col-sm-9">
-              <input
-                id="imageRegistryInput"
-                type="text"
-                class="form-control"
-                name="imagePath"
-                [(ngModel)]="tool.imagePath"
-                minlength="3"
-                maxlength="128"
-                [pattern]="validationPatterns.imagePath"
-                required
-                matTooltip="Docker Image Registry path."
-                placeholder="e.g. cancercollaboratory/dockstore-tool-liftover"
-              />
-              <div *ngIf="formErrors.imagePath" class="alert alert-danger">{{ formErrors.imagePath }}</div>
+              <span *ngIf="tool.irProvider === 'Amazon ECR'">
+                <input
+                  id="imageRegistryInput"
+                  type="text"
+                  class="form-control"
+                  name="repoNameWithSlashesImagePath"
+                  [(ngModel)]="tool.imagePath"
+                  minlength="3"
+                  maxlength="128"
+                  [pattern]="validationPatterns.repoNameWithSlashesImagePath"
+                  required
+                  matTooltip="Docker Image Registry path."
+                  placeholder="e.g. cancercollaboratory/dockstore-tool-liftover"
+                />
+                <div *ngIf="formErrors.repoNameWithSlashesImagePath" class="alert alert-danger">{{ formErrors.repoNameWithSlashesImagePath }}</div>
+              </span>
+              <span *ngIf="tool.irProvider !== 'Amazon ECR'">
+                <input
+                  id="imageRegistryInput"
+                  type="text"
+                  class="form-control"
+                  name="imagePath"
+                  [(ngModel)]="tool.imagePath"
+                  minlength="3"
+                  maxlength="128"
+                  [pattern]="validationPatterns.imagePath"
+                  required
+                  matTooltip="Docker Image Registry path."
+                  placeholder="e.g. cancercollaboratory/dockstore-tool-liftover"
+                />
+                <div *ngIf="formErrors.imagePath" class="alert alert-danger">{{ formErrors.imagePath }}</div>
+              </span>
             </div>
           </div>
           <div class="form-group form-group-sm">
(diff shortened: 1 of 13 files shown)


denis-yuen:
>     * Private images don't have a registry alias (because it's already uniquely identified by the account id), so in our case, the namespace should always be `_`.
>       
>       * Any thoughts on whether this should this be enforced in the UI? It previously wasn't.
>       * EDIT: We specify in the docs that "Amazon ECR images are treated in Dockstore as a custom Docker registry path and an empty namespace": https://docs.dockstore.org/en/develop/advanced-topics/docker-registries.html?highlight=amazon%20ecr#amazon-ecr-tools

I forget, is the `_` just a placeholder or is it functional? 
I'm guessing its a placeholder and that one actually does `docker pull 467982390456.dkr.ecr.us-east-1.amazonaws.com/kathy-test`  Since its just our convention, it sounds like we should enforce it.

src/app/container/register-tool/register-tool.component.html
```diff
@@ -59,280 +59,274 @@ <h4 mat-dialog-title>Register Tool</h4>
...
+              name="toolName"
+              [(ngModel)]="tool.toolname"
+              maxlength="256"
+              [pattern]="validationPatterns.toolName"
+              matTooltip="Optional Dockstore tool path suffix"
+              placeholder="e.g. liftover-fast (optional)"
+            />
+            <mat-error *ngIf="formErrors.toolName">{{ formErrors.toolName }}</mat-error>
+            <mat-hint>Name to distinguish between multiple tools within the same repository</mat-hint>
+          </mat-form-field>
```
coverbeck, line 329: This comment for line 345/339 because I can't comment there directly -- was like this, but opened in WebStorm and `savingActive` doesn't exist.

coverbeck (approved):
Opened the HTML file in WebStorm out of curiousity, and it flagged one reference to a non-existent variable. It was like that already, and it probably isn't a big deal, but if you want to fix, your call.

kathy-t:
Changed it so only one error messaged is shown at a time. Made sure each hint/error message fits on one line instead of using the style I added because the style was pushing the content below further down when the multi-line error showed up, which is something that the material design guide warns against (https://material.io/components/text-fields#anatomy under Error Text).

Screenshot 2021-08-30 at 10-41-02 Dockstore My Tools

Screenshot from 2021-08-30 11-01-07

Some of the error messages I edited are also used by the manual registration of workflows, so while I was at it, I also edited a few of the error messages that exclusively showed up for the manual registration of workflows because they were also overflowing into the next input field.
Before:
Screenshot 2021-08-30 at 11-07-47 Dockstore My Workflows
After:
Screenshot 2021-08-30 at 11-08-27 Dockstore My Workflows

garyluu (approved):

src/app/shared/validationMessages.model.ts
```diff
@@ -133,19 +133,19 @@ export const validationMessages = {
...
     maxlength: 'Image Path is too long (max 128 characters).',
-    pattern: `The namespace and name of the image repository, separated by a '/'. The name of the image repository may contain additional '/'.`,
+    pattern: `Must have the form <namespace>/<name>. The <name> may contain additional '/'.`,
   },
   privateAmazonImagePath: {
     required: 'This field cannot be empty.',
     minlength: 'Image Path is too short (minimum 3 characters).',
     maxlength: 'Image Path is too long (max 128 characters).',
-    pattern: `The namespace and name of the image repository, separated by a '/'. The namespace must be empty (use '_'). The name of the image repository may contain additional '/'.`,
+    pattern: `Must have the form <namespace>/<name>. The namespace must be empty (use '_').`,
```
coverbeck, line 148: I find this a little confusing: it must be of the form `<namespace>/<name>`, but leave the namespace empty, but don't leave it empty, use `_`.

How about `Must have the form _/<name>`, maybe with an explanation.

src/app/container/register-tool/register-tool.component.html
```diff
@@ -378,16 +379,23 @@ <h4 mat-dialog-title>Register Tool</h4>
...
+                  : validationPatterns.imagePath
+              "
               required
               placeholder="e.g. cancercollaboratory/dockstore-tool-liftover"
               matTooltip="Docker Image Repository"
             />
-            <mat-hint *ngIf="hostedTool.registryProvider === 'Amazon ECR'">Has the form &lt;namespace&gt;/&lt;name&gt;. An empty namespace, represented by '_', must be used if the registry is private.</mat-hint>
-            <mat-hint *ngIf="hostedTool.registryProvider !== 'Amazon ECR'">Has the form &lt;namespace&gt;/&lt;name&gt;</mat-hint>
+            <mat-hint *ngIf="hostedTool.registryProvider === 'Amazon ECR'"
+              >Has the form &lt;namespace&gt;/&lt;name&gt;. If the registry is private, use '_' as the namespace.</mat-hint
```
coverbeck, line 392: I think this message is very clear, and maybe you could use it or a variant for the error message (see my other comment).

coverbeck (approved):

NatalieEO (approved):

denis-yuen (approved):
With above thumbed comment

PR #1345, opened 2021-09-02
Feature/dock 1795/use sandbox orcid on dev

svonworl:
Changed the ORCID accounts box to use the root of Dockstore.ORCID_AUTH_URL to create the link to the user's ORCID.
https://github.com/dockstore/dockstore/issues/4253
https://ucsc-cgl.atlassian.net/browse/DOCK-1795

This was hard to test: the better testing methods would consume much more effort to implement than the fix itself.  Gary suggested that user testing on dev might be easiest, and, in lieu of that, Charles suggested I move the helper function getRoot() into AccountsService and write some tests for it since it contained a regexp, and submit a PR.  And here we are.  I tested the modified clause in the HTML template by copying it elsewhere in the template, where it rendered successfully, so I'm 98%+ sure it'll work whenever we deploy it on dev/staging.  Someone will need to check.

I found the core of getRoot() while checking if Javascript had a quick way to get the root of a URL.  I've noted the source in the code comments.  I might have ended up with the exact same code after writing it myself.  If the sourcing is not kosher, let me know and I'll replace it with another method. 

An aside: since getRoot() is a generic helper function, it really should be defined in a central place where it can be reused.  However, I did a quick Internet survey of Angularland regarding helper functions and best practices to manage them, and found two basic recommendations.  One was "helper functions wut?!" and the other was "create a HelperService and put them there".  The latter seemed pretty heavyweight compared to this tiny little bug fix, so I did not.

### DIFF
--- a/src/app/loginComponents/accounts/external/accounts.component.html
+++ b/src/app/loginComponents/accounts/external/accounts.component.html
@@ -83,7 +83,7 @@
                 <strong>Username: {{ row.source | getTokenUsername: tokens }} </strong>
                 <span *ngIf="row.name === 'ORCID'">
                   <img src="../../assets/images/account-logos/{{ row.logo }}" alt="ORCID iD logo" class="orcid-id-logo site-icons-small" />
-                  <a href="{{ this.accountsService.getRoot(Dockstore.ORCID_AUTH_URL) }}/{{ orcidId$ | async }}"> {{ orcidId$ | async }}</a>
+                  <a href="{{ orcidRootUrl }}/{{ orcidId$ | async }}"> {{ orcidId$ | async }}</a>
                 </span>
               </mat-card-subtitle>
               <ng-template #placeholder><br /></ng-template>
--- a/src/app/loginComponents/accounts/external/accounts.component.html
+++ b/src/app/loginComponents/accounts/external/accounts.component.html
@@ -83,7 +83,7 @@
                 <strong>Username: {{ row.source | getTokenUsername: tokens }} </strong>
                 <span *ngIf="row.name === 'ORCID'">
                   <img src="../../assets/images/account-logos/{{ row.logo }}" alt="ORCID iD logo" class="orcid-id-logo site-icons-small" />
-                  <a href="https://orcid.org/{{ orcidId$ | async }}"> {{ orcidId$ | async }}</a>
+                  <a href="{{ getRoot(Dockstore.ORCID_AUTH_URL) }}/{{ orcidId$ | async }}"> {{ orcidId$ | async }}</a>
                 </span>
               </mat-card-subtitle>
               <ng-template #placeholder><br /></ng-template>
--- a/src/app/loginComponents/accounts/external/accounts.component.html
+++ b/src/app/loginComponents/accounts/external/accounts.component.html
@@ -83,7 +83,7 @@
                 <strong>Username: {{ row.source | getTokenUsername: tokens }} </strong>
                 <span *ngIf="row.name === 'ORCID'">
                   <img src="../../assets/images/account-logos/{{ row.logo }}" alt="ORCID iD logo" class="orcid-id-logo site-icons-small" />
-                  <a href="{{ getRoot(Dockstore.ORCID_AUTH_URL) }}/{{ orcidId$ | async }}"> {{ orcidId$ | async }}</a>
+                  <a href="{{ this.accountsService.getRoot(Dockstore.ORCID_AUTH_URL) }}/{{ orcidId$ | async }}"> {{ orcidId$ | async }}</a>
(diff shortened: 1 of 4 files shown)


src/app/loginComponents/accounts/external/accounts.component.html
```diff
@@ -83,7 +83,7 @@
                 <strong>Username: {{ row.source | getTokenUsername: tokens }} </strong>
                 <span *ngIf="row.name === 'ORCID'">
                   <img src="../../assets/images/account-logos/{{ row.logo }}" alt="ORCID iD logo" class="orcid-id-logo site-icons-small" />
-                  <a href="https://orcid.org/{{ orcidId$ | async }}"> {{ orcidId$ | async }}</a>
+                  <a href="{{ this.accountsService.getRoot(Dockstore.ORCID_AUTH_URL) }}/{{ orcidId$ | async }}"> {{ orcidId$ | async }}</a>
```
coverbeck, line 86: I should have noticed this when we spoke yesterday... You generally don't want to call functions like this from Angular templates, because they'll keep re-executing (because Angular doesn't know if it's changed or not). In the component, you should create a property that calls your service method, and then reference the property here.

coverbeck (changes requested):

src/app/loginComponents/accounts/external/accounts.component.html
svonworl: The official "Dockstore Frontend Onboarding" guide says:

"No complicated functions or logic in templates or else it's called a bajillion times"

And, irregardless of whether or not the PR runs afoul of that, the ORCID root URL is static and can be precomputed.  Waste not, want not.  Changed!

garyluu: hmm...maybe i should remove "complicated" from the wiki 🤔

src/app/loginComponents/accounts/external/accounts.service.ts
```diff
@@ -76,4 +76,12 @@ export class AccountsService {
         break;
     }
   }
+
+  /**
+   * Compute the root portion of a URL.  Adapted from:
+   * {@link https://stackoverflow.com/questions/1368264/how-to-extract-the-hostname-portion-of-a-url-in-javascript}
+   */
+  getRoot(url: string): string {
+    return url.replace(/^(.*\/\/[^\/?#]*).*$/,"$1");
```
garyluu, line 85: 3rd option of the link seems cleaner. if you're worried about IE11, we already have a polyfill for it

garyluu (approved):
ok either way

src/app/loginComponents/accounts/external/accounts.component.html
svonworl: I like the current guideline.  A function in a template could be useful sometimes, and as long as the repeated invocations have negligible computational overhead, it shouldn't be a problem.  My understanding (yet to be confirmed by looking at the generated code) is that the Angular change detection logic will evaluate each chunk of embedded code but will only re-render if the result changes.

svonworl: Quick benchmark: the getRoot() replace() invocation runs at about 2,000,000 ~24-char-long URLs per second on my Mac, node v14.17.0.

src/app/loginComponents/accounts/external/accounts.service.ts
svonworl: Yeah, I went with the regexp solution vs URL because it seemed more resistant to browser oddities, although more opaque.

src/app/loginComponents/accounts/external/accounts.component.html
coverbeck: Will the requested change affect the user perception of performance in any noticeable way? Almost certainly not.

It's just a best practice. If 20 function calls  crept into the page, then it might start to be noticeable. And it's a learning PR for you, so when you write more complex pages, you should leverage the techniques you learn here.

A poor Java analogy: if it's called only once in an app, one could use StringBuffer instead of StringBuilder, and there would be no noticeable difference in performance. But if I saw a PR, I would still recommend it be changed -- maybe in the future the code gets copied as a model, invoked from a loop, etc.

coverbeck (approved):

denis-yuen (approved):
>  Someone will need to check. 

Will get a review so should be fine. 

> This was hard to test: the better testing methods would consume much more effort to implement than the fix itself.  

I would have thought so!

src/app/loginComponents/accounts/external/accounts.component.html
svonworl: Yes, Charles, your recommendation was good, thanks!

PR #1348, opened 2021-09-08
redirect failures to specific channels

denis-yuen:
Not sure if the syntax for the channel works.
Needed 4.4.4 for it to understand channel names as well as IDs


denis-yuen:
too many moving parts

PR #1350, opened 2021-09-09
Display multiple authors for workflows

kathy-t:
For https://github.com/dockstore/dockstore/issues/4295
zeplin: 
https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60219b1cf92a1d19c95a33a2
https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/602c16fafaa1a71a3aac49e7

Added a table that displays multiple authors for workflows. Note that Orcid Authors only have the orcid id because the work is not complete yet (there's a ticket to grab orcid author info via the orcid public api https://ucsc-cgl.atlassian.net/browse/SEAB-3220).

I made the styling for the table global because I browsed through zeplin and it looks like all the tables have the same style as the Authors table. I went through all the other tables and made sure they looked acceptable (mainly just removed the elevation style).

Removed some `td` and `th` styles that were being overriden by mat-table so they weren't being used anywhere.

**Screenshots of Authors Table**
My Workflows with authors
Screenshot 2021-09-09 at 14-14-44 Dockstore My Workflows

Published workflow with authors
Screenshot 2021-09-09 at 14-15-35 Dockstore Workflow
- The published workflow page has a fixed width which makes the Authors table quite squished if all the fields are filled out. Not sure what to do about this. Are there plans to make the published workflow page fit the width of the screen, like the my workflows page?

My workflows with no authors
Screenshot 2021-09-09 at 13-56-32 Dockstore My Workflows
Published workflows with authors
Screenshot 2021-09-09 at 13-56-43 Dockstore Workflow

Misc. screenshots of tables across the site, since the styling changed a bit
Screenshot 2021-09-09 at 12-04-32 Dockstore Search

Screenshot 2021-09-09 at 12-05-37 Dockstore Workflow

Screenshot 2021-09-09 at 12-06-11 Dockstore Workflow

Screenshot 2021-09-09 at 12-06-27 Dockstore Workflow

Screenshot 2021-09-09 at 12-06-49 Dockstore Workflow

Screenshot 2021-09-09 at 12-15-35 Dockstore Tools

Screenshot 2021-09-09 at 12-15-17 Dockstore Workflows


src/materialColorScheme.scss
```diff
@@ -136,3 +136,6 @@ $service-color: #ff6c44;
 
 // Color used to indicate success in general (of an operation, verified badge, etc).
 $accent-3-dark: #00bfa5 !important;
+
+// This color is commonly used by headers, titles, and names
+$header-color: #151c3d;
```
garyluu, line 141: is the in the color palette in zeplin?

src/app/organizations/collections/collections.component.scss
```diff
@@ -1,11 +1,14 @@
+@use '~@angular/material' as mat;
```
garyluu, line 1: double-check if this is necessary

coverbeck (approved):

src/materialColorScheme.scss
kathy-t: No it's not, but with how often it's used, I wonder if it should be

denis-yuen (approved):
> Published workflow with authors 
> The published workflow page has a fixed width which makes the Authors table quite squished if all the fields are filled out. Not sure what to do about this. Are there plans to make the published workflow page fit the width of the screen, like the my workflows page?

I see what you mean. We don't have plans yet, but it might be worth keeping an eye on/considering with the way the mail and orcid icons are wrapping. 

For this specific case I wonder if it is worth moving the ORCID icon to the header since the number of ORCID characters should be fixed? (i.e. it seems like it will always wrap if the icon is inline)

Thanks for checking out all the other tables as well.

garyluu (approved):

NatalieEO (approved):

kathy-t:
> For this specific case I wonder if it is worth moving the ORCID icon to the header since the number of ORCID characters should be fixed? (i.e. it seems like it will always wrap if the icon is inline)

I'm not sure if it's worth it because if we do it for the ORCID icon, the wrapping can still happen with the email

PR #1358, opened 2021-09-23
Add PR template

denis-yuen:
https://github.com/dockstore/dockstore/pull/4475


garyluu (approved):

coverbeck (approved):
It seems like the opinionated style guide should be in dockstore-ui2, but that's beyond the scope of this.

denis-yuen:
> It seems like the opinionated style guide should be in dockstore-ui2, but that's beyond the scope of this.

Maybe because issues are consolidated

PR #1361, opened 2021-09-29
use a file download package instead of HREF

wshands:
**Description**
Descriptor file download was being done by packing descriptor contents into an HREF URL and by using 'bypassSecurityTrustUrl' to avoid angular sanitization. Instead this change will utilize a package instead of HREF to download the descriptor file. 

**Issue**
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-3092


garyluu (approved):

src/app/shared/file.service.ts
```diff
@@ -91,21 +92,16 @@ export class FileService {
...
+   * @param {string} fileName The name for the file that will be downloaded
+   * @returns {void}
    * @memberof FileService
    */
-  getFileData(content: string): SafeUrl {
-    if (content) {
-      // This uses the Data URI scheme, 'data:[<media type>][;base64],<data>'. For security
-      // purposes, it is important that the prefix is 'data:text/plain,' as this forces the user-input content to
-      // only be treated as plain text data. Other Data URI media types (such as text/html) could lead to potential XSS vulnerabilities.
-      return this.sanitizer.bypassSecurityTrustUrl('data:text/plain,' + encodeURIComponent(content));
```
coverbeck, line 106: Devil's advocate... what was wrong with this? I know SonarCloud is probably triggering off any usage off of `bypassSecurityTrustUrl`, but there are ways to use it safely.

Instead we are now adding a new 3rd party dependency, which at a brief glance doesn't seem to mention any anti-XSS capabilities, and just adds more code to our bundles.

If it's necessary, this seems fine, but just checking that the response is not overkill.

@Richard-Hansen Since you added the comments, any thoughts about how safe the current approach is?

@david4096 , since you logged the ticket?

denis-yuen (commented):
Curious too.
As discussed, I prefer that descriptors that are published just be links directly to TRS as in https://ucsc-cgl.atlassian.net/browse/DOCK-1868 

I'm not necessarily opposed to this for unpublished content or as a stop-gap, deferring to security team for their opinion.

src/app/shared/file.service.ts
Richard-Hansen: As long as the `text/plain` MIME type is explicit, I don't think this will be a security issue. I remember digging in to ways to try and override or modify the MIME type (or impact browser MIME type sniffing), and wasn't able to locate any vulnerabilities. 

That being said, different browsers support different data URI lengths ([Data URI docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs)), so if someone has an excessively large project this may start failing anyways?

david4096: On the topic of the original ticket, I didn't take a deep dive when I logged it and agree that there isn't an existing exploit.

If it solves the sonarcloud complaint by using the filesaver API I think that's good. Reduced LOC we are maintaining and making better use of modern browser features. I understand where Charles is coming from and agree that we will now be monitoring another package for vulnerabilities.

I could go either way. I would be OK with adding a sonarcloud exception here instead of the library. I think most workflows are not going to generate a download that gets to the Data URI limits.

david4096 (approved):

david4096:
> Curious too. As discussed, I prefer that descriptors that are published just be links directly to TRS as in https://ucsc-cgl.atlassian.net/browse/DOCK-1868
> 
> I'm not necessarily opposed to this for unpublished content or as a stop-gap, deferring to security team for their opinion.

I think this is worth doing while we are still generating downloads on the frontend. If we can move to not doing that (by serving everything from TRS and adding endpoints for unpublished as needed), I think that would be preferable longterm.

denis-yuen (approved):

Richard-Hansen (approved):

NatalieEO (approved):

coverbeck (changes requested):
Everybody else has approved, so I'll withdraw my objections, although I still feel this was a lot of work for something that was already safe.

I did flag a couple of nits, so I'm requesting changes, but after that I'll approve.

src/app/shared/code-editor-list/code-editor-list.component.ts
```diff
@@ -103,4 +104,9 @@ export class CodeEditorListComponent {
     this.sourcefiles[index].absolutePath = newPath;
     this.sourcefiles[index].path = newPath;
   }
+
+  downloadFileContent(content: string, filePath: string) {
+    let fileName = this.fileService.getFileName(filePath);
```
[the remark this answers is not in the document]
wshands: fixe

src/app/shared/file.service.ts
```diff
@@ -91,21 +92,16 @@ export class FileService {
...
-    } else {
-      return null;
+  downloadFileContent(content: string, fileName: string) {
+    if (content && fileName) {
+      var data = new Blob([content], { type: 'text/plain;charset=utf-8' });
```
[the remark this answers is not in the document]
wshands: fixed

PR #1366, opened 2021-10-14
Multiple users orcid export

kathy-t:
**Description**
Webservice PR: https://github.com/dockstore/dockstore/pull/4488
- Allows multiple users to export a workflow to ORCID.
- Allows one user to export to ORCID again if they already exported before (performs an update). The dialog box looks like this if they've exported before:
image

- Don't close the snapshot/export dialog if there's an error. Example vid:

https://user-images.githubusercontent.com/25287123/137328825-76035b6b-df9e-48d2-afd5-2364b1574fea.mp4

- Removed `.only` from tests

**Issue**
dockstore/dockstore#4293


coverbeck:
>     * Allows one user to export to ORCID again if they already exported before (performs an update). The dialog box looks like this if they've exported before:

You can only export snapshotted versions, so does "performs an update" actually do anything? If not, I wonder if shouldn't block the operation.

kathy-t:
The ORCID export button is now unavailable once you export to ORCID.

https://user-images.githubusercontent.com/25287123/137748984-99c82bcb-026c-4493-ad99-403dccbe0f67.mp4

I needed to add PR https://github.com/dockstore/dockstore/pull/4521 in the webservice to return the entry after the export is complete so the UI can update the workflow. Without this, the ORCID export button is still there if the user doesn't refresh the page.

This is currently using the feature branch JAR, but I'll switch it over to the develop JAR once the webservice PR is merged.

PR #1367, opened 2021-10-19
Avoid refreshing .dockstore.yml workflows

garyluu:
**Description**
Ignore the errors about .dockstore.yml. Decided against 304s due to weird httpclient behavior (and java swagger client)

**Issue**
dockstore/dockstore#3811

### DIFF
--- a/src/app/workflow/refresh-workflow-organization/refresh-workflow-organization.component.ts
+++ b/src/app/workflow/refresh-workflow-organization/refresh-workflow-organization.component.ts
@@ -56,16 +56,16 @@ export class RefreshWorkflowOrganizationComponent extends RefreshOrganizationCom
 
   refreshOrganization(): void {
     if (this.orgWorkflowObject) {
-      const workflows = this.orgWorkflowObject.published.concat(this.orgWorkflowObject.unpublished);
+      const workflows = this.orgWorkflowObject.published
+        .concat(this.orgWorkflowObject.unpublished)
+        .filter((workflow) => workflow.mode !== Workflow.ModeEnum.DOCKSTOREYML);
       from(workflows)
         .pipe(
           concatMap((workflow) => {
             this.alertService.start(`Refreshing ${workflow.full_workflow_path}`);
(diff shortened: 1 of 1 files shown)


coverbeck:
TLDR; What about 302?

Looking at Denis' httpstatues.com, which I hadn't seen before, lead me to https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html, where they both say that a 304 can be issued for a conditional GET request. [Conditional requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests) should have conditional headers. The RFC is also more emphatic about a 304 not having a body ("MUST NOT" in caps).

Since we're not issuing a conditional request (?), I'd say it shouldn't return a 304.

I'd lean towards a 302 with the location header set to the workflow, although we'd have to make sure the Angular code doesn't automatically follow because the point is to avoid the extra payload.

Or, I don't know think this is standard, 204 with no body if nothing changed and 200 if something did. But I don't know if you can preserve the types if you do that way in Jersey.

Also, I don't think these refreshes should be GETs, but that's another discussion.

coverbeck:
> Not sure if I'm following correctly, but if Jersey is rewriting 304s without body to 200s. I think it is likely that a 302 would get rewritten too if it had no body. I have no objection otherwise

I'm confused on this one by our Slack convo, where  we said UI sees 304 but Java client sees 200. I was thinking of the former in my comment. It doesn't make sense, but agreed with Gary's other comment that 3xx doesn't make sense, so it's moot.

> Right now, I'm thinking 200 (if you think refresh is akin to a very complicated PUT) or 400

Despite my earlier comments, I now lean towards 400. You cannot refresh a .dockstore.yml repo; it's a bad request.

That makes me realize, we could stop the UI from making the request in the first place with a filter [here](https://github.com/dockstore/dockstore-ui2/blob/e0665c11ed4802156aeff80bd07c3a297deafc88/src/app/workflow/refresh-workflow-organization/refresh-workflow-organization.component.ts#L59-L58).

PR #1381, opened 2021-11-04
Feature/dock 1765/categories search

svonworl:
**Description**
This PR adds Categories support to the search UI, and a small set of Category-related integration tests.

I needed to insert three lines to add the Category facet to the search interface.  If that's truly the case, and I didn't miss anything, a pat on the back to the author(s).   Good job!

The first two integration tests are borrowed from the organizations IT.  I wrote the other two ITs before I learned that we don't currently run ES in our CI tests.  I commented the latter two out, but left them in the file, in case they might someday be of use.  They pass if ES is running and has the current mappings.  It's easy to remove them if they're objectionable.

I made a few changes to commands.ts, some of which make it easier to tweak for testing against a dockerized postgresql/ES.  I don't think I broke anything for non-dockerized developers, because everything passes the CI tests, but please try it out.

**Issue**
https://github.com/dockstore/dockstore/pull/4476
https://ucsc-cgl.atlassian.net/browse/DOCK-1765


cypress/integration/group2/categories.ts
```diff
@@ -0,0 +1,84 @@
...
+      cy.get('#addEntryToCollectionButton').should('be.disabled');
+      cy.get('#selectCollection').click();
+      cy.get('mat-option').contains(categoryName).click();
+      cy.get('#addEntryToCollectionButton').should('not.be.disabled').click();
+    });
+  });
+
+  /*
+   * The following ITs were written before I discovered that the current
+   * UI2 CI process does not run Elasticsearch.  These tests will pass
```
garyluu, line 61: You can mock it https://docs.cypress.io/api/commands/intercept#Usage if you deem it worth the hassle, problably not though

garyluu (approved):

cypress/support/commands.ts
```diff
@@ -47,8 +51,7 @@ export function assertNoTab(tabName: string): any {
 export function resetDB() {
   before(() => {
     cy.exec('java -jar dockstore-webservice.jar db drop-all --confirm-delete-everything travisci/web.yml');
-    // cy.exec('docker exec -i postgres1 psql webservice_test -U postgres < travisci/db_dump.sql');
-    cy.exec('PGPASSWORD=dockstore psql -h localhost -f travisci/db_dump.sql webservice_test -U dockstore');
```
denis-yuen, line 51: this change may need more explanation, loading as postgres (the superuser) instead of dockstore can break certain features like frozen versions.

svonworl: changed

garyluu: for this PR, it was already like this. i think i originally put that in for local testing with Docker and didn't really put too much effort in making it correct

coverbeck (approved):

denis-yuen (approved):

PR #1388, opened 2021-11-05
Feature/seab 3108/display GitHub app tools

NatalieEO:
**Description**
This displays GitHub app tools by basically using the already existing workflow components under the tool routing. Please check this out locally. 

https://github.com/NatalieEO/test-workflows-and-tools

Some remaining issues are as follows:

1. The Launch tab doesn't work correctly right now. Displays the wrong text.
2. I'm not sure how the redirect back to Dockstore works when setting up the GitHub app. Does it always go back to the my-workflows page?
3. The GitHub app logs will display all the logs for workflows and tools
4. CLI won't be able to find GitHub App tools since they are workflows in the backend.
5. Haven't written tests yet.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3108

### DIFF
--- a/src/app/container/register-tool/register-tool.component.html
+++ b/src/app/container/register-tool/register-tool.component.html
@@ -41,6 +41,7 @@
                 {{ selectedOption.extendedLabel }}
               </div>
             </span>
+            {{ selectedOption.value }}
           </mat-radio-button>
         </mat-radio-group>
       </div>
@@ -53,6 +54,29 @@
       <ng-template matStepLabel>Create a tool</ng-template>
       <div *ngIf="selectedOption?.value === 0">
         <app-refresh-wizard></app-refresh-wizard>
+      </div>
+      <div *ngIf="selectedOption?.value === 3">
+        <p>
+          Navigate to GitHub to install our GitHub app on your repositories/organizations. See our
+          <a
+            href="{{ Dockstore.DOCUMENTATION_URL }}/getting-started/github-apps/github-apps.html"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            documentation <mat-icon>open_in_new</mat-icon>
+          </a>
+          for how to use GitHub Apps to automatically sync your workflows with GitHub. A .dockstore.yml file is required.
+        </p>
+        <div fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="2rem" class="mb-2">
+          <a
+            mat-flat-button
+            type="button"
+            color="primary"
+            [href]="gitHubAppInstallationLink$ | async"
+            matTooltip="Manage Dockstore Installations on GitHub"
+            ><mat-icon>add</mat-icon> Manage Dockstore Installations on GitHub</a
+          >
+        </div>
       </div>
       <form
         *ngIf="selectedOption?.value === 1"
--- a/src/app/container/register-tool/register-tool.component.html
+++ b/src/app/container/register-tool/register-tool.component.html
@@ -65,7 +65,7 @@
           >
             documentation <mat-icon>open_in_new</mat-icon>
           </a>
-          for how to use GitHub Apps to automatically sync your workflows with GitHub. A .dockstore.yml file is required.
+          for how to use GitHub Apps to automatically sync your tools with GitHub. A .dockstore.yml file is required.
         </p>
         <div fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="2rem" class="mb-2">
           <a
@@ -73,8 +73,8 @@
             type="button"
             color="primary"
             [href]="gitHubAppInstallationLink$ | async"
-            matTooltip="Manage Dockstore Installations on GitHub"
-            ><mat-icon>add</mat-icon> Manage Dockstore Installations on GitHub</a
+            matTooltip="Manage Dockstore installations on GitHub"
+            ><mat-icon>add</mat-icon> Manage Dockstore installations on GitHub</a
           >
         </div>
       </div>
--- a/src/app/mytools/my-tool/my-tool.component.ts
+++ b/src/app/mytools/my-tool/my-tool.component.ts
@@ -141,15 +141,22 @@ export class MyToolComponent extends MyEntry implements OnInit {
 
     this.getMyEntries();
 
-    this.containerService.tools$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((tools) => {
-      this.tools = tools;
-      if (this.apptools) {
-        this.allTools = this.apptools.concat(this.tools);
-      } else {
-        this.allTools = this.tools;
-      }
-      this.selectEntry(this.mytoolsService.recomputeWhatEntryToSelect(this.allTools));
-    });
+    combineLatest([this.containerService.tools$, this.workflowService.workflows$])
+      .pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe(([tools, workflows]) => {
+        this.tools = tools;
+        this.apptools = workflows;
+        if (this.tools) {
+          if (this.apptools) {
+            this.allTools = this.tools.concat(this.apptools);
+          } else {
+            this.allTools = this.tools;
+          }
+        } else {
+          this.allTools = this.apptools;
+        }
+        this.selectEntry(this.mytoolsService.recomputeWhatEntryToSelect(this.allTools));
+      });
 
     this.groupEntriesObject$ = combineLatest([this.containerService.tools$, this.toolQuery.tool$]).pipe(
       map(([tools, tool]) => {
@@ -157,15 +164,6 @@ export class MyToolComponent extends MyEntry implements OnInit {
       })
     );
 
-    this.workflowService.workflows$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((appTools) => {
-      this.apptools = appTools;
-      if (this.tools) {
-        this.allTools = this.apptools.concat(this.tools);
-      } else {
-        this.allTools = this.apptools;
-      }
-      this.selectEntry(this.mytoolsService.recomputeWhatEntryToSelect(this.allTools));
-    });
     this.groupAppToolEntryObjects$ = combineLatest([this.workflowService.workflows$, this.workflowQuery.selectActive()]).pipe(
       map(([workflows, workflow]) => {
         return this.myWorkflowsService.convertEntriesToOrgEntryObject(workflows, workflow);
@@ -177,6 +175,7 @@ export class MyToolComponent extends MyEntry implements OnInit {
         return orgToolObjects && orgToolObjects.length !== 0;
       })
     );
+
     this.registerToolService.tool.pipe(takeUntil(this.ngUnsubscribe)).subscribe((tool) => (this.registerTool = tool));
   }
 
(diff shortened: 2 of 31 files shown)


src/app/workflow/workflow.component.html
```diff
@@ -206,6 +207,7 @@ <h3>
               </ng-template>
             </mat-tab>
           </div>
+          <div *ngIf="entryType === EntryType.Tool"></div>
```
garyluu, line 210: placeholder or...?

NatalieEO: Ah nope, forgot to delete

src/app/mytools/my-tool/my-tool.component.ts
```diff
@@ -116,22 +129,49 @@ export class MyToolComponent extends MyEntry implements OnInit {
...
     this.groupEntriesObject$ = combineLatest([this.containerService.tools$, this.toolQuery.tool$]).pipe(
       map(([tools, tool]) => {
         return this.mytoolsService.convertEntriesToOrgEntryObject(tools, tool);
       })
     );
+
+    this.workflowService.workflows$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((appTools) => {
+      this.apptools = appTools;
+      if (this.tools) {
+        this.allTools = this.apptools.concat(this.tools);
```
garyluu, line 163: this.allTools is also set above in the containerService.tools$ subscribe.  forkjoin/combinelatest is more suitable

src/app/mytools/mytools.service.ts
```diff
@@ -95,21 +115,49 @@ export class MytoolsService extends MyEntriesService<DockstoreTool, OrgToolObjec
     return entry.tool_path || '';
   }
 
-  sortEntry(entryA: DockstoreTool, entryB: DockstoreTool): number {
-    const registryA = entryA.registry_string.toLowerCase();
-    const registryB = entryB.registry_string.toLowerCase();
-    const compareRegistry = registryA.localeCompare(registryB);
+  sortEntry(entryA: DockstoreTool | Workflow, entryB: DockstoreTool | Workflow): number {
```
garyluu, line 118: wow

src/app/shared/entry.ts
```diff
@@ -95,7 +95,8 @@ export abstract class Entry implements OnDestroy {
   }
 
   init() {
-    this.clearState();
+    // Getting rid of this line makes the linking work again and I didn't notice any weird behaviour, but I'm not sure.. Needs more testing
```
garyluu, line 98: what's the linking work again?

src/app/workflow/launch/launch.component.html
```diff
@@ -71,16 +71,26 @@
...
           </div>
-          -->
+        </mat-card>
+        <mat-card
+          class="m-2"
+          matTooltip="Commands for launching tool through Consonance: a DIY open-source framework for
+      spinning up VMs and running tools on AWS EC2"
+        >
+          <a href="https://github.com/Consonance/consonance/wiki">Run on AWS VMs with Consonance</a>
+          <pre>{{ consonance }}</pre>
```
garyluu, line 93: not sure what this has to do with github apps

src/app/workflow/workflow.component.html
```diff
@@ -346,7 +347,11 @@ <h3>
       </div>
     </div>
     <div fxLayout="column" fxLayoutGap="1rem" class="p-0 col-md-2 col-lg-3 ng-scope" *ngIf="isWorkflowPublic && workflow">
-      <app-launch-third-party [workflow]="workflow" [selectedVersion]="selectedVersion"></app-launch-third-party>
+      <app-launch-third-party
+        *ngIf="entryType !== EntryType.Tool"
```
garyluu, line 351: certain launch-with doesn't work?

src/app/shared/entry.ts
NatalieEO: Trying to switch between different tools on the my-tools page is broken with this line. This line clears the tool after it's been selected.

src/app/container/register-tool/register-tool.component.html
```diff
@@ -51,6 +52,29 @@ <h4 mat-dialog-title>Register Tool</h4>
...
+        <p>
+          Navigate to GitHub to install our GitHub app on your repositories/organizations. See our
+          <a
+            href="{{ Dockstore.DOCUMENTATION_URL }}/getting-started/github-apps/github-apps.html"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            documentation <mat-icon>open_in_new</mat-icon>
+          </a>
+          for how to use GitHub Apps to automatically sync your workflows with GitHub. A .dockstore.yml file is required.
```
denis-yuen, line 65: In this case, `sync your tools` maybe

src/app/workflow/launch/launch.component.html
denis-yuen: Not sure either, we should comment this out if it hasn't already though.

denis-yuen (approved):
Probably a bigger gap than normal, but don't want to be a blocker while OOO

src/app/mytools/mytools.service.ts
```diff
@@ -57,13 +60,30 @@ export class MytoolsService extends MyEntriesService<DockstoreTool, OrgToolObjec
...
+      });
   }
+
   selectEntry(tool: DockstoreTool | Workflow | null): void {
     if (tool && tool.id) {
-      this.containersService.getContainer(tool.id, includesValidation).subscribe((result) => {
-        this.location.go('/my-tools/' + result.tool_path);
-        this.containerService.setTool(result);
-      });
+      if (tool.path.includes('github.com')) {
```
coverbeck, line 74: Suggest making a method out of this, like `isAppTool`. I had a hard time figuring out what this does at first.

You could go full bore and create a [user defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards), although the link shows as deprecated and TypeScript may have a new way of doing it.

src/app/mytools/mytools.service.ts
coverbeck: This method is pretty complex. Suggestions:

1. Add test(s) in the spec file
2. Use methods/user guards for your if statements (`if (isAppTool(entryA))` is easier to understand than `if ('full_workflow_path' in entryA)`)

src/app/workflow/workflow.component.html
NatalieEO: Idk if it doesn't work, but we don't ever display the launch-with buttons for tools

garyluu (approved):
Just assuming this works and other issues will be created for what's left

src/app/mytools/my-tool/my-tool.component.ts
```diff
@@ -116,27 +129,53 @@ export class MyToolComponent extends MyEntry implements OnInit {
...
-    this.containerService.tools$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((tools) => {
-      this.tools = tools;
-      this.selectEntry(this.mytoolsService.recomputeWhatEntryToSelect(tools));
-    });
+    combineLatest([this.containerService.tools$, this.workflowService.workflows$])
+      .pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe(([tools, workflows]) => {
+        this.tools = tools;
+        this.apptools = workflows;
+        if (this.tools) {
```
coverbeck, line 149: I think you could simply this block to, unless there is code relying on `this.tools` or `this.workflows` being `null` further on:

```
this.tools = tools || [];
this appTools = workflows || [];
this.allTools = this.tools.concat(this.appTools)
```

src/app/mytools/mytools.service.ts
```diff
@@ -57,15 +61,45 @@ export class MytoolsService extends MyEntriesService<DockstoreTool, OrgToolObjec
...
+        this.containersService.getContainer(tool.id, includesValidation).subscribe((result) => {
+          this.location.go('/my-tools/' + result.tool_path);
+          this.containerService.setTool(result);
+          this.workflowService.setWorkflow(null);
+        });
+      }
+    }
+  }
+
+  static isAppTool(tool: DockstoreTool | AppTool | OrgToolObject<DockstoreTool> | OrgWorkflowObject<Workflow>): tool is AppTool {
```
coverbeck, line 91: We generally don't use static methods like this, although I know it's a hassle to inject the service. Gary didn't object, so I guess it's OK. :)

coverbeck, line 91: It doesn't sound right that an OrgToolObject is an AppTool

src/app/shared/entry.ts
```diff
@@ -124,6 +125,13 @@ export abstract class Entry implements OnDestroy {
     }
   }
 
+  protected isAppTool(url: String): boolean {
+    if (url.includes('containers') && url.includes('github.com')) {
```
coverbeck, line 129: If I have a workflow with `containers` in its name, will this return `true`?

src/app/shared/myentries.service.ts
```diff
@@ -16,13 +16,13 @@
...
 import { EntryType } from './enum/entry-type';
-import { DockstoreTool, Workflow } from './swagger';
+import { AppTool, DockstoreTool, Workflow } from './swagger';
 import { UrlResolverService } from './url-resolver.service';
 
 export abstract class MyEntriesService<E extends DockstoreTool | Workflow, O extends OrgToolObject<E> | OrgWorkflowObject<E>> {
   constructor(protected urlResolverService: UrlResolverService) {}
 
-  recomputeWhatEntryToSelect(entries: E[]): E | null {
+  recomputeWhatEntryToSelect(entries: Array<DockstoreTool | Workflow>): DockstoreTool | Workflow | null {
```
coverbeck, line 25: Why did this have to change from `E`? And elsewhere in this file?

src/app/workflow/launch/launch.component.html
```diff
@@ -71,17 +71,27 @@
...
+            dockstore.org and other
+            <a href="https://github.com/common-workflow-language/cwltool#use-with-ga4gh-tool-registry-api">GA4GH tool registries</a> when
+            all inputs and outputs are available on the local filesystem or via HTTP(S):
+            <pre>{{ dockstoreSupportedCwlLaunch }}</pre>
+            cwltool can also be used to create the input template:
             <pre>{{ dockstoreSupportedCwlMakeTemplate }}</pre>
           </div>
-          -->
         </mat-card>
+        <!--        <mat-card-->
```
coverbeck, line 87: We don't expect this to come back, I would just delete it.

src/app/shared/entry.ts
NatalieEO: yes, it will return true

src/app/shared/myentries.service.ts
NatalieEO: I think I went through so many iterations I didn't realize this wasn't necessary anymore. Changed it back

src/app/shared/entry.ts
coverbeck: Wouldn't that be wrong? It's a real edge case (I don't see any workflows with "containers" in their giturl or workflow name in the current DB), but it doesn't seem like an unreasonable word to use either. I don't think it would be that hard to catch the case -- not sure what the exact url is, but either a regex or `url.includes('/containers/')` or something like that.

coverbeck (approved):

src/app/shared/entry.ts
NatalieEO: Yeah, you're right. It would be wrong. I'll change the regex to `url.includes('/containers/github.com')`

PR #1391, opened 2021-11-09
need to use git ref name

NatalieEO:
**Description**
Use the name of the github reference, not the docker image tag name. They aren't always the same. This will actually update the links on a number of tools which may or may not have been broken links before. There were some tags that don't have the same reference and name, but the link to github was not broken because they happened to have a tag named the same anyway.

**Issue**
dockstore/dockstore#4554


garyluu (approved):

denis-yuen (commented):
Yeah, should probably be hotfix along with the cwl fix

PR #1392, opened 2021-11-09
Prevent tab change infinite loop

garyluu:
**Description**
It was shown that rapidly changing the tabs before the page has finished loading causes the infinite loop.

For some reason the event gets triggered even though it's not the active tab. Do a quick check.

Note: Rapidly changing tabs is incredibly noticeable by most users. There may be another infinite loop elsewhere that I can't find.

**Issue**
For SEAB-3548


coverbeck (approved):
Nice find!

Confused how it gets in this state, but tried running it locally and I'm no longer able to reproduce the problem.

wshands (approved):

denis-yuen (approved):

NatalieEO (approved):

PR #1394, opened 2021-11-11
Move error message to header

garyluu:
**Description**
Simply snaps the existing error message to the top.

**Issue**
dockstore/dockstore#4538

### DIFF
--- a/src/app/container/register-tool/register-tool.component.html
+++ b/src/app/container/register-tool/register-tool.component.html
@@ -13,10 +13,13 @@
   ~    See the License for the specific language governing permissions and
   ~    limitations under the License.
   -->
-<h4 mat-dialog-title>Register Tool</h4>
+
+<div mat-dialog-title>
+  <h4>Register Tool</h4>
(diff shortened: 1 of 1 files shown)


denis-yuen (approved):
> Wonder if we should do that for all dialogs that potentially has a scroll bar 

Possibly, although that seems like potentially all dialogues if the window is small enough. (which isn't an objection)

NatalieEO:
> > Wonder if we should do that for all dialogs that potentially has a scroll bar
> 
> Possibly, although that seems like potentially all dialogues if the window is small enough. (which isn't an objection)
Can make another ticket for it, but I think we should. I noticed in the original issue, Gary commented that the alert should be visible if the window is sufficiently big, but this hasn't been the case for me. Even on my connected monitor, I can't see any of the error message at the top of this dialog if I'm scrolled down.

NatalieEO (approved):

coverbeck (approved):

wshands (approved):

PR #1395, opened 2021-11-11
Calculate has imports and show warning

garyluu:
**Description**
Show a simple warning in the workflow version info tab

**Issue**
For dockstore/dockstore#4560

### DIFF
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -384,6 +384,11 @@
             </span>
           </div>
         </div>
+        <div *ngIf="selectedVersion?.frozen && hasHttpImports" class="pt-3">
+          <mat-card class="alert alert-warning" role="alert">
+            <mat-icon>warning</mat-icon> This version is snapshotted but contains HTTP imports.
+          </mat-card>
+        </div>
       </div>
     </mat-card-content>
   </mat-card>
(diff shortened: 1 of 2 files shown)


NatalieEO (approved):

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -384,6 +384,11 @@
             </span>
           </div>
         </div>
+        <div *ngIf="selectedVersion?.frozen && hasHttpImports" class="pt-3">
```
denis-yuen, line 387: What does this look like in terms of screenshot?
Thinking it would be nice to link somewhere for more info. We don't have somewhere great for now, but perhaps link to https://docs.dockstore.org/en/stable/advanced-topics/snapshot-and-doi.html?highlight=imports#what-is-included-in-a-snapshot

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -384,6 +384,11 @@
...
+          <mat-card class="alert alert-warning" role="alert">
+            <mat-icon>warning</mat-icon> This version is snapshotted but contains HTTP imports.
```
coverbeck, line 389: It may not be immediately apparent to all end users why this is warning-worthy. I think you should have a fuller explanation here, and/or Denis' suggested link (and that doc page may need updating too).

wshands (approved):

garyluu:
Wonder if it's not prominent enough 🤔 

image

garyluu:
image

Also considering moving the warning up higher 🤔 🤷

denis-yuen (approved):
Yeah, probably higher

garyluu:
image

coverbeck (approved):

PR #1396, opened 2021-11-16
Feature/seab 2928/circleci audit

NatalieEO:
**Description**
This adds a new check on circleci that compares the number of vulnerabilities on develop with the number vulnerabilities on the pull request by counting how many times the word `High` appears in the audit report. If there's more on the PR, the test will fail. But since this is new, the scripts don't exist on develop, so that's why the wgets are there. This test should be made optional in the GitHub settings.

you can see an example of the test failing by looking at 
https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/6754/workflows/06e116c3-9f75-4974-8534-d2288a577685/jobs/22727

https://github.com/dockstore/dockstore-ui2/pull/1396/commits/db198b7ffbd05134bc8a26d9bf1d8eed39f32e14

 
**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-2928


scripts/compare-num-vulns.sh
```diff
@@ -0,0 +1,12 @@
...
+set -o pipefail
+set -o nounset
+
+PR_NUM_VULNS=$(head -n 1 compare-num-vulnerabilities.txt)
+DEV_NUM_VULNS=$(sed -n '2p' compare-num-vulnerabilities.txt)
+
+if [ "$PR_NUM_VULNS" -gt "$DEV_NUM_VULNS" ]; then
+  echo "You have introduced $((PR_NUM_VULNS-DEV_NUM_VULNS)) high vulnerabilities."
+  exit 1
+  fi
```
denis-yuen, line 12: Ridiculous whitespace formatting nitpick

denis-yuen (commented):
Not sure what's going on with the circle ci status for the audit step. 
Github shows passing but when I click-in, it has "Wait for SSH sessions" as a step, clicking through "everything" and back to the audit step it says that the "install dependencies step failed with 
```
found 31 vulnerabilities (16 moderate, 15 high)
  run `npm audit fix` to fix them, or `npm audit` for details
bash: initialize_job_control: no job control in background: Bad file descriptor
npm ERR! code ENOAUDIT
npm ERR! audit Your configured registry (https://registry.npmjs.org/) may not support audit requests, or the audit endpoint may be temporarily unavailable.
...
npm ERR! audit <title>prod-frontdoor-lb-west.internal.npmjs.com | 502: Bad gateway</title>
```

Transient issue combined with some debugging via SSH maybe?

garyluu (commented):
You've checked and there's no GitHub integrations that is able to do this?

NatalieEO:
> Not sure what's going on with the circle ci status for the audit step. Github shows passing but when I click-in, it has "Wait for SSH sessions" as a step, clicking through "everything" and back to the audit step it says that the "install dependencies step failed with
> 
> ```
> found 31 vulnerabilities (16 moderate, 15 high)
>   run `npm audit fix` to fix them, or `npm audit` for details
> bash: initialize_job_control: no job control in background: Bad file descriptor
> npm ERR! code ENOAUDIT
> npm ERR! audit Your configured registry (https://registry.npmjs.org/) may not support audit requests, or the audit endpoint may be temporarily unavailable.
> ...
> npm ERR! audit <title>prod-frontdoor-lb-west.internal.npmjs.com | 502: Bad gateway</title>
> ```
> 
> Transient issue combined with some debugging via SSH maybe?

huh...I don't know why that step is there

NatalieEO:
> You've checked and there's no GitHub integrations that is able to do this?

Okay, I checked the GitHub integrations yesterday and they don't do quite the same thing as this. They both will leave a long PR comment about any vulnerabilities found, not just new ones. One has a handy features of creating tickets for us, but we don't make security issues on GitHub anyway.

NatalieEO:
> Transient issue combined with some debugging via SSH maybe?

Okay, when i did this yesterday, there was some weird transient issue where it wasn't reachable. I think I just chose the wrong rerun step. Although, I'm not totally sure how important `bash: initialize_job_control: no job control in background: Bad file descriptor`

denis-yuen (approved):
output looks sane

garyluu (approved):

.circleci/config.yml
```diff
@@ -84,6 +84,44 @@ jobs:
...
+          name: Run npm audit
+          #Only run 'npm ci' if node_modules was not restored from cache
+          command: |
+            bash -i -c "npm i"
+            bash -i -c "npm audit > vulnerabilities.txt" || true
+            bash scripts/npm-audit.sh
+            git stash
+            git checkout develop
+            bash -i -c 'npm i'
+            bash -i -c 'npm audit > vulnerabilities.txt' || true
```
coverbeck, line 116: Could you instead just commit the vulnerabilities.txt file to develop and compare with that?

On the one hand, this makes our vulnerabilities visible to all. On  the other hand, anybody could run `npm audit` themselves and figure it out very easily.

I'm going to say no to this idea, but I'll leave in the comment in case anybody else likes it. :)

.circleci/config.yml
```diff
@@ -84,6 +84,44 @@ jobs:
...
+          command: |
+            if [[ -n "${CIRCLE_PULL_REQUEST}" ]]
+            then
+              git fetch origin +refs/pull/${CIRCLE_PULL_REQUEST##*/}/merge:
+              git checkout -qf FETCH_HEAD
+            fi
+      - install_container_dependencies
+      - run:
+          name: Run npm audit
+          #Only run 'npm ci' if node_modules was not restored from cache
```
coverbeck, line 108: Confused by this comment; you don't run `npm ci` here ever, but comment implies you might.

And why don't you run `npm ci`? Won't `npm i` skip package-lock.json, which has the actual dependencies we are using?

Not sure on my questions, just asking.

scripts/npm-audit.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -o errexit
+set -o pipefail
+set -o nounset
+
+HIGH_VULN_DEV="$(grep -o -i High vulnerabilities.txt | wc -l)"
```
coverbeck, line 6: Nit-pick, if you both fix one high vulnerability and introduce a new one, it would not get caught, would it?

.circleci/config.yml
denis-yuen: :thinking: you as in circle automatically checks in and commits, or developer managed like the third party licenses file?

coverbeck: I was thinking like the third party licenses file.

scripts/npm-audit.sh
coverbeck: I think you need to grep for Critical as well. Or maybe only critical?

package-lock.json
```diff
@@ -837,7 +837,8 @@
       "dependencies": {
         "ansi-regex": {
           "version": "5.0.0",
-          "resolved": "",
+          "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-5.0.0.tgz",
```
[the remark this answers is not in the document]
garyluu: interesting, i think something may have went wrong with my previous npm audit PR then.

david4096 (approved):

scripts/npm-audit.sh
NatalieEO: Yeah it wouldn't get caught

.circleci/config.yml
NatalieEO: I left that comment in by accident. 

`npm i` doesn't skip the package-lock.json and may even write to that file. You're supposed to use this when you install a new package in the package.json. But you were right, I should have been using `npm ci` anyway. I wasn't before because I added a dependency in order to make the tests fail and `npm ci` will fail when there is a mismatch between the package.json and package-lock.json...and there was a mismatch because I didn't run `npm i`.

package-lock.json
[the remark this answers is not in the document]
NatalieEO: I think this changed because i ran `npm i` locally which `when used without arguments, (npm i) it may write to package-lock.json to lock down the version of some dependencies if they are not already in this file`

https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci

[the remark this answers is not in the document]
coverbeck: Interesting, I don't know how they weren't resolved in the first place. But looking at it closely, it looks like not versions actually changed, which would be my concern.

coverbeck (approved):
I know it's a real edge case, but if it ever happens, it could be bad, so I think it's worth a followup ticket to handle the fix one/add one vulnerability issue some day. It shouldn't be too hard to address.

PR #1397, opened 2021-11-16
Add container classes

garyluu:
**Description**
In order to remove bootstrap, the main hurdles left are the grid and container class usages. The idea of us removing our container usage throughout the site seems...unlikely. Copying the container classes instead (at least until after we actually remove bootstrap 3). Might need to do the same for grid.  Notable differences: 
for 1200px <= screen size width < 1400px, it's now 60px smaller.  
for screen width >= 1400px, it's now 120px wider (1320px, this is what i see on my desktop monitors which should be the same for most users/developers). This is in fact even larger than the search mockup (1275px)

So temporarily, most of the classes added are redundent until we actually remove bootstrap 3.

You may want to check this out and see for yourself. Most of our pages (search, organizations, about, logged out homepage) is wider and have less white space.  In particular, the search table is less cramped and workflow tabs aren't paginated (see test changes) because there's enough room.

**Issue**
Adhoc issue. Though kind of related to https://github.com/dockstore/dockstore/issues/1899

### DIFF
--- a/cypress/integration/immutableDatabaseTests/workflowDetails.ts
+++ b/cypress/integration/immutableDatabaseTests/workflowDetails.ts
@@ -103,7 +103,6 @@ describe('Dockstore Workflow Details', () => {
 
   describe('Change tab to dag', () => {
     it('Change to fullscreen and back', () => {
-      cy.get('.mat-tab-header-pagination-after').click();
       goToTab('DAG');
       goToTab('DAG');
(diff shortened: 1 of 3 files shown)


coverbeck (approved):
I did try locally, nothing obviously different, except no need to scroll the tabs on the workflow page (nice).

denis-yuen (approved):

wshands (approved):

NatalieEO (approved):

PR #1400, opened 2021-11-19
Feature/dock 1765/entry page category buttons

svonworl:
**Description**
Third and final Categories PR of the initial UI implementation, which adds category buttons to the workflow and tool pages.  

The appearance is pretty close to the mockups, except that the indentation of the button row is similar to the existing "Last Modified" text directly above, so it matches.   It looked funny otherwise.

**Update: the below has been edited to reflect a new approach @ 2021/Nov/19 3:35pm PST**

To avoid a race condition bug, `EntryCategoriesService.updateEntryCategories()` is implemented slightly differently than other similar existing methods.  Before the HTTP request, the current update subscription (if it exists) is unsubscribed, and then set to the subscription for the new HTTP request.  This avoids the case where the updates for entry A and B overlap, the responses get reordered, and the categories of A are stored last and displayed on B's page.  Depending on the protocols and software involved, this could happen due to network problems and/or thread scheduling vagaries.  Imagine the user quickly navigating to one entry page and then another, and a server thread receiving the first categories request, and then blocking or not being scheduled again until another thread completes the second categories request.  Or, suppose the user is connected via HTTP/3-QUIC and packet loss delays the first request/response.  Similarly, HTTP/1.1 with multiple connections.  This kind of thing doesn't happen much, but it does happen.

This type of bug probably affects most of the existing Service methods that retrieve and store collections of objects.  

**Issue**
dockstore/dockstore#4476
https://ucsc-cgl.atlassian.net/browse/DOCK-1765

**Screenshots**
Screen Shot 2021-11-18 at 5 35 01 PM

### DIFF
--- a/src/app/categories/state/entry-categories.service.ts
+++ b/src/app/categories/state/entry-categories.service.ts
@@ -35,11 +35,12 @@ export class EntryCategoriesService {
   }
 
   /**
-   * Updates the list of categories.
+   * Updates the list of categories for the specified entry.
    */
   updateEntryCategories(id: number) {
     this.entryCategoriesStore.setLoading(true);
     this.entryCategoriesStore.setError(false);
+    this.entryCategoriesStore.set([]);
     this.entriesService
       .entryCategories(id)
       .subscribe(
--- a/src/app/categories/state/entry-categories.service.ts
+++ b/src/app/categories/state/entry-categories.service.ts
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2021 OICR and UCSC
+ *
+ * Licensed under the Apache License, Version 2.0 (the &quot;License&quot;);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an &quot;AS IS&quot; BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { Category, EntriesService } from '../../shared/openapi';
+import { EntryCategoriesStore } from './entry-categories.store';
+import { EntryCategoriesQuery } from './entry-categories.query';
+
+@Injectable({ providedIn: 'root' })
+export class EntryCategoriesService {
+  public categories$: Observable<Array<Category>>;
+
+  constructor(
+    private entryCategoriesStore: EntryCategoriesStore,
+    private entryCategoriesQuery: EntryCategoriesQuery,
+    private entriesService: EntriesService,
+  ) {
+    /**
+     * An observable list of the categories for a given entry.
+     */
+    this.categories$ = this.entryCategoriesQuery.selectAll();
+  }
+
+  /**
+   * Updates the list of categories.
+   */
+  updateEntryCategories(id: number) {
+    this.entryCategoriesStore.setLoading(true);
+    this.entryCategoriesStore.setError(false);
+    this.entriesService
+      .entryCategories(id)
+      .subscribe(
+        (categories: Array<Category>) => {
+          this.entryCategoriesStore.setLoading(false);
+          this.entryCategoriesStore.setError(false);
+          this.entryCategoriesStore.set(categories);
+        },
+        () => {
+          this.entryCategoriesStore.setLoading(false);
+          this.entryCategoriesStore.setError(true);
+        }
+      );
+  }
+}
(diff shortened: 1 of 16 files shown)


garyluu (commented):
if the edge case is going to be something you want to commonly handle, make a test for it.

additionally, it looks like if it does happen, you're stuck in an infinite loading state?

svonworl:
I don't think it results in an infinite loading state - the loading state should be set by the handlers for the response to the most recent request.  However, I realized...

A simpler solution is to cancel the previous subscription before each successive call.  Same effect, easier to understand.  I will implement it, push, and update the PR description.

garyluu:
> I don't think it results in an infinite loading state - the loading state should be set by the handlers for the response to the most recent request. However, I realized...
> 
> A simpler solution is to unsubscribe the previous subscription before each successive call. Same effect, easier to understand. I will implement it, push, and update the PR description.

I misread a statement in the code. shouldn't have an infinite loading state.

coverbeck (commented):
Was here, going to wait for

> A simpler solution is to unsubscribe the previous subscription before each successive call. Same effect, easier to understand. I will implement it, push, and update the PR description.

You may mean this, but I think the `switchMap` suggested by Gary would accomplish this.

More generally, maybe I'm wrong, but I feel this is pretty edge-casey and sometimes the nature of the web. What if you have two tabs open; the first tab won't reflect changes made in the second tab, etc.

Unless we move to web sockets, I don't think there is an all-encompassing solution, and I wouldn't  recommend spending  a lot of time at this point trying to figure this out.

But if your pattern works, isn't too hard to do, and can easily be applied elsewhere, I don't object.

svonworl:
I do agree it's an edge case, but when it happens, it results in the wrong information being displayed.  Doesn't matter much in this case (the wrong categories get displayed), but if something more important got foobared, it could be a bigger issue.  Maybe it would cause the user to accidentally modify or delete something, or trigger a bug that we spend days to track down, ...

In any case, the revised fix is pretty simple, and understandable enough so that it might not even need a test.  Take a look and let me know what you think.

src/app/container/container.component.html
```diff
@@ -104,9 +104,14 @@ <h3>
     </div>
   </div>
   <div class="row m-1">
-    <div *ngIf="tool" class="col-sm-12" style="margin-bottom: 5px">
-      <span *ngIf="!labelsEditMode && !starGazersClicked">
-        <mat-chip-list>
+    <div *ngIf="tool">
+      <span *ngIf="(categories$ | async).length > 0" class="inline-block mr-5 mt-1 mb-1">
+        <span class="size-12 mr-2">Categories</span>
+        <app-category-button *ngFor="let category of (categories$ | async)" [category]="category" entryType="tool" class="weight-500"></app-category-button>
```
garyluu, line 110: aren't all app-category-button instances visually identical? How come this has a different font-weight?

src/styles.scss
```diff
@@ -851,6 +851,20 @@ body {
   vertical-align: middle;
 }
 
+// Set various font attributes
+.weight-500 {
+  font-weight: 500;
+}
+
+.size-12 {
```
garyluu, line 859: PR checklist, need a comment for this.

From what I can tell, the mockups (not including headings) only use 14px (default) or 12px. There's no 13, 11.9, etc. If true and in order to maintain consistency, the comment should mention the use case for this.  Would hopefully prevent others from using `<small`, or font-size: small, font-size: smaller, and we would use this else where instead.

garyluu (commented):
new solution suprisingly works as well as switchmap as far as i can tell and is simpler unlike the first solution (doesn't cancel http responses)

src/app/categories/state/entry-categories.service.ts
```diff
@@ -0,0 +1,64 @@
...
+  private currentSubscription: Subscription = null;
+
+  /**
+   * Updates the list of categories for the specified entry.
+   */
+  updateEntryCategories(entryId: number) {
+    this.entryCategoriesStore.setLoading(true);
+    this.entryCategoriesStore.setError(false);
+    this.entryCategoriesStore.remove();
+    this.currentSubscription?.unsubscribe();
```
coverbeck, line 49: If `updateEntryCategories` were invoked with two different entry ids in succession, the response for the first entry could get unsubscribed prematurely, and its response ignored. But it looks like the store only has categories for one entry anyway, it may not matter and might actually be the desired behavior.

coverbeck (approved):
Pending addressing of Gary's CSS comments.

src/app/categories/state/entry-categories.service.ts
svonworl: What you describe is the desired behavior.  I implemented this service/store similarly to the existing services/stores, which store a single unit of information which ostensibly corresponds to an item on the current page.  Essentially, my little tweak ignores the first response if it arrives after another request is launched, because we've changed/updated the page and don't want/need the first response any more.

I have some ideas about dynamically-allocated stores which don't suffer from this race condition, and would allow us to do things like retrieve the categories for each of multiple entries on the same page, but that's a discussion for some other day...

src/app/container/container.component.html
svonworl: Currently, the app-category-buttons inherit the font-weight, which was a choice I made after misinterpreting the mockups, and probably a mistake.  I am going to harmonize them all to the same weight, and eliminate the weight-500 style.  Will flag you for re-review after I push.

garyluu (approved):

PR #1402, opened 2021-11-19
Feature/search facet refactoring

garyluu:
**Description**
Refactor the search components to:
- split the search facets to a new sidebar component for modularity
- try to update the state all in one go instead of micro updates which changes the view slightly
- the sidebar and table queries are executed concurrently with their results returned in one block
- put more things in the state (like filters) so that derived values (filterkeys, hasfilters) are never out of sync with the filters

You're going to want to test this out because there's many things that can potentially be broken

Scenario: click on WDL bucket on search page
Before:
image

After:
image

**Issue**
Adhoc issue trying to see if I can find an infinite loop (couldn't but probably has to do with advancedsearch and searchtext subscriptions which will be removed in another issue).


garyluu:
Forgot to remove styles

denis-yuen:
Came here from linked issue, what state is this in? 
Does this need reviewers and was CI cancelled due to an outage?

garyluu:
> Came here from linked issue, what state is this in? Does this need reviewers and was CI cancelled due to an outage?

Need to do some cleanup (when time permits). Also need to check that nothing broke (since we're lacking in tests)

PR #1404, opened 2021-11-29
DOCK-1958: Disable delete menu option for default tag, inform on tag deletion error

svonworl:
**Description**
Prevents the user from deleting a tool's default tag.

The referenced issue noted that the user received no feedback when an attempting to delete the default tag (which is not allowed) failed on the webservice.  This PR implements some UI feedback via the alert service.

However, that's a band-aid fix, because the UI should simply not give the user the option to delete the default tag.

To that end, this PR disables the "Delete" button if the tag is the default tag, using the same disable code as the "Set as Default Version" button directly below it.

I user tested both the "notice" change, and then the "disable delete" change.  The latter makes a code path through the former very rare.

I can write integration tests for either, but IMHO, my time is better invested elsewhere.  Your call.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1958
https://github.com/dockstore/dockstore/issues/4536

### DIFF
--- a/src/app/container/view/view.component.html
+++ b/src/app/container/view/view.component.html
@@ -20,7 +20,7 @@
   <button type="button" mat-menu-item color="accent" *ngIf="isPublic$ | async" (click)="setMode(TagEditorMode.View); (false)">View</button>
   <ng-container *ngIf="(isPublic$ | async) === false">
     <button data-cy="editTagButton" type="button" mat-menu-item color="accent" (click)="setMode(TagEditorMode.Edit); (false)">Edit</button>
-    <button type="button" mat-menu-item color="warn" *ngIf="isManualTool" (click)="deleteTag(); (false)">Delete</button>
+    <button type="button" mat-menu-item color="warn" *ngIf="isManualTool" [disabled]="(isRefreshing$ | async) || version.name === tool.defaultVersion" (click)="deleteTag(); (false)">Delete</button>
     <button
       type="button"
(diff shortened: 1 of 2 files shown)


coverbeck (commented):
Is the underlying behavior that you can't delete the default tag ever, or that you can't delete the default tag if it's the only tag? Both of those seem wrong off-hand, but I may not have the whole context. I vaguely remember for workflows that we would auto select a new default...

If we do disable the button, there should ideally be a visible reason why, but it may be tricky to fit it in.

garyluu:
> Is the underlying behavior that you can't delete the default tag ever, or that you can't delete the default tag if it's the only tag? Both of those seem wrong off-hand, but I may not have the whole context. I vaguely remember for workflows that we would auto select a new default...
> 
> If we do disable the button, there should ideally be a visible reason why, but it may be tricky to fit it in.

I don't think we have a convention for disabled items, but for now a tooltip is probably good enough to show the reason.

garyluu (approved):

NatalieEO (approved):

coverbeck (approved):
If we're saying the web service is going to fail on this (which I'm still not convinced is right), then this PR is fine.

On whether to have tooltip, see our [Slack discussion](https://app.slack.com/archives/C16ET3CF4/p1636063751145400) -- IMO we should consider tooltips nice-to-have at best and not essential purveyors of information.

PR #1406, opened 2021-12-01
Feature/seab 3656/topic sentence

garyluu:
**Description**
Add topic sentence display to workflow information card. Also add the ability to update it (like workflow path).

Forgot to mention...it's slightly different than existing info tab buttons and such because it's closer to mocks. Presumably the rest of the info tab is converted to the mock ups before the next release.

Soon to be replaced by phase 2

**Issue**
For SEAB-3656

### DIFF
--- a/src/app/container/info-tab/info-tab.component.html
+++ b/src/app/container/info-tab/info-tab.component.html
@@ -251,7 +251,6 @@
               <span fxFlex>
                 <input
                   *ngIf="topicEditing"
-                  minlength="3"
                   maxlength="256"
                   type="text"
                   name="topic"
--- a/src/app/container/info-tab/info-tab.component.html
+++ b/src/app/container/info-tab/info-tab.component.html
@@ -258,7 +258,7 @@
                   class="w-100"
                   data-cy="topicInput"
                   [(ngModel)]="tool.topic"
-                  placeholder="Short description of the workflow"
+                  placeholder="Short description of the tool"
                 />
               </span>
               <button
--- a/src/app/container/info-tab/info-tab.component.html
+++ b/src/app/container/info-tab/info-tab.component.html
@@ -246,7 +246,7 @@
           <!-- This looks different than the others ones because it's new and doesn't use bootstrap 3 classes -->
           <li>
             <form fxLayout="row" fxLayoutAlign="start center" class="w-100">
-              <span><strong matTooltip="Short descripton of the tool">Topic</strong>:</span>
+              <span><strong matTooltip="Short descripton of the tool">Topic</strong>:&nbsp;</span>
               <span *ngIf="!topicEditing">{{ tool?.topic || 'n/a' }}</span>
               <span fxFlex>
                 <input
--- a/src/app/container/info-tab/info-tab.component.html
+++ b/src/app/container/info-tab/info-tab.component.html
@@ -243,6 +243,49 @@
               </button>
             </form>
           </li>
+          <!-- This looks different than the others ones because it's new and doesn't use bootstrap 3 classes -->
+          <li>
+            <form fxLayout="row" fxLayoutAlign="start center" class="w-100">
+              <span><strong matTooltip="Short descripton of the tool">Topic</strong>:</span>
+              <span *ngIf="!topicEditing">{{ tool?.topic || 'n/a' }}</span>
+              <span fxFlex>
+                <input
+                  *ngIf="topicEditing"
+                  minlength="3"
+                  maxlength="256"
+                  type="text"
+                  name="topic"
+                  class="w-100"
+                  data-cy="topicInput"
+                  [(ngModel)]="tool.topic"
+                  placeholder="Short description of the workflow"
+                />
+              </span>
+              <button
+                data-cy="topicCancelButton"
+                *ngIf="!isPublic && topicEditing"
+                type="button"
+                mat-button
+                class="accent-1-dark"
+                (click)="cancelEditing()"
+              >
+                <mat-icon>cancel</mat-icon> Cancel
+              </button>
+              <button
+                *ngIf="!isPublic"
+                type="button"
+                [disabled]="!(!somethingIsBeingEdited() || topicEditing)"
+                mat-button
+                class="accent-1-dark"
+                data-cy="topicEditButton"
+                (click)="toggleEditTopic()"
+              >
+                <mat-icon *ngIf="topicEditing">save</mat-icon>
+                <mat-icon *ngIf="!topicEditing">edit</mat-icon>
+                {{ topicEditing ? 'Save' : 'Edit' }}
+              </button>
+            </form>
+          </li>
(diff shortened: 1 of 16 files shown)


src/app/container/info-tab/info-tab.component.html
```diff
@@ -243,6 +243,49 @@
...
+                <input
+                  *ngIf="topicEditing"
+                  minlength="3"
+                  maxlength="256"
+                  type="text"
+                  name="topic"
+                  class="w-100"
+                  data-cy="topicInput"
+                  [(ngModel)]="tool.topic"
+                  placeholder="Short description of the workflow"
```
coverbeck, line 261: workflow -> tool

src/app/container/info-tab/info-tab.service.ts
```diff
@@ -85,6 +86,29 @@ export class InfoTabService extends Base {
...
   }
 
+  setTopicEditing(editing: boolean) {
+    this.topicEditing$.next(editing);
+  }
+
+  saveTopic(tool: DockstoreTool, callback: () => void) {
+    this.alertService.start('Updating topic');
+    const partialTool = this.getPartialToolForUpdate(tool);
+    console.log(tool.topic);
```
coverbeck, line 96: Suggest removing.

src/app/container/info-tab/info-tab.service.ts
```diff
@@ -85,6 +86,29 @@ export class InfoTabService extends Base {
...
+    this.containersService.updateContainer(this.tool.id, partialTool).subscribe(
+      (response) => {
+        this.alertService.detailedSuccess();
+
+        const newTopic = response.topic;
+        console.log(newTopic);
```
coverbeck, line 102: I'd remove.

src/app/container/info-tab/info-tab.service.ts
```diff
@@ -85,6 +86,29 @@ export class InfoTabService extends Base {
     this.wdlTestPathEditing$.next(editing);
   }
 
+  setTopicEditing(editing: boolean) {
+    this.topicEditing$.next(editing);
+  }
+
+  saveTopic(tool: DockstoreTool, callback: () => void) {
```
coverbeck, line 93: Suggest renaming `callback` to something like `errorCallback`.

src/app/workflow/info-tab/info-tab.service.ts
```diff
@@ -80,6 +85,26 @@ export class InfoTabService {
     this.forumUrlEditing$.next(editing);
   }
 
+  /**
+   * Warning, this could potentially update a few other properties
+   * @param workflow
+   */
+  saveTopic(workflow: Workflow, callback: () => void) {
```
coverbeck, line 92: Same suggestion on renaming `callback`.

coverbeck (approved):

NatalieEO (approved):

denis-yuen:
> Forgot to mention...it's slightly different than existing info tab buttons and such because it's closer to mocks. Presumably the rest of the info tab is converted to the mock ups before the next release. 

:+1:

denis-yuen:
> Also add the ability to update it (like workflow path). 

Huh, interesting. Missed this in the webservice, who wins if it gets updated in Dockstore and on GitHub?

denis-yuen (commented):
Display looks good. 
Unsure about editing

garyluu:
> Display looks good. Unsure about editing (as opposed to update on refresh + github app events)

Refresh doesn't automatically set it because it potentially wipes out some other source that the user manually sets. So that shouldn't be done at all (unless it's for a new workflow which it currently does).

GitHub apps events are version level, so it generally should not be affecting workflow level information.  It also has the disadvantage of only affecting GitHub Apps workflows so we need some way to set it for new non-GitHub Apps workflows (otherwise it would pull from GitHub repo and the user potentially doesn't want it)

denis-yuen:
> Refresh doesn't automatically set it because it potentially wipes out some other source that the user manually sets. So that shouldn't be done at all (unless it's for a new workflow which it currently does). 

That's kinda why I'm thinking there shouldn't be any other source, manual or otherwise. It simplifies things if GitHub is the source of truth. Otherwise, we end up with topic sentences being up-to-date after migration but quickly falling out of date if people don't notice they have to edit themselves after editing the topic on GitHub.

> GitHub apps events are version level, so it generally should not be affecting workflow level information. 

I think this is ok. We have `publish` and  `latestTagAsDefault` already

> It also has the disadvantage of only affecting GitHub Apps workflows so we need some way to set it for new non-GitHub Apps workflows  

This would be refresh or if we build out "bitbucket apps" someday (unlikely though based on usage).

garyluu:
> > Refresh doesn't automatically set it because it potentially wipes out some other source that the user manually sets. So that shouldn't be done at all (unless it's for a new workflow which it currently does).
> 
> That's kinda why I'm thinking there shouldn't be any other source, manual or otherwise. It simplifies things if GitHub is the source of truth. Otherwise, we end up with topic sentences being up-to-date after migration but quickly falling out of date if people don't notice they have to edit themselves after editing the topic on GitHub.
> 
> > GitHub apps events are version level, so it generally should not be affecting workflow level information.
> 
> I think this is ok. We have `publish` and `latestTagAsDefault` already
> 
> > It also has the disadvantage of only affecting GitHub Apps workflows so we need some way to set it for new non-GitHub Apps workflows
> 
> This would be refresh or if we build out "bitbucket apps" someday (unlikely though based on usage).

GitHub Apps can't act as the source of truth for updates because once the admin endpoint that pulls info from the GitHub repo's description is ran, half of the workflows and all of the tools on Dockstore will potentially have unwanted topics with no way to change it. 

So for simplicity, I agree that there should ideally be one source for updates (the update endpoint) which works for all types of workflows (hosted, github apps, other) and tools. The topic shouldn't change often enough for it to be a hassle. Also changing the topic on the site is faster than making a new commit with a modified .dockstore.yml and merging it to develop (to prevent the below)

GitHub Apps topic (even as an alternative method) suffers from the same issue as publish. Two developers are making changes to the same workflow on GitHub on different branches. Dockstore just bounces back and forth between different publish and topic states.

Just like how GitHub Apps has the publish feature, the Dockstore website also has a publish button which would mirror topics if we decided to add topic to the .dockstore.yml.

I think it's one thing to favor GitHub Apps by adding more features to it, but it's another thing to cripple all the other non-GitHub Apps workflows to do so.

wshands (approved):

garyluu:
This has major problems with workflows (which was implemented with existing bootstrap classes)

garyluu:
Changed a few things. Truncate on two lines in search. Made workflow info tab component not use bootstrap which fixes the wrapping issue

NatalieEO (approved):

coverbeck (commented):
Does this need to be updated given our discussions about the backend storing both the GitHub description and the user's override?

PR #1414, opened 2021-12-14
Feature/3142/automated smoke tests

Richard-Hansen:
**Description**
This PR executes the UI smoke tests as a part of the CircleCI test suite. This will allow us to identify UI changes that would break smoke tests before they've been merged.

**To be able to run the smoke tests in CircleCI the following was added:**
1. Elasticsearch is now run in CircleCI.
2. A new DB dump was created. This is full of dummy data and workflows.
3. Some minor changes to the existing smoke tests, in case they accessed 'live' data, or expected all workflows to be registered on GitHub.

**Some notes:**
1. Since we don't want any real tokens in the DB, I manually added fake tokens for various services.
2. All the current tool entries are 'fake' (don't link to anywhere real) and workflows are all hosted. 
3. In the future, it may be nice to have some helper scripts for creating dummy data. Specifically, manually inserting entries into the database can end up causing constraint problems (especially when adding more data later) and makes working with the database very difficult. This is why I created a fresh `smoke_test_db.sql` rather than replace `db_dump.sql`.
4. Since there aren't any real users in the DB, the way I logged in was by modifying the `ng2-ui-auth_token` in the browser, and set it to a token in the database, e.g. `fake-admin-token`.

**What to do if smoke tests fail during a PR:**
If your PR fails the CircleCI smoke tests, you may need to update the tests and/or update `smoke_test_db.sql`, depending on what your PR is changing. 
1. Running `docker exec -t postgres1 pg_dump -U postgres --column-inserts webservice_test > smoke_test_db.sql` will create a new DB dump.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3142


scripts/run-webservice-script.sh
```diff
@@ -15,6 +15,6 @@ chmod u+x dockstore-webservice.jar
 psql -h localhost -c "create user dockstore with password 'dockstore' createdb;" -U postgres
 psql -h localhost -c "ALTER USER dockstore WITH superuser;" -U postgres
 psql -h localhost -c 'create database webservice_test with owner = dockstore;' -U postgres
-psql -h localhost -f travisci/db_dump.sql webservice_test -U postgres
+psql -h localhost -f travisci/${DB_DUMP:-db_dump.sql} webservice_test -U postgres
```
coverbeck, line 18: Ideally we would get rid of the `travisci` directory, moving the files in it somewhere else -- I was confused by it at first.

scripts/run-webservice-script.sh
```diff
@@ -15,6 +15,6 @@ chmod u+x dockstore-webservice.jar
...
 java -jar dockstore-webservice.jar db migrate -i 1.5.0,1.6.0,1.7.0,add_service_1.7.0,1.8.0,1.9.0,1.10.0,alter_test_user_1.10.2,1.11.0,1.12.0 travisci/web.yml
```
coverbeck, line 19: Do you need most of these? I know they don't hurt, but if if most of the migrations have already been applied...

coverbeck (commented):
Looks really good.

The obvious thing that could still happen is that the data mismatch causes an issue, but there's not an easy solution for that.

NatalieEO (approved):

garyluu (approved):

svonworl (approved):

kathy-t (approved):

scripts/run-webservice-script.sh
Richard-Hansen: Moved to `test/` https://github.com/dockstore/dockstore-ui2/pull/1414/commits/a5bfaf92a225d281531050e9a7326bbf722c136f

scripts/run-webservice-script.sh
Richard-Hansen: `db_dump.sql` needs them, `smoke_test_db.sql` doesn't, so it's easier to leave as-is for the moment.

coverbeck (approved):

PR #1417, opened 2021-12-16
Feature/dock 510/enforce default tag on publish

svonworl:
**Description**
This PR prevents the user from publishing a workflow/tool without a default version.  If the user attempts to publish an entry without a default version, an informational dialog pops up, instructing them to select a default version.  To facilitate that action, when the dialog is closed, the interface switches to the Versions tab.

In support, I have created a new informational dialog component, which displays a short message and a single button that closes it.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-510
https://github.com/dockstore/dockstore/issues/970

**Screenshot**
Screen Shot 2021-12-16 at 10 04 39 AM


src/app/container/container.component.ts
```diff
@@ -223,6 +223,15 @@ export class ContainerComponent extends Entry implements AfterViewInit, OnInit {
     }
   }
 
+  /**
+   * Select the Versions tab
+   *
+   * @memberof WorkflowComponent
+   */
+  public selectVersionsTab() {
+    this.selectTab(this.validTabs.indexOf('versions'));
```
coverbeck, line 232: I have mixed feelings about this. I don't like unexpected navigation by a site, it can confuse me. And I don't think we do this sort of thing elsewhere in Dockstore.

On the other hand, it's not changing the whole page, nor the URL, just the active tab. And if as a user you wanted to address the issue right away, you would go to the versions tab, assuming you understood what a default version is.

Ideas:

1. Leave it as is
2. Don't do it at all
3. Update the dialog message saying you're going to be redirected to the versions tab.

src/app/shared/entry-actions/entry-actions.service.ts
```diff
@@ -100,11 +104,31 @@ export class EntryActionsService {
...
+
+  openNoDefaultDialog(entry: Entry, showVersions: EventEmitter<void>): void {
+    const informationDialogData: InformationDialogData = {
+      title: 'Default Version Required',
+      message: 'Your tool/workflow must have a default version to be published.  Please use the the Actions menu in the Versions tab to select a default version.',
+      closeButtonText: 'OK',
+    };
+    const observable = this.informationDialogService.openDialog(informationDialogData, bootstrap4mediumModalSize);
+    if (showVersions != null) {
+      observable.subscribe(() => { showVersions.emit(); });
```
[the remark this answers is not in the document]
svonworl: I think it's ok as is: "Any notification Observables will complete when the dialog closes."
https://material.angular.io/components/dialog/overview

src/app/container/container.component.ts
svonworl: My rationale for switching to the version tab: to increase the chances that the user would successfully select a default version.  I'd recommend firing up this PR locally and trying it both ways, I feel like switching to the Versions tab reduces the friction quite a bit.

That said, I will implement it however the team wants...

coverbeck: My comments were based on having tried it locally -- it hadn't clicked at first from reading the code that it would change my active tab, and then I saw it in action.

As I said, I have mixed feelings about it, not strong enough to actively oppose it -- I was proposing some options for consideration.

cypress/integration/group2/myworkflows.ts
```diff
@@ -444,6 +444,23 @@ describe('Dockstore my workflows', () => {
...
+      cy.get('#publishButton').should('contain', 'Unpublish').should('be.visible').click();
+      cy.get('#publishButton').should('contain', 'Publish').should('be.visible').click();
+      cy.get('[data-cy=close-dialog-button]').should('be.visible').click();
+      cy.get('#publishButton').should('contain', 'Publish').should('be.visible');
+    });
+    it('should be able to publish after setting default version', () => {
+      goToTab('Versions');
+      cy.contains('button', 'Actions').should('be.visible').click();
+      cy.get('[data-cy=set-default-version-button]').should('be.visible').click();
+      cy.wait(1000);
```
garyluu, line 459: See if the wait could be removed, try replacing it with an assertion that checks if the default version has been switched. if that doesn't work, leave it.

garyluu (approved):
Don't really have a preference for the tab switching. we don't have a set practice for it. could look at external sites/sources for a 2nd opinion.

denis-yuen (approved):
ok with tab switching, probably would feel differently about navigating to an entirely new page

coverbeck (approved):

PR #1418, opened 2021-12-16
Feature/dock 1990/harmonize label category etc elements

svonworl:
**Description**
This PR refactors app-category-button so that, under the hood, it is a mat-chip, instead of a mat-button <a> tag.  Also, some of the associated styles are refactored.

Initially, I had grand aspirations creating a single component that would be used to render all those little rounded "chippy" boxes that appear throughout our UI.  It was going to render to a mat-button anchor tag, because my reading of the Material docs seemed to confirm that mat-chips are for non-navigational actions, and a regular button should be used otherwise.  Then, I happened upon a Material demo that featured "static" mat-chips.  If a mat-chip could be static, then certainly, if a mat-chip  navigated to another page, that would be fine...

So,, I took the path of least resistance and made an app-category-button a mat-chip under the hood, thus harmonizing it with other similar elements.

I've refactored the styles and selected a term for those little roundish chippy rectangles: "bubbles".  The class "bubble" should be added to each bubble.  The class "bubble-list" can be attached to a surrounding mat-chip-list, and will space the bubbles consistently.  A bubble defaults to a gray background, and I've separated out the styles that can be used to change the background color from the bubble style itself.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1990
https://github.com/dockstore/dockstore/issues/4580

**Screenshot**
The UI looks pretty much the same as before.

- [x ] Do not bypass Angular sanitization (bypassSecurityTrustHtml, etc.), or justify why you need to do so


coverbeck (approved):
One optional suggestion... Going to defer to Gary on the styling details.

garyluu (approved):

denis-yuen (approved):

PR #1425, opened 2022-01-14
prevent edit collection click event from propagating to parent

svonworl:
**Description**
This PR prevents a click on the "edit collection" button from propagating to the parent collection summary, which triggered the page to change to the collection page behind the "edit collection" dialog.  Fix is similar to that used to prevent the same type of problem on the "remove entry from collection" button.

Also, implemented the "nice to have" goal: in the "edit collection" dialog, changed the save button label from "Edit Collection" to "Save".

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1756
https://github.com/dockstore/dockstore/issues/4166

**Screenshot**
Screen Shot 2022-01-14 at 9 22 14 PM

### DIFF
--- a/src/app/organizations/collections/create-collection/create-collection.component.html
+++ b/src/app/organizations/collections/create-collection/create-collection.component.html
@@ -14,49 +14,49 @@
    limitations under the License.
 -->
 <ng-container *ngIf="title$ | async as title">
-<ng-container *ngIf="saveLabel$ | async as saveLabel">
-<div>
-  <h1 mat-dialog-title>{{ title }}</h1>
-  <div mat-dialog-content>
-    <app-alert></app-alert>
-    <form [formGroup]="createCollectionForm" fxLayout="column">
-      <mat-form-field>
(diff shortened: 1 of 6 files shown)


src/app/organizations/collections/create-collection/create-collection.component.html
```diff
@@ -13,7 +13,9 @@
    See the License for the specific language governing permissions and
    limitations under the License.
 -->
-<div *ngIf="title$ | async as title">
+<ng-container *ngIf="title$ | async as title">
+<ng-container *ngIf="saveLabel$ | async as saveLabel">
```
coverbeck, line 17: Shouldn't this be indented as it's a nested block?

coverbeck (approved):

src/app/organizations/collections/create-collection/create-collection.component.html
svonworl: done

garyluu (approved):

denis-yuen (approved):

PR #1427, opened 2022-01-14
remove trailing hyperlink underlines

denis-yuen:
**Description**
Replacing hyperlinked trailing spaces with non-hyperlinked non-breaking spaces

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2007

### DIFF
--- a/src/app/workflow/workflow.component.html
+++ b/src/app/workflow/workflow.component.html
@@ -106,7 +106,11 @@
       <span *ngIf="(categories$ | async).length > 0" class="inline-block mr-5 mt-1 mb-1">
         <span class="size-small mr-2">Categories</span>
         <mat-chip-list class="bubble-list">
-          <app-category-button *ngFor="let category of (categories$ | async)" [category]="category" entryType="workflow"></app-category-button>
+          <app-category-button
+            *ngFor="let category of categories$ | async"
+            [category]="category"
+            entryType="workflow"
+          ></app-category-button>
         </mat-chip-list>
       </span>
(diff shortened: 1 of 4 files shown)


denis-yuen:
for Monday: husky doing some extra formatting

svonworl:
Found more trailing spaces in the "Source Repositories" box on both the tool and container pages.

src/app/workflow/workflow.component.html
```diff
@@ -106,7 +106,11 @@ <h3>
       <span *ngIf="(categories$ | async).length > 0" class="inline-block mr-5 mt-1 mb-1">
         <span class="size-small mr-2">Categories</span>
         <mat-chip-list class="bubble-list">
-          <app-category-button *ngFor="let category of (categories$ | async)" [category]="category" entryType="workflow"></app-category-button>
+          <app-category-button
```
svonworl, line 109: For the sake of knowledge, was there anything wrong with this tag, besides the extra parens?

garyluu (approved):

src/app/workflow/workflow.component.html
denis-yuen: I doubt it, this is one of the automatically husky reformatings

denis-yuen:
> Found more trailing spaces in the links in the "Source Repositories" box on the tool and container pages.

That's what I thought too at first, then I noticed that the "pop-out" icon is part of the same hyperlink. So instead of having "link space same-link", I just left it alone.

NatalieEO (approved):

PR #1428, opened 2022-01-14
some github app tool fixes.

NatalieEO:
**Description**
Uses tool icon now
Fixes the navigation bug when there are unpublished tools
Not a ticket, but also fixes an api call

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3827
https://ucsc-cgl.atlassian.net/browse/SEAB-3780

### DIFF
--- a/src/app/mytools/mytools.service.ts
+++ b/src/app/mytools/mytools.service.ts
@@ -128,8 +128,12 @@ export class MytoolsService extends MyEntriesService<DockstoreTool, OrgToolObjec
     });
   }
 
-  getPath(entry: DockstoreTool): string {
-    return entry.tool_path || '';
+  getPath(entry: DockstoreTool | Workflow): string {
+    if (MytoolsService.isWorkflowBasedObject(entry)) {
+      return entry.full_workflow_path;
+    } else {
+      return entry.tool_path || '';
+    }
   }
 
   sortEntry(entryA: DockstoreTool | Workflow, entryB: DockstoreTool | Workflow): number {
--- a/src/app/mytools/mytools.service.ts
+++ b/src/app/mytools/mytools.service.ts
@@ -130,7 +130,7 @@ export class MytoolsService extends MyEntriesService<DockstoreTool, OrgToolObjec
 
   getPath(entry: DockstoreTool | Workflow): string {
     if (MytoolsService.isWorkflowBasedObject(entry)) {
-      return entry.full_workflow_path;
+      return entry.full_workflow_path || '';
     } else {
       return entry.tool_path || '';
     }
--- a/src/app/workflow/workflow.component.html
+++ b/src/app/workflow/workflow.component.html
@@ -46,9 +46,14 @@
     <ngx-json-ld [json]="schema"></ngx-json-ld>
     <div fxFlex="80">
       <h3 id="workflow-path" data-cy="workflowTitle" fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="1rem">
-        <span>
+        <span *ngIf="entryType === EntryType.BioWorkflow || entryType === EntryType.Service; else gitHubAppTool">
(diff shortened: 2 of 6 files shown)


src/app/mytools/mytools.service.ts
```diff
@@ -128,8 +128,12 @@ export class MytoolsService extends MyEntriesService<DockstoreTool, OrgToolObjec
     });
   }
 
-  getPath(entry: DockstoreTool): string {
-    return entry.tool_path || '';
+  getPath(entry: DockstoreTool | Workflow): string {
+    if (MytoolsService.isWorkflowBasedObject(entry)) {
+      return entry.full_workflow_path;
+    } else {
+      return entry.tool_path || '';
```
coverbeck, line 135: I think it's harmless the way you have it, but can tool_path be null/undefined?

src/app/workflow/workflow.component.ts
```diff
@@ -429,10 +433,16 @@ export class WorkflowComponent extends Entry implements AfterViewInit, OnInit {
...
-      this.gA4GHFilesService.updateFiles(prefix + this.workflow.full_workflow_path, this.selectedVersion.name, [
-        this.descriptorTypeCompatService.stringToDescriptorType(this.workflow.descriptorType),
-      ]);
+      if (this.entryType === EntryType.AppTool) {
+        this.gA4GHFilesService.updateFiles(this.workflow.full_workflow_path, this.selectedVersion.name, [
+          this.descriptorTypeCompatService.stringToDescriptorType(this.workflow.descriptorType),
+        ]);
+      } else {
+        let prefix = this.entryType === EntryType.BioWorkflow ? ga4ghWorkflowIdPrefix : ga4ghServiceIdPrefix;
+        this.gA4GHFilesService.updateFiles(prefix + this.workflow.full_workflow_path, this.selectedVersion.name, [
```
coverbeck, line 442: This a small thing, but the only thing different between this block and line 437 is the first param, so I would only have `updateFiles` once in the code; it makes it easier to read.

The if/else could just set up the first parameter, either as is, or in a method.

src/app/workflow/workflow.component.html
```diff
@@ -46,9 +46,14 @@ <h2 class="text-center">The workflow linked to could not be found!</h2>
     <ngx-json-ld [json]="schema"></ngx-json-ld>
     <div fxFlex="80">
       <h3 id="workflow-path" data-cy="workflowTitle" fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="1rem">
-        <span>
+        <span *ngIf="entryType === EntryType.BioWorkflow || entryType === EntryType.Service; else gitHubAppTool">
```
coverbeck, line 49: I suggest an explicit check rather than relying on an else. If another type ever gets added, we don't want it to default to a tool icon.

src/app/workflow/workflow.component.ts
```diff
@@ -312,18 +311,23 @@ export class WorkflowComponent extends Entry implements AfterViewInit, OnInit {
     this.selectTab(this.validTabs.indexOf('versions'));
   }
 
+  public getWorkflowSubclass(entryType: EntryType): WorkflowSubClass {
+    let subclass: WorkflowSubClass;
```
coverbeck, line 315: If you make this a switch statement, and then if we're ever able to turn on StrictNullCheck we can have the compiler catch any missing values: https://stackoverflow.com/a/55532967/257235

coverbeck (approved):

src/app/mytools/mytools.service.ts
NatalieEO: I'm not sure how it would be tbh.. But I should have put entry.full_workflow_path || '' to be consistent since that is what the other getPath function has. Changing

denis-yuen (approved):

garyluu (approved):

PR #1429, opened 2022-01-17
Versions Tab UI Fixes

KimberleyChong:
**Description**
JIRA Ticket: [Dock-2015](https://ucsc-cgl.atlassian.net/browse/DOCK-2015?atlOrigin=eyJpIjoiM2M1MzIzNWRmYjE2NDRhZGI1NGNiYTRhNDIzMWM0NjYiLCJwIjoiaiJ9)

### Without changes:
image

### With changes:
1. Added padding to left of first column.
2. Changed selection colour to lavender from mock ups. (not seen: hover colour was also changed to a more translucent version of the same lavender colour, transparency was not specified in mock ups)
3. This was not specified as a fix in the ticket but I noticed the “Actions” header was off center from the actions drop downs which looked odd.  I decided to align the header similar to the mock up, if we don’t want to include this change I can remove it.

image

From the mock ups it appears this page has a refresh that needs to be implemented so this would just be a temporary fix before the full refresh:
image


NatalieEO (approved):
Thanks for the arrows

garyluu:
Adding Kathy because she did table related changes recently

KimberleyChong:
> Adding Kathy because she did table related changes recently

Sounds good. I pulled Kathy's changes earlier today and doubled checked, my changes should be unaffected

denis-yuen (commented):
2 and 3 make sense to me
In the screenshot, it seems like 1 is adding whitespace before every row in the table (rather than just the selected version) and so they aren't lined up with the header on the left.

kathy-t:
> 3. This was not specified as a fix in the ticket but I noticed the “Actions” header was off center from the actions drop downs which looked odd.  I decided to align the header similar to the mock up, if we don’t want to include this change I can remove it.

I think it only looks off because the button doesn't have an outline. If you hover over the button or change it to a solid colour button like below, it's aligned with the Actions header

Screenshot 2022-01-18 at 14-06-34 Dockstore Workflow github com DataBiosphere topmed-workflows UM_variant_caller_wdl

We could change the button type to one that shows the button outline, leave it as it was before, or change the padding of the button?

KimberleyChong:
> 2 and 3 make sense to me In the screenshot, it seems like 1 is adding whitespace before every row in the table (rather than just the selected version) and so they aren't lined up with the header on the left.

Okay, I misinterpreted 1. as adding padding to each entry but I see now it's supposed to be just the selected one

KimberleyChong:
> > 3. This was not specified as a fix in the ticket but I noticed the “Actions” header was off center from the actions drop downs which looked odd.  I decided to align the header similar to the mock up, if we don’t want to include this change I can remove it.
> 
> I think it only looks off because the button doesn't have an outline. If you hover over the button or change it to a solid colour button like below, it's aligned with the Actions header
> 
> Screenshot 2022-01-18 at 14-06-34 Dockstore Workflow github com DataBiosphere topmed-workflows UM_variant_caller_wdl
> 
> We could change the button type to one that shows the button outline, leave it as it was before, or change the padding of the button?

That makes sense, should we go ahead and update the button to the updated one in the mockups? The button outline will show and align with the original header like you suggested @kathy-t 

image

kathy-t:
> That makes sense, should we go ahead and update the button to the updated one in the mockups? The button outline will show and align with the original header like you suggested @kathy-t
> 
> image

Note that the screenshot of the mockup with the Info button is for public workflows: https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6021a99cc016953e3ceafa8c

Private workflows still have an Actions button that looks a little different: https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/602c1be9c7ca5718b6e31377

@denis-yuen Thoughts on whether to change the buttons now?

denis-yuen:
I think it makes sense, it does have to be done sometime or later to match the mock-up.

garyluu:
Reminder: If this is the first time you're submitting a PR or even if you just need a refresher, consider reviewing our [style guide](https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide#pr-checklist)

KimberleyChong:
Trying to update CircleCi tests for button changes, failing due to redesign with outdated checks. I don't know if I can test them locally, so I am pushing changes to test

KimberleyChong:
### Updated PR Changes:

Updated "Actions" buttons according to current mock ups for public and private workflows. Reverted "Actions" header cell to original position, fixed padding, and updated CircleCi tests.

### Public Workflow Changes:
image

1. Padding for selected version (5px)
2. "Info" button (changed from "Action" button)
Zeplin Mock up: https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6021a99cc016953e3ceafa8c

### Private Workflow Changes:
image

1. Padding for selected version (5px)
2. Updated "Actions" button and icons
3. Updated drop down menu style
Zeplin Mock up: https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/602c1be9c7ca5718b6e31377

### Updates to CircleCi Tests
I needed to update the CircleCi tests since some became outdated with the redesign which was causing failures.  I spoke with Charles and know now that I can test them locally, sorry for that! If test changes need to be made in the future I will run them locally first.

garyluu:
Spun out mat-raised-button style changes to another issue because it requires a bit of work

KimberleyChong:
### Updates:

- After speaking with Gary, the info button will not be in its final form until we do a global change of the mat-raised-button style to match the new design.  It will be left with the current mat-raised-button style, without the bolded text, lighter shadow and smaller text/button size of the new designs.

image

- The new purple button style now has a global style under button.private-btn (I named it private-btn since the purple buttons only seem to appear on the pages under "My Dockstore" and on no public pages)
- There is also a new global style structure under button.small-btn-structure that follows the new smaller button design of the mock ups 

image

There are a few merge conflicts that'll need to be resolved...

coverbeck (commented):
I just ran this locally; the public view looks a lot cleaner with your changes.

However, the Info button on the public page acts very weird, IMHO:

Screen Shot 2022-01-25 at 11 20 58 AM

It looks like a button, but brings up a one menu item drop-down.

I'm not sure how the mocks have it and I see there was some conversation about the buttons in the PR, i.e., it may already have been discussed, but I find GitHub PR convos hard to follow at times), but I think:

1. Clicking Info should take you directly to the dialog
2. If there are going to be cases where there is more than one menu choice possible, then the button should revert to a dropdown with the arrow.

KimberleyChong:
> I just ran this locally; the public view looks a lot cleaner with your changes.
> 
> However, the Info button on the public page acts very weird, IMHO:
> 
> Screen Shot 2022-01-25 at 11 20 58 AM
> 
> It looks like a button, but brings up a one menu item drop-down.
> 
> I'm not sure how the mocks have it and I see there was some conversation about the buttons in the PR, i.e., it may already have been discussed, but I find GitHub PR convos hard to follow at times), but I think:
> 
> 1. Clicking Info should take you directly to the dialog
> 2. If there are going to be cases where there is more than one menu choice possible, then the button should revert to a dropdown with the arrow.

There hasn't been a discussion concerning the drop-down and Zeplin doesn't seem to specify a drop-down.  Looking at the logic for the public button it appears there can only be one option, which is "view", it does make sense that because it looks like a button now it should probably go directly to the pop up. As of right now I left the drop-down as is, not sure if this is something to change in this PR or in the next @denis-yuen 

image

denis-yuen:
I think that if there is only the one choice and it looks like a button, it should probably just be a button. 
The zeplin mock-up is unclear, but I think the intent is a button

KimberleyChong:
**Update:** Removed drop-down for public button, clicking "Info" will immediately bring up pop up.

image

src/app/workflow/versions/versions.component.css
```diff
@@ -39,5 +39,10 @@ th {
 /* Highlight row with the selected version */
 /* The mat-row is so that this is more specific and able to override the hover style above */
 tr.mat-row.highlight {
-  background-color: rgba(0, 0, 0, 0.12);
+  background-color: rgba(232, 234, 246);
```
kathy-t, line 42: This colour is #e8eaf6 which is in our material colour scheme so you could use the color from the palette. Not sure if you can use it for the `.mat-row:hover` though because you're specifying the transparency. If you can't then I would just add a comment saying that the colour is from the colour scheme.

src/styles.scss
```diff
@@ -1230,3 +1230,21 @@ $truncate-text-values: 1, 2, 3;
 .mat-header-cell:not(:last-of-type) {
   padding-right: 1.6rem;
 }
+
+// This style should be used for buttons and drop downs in the private "My Dockstore" pages, they do not seem to exist in public dockstore pages
+// These redesigned buttons have purple text and white backgrounds
+button.private-btn {
+  font-weight: bold;
```
kathy-t, line 1237: The private button has bolded text, but the public Info button doesn't. Should they both be bolded?

src/app/workflow/versions/versions.component.css
KimberleyChong: mm thanks for pointing that out, I made that change before I knew about the colour palette. I'm updating that now, I'll also be converting the file to scss.  The transparency is also doable so that'll be changed as well

src/styles.scss
KimberleyChong: The public Info button will be bolded eventually, it falls under the mat-raised-button style which will have a global change with this [ticket](https://github.com/dockstore/dockstore/issues/4680) to update them with bolded text and etc. So to avoid redundant styling it'll be left so until then

KimberleyChong:
Updated colour codes:
image
image

coverbeck (approved):
Actions button looks real nice.

kathy-t (approved):

src/app/workflow/view/view.component.css
```diff
@@ -13,3 +13,11 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
+
+/* Drop-down menu restyling to match new design: bolded text & smaller height */
```
garyluu, line 17: Oddly enough, this does appear to be a unique dropdown (different than navbar top right, different than search)

garyluu (approved):

denis-yuen (approved):

PR #1437, opened 2022-01-26
Commenting out DNAStack launch with button

Richard-Hansen:
**Description**
This PR comments out the DNAStack launch-with button:
Screen Shot 2022-01-26 at 9 18 49 AM

But keeps the multi-cloud launch (For when that config option is set):
Screen Shot 2022-01-26 at 9 32 29 AM

This ticket also requests a change to the documentation. The documentation already includes the latest launch-with button images (shown in the second image above), so there may not be anything to change. That being said, will we be using the new launch-with buttons in the next release? If not, I can revert to the older launch with images and adjust the documentation accordingly. (DNAStack launch-with documentation, if curious: https://docs.dockstore.org/en/stable/launch-with/dnastack-launch-with.html)

**Issue**
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-3800

### DIFF
--- a/src/app/workflow/launch-third-party/launch-third-party.component.html
+++ b/src/app/workflow/launch-third-party/launch-third-party.component.html
@@ -89,18 +89,19 @@
           </div>
           <ng-template #oldWdlLaunch>
             <div fxLayout="column" fxLayoutAlign="space-between center" fxLayoutGap="0.625rem">
-              <a
-                mat-raised-button
-                class="old-launch-with-button"
-                [matTooltip]="(hasContent$ | async) ? '' : 'The WDL workflow has no content.'"
-                target="_blank"
-                rel="noopener"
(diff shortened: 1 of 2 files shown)


denis-yuen (commented):
Could you cherry-pick this for hotfix as well? With a deadline of roughly the 12th, I think this will need to be a hotfix

> This ticket also requests a change to the documentation. The documentation already includes the latest launch-with button images (shown in the second image above), so there may not be anything to change. That being said, will we be using the new launch-with buttons in the next release? If not, I can revert to the older launch with images and adjust the documentation accordingly. (DNAStack launch-with documentation, if curious: https://docs.dockstore.org/en/stable/launch-with/dnastack-launch-with.html)

It does look like DNAstack will not be committing to starting multi-cloud launch when they shutdown their global endpoint. So I'm thinking that that we can leave the launch-with page in our backpocket (for later use) but unlink it from the language support table and the general index of pages.

Richard-Hansen:
Closing in favor of hotfix PRs

PR #1443, opened 2022-02-01
Feature/dock 1106/fix broken org logo images

svonworl:
**Description**
Currently, if the admin-specified org logo image is not loadable, in its place, the UI displays a broken image icon and/or the img tag's alt text, depending upon the browser.  This PR displays a default org logo image, instead of the broken image/alt text, using an entirely client-side mechanism.  An org with no admin-specified org logo image is rendered as before.

When a browser cannot successfully load an image, it fires an `error` event on the corresponding `img` tag:
https://developer.mozilla.org/en-US/docs/Web/API/Element/error_event

Edit: functionally, the PR has changed substantially, the description below is updated:
This PR adds an attribute directive on `img` tags which handles the load error event by setting `img.src` to the fallback image URL specified by the `appFallback` attribute.  The directive includes logic to set `img.src` no more than once, to avoid an infinite load loop if the fallback image is also not loadable.

I made the change in four places in the UI where the org logo appears:
* organizations page
* organization page
* collection page
* my organizations page

Did I miss anywhere?  Possibly.

I selected the default org logo, which is a modified version of this CC0 (public domain) svg:
https://www.svgrepo.com/svg/70462/organization

The goal for the default org logo was something generic that suggested the notion of an organization and was simple and non-descript enough to blend into the UI.  I think it meets those goals, but I'd recommend we engage our UI person regarding its suitability.  It is easy to change.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1106
dockstore/dockstore#3070

**Screenshot**
Screen Shot 2022-01-31 at 1 38 48 PM
Screen Shot 2022-01-31 at 6 51 21 PM


src/app/shared/org-logo.service.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class OrgLogoService {
+  setDefault(img: any) {
+    // Check/set a flag to ensure the default logo is set only once, to avoid an infinite load loop if the default org image is not loadable
+    // IMPORTANT: BEFORE YOU CHANGE THE BELOW CODE, MAKE SURE YOU COMPLETELY UNDERSTAND IT!
```
denis-yuen, line 7: this seems a bit more alarming than normal, is that intentional?

svonworl: Yup, intentionally scary.  If the following code is incorrectly modified (that is, the "change" check is removed/broken) and the default image is not loadable for whatever reason, the browser will request the default image over and over again, hitting our server many times per second, until the page changes.  I can tone down the warning if you want. :)

garyluu (approved):

coverbeck (commented):
What if you did this instead as a component or directive? Then its usage could something like (my Angular component syntax is rusty, could be off):

```
<app-img-with-fallback [src] ="gravatarUrl$ | async" class="big-org-logo" 
alt="organization logo" 
fallbackImage="../../../../assets/images/dockstore/default-org-logo.svg">
</app-img-with-fallback>
```

You could also have a `fallbackText` attribute and use the same component for your collections feature. And maybe the fallbackImage is an enum instead of a path, like DEFAULT_ORG_LOGO.

* Makes it reusable for non-org use cases, if ever have any
* Makes it clear what the fallback is by looking at the HTML.
* Mainly hides `$event.target`, which I think you should in Angular

If you think you could/should reuse the component for the collections, then it's probably worth the change; if not, then borderline.

svonworl:
> What if you did this instead as a component or directive? Then its usage could something like (my Angular component syntax is rusty, could be off):
> 
> ```
> <app-img-with-fallback [src] ="gravatarUrl$ | async" class="big-org-logo" 
> alt="organization logo" 
> fallbackImage="../../../../assets/images/dockstore/default-org-logo.svg">
> </app-img-with-fallback>
> ```
> 
> You could also have a `fallbackText` attribute and use the same component for your collections feature. And maybe the fallbackImage is an enum instead of a path, like DEFAULT_ORG_LOGO.
> 
> * Makes it reusable for non-org use cases, if ever have any
> * Makes it clear what the fallback is by looking at the HTML.
> * Mainly hides `$event.target`, which I think you should in Angular
> 
> If you think you could/should reuse the component for the collections, then it's probably worth the change; if not, then borderline.

Before wrapping the img in an Angular component, I would suggest the other ticket be completed, and we can refactor if desired.  The component solution is more complex and, in my experience, introduces annoying quirks/issues.  Not sure if it's worth it in this case.

Honestly, I probably would have implemented this with an attribute directive:
https://angular.io/guide/attribute-directives

ala
``
or
``
or whatever.

But, the `img[default]` selector would add a bit of overhead to the templates, and given the spirit of the "no functions in templates" mandate, I thought that it might elicit non-positive reactions, so I did not.

coverbeck:
> But, the img[default] selector would add a bit of overhead to the templates, and given the spirit of the "no functions in templates" mandate, I thought that it might elicit non-positive reactions, so I did not.

Can you please elaborate on that? Where would the function in the template be coming from?

svonworl:
> > But, the img[default] selector would add a bit of overhead to the templates, and given the spirit of the "no functions in templates" mandate, I thought that it might elicit non-positive reactions, so I did not.
> 
> Can you please elaborate on that? Where would the function in the template be coming from?

No explicit function call, but computational overhead that's probably greater than that of calling very simple functions sparingly in templates.

NatalieEO (approved):

kathy-t (approved):

svonworl:
This PR has changed significantly, see the updated description at top.

src/app/organizations/collection/collection.component.ts
```diff
@@ -69,6 +70,7 @@ export class CollectionComponent implements OnInit {
     public dialog: MatDialog,
     private activatedRoute: ActivatedRoute,
     private userQuery: UserQuery,
+    private orgLogoService: OrgLogoService,
```
coverbeck, line 73: This feels like it should be public, because you access it from the template. I was surprised this built at all, but it looks like its enforcement depends on settings that I didn't bother to figure out.

src/app/shared/org-logo.service.ts
```diff
@@ -0,0 +1,7 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class OrgLogoService {
+  // default org logo image URL, including enough parent directory references to ensure that '/assets' is always relative to the root
+  readonly DEFAULT_URL = '../../../../assets/images/dockstore/default-org-logo.svg';
```
coverbeck, line 6: This may be made moot in prod by the string substitution we do for the CDN (?), but this string is used by different components with different urls that may not be at the same depth level. So `../../../../assets` may take you to the assets directory in some cases but not in others? Just an untested theory, but the relative path should be resolved relative to the current url in the browser.

coverbeck (approved):

src/app/shared/org-logo.service.ts
svonworl: The initial goal here was include enough `../`s to crawl up the url path so that, during development, we'd reach root in all cases (four levels), and to provide the deploy script with the pattern it latches onto when it does the rewrite (which, as you mention, makes the issue moot):

https://github.com/dockstore/dockstore-ui2/blob/d10573defb7d9dfc106cb19e418145746a43eebe/.circleci/config.yml#L218

However, upon further inspection, we appear to always render with `<base href="/">` in the header, which means that all relative urls are resolved relative to the root url, and the `../`s don't matter (except to the deploy script).  Indeed, if you add or subtract any number of `../`s from the images on your local setup, you will find they still work (which is good, because a lot of them don't include the correct number of `../`s in the first place).

In summary, might as well leave it as is.

denis-yuen (approved):

PR #1448, opened 2022-02-04
Feature/3886/flakey auth tests round two

Richard-Hansen:
(The reason the test was failing was due to a leftover registered tool from a past run. Deleting this tool seemed to fix the issue, but this is the second time I've made a PR for these tests, so I took another stab at making them more reliable...)

**Description**
This PR fixes broken dev auth tests. To verify they were actually fixed, a CI job was run using the dev auth tests on this branch: https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/7138/workflows/b583cade-3dbf-4c60-b30e-83254662e7d4/jobs/24606

It looks like the `cy.wait()`s we had hanging around weren't waiting long enough to be reliable, and some of the `click()` actions were grabbing the DOM too early, so the buttons never became actionable. Removing the waits, and explicitly waiting for buttons to be visible before attempting a click seemed to fix the flakiness issue.

That being said, these auth tests create active data on the dev account, so:
1. If they fail once, they will keep failing until fixed, as there will be leftover tools/workflows hanging around in the test account.
2. If the job is canceled before it's finished, there will most likely be leftover tools/workflows, so subsequent runs will fail.

---

The Cypress docs discourage conditional DOM testing, but perhaps a followup ticket to conditional cleanup resources from previous runs (i.e. `if a tool exists unpublish and delete, otherwise do nothing`) would prevent further flakiness.

---

**Issue**
https://github.com/dockstore/dockstore/issues/4589

### DIFF
--- a/cypress/integration/smokeTests/sharedTests/auth-tests.ts
+++ b/cypress/integration/smokeTests/sharedTests/auth-tests.ts
@@ -19,6 +19,8 @@ const workflowTuple = ['github.com', username, 'hello-dockstore-workflow'];
 // tuple of organization name, collection name
 const collectionTuple = ['test', 'testcollection'];
 
+const HARDCODED_WAIT_TIME = 10000;
+
 // get the dockstore token from env variable and put it in local storage
 function storeToken() {
   window.localStorage.setItem('ng2-ui-auth_token', Cypress.env('TOKEN'));
@@ -38,7 +40,7 @@ function deleteTool() {
     storeToken();
     cy.server();
     cy.route('delete', '**/containers/**').as('containers');
-    cy.wait(2000); // hardcoded 2s wait is least flaky option right now, revisit in future
+    cy.wait(HARDCODED_WAIT_TIME); // hardcoded 2s wait is least flaky option right now, revisit in future
     cy.contains('#deregisterButton', 'Delete').should('be.enabled');
     cy.contains('#deregisterButton', 'Delete').click();
     // cy.get('#deregisterButton').click();
@@ -66,7 +68,7 @@ function registerQuayTool(repo: string, name: string) {
     cy.visit('/my-tools');
     cy.wait('@tokens');
     // click thru the steps of registering a tool
-    cy.wait(2000); // hardcoded 2s wait is least flaky option right now, revisit in future
+    cy.wait(HARDCODED_WAIT_TIME); // hardcoded 2s wait is least flaky option right now, revisit in future
     cy.get('#register_tool_button').click();
     cy.wait('@orgs');
     // cy.wait(1000);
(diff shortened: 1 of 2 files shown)


cypress/integration/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -99,16 +96,15 @@ function registerRemoteSitesTool(repo: string, name: string) {
 
     cy.visit('/my-tools');
     cy.wait('@tokens');
-    cy.wait(2000); // hardcoded 2s wait is least flaky option right now, revisit in future
-    cy.get('#register_tool_button').click();
+    cy.get('#register_tool_button').should('be.visible').should('be.visible').click();
```
svonworl, line 99: accidental double be.visible?

cypress/integration/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -38,10 +38,7 @@ function deleteTool() {
     storeToken();
     cy.server();
     cy.route('delete', '**/containers/**').as('containers');
-    cy.wait(2000); // hardcoded 2s wait is least flaky option right now, revisit in future
-    cy.contains('#deregisterButton', 'Delete').should('be.enabled');
-    cy.contains('#deregisterButton', 'Delete').click();
-    // cy.get('#deregisterButton').click();
+    cy.contains('#deregisterButton', 'Delete').should('be.visible').click();
```
svonworl, line 41: would this be even safer if we also included a "be.enabled" (per the replaced code)?  not sure, genuinely curious. :)

cypress/integration/smokeTests/sharedTests/auth-tests.ts
Richard-Hansen: Reduced visibility: https://github.com/dockstore/dockstore-ui2/pull/1448/commits/bd22ad2d548e2f837c486a0b101aae8c567f1a73

garyluu (approved):
TIme will tell I guess

cypress/integration/smokeTests/sharedTests/auth-tests.ts
Richard-Hansen: TLDR: As long as the correct DOM element is passed to `.click()`, there shouldn't be a difference between `be.enabled` and `be.visible`, because `.click()` checks for visibility and being enabled anyways.

Ideally, `.should('be.visible')` wouldn't be required at all. This is because the Cypress `.click()` action checks for [actionability](https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Actionability), which includes making sure the element is enabled, visible, etc... But, for an unknown reason, the command `cy.contains('#deregisterButton', 'Delete').click();` would fail, while `cy.contains('#deregisterButton', 'Delete').should('be.visible').click();` succeeds.

Just hypothesizing, but I expect the DOM element yielded from `cy.contains('#deregisterButton', 'Delete').click();` can be grabbed too quickly, and that DOM element is never updated to the actionable state (perhaps the button is reloaded after a few moments, and the reloaded button is the one that becomes visible?). By changing the line to `cy.contains('#deregisterButton', 'Delete').should('be.visible').click();`, Cypress will wait for a visible button to be in the DOM before passing it on to `.click()`, and this seems to fix the issue I was having locally.

svonworl (approved):

coverbeck (approved):

denis-yuen (approved):

PR #1449, opened 2022-02-04
Hiding a version should show on table without refreshing page

NatalieEO:
**Description**
The version table wasn't updating before after hiding/unhiding a version for workflows in general. This fixes that.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3847

### DIFF
--- a/src/app/workflow/version-modal/version-modal.service.ts
+++ b/src/app/workflow/version-modal/version-modal.service.ts
@@ -23,6 +23,8 @@ import { WorkflowQuery } from '../../shared/state/workflow.query';
 import { WorkflowsService } from '../../shared/swagger/api/workflows.service';
 import { SourceFile } from '../../shared/swagger/model/sourceFile';
 import { WorkflowVersion } from '../../shared/swagger/model/workflowVersion';
+import { WorkflowService } from '../../shared/state/workflow.service';
+import { AppTool, BioWorkflow, Service } from '../../shared/swagger';
 
 @Injectable()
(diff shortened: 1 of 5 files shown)


src/app/workflow/version-modal/version-modal.service.ts
```diff
@@ -74,6 +78,8 @@ export class VersionModalService {
     if (workflowMode !== 'HOSTED') {
       this.workflowsService.updateWorkflowVersion(workflowId, [workflowVersion]).subscribe(
         (response) => {
+          workflow = { ...workflow, workflowVersions: response };
```
coverbeck, line 81: It looks  like we don't have this [lint rule](https://eslint.org/docs/rules/no-param-reassign) turned on, but I think it's a good practice and would recommend not reassigning the parameter value. But your code will work as is.

src/app/workflow/version-modal/version-modal.service.ts
```diff
@@ -103,6 +109,8 @@ export class VersionModalService {
     } else {
       this.workflowsService.updateWorkflowVersion(workflowId, [workflowVersion]).subscribe(
         (response) => {
+          workflow = { ...workflow, workflowVersions: response };
```
coverbeck, line 112: Same comment about reassigning a parameter value.

coverbeck (approved):

coverbeck (approved):

garyluu (approved):

PR #1450, opened 2022-02-08
Fix relative links in README descriptions

kathy-t:
**Description**
This PR fixes the relative links displayed in an entry's README description by specifying a base URL for relative links. 

ngx-markdown allows options like `baseUrl` (prefix url for relative links) to be specified as [MarkedOptions](https://github.com/jfcere/ngx-markdown#markedoptions). However, MarkedOptions is meant to be defined globally, which isn't ideal for us because relative links only really show up in README descriptions and not in other places where we use markdown. Setting it globally also prevents us from changing the base URL dynamically based on entry repo.

Found a work-around by passing in MarkedOptions in the MarkDownService's `compile` method, which we already use. My PR only specifies base URLs for tool and workflow descriptions.

I found that GitLab and BitBucket also support relative links in their READMEs, so this PR handles those services as well.

**Issue**
[#4453](https://github.com/dockstore/dockstore/issues/4453)

### DIFF
--- a/src/app/shared/entry/base-url.pipe.ts
+++ b/src/app/shared/entry/base-url.pipe.ts
@@ -1,0 +1,26 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'baseUrl',
+})
+export class BaseUrlPipe implements PipeTransform {
+  /**
+   * Constructs a base url for relative links in markdown README files.
+   * @param providerUrl
+   * @param versionName
+   * @returns {string} base url to be used by ngx-markdown
+   */
+  transform(providerUrl: string, versionName: string): string {
+    if (!providerUrl || !versionName) {
+      console.error('providerUrl or versionName is not truthy');
+      return null;
+    }
+    if (providerUrl.startsWith('https://github.com/') || providerUrl.startsWith('https://gitlab.com/')) {
+      return providerUrl + '/blob/' + versionName + '/'; // the last slash is needed by ngx-markdown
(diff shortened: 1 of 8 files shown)


src/app/shared/entry/base-url.pipe.ts
```diff
@@ -0,0 +1,26 @@
...
+export class BaseUrlPipe implements PipeTransform {
+  /**
+   * Constructs a base url for relative links in markdown README files.
+   * @param providerUrl
+   * @param versionName
+   * @returns {string} base url to be used by ngx-markdown
+   */
+  transform(providerUrl: string, versionName: string): string {
+    if (!providerUrl || !versionName) {
+      console.error('providerUrl or versionName is not truthy');
```
coverbeck, line 15: I'm not a fan of logging to the user's browser console, but others disagree. :)

coverbeck (approved):
Added @david4096 as a reviewer for any security implications. Trying to think if there's some way a user could create a malicious README.md. The existing code already prevents scripts from running. I guess there could be evil link, but somebody could do that already with an absolute url.

I can't think of anything.

garyluu (approved):

NatalieEO (approved):

denis-yuen (commented):
No objections on my end, but curious about the security discussion

Richard-Hansen (approved):

david4096 (approved):
I can't think of anything off hand that someone could take advantage of here that the absolute URL wouldn't also affect. Looks good to me

denis-yuen (approved):

PR #1451, opened 2022-02-08
Reduce changeDescriptor calls

garyluu:
**Description**
Reduce the number of changeDescriptor calls from 3 or 4 to 1 or 2 using behavior subjects and combineLatest with an odd trick (debounceTime(0)). Downside of this PR is that it introduces 3 new behavior subjects which is kind of duplicating values

Not entirely sure if it's even worth doing.

**Issue**
dockstore/dockstore#4194

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 orbs:
   aws-s3: circleci/aws-s3@3.0.0
   slack: circleci/slack@3.4.2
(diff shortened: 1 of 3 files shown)


coverbeck (approved):
> Not entirely sure if it's even worth doing.

I lean towards doing it, but I don't feel super strongly. It is a small number of calls in question, but code is easier to understand than I anticipated.

NatalieEO (approved):

denis-yuen (approved):

PR #1457, opened 2022-02-11
Match Redesign Button Colours/Styles

KimberleyChong:
**Description**
JIRA: [dock-2029](https://ucsc-cgl.atlassian.net/browse/DOCK-2029?atlOrigin=eyJpIjoiN2QyODBiMzYyYTA3NDE1MTg0Nzg0MWZmNzJjNjU0YTUiLCJwIjoiaiJ9)

The purpose of this ticket is to match existing buttons on dockstore with the new button colours on Zeplin (accent 1 success and secondary). These changes were done using existing classes (secondary-1, accent-1-dark, accent-1-dark-btn, private-btn, small-btn-structure, small-mat-btn-skin) and applying them on a case by case basis to match the current mock-ups. Felt like this made more sense than changing the Material primary colour since it would not only disrupt the current colour pallet but also affect places that do not need colour changes, additionally this method allows us to just reuse existing styles.

**Most button changes in this PR follow this common style in Zeplin mock-ups:**
image

**Here are some screenshots of the changes, I haven't included them all since there are quite a few, but in each case it is applying the same design from Zeplin:**
image
image
image
image
image
image
image
image
image
image
image
image
image

**More unique changes to match additional redesigned buttons:**
image
image
image
image
image
image
image

Unrelated margin fix, the starred organizations page did not have any space between the tops and bottoms of cards, I looked into the existing organizations page and found it was missing `fxLayoutGap` attribute so I decided to add that in while I was there.
image
image

### Notes:
image
- there is a change in the cypress test for updating labels in a workflow/tool, for some reason the test is failing to click the save button and actually save the labels
- gary and I looked into it and he found that changing the line height not to 26px allowed the tests to pass... which is weird considering other buttons I have changed use the same line height and pass their tests...
- we're thinking its a weird bug/issue with cypress, gary found a fix to force:true on the button click which forces the event to be fired (submitWorkflowEdits() or submitContainerEdits()) and the label is added
- from a user point of view the button is still functioning as it did before so no concerns there, but just a note about the weird cypress behaviour 
- related to the add labels buttons I did make a [github issue](https://github.com/dockstore/dockstore/issues/4719) about some functionality that is a little buggy and exists in dev + prod

- **npm audit** -gary mentioned this on slack but it is failing since it is comparing release changes with develop so yeahhh


coverbeck (approved):

src/app/container/container.component.css
```diff
@@ -0,0 +1,20 @@
+/*
+*    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
denis-yuen, line 2: The `WITHOUT WARRANTIES` bit seems extra, although true I guess. Might be an IDE setting, might want to grab something more standard like the license on https://github.com/dockstore/dockstore/blob/4b59cbb17be115288529314d817aed3aed72829d/dockstore-webservice/src/test/java/io/dockstore/webservice/resources/proposedGA4GH/package-info.java

PR #1458, opened 2022-02-14
Collection description table resizing bug

KimberleyChong:
**Description**
JIIRA: [dock-1886](https://ucsc-cgl.atlassian.net/browse/DOCK-1886?atlOrigin=eyJpIjoiYWZiODlhOWE2Njg5NDJiNDgwYjdhZDhhZjczMDI4NjQiLCJwIjoiaiJ9)

(for 1.12 so audit test failure)

When viewing Biodata Catalyst's Utilities collection in a smaller window there is a squishing issue on the left due to the markdown table in their description.

**Changes:**
- screen width ratio of the left an right columns have been set to 2:3 (Denis felt this size made sense) which prevents the left side from being squished too much
- when resizing window smaller, any overflow from the markdown portion that cannot be resized smaller (ie. table) will trigger a horizontal scroll bar which prevents the contents (table) from falling out of the page

### Full Screen view of 2:3 ratio:
image

### Resizing Before:
image

### Resizing with no scrollbar triggered:
image

### Resizing with scrollbar triggered:
image

**Note the scrollbar will only be triggered in rare cases like these for ex. when they insert a table in the description, but in most cases the collections page will look like:**
image


denis-yuen:
@garyluu  Hmmm, maybe an addition for https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide once we decide. I can't see anything relevant in my Slack searches, do you have something to link to?

garyluu:
> @garyluu Hmmm, maybe an addition for https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide once we decide. I can't see anything relevant in my Slack searches, do you have something to link to?

It was during a presentation, so it wasn't in Slack

KimberleyChong:
@garyluu so word-break works well up until here:
image

but once you get smaller than that it needs to break every word to fit...
image

if we don't want a scroll bar we could do something I previously tried which was to go into the "column" view earlier to prevent the table from getting so small, but the next largest angular breakpoint is ~1270px so it remains in a column for a pretty large window size
image

garyluu:
Without actually deciding how things should look (others as well as ui designer should figure it out and have a consistent way), my personal preference would be to flexbox it so that if the markdown needs more space for tables than the alotted amount, it will flex wrap to the next row.

coverbeck (commented):
Hi @KimberleyChong , are you still responding to Gary's feedback? I was going to wait until you did that before reviewing... If you want feedback now, or whenever you're ready, please request re-review. Thanks!

KimberleyChong:
> Hi @KimberleyChong , are you still responding to Gary's feedback? I was going to wait until you did that before reviewing... If you want feedback now, or whenever you're ready, please request re-review. Thanks!

Hey! Sorry we've been communicating on slack about this, I'm gonna update this PR today and re-request.

KimberleyChong:
### Updates:

So after many discussions this is the solution we've come up with.  Kim (other one) recommends against horizontal scroll. It's not perfect, but in the **rare** cases someone does have a table, we've agreed  that it is an improvement over how it was being handled before.

**Solution:**
- If the contents in the right column cannot fit within its div while allowing both columns to maintain a 1:1 ratio with min-width of 500px, we will wrap the contents into a singular column
- This prevents squishing and overflowing when there is a table in the right column. and when wrapped it will have full screen width available
- Additional changes: Gary recommended adding table padding and removing word wrapping  to the markdown-wrapper component to make the tables cleaner/easier to read and also keep it consistent with other tables on dockstore

**Where the solution falls short:**
- When resizing with a table, if the window gets smaller than the size of the table, there is no wrapping to decrease the width and no scroll bar since the logic behind `overflow:auto` conflicts with the nature of flex box wrapping... if the overflow creates a horizontal scroll, then there is not need to row wrap to _prevent_ the overflow (if you see what I mean)
- Denis, Gary and I don't have a way to prevent this right now, and it's difficult to handle since the user has the ability to create a table of any size that we can't predict
- But again, this will only occur in **rare** cases, I've looked around and I've only found 2 collections with tables in them (both from BioData Catalyst) so most will not encounter this issue, Denis and I think its still a good improvement to have

### "Normal" Collection Resizing: (aka no table)

https://user-images.githubusercontent.com/97123241/154718664-6d6383c8-235f-4e82-9f79-301b7f21caa8.mp4

### Small Table Resizing:
Small overflow when in small window:

https://user-images.githubusercontent.com/97123241/154718774-512c8ca2-5cd2-47ee-a253-38d28540abd7.mp4

### Large Table Resizing:
This is a more severe case where the table has a very large width:

https://user-images.githubusercontent.com/97123241/154718814-2da7c021-9dc0-46d4-ba73-018286753846.mp4

### Large Table Resizing Before:

The squished cards on the left side is what the user mentioned in the Jira ticket as the primary issue:

https://user-images.githubusercontent.com/97123241/154718843-58bb5b8a-ce90-4e79-abe7-24ad04911a25.mp4

### Additional Details on Changes:

image
image
image

NatalieEO (approved):
Looks much better!

kathy-t (approved):

denis-yuen (approved):
Looks good and thanks!
Please add a line in the https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide now that we have a consensus to use this strategy in the future when the size of user-generated content is variable

PR #1460, opened 2022-02-15
Feature/seab 3679/fix circle ci npm audit

NatalieEO:
**Description**
Moved the scripts into one script and run using an npm command in the config.yml and added a new config, `base_branch` to the package.json so that the tests are not always being hard coded to run against develop.

See failure example at `[show failure](https://github.com/dockstore/dockstore-ui2/pull/1460/commits/d1199c9370a9beba64fbac1c07bf600e321da818)
[d1199c9](https://github.com/dockstore/dockstore-ui2/pull/1460/commits/d1199c9370a9beba64fbac1c07bf600e321da818)

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3679


scripts/npm-audit-comparison.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+set -o errexit
+set -o pipefail
+set -o nounset
+set -o xtrace
+
+# Run npm audit on current branch and compare it with the results of running npm audit on the base branch that is set in the package.json. If there are more
+# high or critical findings in the current branch, then the test fails. If the same number of findings are found, then check that the vulnerabilities are the same.
+# If they are are different, then the test fails.
+# Test is optional to merge. Use your and the reviewer's best judgement
```
denis-yuen, line 10: Specific comment feels weird to me since we decide this and set it in the GitHub UI anyway

scripts/run-webservice-script.sh
```diff
@@ -18,3 +18,5 @@ psql -h localhost -c 'create database webservice_test with owner = dockstore;' -
 psql -h localhost -f test/${DB_DUMP:-db_dump.sql} webservice_test -U postgres
 java -jar dockstore-webservice.jar db migrate -i 1.5.0,1.6.0,1.7.0,add_service_1.7.0,1.8.0,1.9.0,1.10.0,alter_test_user_1.10.2,1.11.0,1.12.0 test/web.yml
 
+
```
denis-yuen, line 21: ditto

scripts/npm-audit-comparison.sh
```diff
@@ -0,0 +1,83 @@
...
+
+git stash
+
+# Save high and critical vulns from the base branch (e.g. develop, hotfix/1.11.2)
+git checkout "$npm_package_config_base_branch"
+npm ci
+npm audit | grep -E "(High)" -B3 -A10 > base-branch-high-vulnerabilities.txt || true
+npm audit | grep -E "(Critical)" -B3 -A10 > base-branch-critical-vulnerabilities.txt || true
+
+HIGH_VULN="$(grep -o High base-branch-high-vulnerabilities.txt | wc -l || true)"
```
denis-yuen, line 30: unsure about something, might have missed something might be worth a code comment. 
if all that `HIGH_VULN` and `CRITICAL_VULN` are storing are line counts
... ditto for the 'current_branch` / `base_branch` counts, how does this deal with the ticket example of removing one vulnerability but adding a different one?

NatalieEO: Yeah, so lines 14-15 save truncated version of the npm audit report to files. The truncated version has info about each vuln. Then just the count from the truncated version gets saved separately in new files (line 32-33). Further down I compare the counts first. If the current branch has a larger number of vulns, then it fails. But after that comparison, there's another check that compares the content of the files created from lines 14-15 if the count between the base and current branches are equal. Does that make sense?

denis-yuen: Ah, I missed the `cmp`, so its a "compare the number, but if equal diff/cmp to see if there is movement"
works for me, thanks!

denis-yuen (approved):

scripts/npm-audit-comparison.sh
```diff
@@ -0,0 +1,73 @@
...
+
+HIGH_VULN="$(grep -o High current-branch-high-vulnerabilities.txt | wc -l || true)"
+CRITICAL_VULN="$(grep -o Critical current-branch-critical-vulnerabilities.txt | wc -l || true)"
+echo $HIGH_VULN > compare-num-vulnerabilities.txt
+echo $CRITICAL_VULN >> compare-num-vulnerabilities.txt
+
+git stash
+
+# Save high and critical vulns from the base branch (e.g. develop, hotfix/1.11.2)
+git checkout "$npm_package_config_base_branch"
```
coverbeck, line 24: If you run this locally, won't you:

1. End up with your branch switched to base branch
2. Any local changes stashed (from line 21)

Normally only the CI server will run this, but presumably one day we'll introduce a vulnerability that this will catch, and then we'll want to see what's going on locally, and then it might be confusing.

Maybe just print out a warning message at least? I was going to say always do `git stash pop`, but if `git stash` didn't have anything to save, that will be a mess.

Or maybe don't do the `git stash` at all. I think that should work on CI; and then leave it up to the developers to stash themselves if they run into issues.

coverbeck (approved):

garyluu (approved):

.circleci/config.yml
```diff
@@ -117,22 +117,9 @@ jobs:
       - run:
           name: Run npm audit
           # Run npm audit on current branch and compare it with the results of running npm audit on the develop branch. If there are more
-          # high or critical findings, then the test fails (Test is optional)
+          # high or critical findings, then the test fails. If the same number (Test is optional)
```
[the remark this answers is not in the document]
NatalieEO: Fixed

scripts/npm-audit-comparison.sh
NatalieEO: Yup, you end up in the base branch. Ended up getting rid of the git stash and added a warning at the top that you'll end up on the base branch and that you may want to stash their local changes.

PR #1461, opened 2022-02-15
Flakey Auth Tests Round Three

Richard-Hansen:
**Description**
This PR re-introduces `cy.waits()` that'd I'd previously removed in: https://github.com/dockstore/dockstore-ui2/pull/1448, because that PR didn't really fix anything... Any suggestions appreciated.

The UI uses javascript to update asynchronously, Cypress is grabbing DOM elements that, after the UI updates, are considered detached, causing tests to fail. After spending a while trying to identify when the UI had updated sufficiently, I was still hitting flakey tests, and succumbed to re-introducing `cy.wait()` at a higher timeout than before.

Hopefully this PR will reduce the flakiness of the auth tests, but it isn't guaranteed. If the auth tests fail again, I'll create a followup ticket to refactor the `my-tools` and `my-workflows` components to see if the flakiness is better addressed from that side. Or, we could reject this PR in favor of that approach in the first place.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1997

### DIFF
--- a/cypress/integration/smokeTests/sharedTests/auth-tests.ts
+++ b/cypress/integration/smokeTests/sharedTests/auth-tests.ts
@@ -14,10 +14,12 @@ import { goToTab } from '../../../support/commands';
 // TODO: for clarity and future debugging refactor variable names to use 'namespace', 'entry-name', call those variable names in methods
 const username = 'dockstoretestuser4';
 // tuples of registry, repo namespace (username), and entry-name (repo name)
-const toolTuple = ['github.com', username, 'dockstore-tool-md5sum'];
+const toolName = 'dockstore-tool-md5sum';
+const toolTuple = ['github.com', username, toolName];
 const workflowTuple = ['github.com', username, 'hello-dockstore-workflow'];
 // tuple of organization name, collection name
 const collectionTuple = ['test', 'testcollection'];
+const hardcodedWaitTime = 8000;
 
 // get the dockstore token from env variable and put it in local storage
 function storeToken() {
@@ -27,6 +29,8 @@ function storeToken() {
 function unpublishTool() {
   it('unpublish the tool', () => {
     storeToken();
+    cy.visit('/my-tools');
+    cy.wait(hardcodedWaitTime);
     cy.contains('#publishToolButton', 'Unpublish').should('be.enabled').click();
     cy.contains('#publishToolButton', 'Publish').should('be.enabled');
     cy.contains('#deregisterButton', 'Delete').should('be.enabled');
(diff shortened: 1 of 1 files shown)


coverbeck:
1. https://filiphric.com/waiting-in-cypress-and-how-to-avoid-it?utm_campaign=Software%2BTesting%2BWeekly&utm_medium=email&utm_source=Software_Testing_Weekly_106 -- "Just use the wait and be done with it" :(.
2. I noticed Cypress is up to 9.x, and there are newer 8.x versions than the 8.0.0 we have installed. Given the above article, it probably won't make a difference, but maybe worth a shot?

svonworl:
I found this open cypress issue, basically a lot of people begging for a fix to our problem:

https://github.com/cypress-io/cypress/issues/7306

No one seems to have a good workaround.

One possibility is to figure out some way to repeat the query if we trigger the "detached DOM" error.  A crude but effective solution might be to create a function called `complete` that would invoke its function argument, catching the error and retrying up to N times, pausing a geometrically-increasing time between retries (to keep the tests running quickly if possible).

You'd use it like:
`complete(() => cy.get('whatever').etc().click());`

Possibly, there's a way to graft it into the cypress system to make it more convenient to use.

Richard-Hansen:
> I found this open cypress issue, basically a lot of people begging for a fix to our problem:
> 
> [cypress-io/cypress#7306](https://github.com/cypress-io/cypress/issues/7306)
> 
> No one seems to have a good workaround.
> 
> One possibility is to figure out some way to repeat the test if we trigger the "detached DOM" error. A crude but effective solution might be to create a function called `complete` that would invoke its function argument, catching the error and retrying up to N times (or until a specified amount of time has elapsed), pausing an initially-small-but-geometrically-increasing amount between retries (to keep the tests running as quickly as possible, but not hog a bunch of cpu).
> 
> You'd use it like: `complete(() => cy.get('whatever').etc().click());`
> 
> Possibly, there's a way to graft it into the cypress system to make it more convenient to use.

I gave this a whirl, but unfortunately it didn't work either. This is because some Cypress requests fail due to a detached DOM, but others get stuck waiting for an element to appear in whatever DOM they are currently looking at, causing an assertion error. I think this is because a detached DOM can cause two separate issues:
1. If we are waiting for an element (that we have located) to reach a certain state, but the DOM is changed, this throws an assertion error, as the element never gets updated. In other words, the test times out
2. We get an element in one request, and then in a subsequent request/chain we interact with it. When we attempt to interact with a detached element, the detached DOM exception is thrown.

In any case though, we have Cypress set to wait 30 seconds for assertions to become true, which takes longer to re-capture the DOM than adding the `cy.wait(8000)`s. Lowering the default Cypress retry may make more tests flakey as well...

Initially, I planned on testing Cypress 9, but the issue linked above has some recent responses that it's still broken :( If we still are interested in Cypress 9, I can create a followup ticket.

For moving forward, I suggest we look to refactoring the UI pages in question to make them more test friendly, as that seems more scalable than trying to hack Cypress to function as desired. But, I'm open for suggestions.

svonworl:
I wonder if, in case #1, we could drastically shorten the wait timeout, catch that error also, and retry similarly?

Or maybe Charles' "just add the waits" is the most pragmatic solution.

PR #1462, opened 2022-02-15
LGTM fix - Use regex replaceAll until EC2021

david4096:
**Description**

A fix suggested by LGTM to avoid missing other `&` in the parameters.

**Issue**

https://ucsc-cgl.atlassian.net/browse/SEAB-3876

### DIFF
--- a/src/app/loginComponents/accounts/accounts.component.ts
+++ b/src/app/loginComponents/accounts/accounts.component.ts
@@ -57,6 +57,6 @@ export class AccountsComponent extends Base implements OnInit {
   setAccountsTab(tabName: string) {
     // Adding the & symbol causes the URL to be parsed in an unexpected way.
     // TODO: Change replace to replaceAll once we use EC2021
-    this.location.replaceState('accounts?tab=' + tabName.replace('&', 'and'));
+    this.location.replaceState('accounts?tab=' + tabName.replace(/&/g, 'and'));
   }
 }


NatalieEO (approved):

coverbeck (approved):

src/app/loginComponents/accounts/accounts.component.ts
```diff
@@ -57,6 +57,6 @@ export class AccountsComponent extends Base implements OnInit {
   setAccountsTab(tabName: string) {
     // Adding the & symbol causes the URL to be parsed in an unexpected way.
     // TODO: Change replace to replaceAll once we use EC2021
-    this.location.replaceState('accounts?tab=' + tabName.replace('&', 'and'));
+    this.location.replaceState('accounts?tab=' + tabName.replace(/&/g, 'and'));
```
svonworl, line 60: This works.  Technically, if we wanted to be 100% robust and allow full freedom of naming, instead of the bandaid `&` -> `and`, we should fully urlencode `tabname` via `encodeURIComponent()` and change the `and`s back to `&`s in `validTabs`.  But I'm not sure it's worth the trouble in this case.

svonworl (approved):

kathy-t (approved):

garyluu (approved):

src/app/loginComponents/accounts/accounts.component.ts
david4096: I was trying to think of a bug but since we constrain the input it shouldn't happen? If you think there's a bug we should open it though.

svonworl: Yeah, I think it's ok for now, no bugs in the current configuration (that I can see).

PR #1464, opened 2022-02-17
Org logo not displaying on Orgs page

coverbeck:
**Description**
Problem was due to strict-origin-when-cross-origin policy on the
image. Using gravatar fixes it (and is why it worked on the single Org page).

Consolidated all the gravatar calls into one service.

**Issue**
#4700

- [] Run `npm audit` and ensure you are not introducing new vulnerabilities
- [x] Don't allow user-uploaded images to be served from the Dockstore domain (fixes this!)

### DIFF
--- a/src/app/gravatar/gravatar.service.ts
+++ b/src/app/gravatar/gravatar.service.ts
@@ -10,7 +10,7 @@ export class GravatarService {
 
   public gravatarUrlForEmail(email: string, defaultImage: string) {
     const hash = Md5.hashStr(email);
-    return `${this.gravatarBaseUrl}${hash}}?id=${defaultImage}&s=500`;
+    return `${this.gravatarBaseUrl}${hash}}?id=${encodeURIComponent(defaultImage)}&s=500`;
   }
 
   public gravatarUrlForMysteryPerson() {
@@ -19,6 +19,6 @@ export class GravatarService {
   }
 
   public gravatarUrlForImageUrl(imageUrl: string | null) {
-    return imageUrl ? `${this.gravatarBaseUrl}000?d=${imageUrl}` : null;
+    return imageUrl ? `${this.gravatarBaseUrl}000?d=${encodeURIComponent(imageUrl)}` : null;
   }
 }
--- a/src/app/gravatar/gravatar.service.ts
+++ b/src/app/gravatar/gravatar.service.ts
@@ -10,7 +10,7 @@ export class GravatarService {
 
   public gravatarUrlForEmail(email: string, defaultImage: string) {
     const hash = Md5.hashStr(email);
-    return `${this.gravatarBaseUrl}${hash}}?id=${encodeURIComponent(defaultImage)}&s=500`;
+    return `${this.gravatarBaseUrl}${hash}?id=${encodeURIComponent(defaultImage)}&s=500`;
   }
 
   public gravatarUrlForMysteryPerson() {
--- a/src/app/gravatar/gravatar.service.ts
+++ b/src/app/gravatar/gravatar.service.ts
@@ -1,0 +1,24 @@
+import { Injectable } from '@angular/core';
+import { Md5 } from 'ts-md5/dist/md5';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class GravatarService {
+  readonly gravatarBaseUrl = 'https://www.gravatar.com/avatar/';
+  constructor() {}
+
+  public gravatarUrlForEmail(email: string, defaultImage: string) {
+    const hash = Md5.hashStr(email);
+    return `${this.gravatarBaseUrl}${hash}}?id=${defaultImage}&s=500`;
+  }
+
+  public gravatarUrlForMysteryPerson() {
+    // https://en.gravatar.com/site/implement/images/ -- it's "mp", not "mm" there, but we've had it like this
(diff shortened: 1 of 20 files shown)


src/app/gravatar/gravatar.service.ts
```diff
@@ -0,0 +1,24 @@
...
+    return `${this.gravatarBaseUrl}${hash}}?id=${defaultImage}&s=500`;
+  }
+
+  public gravatarUrlForMysteryPerson() {
+    // https://en.gravatar.com/site/implement/images/ -- it's "mp", not "mm" there, but we've had it like this
+    return `${this.gravatarBaseUrl}?d=mm&s=500`;
+  }
+
+  public gravatarUrlForImageUrl(imageUrl: string | null) {
+    return imageUrl ? `${this.gravatarBaseUrl}000?d=${imageUrl}` : null;
```
svonworl, line 22: We should urlencode any uncontrolled query string values here and elsewhere via `encodeURIComponent`, so that a value with sensitive characters (for example, an `imageUrl` containing `&` ala `http://www.site.com/images?p=foo&q=bar`) gets through correctly.

svonworl (changes requested):

garyluu (approved):

NatalieEO (approved):

src/app/gravatar/gravatar.service.ts
coverbeck: Good catch; we'd been doing it this way all along, and surprisingly it worked. I guess the browser was encoding it. Changed and fixed in 22934034d46cae1dafc121d86bd77b5d771c6c53

svonworl (approved):

coverbeck:
> In general, I think this is the best approach that conforms to our "no functions in templates" policy. Adding a `gravatizeUrl` function call at each site would have been much simpler, and arguably more understandable. Someday, I will collect some data regarding the overhead of Angular change detection in our UI, and present it and some recommendations at a storytime.

As an experiment, I changed it to a function on the organizations page; to load that page, the underlying method gets called 9 times with a pipe, once for each org on the page, vs. 126 times with a function. And the pipe is a common and recommended pattern in Angular.  But if you have other findings I'd be curious to hear them.

svonworl:
> > In general, I think this is the best approach that conforms to our "no functions in templates" policy. Adding a `gravatizeUrl` function call at each site would have been much simpler, and arguably more understandable. Someday, I will collect some data regarding the overhead of Angular change detection in our UI, and present it and some recommendations at a storytime.
> 
> As an experiment, I changed it to a function on the organizations page; to load that page, the underlying method gets called 9 times with a pipe, once for each org on the page, vs. 126 times with a function. And the pipe is a common and recommended pattern in Angular. But if you have other findings I'd be curious to hear them.

Assuming each of those 126 calls consumed 1000 cycles, they consumed 126000 cycles total, which, assuming 3e9 cycles per second, is (126 * 1000) / 3e9 = 42 millionths of a second = 1 / 23809th of a second
I understand why it's important to be careful with functions in templates, but I'm not sure it's a good idea to completely ban them in all cases...
Keep in mind that I'm not trying to be hostile, just advocating for what I think is best for the codebase... :)

coverbeck:
> Assuming each of those 126 calls consumed 1000 cycles, they consumed 126000 cycles total, which, given 3e9 cycles per second, is (126 * 1000) / 3e9 = 42 millionths of a second = 1 / 23809th of a second
I understand why it's important to be careful with functions in templates, but I'm not sure it's a good idea to completely ban them in all cases...
Keep in mind that I'm not trying to be hostile, just advocating for what I think is best for the codebase... :)

I'm not trying to be hostile either, but to respond to your points. And I'm happy to be proven wrong.

I get your point, but we've had pages in the past that were noticeably sluggish or even brought up the JavaScript busy dialog due to variants of this. If this is the only function being invoked from the page and it's fast, then yeah, it may not make a noticeable difference to the end user, but add in other functions that take more time and with a multiplicative effect... Why start going down the road towards danger which somebody one year from now may inadvertently accelerate by adding some more functions, when there's a standard Angular pattern that avoids it?

svonworl:
Oh, I forgot to check, we need to add the `gravatar` pipe here:
https://github.com/dockstore/dockstore-ui2/blob/22934034d46cae1dafc121d86bd77b5d771c6c53/src/app/starredentries/starredentries.component.html#L196
and here:
https://github.com/dockstore/dockstore-ui2/blob/22934034d46cae1dafc121d86bd77b5d771c6c53/src/app/entry/current-collections/current-collections.component.html#L26

Also, the existing `gravatarUrl$` observables in the organization and collection components are now redundant, we might as well replace them with a pipe.

kathy-t (approved):

src/app/gravatar/gravatar.service.ts
```diff
@@ -0,0 +1,24 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class GravatarService {
+  readonly gravatarBaseUrl = 'https://www.gravatar.com/avatar/';
+  constructor() {}
+
+  public gravatarUrlForEmail(email: string, defaultImage: string) {
+    const hash = Md5.hashStr(email);
+    return `${this.gravatarBaseUrl}${hash}}?id=${encodeURIComponent(defaultImage)}&s=500`;
```
svonworl, line 13: Extra `}` after hash?

PR #1468, opened 2022-02-23
Feature/dock 1878/search on enter

svonworl:
**Description**
This PR changes the search UI so that a search is initiated when the user presses the enter key or clicks a search control (rather than the current behavior, which searches after each character in the search text is entered, debounced).  Text prompting the user to "Press &lt;Enter&gt; to search" was added below the search box.  Also, the autosuggest dropdown now closes upon click/enter, rather than staying open and obscuring the controls underneath.

Ths PR alters the "basic search" component to pass three events to the containing "search" component:
- a "changed" event upon every change to the search text,
- a "changedDebounced" event, similar to the "changed" event, except debounced, and
- a "submitted" event, denoting that the user has pressed "enter".

The parent "search" component then decides how to handle these events.
 
**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1878
dockstore/dockstore#4404


NatalieEO (approved):

coverbeck (commented):
I ran it locally and it looks good! A lot less flashing going on.

I know this code is gnarly, so no need to change your PR for this, but I'm going to ask anyway: do you need to track the debounced and the non-debounced text separately? Can't you just use the debounced text?

One actual request: can you please sort out the formatting w/Husky issue we discussed in that other PR? I tried changing one of your files locally, ran `npx lint-staged`, and it did do some additional formatting, so I don't think it was run for this PR.

svonworl:
> I ran it locally and it looks good! A lot less flashing going on.
> 
> I know this code is gnarly, so no need to change your PR for this, but I'm going to ask anyway: do you need to track the debounced and the non-debounced text separately? Can't you just use the debounced text?

This PR runs a search on the current text in the search box in two cases:
1. enter is pressed
2. a facet is clicked
To implement #2, we need the non-debounced text, to avoid the case where the entire search text doesn't "register" if the user very quickly types the text and then very quickly clicks a facet.
Unrelated: try saying "very quickly clicks" ten times fast.

denis-yuen (approved):

kathy-t (approved):

svonworl:
Substantially "prettier"ed.

coverbeck (approved):
Did you figure out why this isn't happening automatically for you? Did you have to run it manually?

svonworl:
> Did you figure out why this isn't happening automatically for you? Did you have to run it manually?

Wasn't making much headway figuring out what was going wrong, so I deleted my copy of dockstore-ui2, cloned again, and now it prettiers...

PR #1469, opened 2022-02-23
Global import of Roboto font family

KimberleyChong:
**Description**

When working on my search sidebar ticket I noticed something strange about the fonts used on Dockstore... according to styles.scss we're supposed to be using Roboto as our primary font, but after some investigation I realized that we're actually not.  There isn't an import for the Roboto font family specified anywhere in our repository, so Dockstore has been rendering with fallback fonts for probably this entire time.

image

Depending on your operating system, with use of inspect you can find that for these fonts have been used as fallbacks:
* **Ubuntu:** Liberation Sans (3rd fallback: default sans serif font)
* **Windows:** Arial (3rd fallback: default sans serif font)
* **Mac:** Helvetica Neue (2nd fallback)

Its pretty inconsistent and also doesn't match Zeplin which calls  for Roboto.  The font differences also causes a lot of formatting inconsistencies like the font weight, ex. where font-weight 400 & 500 render the same but with Roboto there is a distinct weight difference that Zeplin calls for.

image

### Solution
Simply import the Roboto font family into our global style sheet so users can render the primary font we've specified.

### Implementation
Quick discussion on implementation, my preferred method (and the most simple method) is just with an `@import` rule in styles.scss. Roboto is a google font so we'll call their API and import it into the style sheet. I've had some discussions about best practice and after some additional searches I'd say this is a pretty recommended method for font importing.

[Stackoverflow answer:](https://stackoverflow.com/questions/44878591/best-practice-style-for-importing-font-into-angular-4-cli-web-application)
image

[Article specific to Angular: Method 2 – @import (Recommended)](https://www.softaox.info/best-way-to-use-google-fonts-in-your-angular-project/)
image

An alternative method would be to download all of the font files from google, create an additional global style sheet (or include it in the original but it would be a large chunk of code) then link each font file and specify individual weights with the `@font-face` rule. Further details [here](https://www.w3schools.com/cssref/css3_pr_font-face_rule.asp) and [here](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face).

Let me know what you think in the comments and if you have any alternative methods. In this PR I have included the `@import` rule as my preferred solution. It'll be merging into develop so that we can keep an eye on any formatting issues as this is a global change. Thanks!

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -23,6 +23,7 @@ $material-design-icons-font-directory-path: '~material-design-icons-iconfont/dis
 @import '~ngx-sharebuttons/themes/default/default-theme';
 @import 'featured-content.scss';
 @import '/src/materialColorScheme.scss';
+@import url('https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap');
 /*@import url(https://fonts.googleapis.com/css?family=Open+Sans:400,600);*/
 
 /* #487980 GREEN */
@@ -208,7 +209,7 @@ angular-tag-cloud > span.w1:hover {
 /* Main Styles */
 
 body {
-  font-family: Roboto, 'Helvetica Neue', sans-serif;
+  font-family: 'Roboto', 'Helvetica Neue', sans-serif;
   font-weight: normal;
   color: #313131;
   font-size: 14px;


coverbeck:
I found another [article](https://rewis-io.translate.goog/urteile/urteil/lhm-20-01-2022-3-o-1749320/?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=wapp) (the original is in German, my link is a Google Translate to English link). It's a bit hard to understand, but I think the idea is somebody was complaining about a site requiring their browser to go to Google, which is evil, because they collect IPs :) .

Without understanding all the issues, I think it would be better if we deliver the fonts ourselves, but I'm willing to be talked out of it.

@svonworl @david4096 Any thoughts on this?

coverbeck (commented):
Was here, left a comment without submitting a review, but I am now so doing. Please re-request review

svonworl:
> I found another [article](https://rewis-io.translate.goog/urteile/urteil/lhm-20-01-2022-3-o-1749320/?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=wapp) (the original is in German, my link is a Google Translate to English link). It's a bit hard to understand, but I think the idea is somebody was complaining about a site requiring their browser to go to Google, which is evil, because they collect IPs :) .
> 
> Without understanding all the issues, I think it would be better if we deliver the fonts ourselves, but I'm willing to be talked out of it.
> 
> @svonworl @david4096 Any thoughts on this?

When you browse the Google Fonts site and select some fonts, the site will recommend some html that you can embed in the header of your site to use them.  You can choose either `link`-based or `@import`-based html.  I use the `link`-based html on a couple of my personal sites (to serve Roboto, coincidentially) with no problems.  This reference seems to lean towards the `link`-based solution: https://stackoverflow.com/questions/12316501/including-google-fonts-link-or-import

The Google-recommended `link` tags look like this:
```
<link rel="preconnect" href="https://fonts.googleapis.com">
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
<link href="https://fonts.googleapis.com/css2?family=Roboto+Serif:ital,wght@0,400;0,700;1,400;1,700&display=swap" rel="stylesheet">
```
It'd be pretty easy to adapt them to serve the files locally.

The caveat is that we'd need to test to make sure there's no weirdo interactions with our UI stack, and that serving the files ourselves may increase the site load time slightly (the font requests will compete with other resource requests).

Re what's correct regarding GDPR, that's a question for the lawyers.  As Charles mentioned, the safest solution is to serve as little from 3rd party sources as possible.

david4096:
Thanks for finding this and such a thorough report on the current status! 

A similar conversation came up [around the material icons](https://github.com/dockstore/dockstore-ui2/pull/1095). We provide [sub-resource integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) checks for included style and script files, basically a hash of the file that is added to the index.html during build.

image

I believe that to get SRI for the web fonts, we could host them ourselves and could include the link rel including the SRI in the existing `index.html`. For the material icons, the build artifacts get generated as part of the rest of the UI assets. 

We could continue this pattern and there are npm packages to help with this (for example [fontsource](https://github.com/fontsource/fontsource)).

Alternatively, we might upload the artifact to a different S3 location so that the URL is stable across builds. This has the benefit of reducing load time when we release a new version if the browser caches the font.

Ultimately, I think the risk of a tainted font file from Google is low. Regarding a privacy leakage, it seems somewhat moot given we are already sending things to Google Analytics. The benefits of downloading a font from a CDN are pretty significant for user experience. For reference, Terra loads fonts directly from the Google API.

david4096 (approved):

denis-yuen (approved):
(convinced)

KimberleyChong:
I agree with David to keep the current proposed implementation I think he made some great points!

But just to follow up with a proposed alternative, I've went and tested to make sure it was feasible.  The alternative is to create a Roboto.scss sheet and link it to the existing styles.scss sheet.  Within Roboto.scss we can use the `@font-face` rule to define each font format:
image

Where the source url is a local copy of the font files from google. We would then repeat this for all font weights: 100, 300, 400, 500, 700, 900 and italics and normal.

Essentially doing what the import rule links to https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap

coverbeck:
> Regarding a privacy leakage, it seems somewhat moot given we are already sending things to Google Analytics.

I thinking we're going to have to give users the option to turn that off, eventually. A bit surprised we've gotten away with it this long.

NatalieEO (approved):

kathy-t (approved):

coverbeck (approved):
I seem to be a minority of one here, but we can always revisit, especially if we do get to allowing users to opt out of Google tracking.

Any reason to not target the release branch with this?

denis-yuen:
Hmmm, thinking at the time was to draw a line somewhere and global changes breaking something unexpected as we work our way toward a release ...

PR #1470, opened 2022-02-23
Compact search sidebar to match redesign

KimberleyChong:
**Description**
JIRA: [Dock-1969](https://ucsc-cgl.atlassian.net/browse/DOCK-1969?atlOrigin=eyJpIjoiNTQyNjZhMmU3ODc5NGU1ZWEwNDIxNWZiYWYzNDcxMGQiLCJwIjoiaiJ9)

Goal of this ticket was to make the search sidebar more compact and closer to the [Zeplin](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/5fc690e874e0c807afeb299b) redesign.

**Changes:**
* Removed spacing between panels 
* Tightened padding(s)
* Header style when collapsed
* Updated style of 3 search buttons
* Updated sort buttons & tool tip icons
* Additional small tweaks to match redesign etc.
* Not changed: drop down arrows- deeply embedded in the Angular expansion panel component with animations called that are difficult to override

### Without changes:
image

### With Changes:
image

### With Changes: Collapsed Headers and tool tips:
image

### Font and Bolding Change
You may notice that there is a small difference with the font weight in the current version compared to Zeplin... this will be resolved with this [PR](https://github.com/dockstore/dockstore-ui2/pull/1469) that I want to merge. Essentially we never imported the font Roboto into our project and we've been using fallback fonts on Dockstore (see PR for more details). But after that PR is merged then the font weight will be corrected.

**Zeplin vs Current vs With Roboto import change:**
image

**Small Note:**
* It says 19 files changed, I only changed 7 of those, the rest appears to be some automatic formatting that took place with some commits from develop.
* Files changed: basic-search.component.html, basic-search.component.scss, basic-search.component.ts, search.component.html, search.component.scss, help.svg, reset.svg

### DIFF
--- a/src/app/search/search.component.scss
+++ b/src/app/search/search.component.scss
@@ -117,7 +117,6 @@ mat-tab-group.homeComponent {
 }
 
 fa-icon.ng-fa-icon {
-  display: inline-block;
   vertical-align: top;
 }
 


KimberleyChong:
Just updated the hover and select colours I missed it on another Zeplin file
image

NatalieEO (approved):

src/app/search/basic-search/basic-search.component.scss
```diff
@@ -0,0 +1,73 @@
...
+@import '/src/materialColorScheme.scss';
+
+// Panel spacing
+.mat-expansion-panel-spacing {
+  margin: 0px;
+}
+
+// Panel header
+mat-expansion-panel mat-expansion-panel-header {
+  padding: 10px;
```
coverbeck, line 27: This may have since changed, but the style guide recommends against pixels and to use rems. Are you using pixelst because you're overriding default Angular classes?

cypress/integration/group2/organizations.ts
```diff
@@ -306,7 +306,7 @@ describe('Dockstore Organizations', () => {
     });
 
     // test the fix for DOCK-1945
-    it('stay on collections page when removing an entry',  () => {
+    it('stay on collections page when removing an entry', () => {
```
coverbeck, line 309: @svonworl These and a lot of the diffs that Kim didn't change seem to be yours.

The husky [pre-commit hook](https://github.com/dockstore/dockstore-ui2/blob/01ed8b6080ec7afd2ab6786df5fe7b20c734a3d1/.husky/pre-commit#L4) is what's doing this. Any idea why it's not happening for your commits, if that indeed is the issue?

svonworl: Yes, it appears to be mangling my well-formatted code :/ (with the exception of deleting the extra space above, which was merited :)   Hmm, maybe it got disabled somehow, will check.  Tough question: do we really like what it does to the HTML, in general?

svonworl:
Overall, this looks great!  I would like to register a beef with the mockups: I feel that the frequency number-bars are too close to the right edge and feel cramped.  But that's not an issue with this PR, nothing needs to be changed, good job!

svonworl (approved):

src/app/search/basic-search/basic-search.component.scss
KimberleyChong: I am overriding default Angular styles, but I used pixels mainly to follow Zeplin's specs. I can convert them to rems but just a bit on the fence about it since on my computer/browser 1rem = 10px for some reason... which is not the standard 16px. I've wondered if its because I'm on linux since my windows laptop follows the standard. I'll update this PR, can you confirm it displays nicely on your browser once its up?

KimberleyChong:
**Update:** converted pixels to rems on the code I added except for the border-radius on buttons, pixels seems to be standard for all other angular buttons.

Bare in mind I used a conversion of 1rem = 10px (not standard 16px), let me know on your browsers if it is still accurate, may need to convert 1rem = 16px if not

cypress/integration/group2/organizations.ts
denis-yuen: Think we're using prettier https://github.com/prettier/prettier 
IIRC, its pretty opinionated though you could try upgrading it to see if it does anything in edge cases.

denis-yuen (approved):

KimberleyChong:
Integration test 2 keeps failing:
* I ran it once and it failed with github app tools
* re-ran and it failed with dockstore my tools
* now its failing again with github app tools.

Both are flaky tests, not sure what to do now, I don't want to re-run again.
image
image

cypress/integration/group2/organizations.ts
coverbeck: 1. I think it's good to be standardized so that PRs like this don't get confusing, so reviewers can focus on the PR proper.
2. I don't always like all the formatting choices either. They're probably configurable via prettier if you want to submit PRs if you have strong opinions on particular formatting.
3. I'm also open to other ideas, but the idea of doing it this way is so non-standardized commits don't slip through. Although apparently your code did, which is confusing. But if you can figure out a different way to ensure it happens reliably. ok.

coverbeck (approved):

coverbeck (commented):
Tried this locally, I am seeing it a little differently. In particular the sort column headers don't seem vertically centered (the buttons with a-z, 0-9, and the arrow).

Seems to happen with rems and pixels. Maybe a quick call can resolve it.

Screen Shot 2022-02-25 at 4 57 18 PM

PR #1475, opened 2022-03-02
Fix null OrcidAuthor errors

kathy-t:
**Description**
The problem is that console shows several messages like `ERROR TypeError: this.selectedVersion.orcidAuthors is null` if a workflow is published/unpublished, refreshed, snapshotted, DOI'd, exported to ORCID, or a new default version is selected. This is happening because ORCID authors are lazy loaded and all of those endpoints mentioned do not initialize ORCID authors.

This PR uses the `workflows/{workflowId}/workflowVersions/{workflowVersionId}` endpoint introduced in PR https://github.com/dockstore/dockstore/pull/4754 to get the selected version's ORCID authors. I introduced a new endpoint instead of initializing ORCID authors in all of the previously mentioned endpoints because after speaking to Denis, this seems more aligned with best practices since those endpoints may not need to have ORCID authors initialized in all scenarios.

I created a `workflows/{workflowId}/workflowVersions/{workflowVersionId}` endpoint instead of a specific getVersionAuthors endpoint because the former is more flexible and can be used to fetch other lazy loaded content if needed in the future.

**Issue**
Part of [SEAB-3220](https://ucsc-cgl.atlassian.net/browse/SEAB-3220)


denis-yuen (commented):
Looks good but we should make sure that this goes in after the associated webservice PR

src/app/workflow/info-tab/info-tab.component.ts
```diff
@@ -70,7 +71,7 @@ export class InfoTabComponent extends EntryTab implements OnInit, OnChanges {
   trsLink: string;
   displayTextForButton: string;
   displayedColumns: string[] = ['name', 'role', 'affiliation', 'email', 'orcid_id'];
-  authors: (Author | OrcidAuthor)[];
+  authors: any[] = [];
```
svonworl, line 74: Can the elements of `authors` be of a type other than `Author` or `OrcidAuthor`?

kathy-t: The elements should only be `Author` and `OrcidAuthor`. I was getting errors like `Type 'Author' is not assignable to type 'Author | OrcidAuthor'. Property 'orcid' is missing in type 'Author' but required in type 'OrcidAuthor'.` so I changed it to `any`. Checking again, the error disappears if I use the models from openapi so I changed it back

NatalieEO (approved):

src/app/workflow/info-tab/info-tab.component.ts
```diff
@@ -121,7 +122,11 @@ export class InfoTabComponent extends EntryTab implements OnInit, OnChanges {
       const found = this.validVersions.find((version: WorkflowVersion) => version.id === this.selectedVersion.id);
       this.isValidVersion = found ? true : false;
       this.downloadZipLink = Dockstore.API_URI + '/workflows/' + this.workflow.id + '/zip/' + this.currentVersion.id;
-      this.authors = [...this.selectedVersion.authors, ...this.selectedVersion.orcidAuthors];
```
coverbeck, line 124: It's low odds, but the next line could either fail, e.g., due to flaky network, or be very slow, so I think you should clear the authors here so the previous ones aren't displayed.

coverbeck (approved):

kathy-t:
New changes:
- Used the new getWorkflowVersionOrcidAuthor endpoint to get ORCID author information.
- Changed `ORCID ID` to `ORCID iD` to match what they use in their [docs](https://info.orcid.org/brand-guidelines/#the-orcid-name)

denis-yuen (approved):

coverbeck (approved):
Pending CircleCI artifacts coming from release branch.

svonworl (approved):

NatalieEO (approved):

PR #1479, opened 2022-03-08
Start of My Sidebar Implementation

KimberleyChong:
**Description**
JIRA: [SEAB-3964](https://ucsc-cgl.atlassian.net/browse/SEAB-3964?atlOrigin=eyJpIjoiZDAyMmZmYjMzZmJjNDZjZjliZWViNjdmYjk2OTc0YTAiLCJwIjoiaiJ9)

The purpose of this ticket is to start implementing the "my sidebar" which is the first little sidebar on the left of this zeplin [page](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/602c3742b9fe1a21d8fe96da).

image
image

### Updates to header style:
image
image

**Box shadow**
The `#container-header` style already applies a box shadow on the header component, yet we have another style, `.container-fluid`, applying the same shadow to the header component -> which affects the my tools/workflows/services pages. So I removed the duplicate box shadow from the `.container-fluid style`.

**Top padding:**
Also removed the extra 5px of top padding in `container-fluid`, not sure why it was added but it makes the header 5px taller than Zeplin specifies and again it only affects my tools/workflows/services which looks inconsistent when navigating

**Background colour:** 
I also removed background colour from `.container-fluid` because its just white on Zeplin like all the other headers

**On that note** the entire `.container-fluid` style is redundant since position and z-index is also set by `#container-header` so I just removed the rest of the style, headers should be more consistent

### Header placement and sidebar icons:
These both follow what is shown on Zeplin.
image

### Modifications from Zeplin:
image

### Additional styles to accounts page to match Zeplin
image

### Added :host
* Isolate padding style defined in starredentires.component.scss since it is using ::ng-deep without :host and it was affecting places that don't need it like the my tools page
* See right photo, was adding extra 26px of top padding to mat-tab-body-content

image

### Action Buttons Alignment:
Actions buttons were not aligning back properly after resizing as seen on production and develop, should be fixed now
### Before - on production:

https://user-images.githubusercontent.com/97123241/157284748-20926227-f5bb-46c6-accf-bfc41b7e4d51.mp4

### After:

https://user-images.githubusercontent.com/97123241/157284794-47f6965d-2430-47f4-b351-7a8a1794cb10.mp4

### My Workflows/Tools resizing Issue:
* With the addition of the sidebar on the left, for pages like My Tools and Workflows I needed to make layout changes which resulted in some resizing issues, in particular when it got to mobile view there was an issue where the existing pop-out sidebar would get infinitely long after the window resizing.  When played with for too long it would make chrome crash

https://user-images.githubusercontent.com/97123241/157294305-358612d3-1308-44b7-ad85-f218b366c9e8.mp4

* Little back story: we are using the mat-sidenav component which consists of the **drawer content** and the **main content**
image
* With the previous code, when resized small enough it would resize both the drawer and main content to `col-xs-12` (xs window full width) which did not work well with the addition on the new sidenav because it could no longer take up full width and so I believe that is what caused it to resize to infinite width 
* I tried to make the arrangement work with bootstrap but the interaction with the components in the desired layout was not ideal and using predefined column sizes didn't fit well with the design
* I decided to essentially remove most of the main bootstrap layout code and replace it with angular fxLayout and fxFlex, I left .container-fluid since there are still inner components that rely on it
* Another issue I ran into with getting the drawer to cover the main content, angular has "modes" for this component and it is set to "side" instead of "over",  meaning that the drawer remains next to the main content at all times, which does not allow for it to cover when resized. But when the mode is set to "over" it can only ever be on top the content and never beside
* I don't want to go too much into the details since this is already so long but here's the layout solution I came up with, it works well, only issue is when resized to the view where the drawer covers full width there is a lot of empty white space which is due to the main content being hidden
* **Current Solution:**

https://user-images.githubusercontent.com/97123241/157309683-905c44b9-75dc-453c-808b-6504b237ab18.mp4
* I spoke with Denis and he said the solution works well for now, since this is going into develop and the drawer itself needs to be redesigned we can look into other solutions or removing mat-side-nav
* An alternative solution I tried was using an event listener to check the width of the screen when resizing, when the window is less than a certain width update the mode from side to over to allow for overlap -> only problem was the mode was updated appropriately but the component itself would remain the same mode it started with. If you have any ideas let me know, else, when the sidenav is redesigned we can look into alternatives

### Last side note:
Kathy helped me update `package.json` and `run-webservice-script.sh` to use the latest circleci build since the previous one here was expired, pushing changes to keep it updated


KimberleyChong:
> This pull request **introduces 3 alerts** when merging [cc37800](https://github.com/dockstore/dockstore-ui2/commit/cc378008f2af2aedae8c76e60bb506f04caac485) into [2662a26](https://github.com/dockstore/dockstore-ui2/commit/2662a269a18c2fd615416d7619b231d5a85aa0bf) - [view on LGTM.com](https://lgtm.com/projects/g/dockstore/dockstore-ui2/rev/pr-43c686fdf8826c43e450546e0a11d89795b97d95)
> 
> **new alerts:**
> 
> * 3 for Unused variable, import, function or class

Uh both imports are default when you create a new component and I couldn't leave ngOnInit() empty or it would fail the lint_license_unit_test_coverage test for having an empty lifecycle.

denis-yuen:
quick tip for reviewers, the videos didn't work in Firefox for me but they work in Chrome

NatalieEO (approved):
had never seen the infinite scrolling before!

coverbeck:
At first glance, running locally, this looks great! Some random thoughts/questions before even looking at the code:

* Formatting for one of my tools for me seems off:

Screen Shot 2022-03-09 at 9 56 50 AM

* I find it odd that `Dashboard` takes you to the home page, dropping the sidebar. Is that the intent?
* I don't know about the sliding effect of starred -- it looks a little noisy to me.
* Why does starred default to starred organizations? Or maybe it's just defaulting to the most recently starred, which for me happens to be an organization?
* I think a starred workflow should be rendered using markdown, the same way it's rendered on the public page:
Screen Shot 2022-03-09 at 10 18 29 AM

src/app/my-sidebar/my-sidebar.component.html
```diff
@@ -0,0 +1,77 @@
...
+  ~
+  ~    Unless required by applicable law or agreed to in writing, software
+  ~    distributed under the License is distributed on an "AS IS" BASIS,
+  ~    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~    See the License for the specific language governing permissions and
+  ~    limitations under the License.
+  -->
+
+<div fxFlex="0 0 auto">
+  <!-- Dashboard links to homepage since page doesn't exist yet -->
```
coverbeck, line 18: I should have read this before my other comment about the dashboard. :)

src/app/starredentries/starredentries.component.html
```diff
@@ -1,253 +1,269 @@
...
+                <div fxLayout="row" fxLayoutAlign="space-between stretch">
+                  <div fxLayout="row" fxLayoutAlign="flex-start" fxLayoutGap="1rem">
+                    <small class="spacing">Last updated {{ workflow?.lastUpdated | date }}</small>
+                    <div fxLayout="row wrap" fxLayoutGap="1rem">
+                      <span>
+                        <mat-chip-list>
+                          <mat-chip class="bubble workflow-background" *ngFor="let label of workflow?.labels | slice: 0:3">{{
+                            label.value
+                          }}</mat-chip>
+                          <mat-chip class="bubble" *ngIf="workflow.versionVerified === true">
```
coverbeck, line 86: Inconsistent use of `?` -- used on lines 79 and 83 but not here and line 95 (probably inherited). If it's needed, it's needed everywhere; if it's not needed anywhere. Not just here, but portions of the code as well.

If this is existing code that just got moved, don't worry about it.

coverbeck (commented):
Looks good, going to wait for you to address/dismiss my other general comments; please request review again after you've had a chance to go through those.

coverbeck:
>     * I think a starred workflow should be rendered using markdown, the same way it's rendered on the public page:
>       Screen Shot 2022-03-09 at 10 18 29 AM

Looking at the screenshots in #1481, it looks like this is an existing issue.

KimberleyChong:
> At first glance, running locally, this looks great! Some random thoughts/questions before even looking at the code:
> 
> * Formatting for one of my tools for me seems off:
> 
> Screen Shot 2022-03-09 at 9 56 50 AM
> 
> * I find it odd that `Dashboard` takes you to the home page, dropping the sidebar. Is that the intent?
> * I don't know about the sliding effect of starred -- it looks a little noisy to me.
> * Why does starred default to starred organizations? Or maybe it's just defaulting to the most recently starred, which for me happens to be an organization?
> * I think a starred workflow should be rendered using markdown, the same way it's rendered on the public page:
>   Screen Shot 2022-03-09 at 10 18 29 AM

1. Actions buttons: Not sure what to do here, let me know what you think. The main title div is set to 80% of the container width (in red) and the buttons on the right just get the remaining 20%. When we have our tool/workflow published there are now 4 buttons all in the same div that can't align in the same row since there is not enough space and so it row wraps.
image
* We could remove the row wrapping from the buttons but my only conern would be that the row is quiet long and it will limit the amount of room the title will have, especially if you resize to a small window
image
* This is not the final design on Zeplin though, these buttons will actually be above the title so they will have more room to staty in a row, so if we leave it as so it should be updated in the future
image

2. Dashboard Link: temporary, but I agree it does feel like a weird redirect
3&4. Starred Organizations: I didn't create the starred page I just did some layout reformatting when I added the navbar which is why it says I edited it. Not sure why it defaults to organizations first but someone must have chosen that
5. Workflow card: Sorry again I didn't work on this page but we can probably create a ticket to update how the card is being displayed?

src/app/starredentries/starredentries.component.html
KimberleyChong: Yeah this is just code that got moved!

coverbeck (approved):

svonworl:
I ran this locally, and overall, it looks great!  A few comments follow...

svonworl:
Regarding the "Dashboard" tab, I feel like the "logged-in" homepage and the dashboard might actually be synonymous: they both seem to have the same intent and contain similar information (but laid out differently, we need to overhaul the logged-in page in a later PR to match the dashboard mockups).  If that's true, maybe we keep the sidebar up when someone clicks "Dashboard" and display the "logged-in" homepage?  Thoughts, everyone?

svonworl:
Little bug: when I select some of the "tabs" in the sidebar, in the selected tab, the label text below the icon moves down one pixel, then moves back up into the original position when another tab is selected.  Nothing else moves, and the tabs don't change in height.  The afflicted tabs are "Workflows", "Tools", and "Services", it doesn't happen to the others (unsure about "Dashboard" since it disappears when I click it).  I see the bug using Mac Chrome, and I can make a video if you can't reproduce.

svonworl:
I wonder if the sidebar should have some sort of mouseover effect, kinda like the top menubar does, to help people notice that it's interactive and make it feel even more slick.

src/app/my-sidebar/my-sidebar.component.html
```diff
@@ -0,0 +1,77 @@
...
+
+  <!-- My Starred -->
+  <button mat-flat-button id="purple" routerLink="/starred" routerLinkActive="active">
+    <div><img src="../assets/svg/my-nav/starred.svg" class="pb-1" alt="starred" /></div>
+    Starred
+  </button>
+  <hr routerLink="/starred" routerLinkActive="active" />
+
+  <!-- My Accounts -->
+  <button mat-flat-button id="purple" routerLink="/accounts" routerLinkActive="active">
```
svonworl, line 63: Duplicate id "purple" here, not sure if it matters, but HTML spec says it's a no no. https://stackoverflow.com/questions/48240240/why-are-duplicate-id-values-not-allowed-in-html

KimberleyChong: Didn't know that thank you! Updating

denis-yuen:
> Regarding the "Dashboard" tab, I feel like the "logged-in" homepage and the dashboard might actually be synonymous: they both seem to have the same intent and contain similar information (but laid out differently, we need to overhaul the logged-in page in a later PR to match the dashboard mockups). If that's true, maybe we keep the sidebar up when someone clicks "Dashboard" and display the "logged-in" homepage? Thoughts, everyone?

the one is definitely the replacement for the other, makes sense to me

KimberleyChong:
**Updates:**
* Fixed the little 1px movement when clicked - really great find, thanks Steve
* Added the sidebar to the "home" (dashboard) page
image
* There was a small space on the right side of the page for "my X" pages which causes a horizontal scroll when resizing, i finally found what was causing it, the header component uses bootstrap class "row" and adds "negative" margin. Replaced it with fxLayout="row" to get rid of the space
image

denis-yuen (approved):

svonworl:
Tried it out, looks super-duper!

svonworl (approved):

PR #1483, opened 2022-03-10
Smoke tests structure for staging and prod

Richard-Hansen:
**Description**
This PR introduces staging tests to the nightly CI job. The current auth smoke tests do not function against prod, but do against dev and staging. After the release is complete, the smoke tests should work everywhere, and the prod tests can be turned on.

Working dev auth tests: https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/7371/workflows/dd3967ca-4b86-40f0-9dff-d039f13ef6e6/jobs/25898
Working staging auth tests: https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/7371/workflows/dd3967ca-4b86-40f0-9dff-d039f13ef6e6/jobs/25897

Setup process (doesn't need to be repeated, just documenting):
1. The user `DockstoreTestUser4` was added to each stage manually.
2. Registered this workflow: https://github.com/DockstoreTestUser4/hello-dockstore-workflow
3. Linked quay accounts
4. Created an organization named `DockstoreAuthTestOrg` with empty collection `SimpleCollection`. This organization doesn't need to be approved, but still must be accessible by `DockstoreTestUser4`.

Created 3 CircleCI env variables to house the Dockstore tokens, these are accessed during the auth tests:
1. `CYPRESS_DEV_TOKEN`
2. `CYPRESS_STAGING_TOKEN`
3. `CYPRESS_PROD_TOKEN`

This PR also means that we can replace the dev database with a newer version from prod without breaking the auth tests (from a data standpoint, if the user tokens change then the tests will break).

Added a note a ticket for updating staging nightly with a censored copy of the prod DB to also account for tokens: https://github.com/dockstore/dockstore/issues/3982

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-1871

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -410,10 +410,13 @@ workflows:
           stack: "staging"
           context:
             - dockerhub
-      - uptime_monitor_auth:
-          stack: "prod"
-          context:
-            - dockerhub
+
+# TODO: wait until 1.12 release
+#      - uptime_monitor_auth:
+#          stack: "prod"
+#          context:
+#            - dockerhub
+
 commands:
   install_container_dependencies:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -401,12 +401,19 @@ workflows:
 #          stack: "prod"
 #          context:
 #            - dockerhub
-      # TODO: Add auth tests for staging/prod once https://ucsc-cgl.atlassian.net/browse/SEAB-2071 is done.
+
       - uptime_monitor_auth:
           stack: "dev"
           context:
             - dockerhub
-
+      - uptime_monitor_auth:
+          stack: "staging"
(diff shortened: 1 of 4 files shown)


coverbeck (approved):
I'm not sure we want the smoke tests run nightly against staging and prod, although I can see it both ways.

We definitely want the auth scripts so we can reduce manual testing during the release process.

Nightly runs against software that's not changing though makes it more of a monitoring software. Which is not a bad thing necessarily, but is this what we want the smoke tests to be?

I guess there are other changes we might make to the infrastructure, e.g., IAM permissions, that have the potential to break things when we think they won't, so maybe it's just as well to have them. I talked myself into it :)

denis-yuen (approved):

PR #1484, opened 2022-03-14
My Sidebar for Homepage - Fixed Small Window Layout

KimberleyChong:
**Description**
So I added the my sidebar to the home/dashboard page as a final change on Friday and didn't to check for resizing… forgot to enclose all of the components into a new “row” `fxLayout` div. Went and updated that. Also added z-indexes so that the sidebar shadow can fall on top of the component next to it.

image


coverbeck (approved):
Approved, assuming the dashboard page is going to further modified to be like the mocks.

svonworl (approved):

PR #1485, opened 2022-03-14
Feature/seab 4047/published app tools

coverbeck:
**Description**
Add browse view of app tools.

This makes the term "App tool" sort of public, but this is a fallback UI if ES is not working, and it page is not linked to directly on the main Dockstore pages (there is a link to it on the sitemap).

Screen Shot 2022-03-14 at 3 04 05 PM

**Issue**
[SEAB-4047](https://ucsc-cgl.atlassian.net/browse/SEAB-4047)

### DIFF
--- a/cypress/integration/group2/githubAppTools.ts
+++ b/cypress/integration/group2/githubAppTools.ts
@@ -164,7 +164,7 @@ describe('GitHub App Tools', () => {
       cy.visit('/apptools');
       cy.url().should('contain', 'apptools');
       cy.contains('Search app tools');
-      cy.get('[data-cy=entry-link]').should('contain', 'A/l');
+      cy.get('[data-cy=entry-link]').should('contain', 'test-github-app-tools');
     });
   });
 });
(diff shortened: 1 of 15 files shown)


denis-yuen (approved):

denis-yuen:
FYI @KimberleyChong since you were working on routing from the user pages

svonworl:
I'm a bit wary, because I felt like we were trying pretty hard to make apptools appear like regular tools to the end user.  But if everyone else is down with it, I'm game.

coverbeck:
> I'm a just a little bit wary, because I felt like we were trying pretty hard to make apptools appear like regular tools to the end user. But if everyone else is down with it, I'm game.

I have a [comment](https://ucsc-cgl.atlassian.net/browse/SEAB-4047?focusedCommentId=40917) that slightly touches on this, but the complexity is that we're using Angular's Paginator, and it would be a bit of work to handle two API calls that return different response types, but treat them as one for the purposes of pagination. It's doable, but given the timeframe, this seemed like the lesser evil approach, at least for now.

Richard-Hansen (approved):

PR #1487, opened 2022-03-15
Fix Copy TRS ID

coverbeck:
**Description**
We were appending `tool` to the TRS ID, which we shouldn't.

I could have gone for a simpler fix, but tried to cut down on duplication.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3828

### DIFF
--- a/src/app/workflow/info-tab/info-tab.service.ts
+++ b/src/app/workflow/info-tab/info-tab.service.ts
@@ -265,13 +265,36 @@ export class InfoTabService {
    * @memberof InfoTabService
    */
   getTRSLink(path: string, versionName: string, descriptorType: string, descriptorPath: string, entryType: EntryType): string {
-    const prefix: string = entryType === EntryType.BioWorkflow ? ga4ghWorkflowIdPrefix : ga4ghServiceIdPrefix;
     return (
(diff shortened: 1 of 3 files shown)


src/app/workflow/info-tab/info-tab.service.ts
```diff
@@ -265,13 +265,36 @@ export class InfoTabService {
...
+  private getTRSPrefix(entryType: EntryType): string {
+    switch (entryType) {
+      case EntryType.BioWorkflow:
+        return ga4ghWorkflowIdPrefix;
+      case EntryType.Service:
+        return ga4ghServiceIdPrefix;
+      case EntryType.Tool: // This one shouldn't get invoked from this code
+      case EntryType.AppTool:
+      default:
+        return '';
```
denis-yuen, line 297: starting to lean toward these cases dying more dramatically to warn us when working on something new, like when we work on notebooks maybe

denis-yuen (approved):

svonworl (approved):

src/app/workflow/info-tab/info-tab.service.ts
svonworl: I wonder if someday, we might be able to go a step farther and eliminate the cases somehow...  Maybe enhance the data objects so that they themselves can compute (or are decorated with) values like their TRS urls?

PR #1493, opened 2022-03-21
Global mat-card style update & star button update

KimberleyChong:
**Description**
Decided to do a global `mat-card` style refresh to match what is present on [Zeplin](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60072419e08ae47561013a80). Also updated a few other things like the star button and some card alignment.

Changes:
* Added a border style
* Updated the shadow and removed previous shadow styles
* Updated star button
* Additional minor tweaks

image
### Before - Accounts page
image
### After - Accounts Page
image
### Before - Organizations page
image
### After - Organizations page
image

### New colour definition
* `#cfd1d7` is used often on Zeplin yet the Zeplin colour palete (and resulting materialColorScheme.scss file) does not have it defined
* It's used in several places including as the `mat-card` border colour, I talked about this with gary before he left and he suggested just defining it in our palette so I added it under `$kim-gray` and replaced previous instances where the colour code was being used

image
### Mat-card global style and shadow cleanup
* Added the global style to include the new border and updated  the shadow (needed to use important to override the material shadow style)
* Also removed instances of `mat-elevation-z(2,3,4)` these all define different shadow styles for different `mat-cards` but from all the cards I checked on Zeplin they seem to use the new shadow style, keeps consistency

image

### Stargazers minor update - stargazers.component.html:
* There's no redesign for this page but all of the cards were enclosed in another `mat-card` which I didn't think made sense
* Replaced the card with a div and padding - I think it looks cleaner the card wasn't necessary
### Before - Stargazers
image
### After - Stargazers
image

### Launch cards alignment - launch.component.html
* As I was verifying the changes I noticed the launch cards had some interesting misalignment, turns out some cards had additional bootstrap margin classes added
* Removed the extra margins and fixed alignment
* Applied to both Tools and Workflows
### Before - Launch Cards
image
### After - Launch Cards (Workflows)
image
### After - Launch Cards (Tools)
image

### Star button refresh - starring.component.html
* Small update but now it matches Zeplin (updated colour, removed grey background when clicked, updated shadow)

image
image

### search.component.html
* From the release 2.9 clone denis merged into develop I left a comment about the search button changes needing to be reverted, I actually don't know if that comment went through since I commented that during the github outage... either way I reverted the search button changes, just re-merging those back into develop 


svonworl (approved):

svonworl:
Looks good!

The following is a general comment, no need to make any changes to this PR...

Regarding the "star button box", even though it's not what's in the mockups:
I feel like when the user stars an entry, the transition on the star from hollow to filled is good, but changing the background color is appropriate and gives it more "weight".  I think we should consider continuing to do it, to give the user additional visual feedback that the action has been taken.  If we did so, maybe use the light "accent" color, rather than gray as before?

NatalieEO (approved):

denis-yuen (approved):

coverbeck (approved):

PR #1497, opened 2022-03-30
My Tools/Workflows Second Sidebar Refresh

KimberleyChong:
**Description**
JIRA: seab-4091

The purpose of this ticket is to update the existing my tools and my workflows sidebars to match the Zeplin redesigns (I really like the new design).
* [Tools sidebar design](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6022f90accad542f52f5acfd)
* [Workflows sidebar design](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/602c22beaeaa3024ad0bae18)

**Main design change:**
* Entries are now sorted and displayed by their registry/sourceControl - you'll see that for tools amazonECR and seven bridges are commented out for the time being, all additional tools will fall into an "Additional" section at the bottom

### My Tools Sidebar - Before
image
image

### My Tools Sidebar - After
image
image

### My Workflows Sidebar - Before
image
image

### My Workflows Sidebar - After
image
image

### Scrolling - Structural Change
* Needed to make right padding of sidebar smaller, but I noticed when you remove the padding it reveals a scroll bar which was hidden with the 17px of padding to allow functionality without seeing it
* What if we kept the scroll functionality, but instead of hiding the scroll bar outside of its container, we made it `display:none`?
* Turns out it's actually possible to hide the bar, so I removed the previous padding and hid it instead
### Previous - with padding removed
image
### Updated scroll functionality
(video doesn't work on ff, please use chrome!)

https://user-images.githubusercontent.com/97123241/160705678-edaefb9f-42c8-470c-9531-c45c10690488.mp4

image

### Services Sidebar
* So the services page doesn't have a zeplin mockup, but it uses the same template as the `my-workflows` page, so when I removed the title style the services page didn't really have one
* Just following the tools and workflows sidebars I updated the top "add" button it to match

image

### Removed Class/Global style
* Removed class `org-accordion-header` - removed so that I could also remove `mat-panel-title`, and because the header style was going to be different depending on its open/close state

image

### My Sidebar Height Update 
* I descovered there is a `push` class in app.component.html that creates 155px of white space before the footer, I've only noticed this problem on a 25" monitor where the `push div` is visible and cuts the sidebar short - only if the content on the rest of the page is < or = the sidebar height
* I found that this can be resolved by updating the `min-height` for the sidebar to be large enough to "push" the `push div` out of view (fitting name?)

image

### Cypress Tests Changes
Quite a few cypress files/tests became outdated with the sidebar changes:
* sharedWorkflows.ts
* myWorkflows.ts, mytools.ts
* githubAppTools.ts
* hostedWorkflows.ts, hostedTools.ts
* checkerWorkflowFromWorkflow.ts, checkerWorkflowFromTools.ts
* myWorkflows.json, commands.ts

**Naming convention change:** 
Most of the problems were caused by the naming convention changes with the redesign, instead of the `mat-expansion-panel-headers` being "source control" / "organization" they're now just "organization" and so every command that was looking for the old naming convention failed

**Change to myWorkflows.json:** 
Because we changed from the old header naming convention it also created a problem with the test data previously having "Dockstore.org/A" and "GitHub.com/A" both being displayed as just "A" after the redesign. In order to have unique identifiers I updated the test data for the Dockstore entry to "A2" - luckily this data was only being used by one test (hostedWorkflows.ts)

**Test for `refreshOrganization` button tool-tip:** 
This test fails every time for tools and workflows (but is not required to pass) because a tool-tip does not _exist_ for the `refreshOrganization` button. I updated this button in this PR as part of the sidebar redesign (refresh-organization.component.html) and yeah, there's not tool-tip specified on zeplin either so the test is commented out for now to prevent the unnecessary failures... can be put back in place if we decide to add one

**Issue**


cypress/integration/group2/myworkflows.ts
```diff
@@ -521,18 +521,19 @@ describe('Dockstore my workflows', () => {
...
-    ).trigger('mouseenter');
-    cy.get('[data-cy=refreshOrganization]:visible').should('be.visible').click();
-    cy.contains('button', 'Cancel').should('be.visible');
-    cy.get('[data-cy=confirm-dialog-button] > .mat-button-wrapper').contains('Refresh').click();
-    cy.get('.error-output').should('be.visible');
-    cy.get('[data-cy=refreshOrganization]:visible').should('be.visible').click();
-    cy.contains('button', 'Cancel').should('be.visible').click();
-    cy.get('.error-output').should('not.be.visible');
-  });
+  // Refresh organization button does not have tooltip
```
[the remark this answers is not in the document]
denis-yuen: Hmmm, a little different thought on my end. 

The current tooltip is kinda useless, but I feel like we could use a more helpful one. This isn't on you @KimberleyChong  but could you turn this off with an if statement. That way, the code is still kept in a compliable state, ready for us when we do add a useful new tooltip.

denis-yuen (commented):
Looks really nice and thanks for the screenshots+video as always

One really minor comment

cypress/integration/group2/myworkflows.ts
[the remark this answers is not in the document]
KimberleyChong: I enclosed the tests in an `if(false)` for both the tools and workflows

cypress/integration/group1/hostedTools.ts
```diff
@@ -23,17 +23,18 @@ describe('Dockstore hosted tools', () => {
...
     cy.visit('/my-tools');
 
-    cy.server();
-
-    cy.route({
-      method: 'GET',
-      url: /containers\/.+\/zip\/.+/,
-      response: 200,
-    }).as('downloadZip');
+    // cy.server and cy.route are depreciated, keeping this code in seems to interfere with cypress behaviour when running tests...
```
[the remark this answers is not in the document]
KimberleyChong: Got it, updated both hostedTools and hostedWorkflows to use `cy.intercept` and check for a `200`, it's working locally!

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -1,24 +1,97 @@
...
+    // }
+  ];
+
+  constructor(private toolQuery: ToolQuery) {}
+
+  /**
+   * Sort tools by registeries to display by groups
+   */
+  public sortByRegisteries() {
+    for (var index in this.groupEntriesObject) {
```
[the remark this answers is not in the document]
KimberleyChong: Updated tools and workflows to iterate directly. Thank you for the suggestion, it's helping me learn more about best practices for ts

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -1,24 +1,97 @@
...
+      if (this.groupEntriesObject[index].registry === 'quay.io') {
+        this.sortedTools[0].groupEntryInfo.push(this.groupEntriesObject[index]);
```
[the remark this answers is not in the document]
KimberleyChong: You can determine their indexes from the sortedTools declaration at the top, but I can see that may not be ideal

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -1,24 +1,97 @@
...
-  constructor(private toolQuery: ToolQuery) {}
   @Input() openOneAtATime;
   @Input() groupEntriesObject: OrgToolObject<DockstoreTool>[];
   @Input() refreshMessage;
-
   public toolId$: Observable<number>;
   activeTab = 0;
 
-  ngOnInit(): void {
+  public sortedTools: groupEntriesBySource[] = [
```
svonworl, line 24: Looks good overall!  Something to consider...

You might use a `Map` for `sortedTools` rather than an array.  When you sort, the map key would be the value of `sourceControl` (normalized, perhaps by stripped of the leading 'registry.' etc) and the map value would be the list of tools.  This would avoid the subsequent big chain of `if`s, need to match up source control values to array indexes, etc.  Then, in your template, you could iterate through the source control values (keys) of interest, in whatever order you wanted, and retrieve the lists of tools by simply getting the value from the map.

If you did this, `sortedTools` probably gets renamed to `sourceControlToTools` or something like that.

KimberleyChong: Thanks for the suggestion Steve! I updated `sortedTools` and `sortedWorkflows` to use a Map instead of an array. Instead of having the array of tools as the `value` I used the `groupEntriesBySource` interface from the original array so that I could still keep the `sourceControlTitle` (since this is used as the display name). I removed the large block of `if else` statements  for Workflows and there is now only one `if else` for Tools (to handle additional tools that do not match with any of the existing keys). Much happier with the code now, this was my first time learning about Maps, thank you!

denis-yuen (commented):
Actual changes look good

Codacy results worth a second look (and/or cutting down on) though, in particular many look like a trivial warning about CSS comments. A couple are Angular specific and might be worth looking at too.

KimberleyChong:
> Actual changes look good
> 
> Codacy results worth a second look (and/or cutting down on) though, in particular many look like a trivial warning about CSS comments. A couple are Angular specific and might be worth looking at too.

hmm I've actually been thinking about this, one issue with Codacy is they don't have support for SCSS, and so it checks our `.scss` files as if they are `.css`.
* 26 warnings are for using `//` comments which are standard in scss but invalid in css
* 6 warnings are for assigning a value of `0rem` - which we need in order to override angular mat. styles

I can resolve the 9 (rule-empty-line-before) and add 9 new lines where necessary but the other tests described above:
*  (no-invalid-double-slash-comments)
* (length-zero-no-unit)

I think we should consider disabling from the Codacy checks. Since Codacy doesn't support scss, comments are always invalid, and assigning 0 length is necessary to override angular material styles. Up for dicussion and better alternatives, let me know what you think https://docs.codacy.com/getting-started/supported-languages-and-tools/

src/app/mytools/sidebar-accordion/sidebar-accordion.component.scss
```diff
@@ -1,7 +1,106 @@
...
+
+  .mat-tab-label-content {
+    font-weight: bold;
+    color: #181f40;
+  }
+}
+
+// Selected tool style
+.selected {
+  background-color: #d0effd;
```
kathy-t, line 90: It looks like all the styles for both the my tools sidebar and my workflows sidebar are the same except for this one line for background color. I would recommend moving the styles used by both to a new scss file and loading the this new style sheet to both components. There are examples of shared style sheets in `src/app/shared/styles/`

kathy-t (commented):
Looks really nice!

Checked it out locally and noticed that GitHub App tools have the same styling as workflows:
Screenshot 2022-04-01 at 14-40-53 Dockstore My Tools
The bottom GitHub App tool has green styling (like workflows) instead of blue.

Also noticed that Additional tools is being listed at the top. Should it be on the bottom?
Screenshot 2022-04-01 at 14-41-16 Dockstore My Tools

KimberleyChong:
> Looks really nice!
> 
> Checked it out locally and noticed that GitHub App tools have the same styling as workflows: Screenshot 2022-04-01 at 14-40-53 Dockstore My Tools The bottom GitHub App tool has green styling (like workflows) instead of blue.
> 
> Also noticed that Additional tools is being listed at the top. Should it be on the bottom? Screenshot 2022-04-01 at 14-41-16 Dockstore My Tools

mm I don't have GitHub App tools but I believe it has the style of a workflow because I noticed in my-tool.component.html it displays the tool sidebar then displays the workflow sidebar with `groupAppToolEntryObjects$` 
https://github.com/dockstore/dockstore-ui2/blob/cd9414fade28e209cd3bdfcb0868d89b2b9af574/src/app/mytools/my-tool/my-tool.component.html#L72-L81
I'm guessing this is because the workflow sidebar component displays the "GitHub App Logs" button for a github entry and the tool sidebar component doesn't
https://github.com/dockstore/dockstore-ui2/blob/cd9414fade28e209cd3bdfcb0868d89b2b9af574/src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.html#L50-L52
This was implemented before I started working on the refresh, since there wasn't a style difference between tools and workflows previously the difference in components wouldn't have been noticeable. To resolve this I guess the question is should we add the "GitHub App Logs" button to the tools sidebar component so we can remove `app-workflow-sidebar-accordion` from my-tool.component.html? And if so, I'm guessing regular github tools shouldn't have the button since they do not currently have one now?

Also, the ordering of the groupings changed when I switched from an array to a map for the sorted tools/workflows... apparently looping through a map goes in key order -> I just added a function to the `keyvalue` pipe so it displays in the default ordering

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -1,24 +1,66 @@
...
-  ngOnInit(): void {
+  public registryToTools: Map<string, groupEntriesBySource> = new Map<string, groupEntriesBySource>([
+    ['quay.io', { groupEntryInfo: [], sourceControlTitle: 'QUAY.IO' }],
+    ['registry.gitlab.com', { groupEntryInfo: [], sourceControlTitle: 'GITLAB.COM' }],
+    ['github.com', { groupEntryInfo: [], sourceControlTitle: 'GITHUB.COM' }],
+    ['dockstore.org', { groupEntryInfo: [], sourceControlTitle: 'DOCKSTORE.ORG' }],
+    ['registry.hub.docker.com', { groupEntryInfo: [], sourceControlTitle: 'HUB.DOCKER.COM' }],
+    // For additional tools, to be removed once amazon and seven bridges have their own groupings
+    ['additional', { groupEntryInfo: [], sourceControlTitle: 'ADDITIONAL' }],
+    // Amazon and seven bridges commented out for now, to be included at a later time
```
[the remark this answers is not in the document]
KimberleyChong: I can't remember the entire reasoning but I had a chat with Denis and he said to leave them commented out for now and we can add them in later... @denis-yuen

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -1,24 +1,66 @@
...
   public toolId$: Observable<number>;
   activeTab = 0;
 
-  ngOnInit(): void {
+  public registryToTools: Map<string, groupEntriesBySource> = new Map<string, groupEntriesBySource>([
+    ['quay.io', { groupEntryInfo: [], sourceControlTitle: 'QUAY.IO' }],
+    ['registry.gitlab.com', { groupEntryInfo: [], sourceControlTitle: 'GITLAB.COM' }],
+    ['github.com', { groupEntryInfo: [], sourceControlTitle: 'GITHUB.COM' }],
+    ['dockstore.org', { groupEntryInfo: [], sourceControlTitle: 'DOCKSTORE.ORG' }],
+    ['registry.hub.docker.com', { groupEntryInfo: [], sourceControlTitle: 'HUB.DOCKER.COM' }],
```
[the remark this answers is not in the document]
coverbeck: Ah good catch on using the API, and agreed. 

If you do that, I was looking at the API response, and wondering if you should just use the friendly names in the API response, perhaps capitalized, even if they don't quite match the mocks.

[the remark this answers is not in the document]
KimberleyChong: Yep, I think friendly names is the best match. So if I'm understanding correctly, I'll call the API and use `friendlyName` and `dockerPath` to initialize my map, then proceed as currently and map the entries from `groupEntriesObject` with the `sortByRegistries` function?

cypress/fixtures/myWorkflows.json
```diff
@@ -1 +1 @@
...
...ent_id":null,"path":"github.com/B/z","repository":"z","sourceControl":"github.com","source_control_provider":"GITHUB","starredUsers":[],"topicId":null,"users":null,"workflowName":null,"workflowVersions":[],"workflow_path":"/Dockstore.cwl"},{"type":"BioWorkflow","descriptorType":"WDL","aliases":null,"author":null,"checker_id":null,"conceptDoi":null,"dbCreateDate":1465409196000,"dbUpdateDate":1465409196000,"defaultTestParameterFilePath":"/test.json","defaultVersion":null,"description":null,"descriptorTypeSubclass":"NOT_APPLICABLE","email":null,"full_workflow_path":"dockstore.org/A2/hosted-workflow","gitUrl":"git@dockstore.org:workflows/dockstore.org/A2/hosted-workflow.git","has_checker":false,"id":21,"input_file_formats":[],"isChecker":false,"is_published":false,"labels":[],"lastUpdated":1480363257315,"last_modified":null,"last_modified_date":null,"mode":"HOSTED","organization":"A","output_file_formats":[],"parent_id":null,"path":"dockstore.org/A2/hosted-workflow","repository":"hosted-workflow","sourceControl":"dockstore.org","source_control_provider":"DOCKSTORE","starredUsers":[],"topicId":null,"users":null,"workflowName":null,"workflowVersions":[],"workflow_path":"/Dockstore.wdl"}]
```
coverbeck, line 1: GitHub complaining about no EOL at EOF

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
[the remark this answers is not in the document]
coverbeck: > Yep, I think friendly names is the best match. So if I'm understanding correctly, I'll call the API and use `friendlyName` and `dockerPath` to initialize my map, then proceed as currently and map the entries from `groupEntriesObject` with the `sortByRegistries` function?

Yes.

coverbeck (commented):
Was here, waiting for use (or not) of API to get registries before reviewing again

KimberleyChong:
So **updates:**

1. `sidebar-accordion.component` for both tools and workflows now use api calls to get a list of registries/sourceControls, then `friendlyName` and `dockerPath` are used to initialize the maps used for sorting the entries - since `friendlyName` is being used as the grouping titles they will look slightly different from mock-ups (no .org or .com)
2. `sidebar-accordion.component` for both tools and workflows now use a shared style sheet (`my-entry-sidebar.scss`) and their individual style sheets now only assign colour differences between the two designs.
3. There are also 4 new colours declared in `materialColorScheme.scss`: `$tool-background-color`, `$tool-selection-color`, `$workflow-background-color`, `$workflow-selection-color`
4. The style for GitHub App Tools has been fixed! It posed some challenges since I learned GitHub App Tools are really workflows being displayed as tools. If you do have a github app tool it will be displayed at the top of your list (this is because it is not included in the regular tools sidebar component but instead it is being displayed as a workflow sidebar component, and so to avoid it from being displayed below the "Additional" section the component needs to be called above).  Within the workflows `sidebar-accordion.component`.html and .scss files there is some updated logic to display different styles for github apps and workflows (thank you kathy for helping me)
image

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -3,38 +3,48 @@ import { DockstoreTool } from 'app/shared/swagger';
...
-  styleUrls: ['./sidebar-accordion.component.scss'],
+  styleUrls: ['./sidebar-accordion.component.scss', '../../shared/styles/my-entry-sidebar.scss'],
 })
 export class SidebarAccordionComponent implements OnInit, OnChanges {
   @Input() openOneAtATime;
   @Input() groupEntriesObject: OrgToolObject<DockstoreTool>[];
   @Input() refreshMessage;
   public toolId$: Observable<number>;
+  private dockerRegistryMap = [];
+  public registryToTools: Map<string, GroupEntriesByRegistry> = new Map<string, GroupEntriesByRegistry>([]);
```
coverbeck, line 25: Nit-pick, do you need to pass an empty array to the Map() constructor?

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -1,24 +1,74 @@
...
+  constructor(private toolQuery: ToolQuery, private metadataService: MetadataService) {
+    this.metadataService.getDockerRegistries().subscribe((map) => {
+      this.dockerRegistryMap = map;
+      this.dockerRegistryMap.forEach((registry) => {
+        // Do not create new keys for amazon and seven bridges, to be put in additional category
+        if (registry.dockerPath !== null && registry.dockerPath !== 'public.ecr.aws')
+          this.registryToTools.set(registry.dockerPath, { groupEntryInfo: [], registryTitle: registry.friendlyName.toUpperCase() });
+      });
+      // For additional tools, to be removed once amazon and seven bridges have their own groupings
+      this.registryToTools.set('additional', { groupEntryInfo: [], registryTitle: 'ADDITIONAL' });
```
coverbeck, line 37: It would be good if `'additional'` were a constant so you could have the compiler enforce that you're spelling it the same way on line 57 (you are spelling it the same way now, but it's not enforced).

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -1,24 +1,74 @@
...
+        this.registryToTools.get(groupEntryObject.registry).groupEntryInfo.push(groupEntryObject);
+      } else {
+        this.registryToTools.get('additional').groupEntryInfo.push(groupEntryObject);
+      }
+    });
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes.groupEntriesObject && this.groupEntriesObject) {
+      for (let key of this.registryToTools.keys()) {
```
coverbeck, line 64: Suggest `const` instead of `let`. FYI, it may seem trivial/nit-picky, and to a degree it is -- but when I saw `let`, I scanned the code to see where you modified it (but you don't). Whereas if I see `const`, I don't scan the code to see where it gets changed, because it's not allowed. It helps readers comprehend the code a little more easily.

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -24,12 +32,51 @@ export class SidebarAccordionComponent implements OnInit {
   entryType$: Observable<EntryType>;
   EntryType = EntryType;
   public isRefreshing$: Observable<boolean>;
+  private sourceControlMap: Array<SourceControlBean> = [];
+  public sourceControlToWorkflows: Map<string, GroupEntriesBySource> = new Map<string, GroupEntriesBySource>([]);
```
coverbeck, line 36: Nit-pick, do you need to pass array to constructor?

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -24,12 +32,51 @@ export class SidebarAccordionComponent implements OnInit {
   entryType$: Observable<EntryType>;
   EntryType = EntryType;
   public isRefreshing$: Observable<boolean>;
+  private sourceControlMap: Array<SourceControlBean> = [];
```
coverbeck, line 35: I don't think you need this. You assign to it on line 46, but then you only use it on line 47... If so, I'd suggest getting rid of the property.

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -24,12 +32,51 @@ export class SidebarAccordionComponent implements OnInit {
...
     private workflowQuery: WorkflowQuery,
     public dialog: MatDialog,
     private sessionQuery: SessionQuery,
-    private alertQuery: AlertQuery
-  ) {}
+    private alertQuery: AlertQuery,
+    private metadataService: MetadataService
+  ) {
+    this.metadataService.getSourceControlList().subscribe((map) => {
+      this.sourceControlMap = map;
```
coverbeck, line 46: See my other comment; I don't think you need this.

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -24,12 +32,51 @@ export class SidebarAccordionComponent implements OnInit {
...
+   */
+  public sortBySourceControl() {
+    this.groupEntriesObject.forEach((groupEntryObject) => {
+      this.sourceControlToWorkflows.get(groupEntryObject.sourceControl).groupEntryInfo.push(groupEntryObject);
+    });
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes.groupEntriesObject && this.groupEntriesObject) {
+      for (let key of this.sourceControlToWorkflows.keys()) {
```
coverbeck, line 72: const

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -3,38 +3,48 @@ import { DockstoreTool } from 'app/shared/swagger';
...
   templateUrl: './sidebar-accordion.component.html',
-  styleUrls: ['./sidebar-accordion.component.scss'],
+  styleUrls: ['./sidebar-accordion.component.scss', '../../shared/styles/my-entry-sidebar.scss'],
 })
 export class SidebarAccordionComponent implements OnInit, OnChanges {
   @Input() openOneAtATime;
   @Input() groupEntriesObject: OrgToolObject<DockstoreTool>[];
   @Input() refreshMessage;
   public toolId$: Observable<number>;
+  private dockerRegistryMap = [];
```
kathy-t, line 24: Similar to Charles' comment for `sourceControlMap`. You can probably get rid of this property since you only use it once on line 31

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -44,8 +54,6 @@ export class SidebarAccordionComponent implements OnInit, OnChanges {
       if (this.registryToTools.has(groupEntryObject.registry)) {
         this.registryToTools.get(groupEntryObject.registry).groupEntryInfo.push(groupEntryObject);
       } else {
-        // If a tool doesn't belong to any of the specified keys then add to additional group
-        // For additional tools, remove once amazon and seven bridges have their own groups
```
kathy-t: I'm getting a console error in my browser's developer tools for line 57: `ERROR TypeError: this.registryToTools.get(...) is undefined`. Probably because `additional` isn't in the map yet when this method is called so could add a check for that too

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -3,38 +3,48 @@ import { DockstoreTool } from 'app/shared/swagger';
...
-    // Amazon and seven bridges commented out for now, to be included at a later time
-    // ["", { groupEntryInfo: [], sourceControlTitle: 'AMAZONECR'} ],
-    // ["", { groupEntryInfo: [], sourceControlTitle: 'SEVENBRIDGES'} ],
-  ]);
+  constructor(private toolQuery: ToolQuery, private metadataService: MetadataService) {
+    this.metadataService.getDockerRegistries().subscribe((map) => {
+      this.dockerRegistryMap = map;
+      this.dockerRegistryMap.forEach((registry) => {
+        // Do not create new keys for amazon and seven bridges, to be put in additional category
+        if (registry.dockerPath !== null && registry.dockerPath !== 'public.ecr.aws')
```
kathy-t, line 33: I don't think this is handling seven bridges like the comment suggests

src/app/shared/styles/my-entry-sidebar.scss
```diff
@@ -0,0 +1,102 @@
...
+  opacity: 1;
+
+  .mat-tab-label-content {
+    font-weight: bold;
+    color: #181f40;
+  }
+}
+
+// Selected entry style
+.selected,
```
kathy-t, line 85: What is `selected` used for? Does it have the same styles as `selectedGitHubAppTool`? If so, could you collapse it into one class, like `selectedTool`?

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -75,7 +91,23 @@ export class SidebarAccordionComponent implements OnInit, OnChanges {
...
   }
+
+  /**
+   * If GitHub app tool, apply tool style (blue theme)
+   */
+  isGitHubAppTool(): boolean {
+    let isAppTool: boolean = false;
+    this.entryType$
+      .subscribe((value) => {
+        if (value === this.EntryType.Tool) {
```
kathy-t, line 106: Does this work with `this.EntryType.AppTool`?

src/app/mytools/sidebar-accordion/sidebar-accordion.component.ts
coverbeck: I don't think you should rely on `dockerPath` being null; some day another registry might get added that also is null.

Can you look at `_enum` instead?

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.ts
KimberleyChong: mm no it doesn't, and when I print "value" it says it's a tool

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.ts
```diff
@@ -75,7 +91,23 @@ export class SidebarAccordionComponent implements OnInit, OnChanges {
...
   }
+
   openGitHubAppsLogs(organization: string) {
     this.dialog.open(GithubAppsLogsComponent, { width: bootstrap4largeModalSize, data: organization });
   }
+
+  /**
+   * If GitHub app tool, apply tool style (blue theme)
+   */
+  isGitHubAppTool(): boolean {
```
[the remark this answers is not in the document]
KimberleyChong: This was very helpful thank you Charles

src/app/shared/session/session.query.ts
```diff
@@ -39,6 +39,7 @@ export class SessionQuery extends Query<SessionState> {
   entryPageTitle$: Observable<string> = this.entryTypeDisplayName$.pipe(map((displayName: string) => displayName + 's'));
   myEntryPageTitle$: Observable<string> = this.entryType$.pipe(map((entryType: EntryType) => 'my ' + entryType + 's'));
   isService$: Observable<boolean> = this.entryType$.pipe(map((entryType) => entryType === EntryType.Service));
+  isAppTool$: Observable<boolean> = this.entryType$.pipe(map((entryType) => entryType === EntryType.Tool));
```
coverbeck, line 42: I think this echoes one of Kathy's earlier comments, but should this be ` === EntryType.AppTool`? Or should `isAppTool$` be named `isTool$`

KimberleyChong: Yeah no I tried using it with `=== EntryType.AppTool` but that doesn't work, GitHub App tools are being returned as type tool at least from this check

KimberleyChong: maybe it makes more sense to change it to isTool$?

coverbeck: > maybe it makes more sense to change it to isTool$?

It sounds like it based on the current code.

Do you have both tools and app tools to try this out?

coverbeck: Asking because I don't think your screenshots have any app tools... There should be a "github.com" section under My Tools.

KimberleyChong: No I don't have any app tools, I tested it out by running the `githubAppTool.ts` test locally and using the test ui. There has some app tools and regular tools

KimberleyChong:
Github App Tool with regular tools:
image

coverbeck (approved):
Pending passing tests

KimberleyChong:
Updated `package.json` `"use_circle": false` and `"webservice_version": "1.13.0-alpha.6"` since some circle artifacts are missing blocking tests, see ticket here: https://ucsc-cgl.atlassian.net/browse/SEAB-4239?atlOrigin=eyJpIjoiMDcxZjA3MjlmODY0NGVmM2E0ZmJlM2I4MjdmYmI3NjgiLCJwIjoiaiJ9

denis-yuen (approved):

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.scss
```diff
@@ -2,38 +2,28 @@
...
   background-color: $workflow-selection-color;
 }
 
-// Header style - Expanded - GitHub App Tool
+// // Expansion panel background - GitHub App Tool
 mat-expansion-panel-header.gitHubAppTool,
 mat-expansion-panel-header.gitHubAppTool[aria-expanded='true']:hover,
 mat-expansion-panel-header.gitHubAppTool[aria-expanded='true']:focus,
 :host ::ng-deep mat-expansion-panel.gitHubAppTool .mat-expansion-panel-content {
   background-color: $tool-background-color;
```
kathy-t, line 23: This looks duplicated from the mytools style sheet. Maybe you can add a class name to these styles in the mytools style sheet so it can also be used in myworkflows by importing the mytools styles to this workflow component (similar to what you did to import the shared styles)

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.scss
```diff
@@ -2,38 +2,28 @@
...
 }
 
-// Expansion body - GitHub App Tool
-:host ::ng-deep mat-expansion-panel.gitHubAppTool .mat-expansion-panel-content {
-  background-color: $tool-background-color;
-}
-
 // Selected GitHub App Tool style
-.selectedGitHubAppTool {
+.selectedTool {
```
kathy-t, line 27: Similar to my last comment, this is duplicated from mytools, so you could import the mytools styles to this component and get rid of this

kathy-t (approved):

svonworl (approved):

PR #1499, opened 2022-04-01
nicely format starred app tool and workflow descriptions

svonworl:
**Description**
This PR changes the "starred" entry pages so that (up to) the first three lines of the formatted description is displayed with each workflow or app tool, rather than the full block of raw Markdown.

See https://ucsc-cgl.atlassian.net/browse/SEAB-4106 for the "before", and the screenshot below for the "after".

Note that the description that's displayed on the "starred" entry pages is the value of the "description" field of the entry object returned by the starred endpoint, which I believe corresponds to the value of the "description" column in the entry's table.  This is not necessarily the same as the description that is displayed in the info tab, which corresponds to the currently-selected version.  Many workflows have versions with descriptions, but no description is stored in the workflow table itself.

I removed this line, because it didn't seem to have any functional effect:
https://github.com/dockstore/dockstore-ui2/blob/cd9414fade28e209cd3bdfcb0868d89b2b9af574/src/app/starredentries/starredentries.component.html#L105
If you understand its purpose, please chime in.

There was an existing css class to limit the number of lines displayed, but it was set on the `mat-card-subtitle` element, which caused it to not work.  Embedding the "limited"-and-formatted description in a `div` within the `mat-card-subtitle` caused some weird overflow and horizontal scrollbar behavior.  So, the `mat-card-subtitle` went away.

Most of the diff is indentation changes.

One IT flaked three successive times, so I added `cy.wait`s to make it pass.  Some of them are probably not necessary.   It looks like this test has failed in 10 out of the 25 most recent CircleCI runs.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4106

**Screenshot**
Screen Shot 2022-03-31 at 5 48 05 PM

### DIFF
--- a/src/app/starredentries/starredentries.component.html
+++ b/src/app/starredentries/starredentries.component.html
@@ -86,7 +86,7 @@
 
         <ng-template #workflowSummary let-workflow="workflow" let-isAppTool="isAppTool">
           <div fxLayout="row" fxLayoutGap="10px" fxLayoutAlign="space-between">
-            <a routerLink="{{ isAppTool ? '/containers/' : '/workflows/' }}{{ workflow.full_workflow_path }}" class="no-underline">
+            <a [routerLink]="(isAppTool ? entryType.AppTool : entryType.BioWorkflow) | routerLink: workflow" class="no-underline">
               <mat-card-header>
                 <mat-card-title fxLayoutAlign="space-between" fxLayoutGap="1rem">
                   <img
@@ -138,7 +138,7 @@
                 </span>
               </div>
             </div>
-            <a routerLink="{{ isAppTool ? '/containers/' : '/workflows/' }}{{ workflow.full_workflow_path }}" class="no-underline">
+            <a [routerLink]="(isAppTool ? entryType.AppTool : entryType.BioWorkflow) | routerLink: workflow" class="no-underline">
               <button mat-raised-button class="small-mat-btn-skin small-btn-structure" color="accent">View</button>
             </a>
           </div>
--- a/src/app/starredentries/starredentries.component.html
+++ b/src/app/starredentries/starredentries.component.html
@@ -101,44 +101,46 @@
             </a>
             <app-starring class="pull-right" [workflow]="workflow" (starGazersChange)="starGazersChange()"></app-starring>
           </div>
-          <mat-card-header>
-            <div *ngIf="workflow?.starredUsers.length !== 0" fxLayout="row" fxLayoutAlign="end center"></div>
-            <mat-card-subtitle class="truncate-text-1">{{ workflow?.description }}</mat-card-subtitle>
-          </mat-card-header>
-          <div>
-            <hr />
-            <div fxLayout="row" fxLayoutAlign="space-between stretch">
-              <div fxLayout="row" fxLayoutAlign="flex-start" fxLayoutGap="1rem">
-                <small class="spacing">Last updated {{ workflow?.lastUpdated | date }}</small>
-                <div fxLayout="row wrap" fxLayoutGap="1rem">
-                  <span>
-                    <mat-chip-list>
-                      <mat-chip class="bubble workflow-background" *ngFor="let label of workflow?.labels | slice: 0:3">{{
-                        label.value
-                      }}</mat-chip>
-                      <mat-chip class="bubble" *ngIf="workflow.versionVerified === true">
-                        <span class="verified-check" fxLayoutGap="0.5rem" fxLayoutAlign=" center">
-                          <mat-icon class="verified-icon">done</mat-icon>
-                          <span>Verified</span>
-                        </span>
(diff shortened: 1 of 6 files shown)


kathy-t (approved):
Looks good! 

The following are somewhat unrelated things that I noticed while looking at the starred entries page. 
- Should we be displaying the full workflow/tool path instead of organization + repo? For the starred tools tab, it's hard to tell if it's a GitHub App tool if we can't see the full path. It seems like we usually use the full entry path, like for listing the entries for a collection or on the search page.
- We're not showing the description for regular tools. I feel like we should to be consistent on the tools tab since GitHub App tools will have descriptions.

These can be follow up tickets, but just wanted to bring it up for discussion.

src/app/starredentries/starredentries.component.html
```diff
@@ -101,44 +101,46 @@ <h4>
...
+                      </span>
+                    </mat-chip>
+                  </mat-chip-list>
+                </span>
               </div>
-              <a routerLink="{{ isAppTool ? '/containers/' : '/workflows/' }}{{ workflow.full_workflow_path }}" class="no-underline">
-                <button mat-raised-button class="small-mat-btn-skin small-btn-structure" color="accent">View</button>
-              </a>
             </div>
+            <a routerLink="{{ isAppTool ? '/containers/' : '/workflows/' }}{{ workflow.full_workflow_path }}" class="no-underline">
```
coverbeck, line 141: There is a [pipe](https://github.com/dockstore/dockstore-ui2/blob/ff13411f0e34898a183263f8ba6a81a4e002d3d5/src/app/entry/router-link.pipe.ts#L6-L5) to calculate the routerLink, but it takes an EntryType as a param, so you'd have to change your invocation of the ng-template, which doesn't look too hard.

src/app/starredentries/starredentries.component.html
```diff
@@ -101,44 +101,46 @@ <h4>
...
-          <mat-card-header>
-            <div *ngIf="workflow?.starredUsers.length !== 0" fxLayout="row" fxLayoutAlign="end center"></div>
-            <mat-card-subtitle class="truncate-text-1">{{ workflow?.description }}</mat-card-subtitle>
-          </mat-card-header>
-          <div>
+          <div class="ml-4 mr-2">
+            <div class="truncate-text-3 mb-3">
+              <app-markdown-wrapper
+                [data]="workflow?.description"
+                [baseUrl]="workflow?.providerUrl | baseUrl: workflow?.defaultVersion"
```
coverbeck, line 108: I see this pattern exists already, but @kathy-t what would it mean to have `workflow.defaultVersion` as the baseUrl? I understand the `providerUrl` -- what assets would ever get resolved against the defaultVersion?

coverbeck (commented):
Looks good.

I know the routerLink was already like that, but I have an abhorrence for duplicate code and since you're in here anyway... But if you don't want to do it, could you please log an issue for it? My other comment is just curiosity.

src/app/starredentries/starredentries.component.html
svonworl: The workflow `defaultVersion` is an argument to the `baseUrl` pipe, used to modify certain source code control site URLs to point at a particular branch/tag. 
https://github.com/dockstore/dockstore-ui2/blob/develop/src/app/shared/entry/base-url.pipe.ts

kathy-t: ^^ what Steve said

It's really only used if the workflow description is from a README that contains relative links

src/app/starredentries/starredentries.component.html
svonworl: I pushed a mod, please take a look and see if that's what you were after...

svonworl: Let me take this opportunity to perform my weekly advocation...
Imagine if all of the tool/workflow/service/apptool objects knew how to calculate links to themselves, and other such quantities.  Or, perhaps precalculated fields were added to the objects, containing such information.  Then, all the big fragile `case` statements go away, and the code gets simpler!
Q: How cool would that be?
A: Pretty darn-tootin' cool.

coverbeck: 1. Yes that looks right
2. Feel free to create a ticket on moving things to a class, but I think in the meantime it makes sense to avoid duplication (best-case it's duplication, worst-case it's not exact duplication).

coverbeck (approved):

denis-yuen (approved):

KimberleyChong (approved):

src/app/starredentries/starredentries.component.html
denis-yuen: Interesting idea, I encourage the ticket. To be clear, were you thinking the TRS objects or the model JPA objects? (Or UI objects?)

PR #1501, opened 2022-04-04
Up the timeout to 45 seconds

Richard-Hansen:
**Description**
Staging auth tests are failing due to a command timeout. Running the tests against staging from my local environment, we can see all the tests are passing, but some our running a bit long:
Screen Shot 2022-04-04 at 10 03 01 AM

This PR just ups the timeout duration for Cypress tests.

**Issue**

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -3,7 +3,7 @@
   "baseUrl": "http://localhost:4200",
   "viewportHeight": 900,
   "viewportWidth": 1440,
-  "defaultCommandTimeout": 30000,
+  "defaultCommandTimeout": 45000,
   "requestTimeout": 30000,
   "retries": {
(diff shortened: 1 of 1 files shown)


Richard-Hansen:
> Grasping at straws, does this indicate some deeper underlying issue of performance with staging? Probably not, but just asking.
> 
> Staging does have more data than dev, but should have slightly less than prod -- is prod seeing the same issue, or are we not running the tests there?

Not sure, this could be a performance issue though, as we aren't seeing the failures on dev. The prod auth tests aren't automated yet (https://github.com/dockstore/dockstore-ui2/pull/1500), but I'd expect to see similar time issues there.

denis-yuen (approved):

PR #1504, opened 2022-04-08
Disable username input field if not allowed

KimberleyChong:
**Description**
JIRA: [DOCK-1959](https://ucsc-cgl.atlassian.net/browse/DOCK-1959?atlOrigin=eyJpIjoiY2ZhOWQ0ODdkZjA1NDljMjg4NDg2NjY2NWRkMTgxN2UiLCJwIjoiaiJ9)

The purpose of this PR is to disable the form field input when a user cannot change their username.

Really the only change is I added a function to disable the form field if `canChangeUsername$` is false.

### Before:
image

### After:
image


src/app/loginComponents/accounts/internal/change-username/change-username.component.ts
```diff
@@ -97,4 +98,17 @@ export class ChangeUsernameComponent implements OnInit {
   updateUsername() {
     this.userService.updateUsername(this.username);
   }
+
+  /**
+   * Enable or disable form input if user can or cannot update username
+   */
+  enableDisableFormControl() {
+    this.canChangeUsername$.subscribe((canChangeUsername: boolean) => {
```
coverbeck, line 106: Not sure, but I think you need to unsubscribe.

Or better, I think, just use `this.canChangeUsername$` directly in HTML like the button does [here](https://github.com/dockstore/dockstore-ui2/blob/ed75de98a8018ae72348a30de59da8c08a3ef05b/src/app/loginComponents/accounts/internal/change-username/change-username.component.html#L48-L47) (I think you can just add disabled to the input).

denis-yuen (commented):
Might not want to invest too much in this, this should be moved to a dialogue long-term anyway. 
Added a zeplin link to the other ticket linked to the ticket that this is linked to

src/app/loginComponents/accounts/internal/change-username/change-username.component.ts
coverbeck: > Yeah I first tried disabling it from the html directly, but for reactive forms it's not possible

Thanks for clarifying, I did not know that.

> I'll update this to unsubscribe?

I think you need to, yes.  And I think the way to it is to use `takeUntil`, like line 61.

coverbeck (approved):

kathy-t (approved):

denis-yuen (approved):

svonworl (approved):

PR #1506, opened 2022-04-12
Upping this timeout as well for Cypress tests

Richard-Hansen:
**Description**
There appear to be smoke test timeouts failing after a 30 second wait. Upping the request timeout to 45 seconds (in a previous PR, the command timeout was increased to 45 seconds).

**Issue**

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -4,7 +4,7 @@
   "viewportHeight": 900,
   "viewportWidth": 1440,
   "defaultCommandTimeout": 45000,
-  "requestTimeout": 30000,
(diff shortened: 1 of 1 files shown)


coverbeck (approved):
I don't like this trend but I guess there's no other option.

Richard-Hansen:
> I don't like this trend but I guess there's no other option.

Agreed, on both points. Not sure of other better approaches. If this continues to occur, I'll create a ticket to investigate what calls are taking the most time, which may give us opportunities to improve efficiency.

PR #1512, opened 2022-04-26
Update entry-actions.service.ts

aofarrel:
**Description**
Fix typo in dialog box informing the user that published entries need a default version. There are two more instances of "the the" in this repo not in this PR.

**Issue**
https://github.com/dockstore/dockstore/issues/4853

Feel free to reject this PR if it's more trouble than it's worth r/e testing -- I was not able to get our CI working correctly on my machine. It's claiming there is a syntax error in the generated file node_modules/@angular/compiler-cli/bundles/ngcc/main-ngcc.js, which I want to flag here in case this is actually a bug... I get the feeling that could just as likely be a consequence of me fumbling the installation of three(?!) different versions of node, so I haven't made a ticket.

### DIFF
--- a/src/app/shared/entry-actions/entry-actions.service.ts
+++ b/src/app/shared/entry-actions/entry-actions.service.ts
@@ -111,7 +111,7 @@ export class EntryActionsService {
   openNoDefaultDialog(entry: Entry, entryType: string, showVersions: EventEmitter<void> | null): void {
     const informationDialogData: InformationDialogData = {
       title: 'Default Version Required',
-      message: `Your ${entryType} must have a default version to be published.  Please use the the Actions menu in the Versions tab to select a default version.`,
+      message: `Your ${entryType} must have a default version to be published.  Please use the Actions menu in the Versions tab to select a default version.`,
       closeButtonText: 'OK',
     };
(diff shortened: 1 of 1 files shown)


coverbeck (approved):
FYI, you should be creating branches starting with `feature`, not `hotfix` (or use Hubflow, e.g., `git hf feature start 4853`). In practice, it doesn't matter that much now, but

1. It's the convention
2. I think Hubflow does get confused if there are multiple hotfix branches, at least locally.

denis-yuen (approved):

aofarrel:
> FYI, you should be creating branches starting with `feature`, not `hotfix` (or use Hubflow, e.g., `git hf feature start 4853`). In practice, it doesn't matter that much now, but
> 
> 1. It's the convention
> 2. I think Hubflow does get confused if there are multiple hotfix branches, at least locally.

Appreciate this advice, will keep it in mind!

svonworl:
Thanks for fixing this.  I am the original author of the "the the" and recall someone noticing it.  I thought I'd singularized it in the PR, but maybe I never pushed?  Hmmm, I need to go back and do some forensics, for self-improvement's sake. :)

svonworl (approved):

denis-yuen:
renamed to https://github.com/dockstore/dockstore-ui2/pull/1640

PR #1515, opened 2022-04-28
Fix CircleCI artifacts URL

kathy-t:
**Description**
This PR fixes the failure to retrieve CircleCI artifacts. They're not missing from CircleCI (can see them in the Artifacts URL of a CircleCI build). The problem is that the artifact download URL changed after this [build](https://app.circleci.com/pipelines/github/dockstore/dockstore/7435/workflows/b025395e-4a6e-4c55-a449-62226dc54729/jobs/18239).

- Old URL: `https://18239-33383826-gh.circle-artifacts.com/0/tmp/artifacts/dockstore-webservice-1.13.0-SNAPSHOT.jar`
- New URL: `https://output.circle-artifacts.com/output/job/94cd379a-e46a-4a0a-b24e-0a644c192a5d/artifacts/0/tmp/artifacts/dockstore-webservice-1.13.0-SNAPSHOT.jar`

This [discuss topic](https://discuss.circleci.com/t/artifact-urls-now-private/43282) has people who ran into this change as well and a [CircleCI employee said that they made some changes internally](https://discuss.circleci.com/t/artifact-urls-now-private/43282/8). It seems like the URL we were using is part of the [v1 API](https://circleci.com/docs/api/#download-an-artifact-file), which is deprecated, so it's probably best to move away from it anyway.

This PR creates a script that gets the artifact URL by calling the [CircleCI endpoint that retrieves all artifacts for a build](https://circleci.com/docs/api/v2/#operation/getJobArtifacts), finding the correct artifact based on its path, and returning its URL. This method should guarantee that any future changes to the artifact URL doesn't affect us. 

Note that the script finds the correct artifact by looking for an artifact with a path that contains the `<artifact-name>` that is provided. The artifacts that we're interested in have paths with the format `tmp/artifacts/<artifact-name>`. An advantage of using a "contains" filter to find the artifact is that we won't need to specify the specific Dockstore webservice jar version in `scripts/run-webservice-script.sh`. It'll look for an artifact that contains `dockstore-webservice` in the CircleCI build.

Somewhat unrelated changes:
- Added 1.13.0 migrations (forgot to do this for the front-end)
- I updated the Java version from 11 to 17 because the latest develop CircleCI webservice JAR failed to start on CircleCI due to this error: `java.lang.UnsupportedClassVersionError: io/dockstore/common/EntryType has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0`.
  - Had to use `openjdk-17-jdk` instead of `openjdk-17-jre` because the jre version failed to install on CircleCI.
  - Tested Java 17 with the current artifactory `1.13.0-alpha.6` JAR and most of the tests passed. The failed tests are due to unrelated flakiness. Can view the CircleCI job [here](https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/7600/workflows/54c20b38-3dc0-44d4-883e-30bfffd476de) if curious.

**Issue**
[SEAB-4239](https://ucsc-cgl.atlassian.net/browse/SEAB-4239)

### DIFF
--- a/scripts/generate-openapi-script.sh
+++ b/scripts/generate-openapi-script.sh
@@ -6,8 +6,6 @@ set -o xtrace
 
 GENERATOR_VERSION="4.3.0"
 BASE_PATH="https://raw.githubusercontent.com/dockstore/dockstore/$npm_package_config_webservice_version"
-CIRCLE_CI_PATH="https://""$npm_package_config_circle_build_id""-33383826-gh.circle-artifacts.com/0/tmp/artifacts"
-
 # DOCKSTORE-2428 - demo how to add new workflow language, generate from local copy of swagger
 # Uncomment this to use your local copy of swagger instead
 # BASE_PATH="../dockstore"
@@ -19,8 +17,8 @@ rm -Rf src/app/shared/openapi
 
 if [ "$npm_package_config_use_circle" = true ]
 then
-        SWAGGER_PATH="${CIRCLE_CI_PATH}""/swagger.yaml"
-        OPENAPI_PATH="${CIRCLE_CI_PATH}""/openapi.yaml"
+        SWAGGER_PATH=$(./scripts/get-circleci-artifact-url.sh $npm_package_config_circle_build_id swagger.yaml)
+        OPENAPI_PATH=$(./scripts/get-circleci-artifact-url.sh $npm_package_config_circle_build_id openapi.yaml)
 else
         SWAGGER_PATH="${BASE_PATH}""/dockstore-webservice/src/main/resources/swagger.yaml"
         OPENAPI_PATH="${BASE_PATH}""/dockstore-webservice/src/main/resources/openapi3/openapi.yaml"
--- a/scripts/generate-openapi-script.sh
+++ b/scripts/generate-openapi-script.sh
@@ -17,8 +17,8 @@ rm -Rf src/app/shared/openapi
 
 if [ "$npm_package_config_use_circle" = true ]
 then
-        SWAGGER_PATH=$(./scripts/get-circleci-artifact-url.sh $npm_package_config_circle_build_id swagger.yaml)
-        OPENAPI_PATH=$(./scripts/get-circleci-artifact-url.sh $npm_package_config_circle_build_id openapi.yaml)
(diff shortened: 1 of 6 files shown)


scripts/generate-openapi-script.sh
```diff
@@ -19,8 +17,8 @@ rm -Rf src/app/shared/openapi
 
 if [ "$npm_package_config_use_circle" = true ]
 then
-        SWAGGER_PATH="${CIRCLE_CI_PATH}""/swagger.yaml"
```
denis-yuen, line 22: codacy seems to be warning about removing the quotes here, didn't read too deeply into it, might be worth a look

denis-yuen (approved):
with minor comment

denis-yuen:
> I updated the Java version from 11 to 17 because the latest develop CircleCI webservice JAR failed to start on CircleCI due to this error 

as noted in DM, I think this makes sense for webservice since we control the JDK there. We'll need to be a little more careful with CLI and lambdas since users may have a harder time updating to Java 17 LTS and there isn't an AWS Java 17 lanbda yet respectively (upvote at https://github.com/aws/aws-lambda-base-images/issues/29 )

coverbeck:
> > I updated the Java version from 11 to 17 because the latest develop CircleCI webservice JAR failed to start on CircleCI due to this error
> 
> as noted in DM, I think this makes sense for webservice since we control the JDK there. We'll need to be a little more careful with CLI and lambdas since users may have a harder time updating to Java 17 LTS and there isn't an AWS Java 17 lambda yet respectively (upvote at [aws/aws-lambda-base-images#29](https://github.com/aws/aws-lambda-base-images/issues/29) )

I vaguely remember some discussion about this... Denis, didn't you change things so that even though the web service is built with 17, it compiles to 1.11 class file format?

coverbeck (approved):

denis-yuen:
> I vaguely remember some discussion about this... Denis, didn't you change things so that even though the web service is built with 17, it compiles to 1.11 class file format?

Kinda, https://github.com/dockstore/dockstore/blob/1.13.0-alpha.6/pom.xml#L185-L187 the `release` property is supposed to ensure that we compile to Java 11, I didn't change it, but I also didn't look into it. Looking now, I do wonder if it is compatible with `forceJavacCompilerUse`

Richard-Hansen (approved):

PR #1524, opened 2022-05-05
DOCK-2105: Render newlines in GitHub app log messages

svonworl:
**Description**
This PR changes the UI so that newlines in the github apps log messages are formatted as newlines, rather than whitespace.  The log message style is changed so it preserves whitespace and is rendered in a small, monospace font, with a tighter-than-usual line height to fit more output in the available space.  The formatting is similar to how the message might look if it was within a `<pre>` tag.

Screen Shot 2022-05-04 at 5 31 28 PM

These changes will improve the readability of multiline messages, and preserve the formatting of useful, preformatted messages, such as [some of] those that emerge via exceptions from deep in the YAML parser.

This PR was motivated by Kathy's review ticket feedback, combined with my earlier realization that it was The Right Thing to Do.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2105

### DIFF
--- a/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.html
+++ b/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.html
@@ -39,9 +39,7 @@
               class="expanded-row-detail"
               [@detailExpand]="element === expandedElement ? 'expanded' : 'collapsed'"
             >
-              <div class="p-2">
-                {{ element.message }}
-              </div>
+              <div class="p-2 log-entry-message">{{ element.message }}</div>
             </div>
           </td>
         </ng-container>
--- a/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.scss
+++ b/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.scss
@@ -25,3 +25,9 @@ tr.element-row:not(.expanded-row):active {
   overflow: hidden;
   display: flex;
 }
+
+.log-entry-message {
+  white-space: pre-wrap;
+  font-family: monospace;
+  line-height: 1.25;
+}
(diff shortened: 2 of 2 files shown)


coverbeck (approved):
Based on the screenshot, it would be nice if you could indent the children of "In version...".  It doesn't matter for one repo->one workflow, but would be useful for one repo -> many workflows.

But I don't think you can easily tell. And this is much better that what it was. So, it's fine, just brainstorming.

What I mean is, now it's:

```
In version foo of workflow goo
A validation error
In version foo of workflow bar
A validation error
Another validation error
```

to:

```
In version foo of workflow goo
  A validation error
In version foo of workflow bar 
  A validation error
  Another validation error
```

Also see my comment on the web service ticket with a suggestion on the wording.

svonworl:
> Based on the screenshot, it would be nice if you could indent the children of "In version...". It doesn't matter for one repo->one workflow, but would be useful for one repo -> many workflows.
> 
> But I don't think you can easily tell. And this is much better that what it was. So, it's fine, just brainstorming.
> 
> What I mean is, now it's:
> 
> ```
> In version foo of workflow goo
> A validation error
> In version foo of workflow bar
> A validation error
> Another validation error
> ```
> 
> to:
> 
> ```
> In version foo of workflow goo
>   A validation error
> In version foo of workflow bar 
>   A validation error
>   Another validation error
> ```
> 
> Also see my comment on the web service ticket with a suggestion on the wording.

The messages are formatted by the webservice, so we can do the indenting pretty easy (in the other PR).

kathy-t (approved):

denis-yuen (approved):
> This PR was motivated by Kathy's review ticket feedback, combined with my earlier realization that it was The Right Thing to Do. 

:grin:

Richard-Hansen (approved):
The recent changes to GitHub App logs have made them far more actionable, thanks for taking the time to make these upgrades!

PR #1531, opened 2022-05-11
Fix ORCID ID font size

eileenzhh:
**Description**
The change of font caused the original defined font-size to be too small, increased font size by a bit (0.1em) to match prod.

**Issue**
[DOCK-2146](https://ucsc-cgl.atlassian.net/browse/DOCK-2146)

### DIFF
--- a/src/app/stargazers/stargazers.component.html
+++ b/src/app/stargazers/stargazers.component.html
@@ -40,7 +40,7 @@
             </a>
             <h6 class="text-break" *ngIf="stargazer.orcid">
               ORCID <img src="../../assets/images/account-logos/orcid.svg" alt="ORCID iD logo" class="orcid-id-logo site-icons-small" />
-              <a href="https://orcid.org/{{ stargazer.orcid }}" style="font-size: 0.9em">{{ stargazer.orcid }}</a>
+              <a href="https://orcid.org/{{ stargazer.orcid }}">{{ stargazer.orcid }}</a>
             </h6>
           </div>
         </mat-card>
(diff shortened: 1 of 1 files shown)


eileenzhh:
Previously:
image

Production:
image

Now: 
image

src/app/stargazers/stargazers.component.html
```diff
@@ -40,7 +40,7 @@ <h5>{{ stargazer.username }}</h5>
             </a>
             <h6 class="text-break" *ngIf="stargazer.orcid">
               ORCID <img src="../../assets/images/account-logos/orcid.svg" alt="ORCID iD logo" class="orcid-id-logo site-icons-small" />
-              <a href="https://orcid.org/{{ stargazer.orcid }}" style="font-size: 0.8em">{{ stargazer.orcid }}</a>
+              <a href="https://orcid.org/{{ stargazer.orcid }}" style="font-size: 0.9em">{{ stargazer.orcid }}</a>
```
coverbeck, line 43: It was like this, but per the [Frontend Guide](https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide), we should avoid inline styles. Also the guideline recommends using `rem`s, although that's more as opposed to using pixels instead of `em`s.

coverbeck (approved):
I left a comment, but don't feel too strongly about it. I think @kathy-t has a better understanding of our styling that I do, so I'll defer to her.

src/app/stargazers/stargazers.component.html
kathy-t: Poked around the places we have ORCID IDs and it seems like this is the only place where we define a specific font-size. It might be better to remove the font size style entirely if the ORCID ID line doesn't overflow to the next line

kathy-t (approved):

PR #1529, opened 2022-05-10
Fix: Selectively add CWL instructions (DOCK-2144)

amarjandu:
**Description**
A description of the PR, should include a decent explanation as to why this change was needed and a decent explanation as to what this change does

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2144

Please make sure that you've checked the following before submitting your pull request. Thanks!

- [ ] Check that your code compiles by running `npm run build`
- [ ] If this is the first time you're submitting a PR or even if you just need a refresher, consider reviewing our [style guide](https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide#pr-checklist)
- [ ] Do not bypass Angular sanitization (bypassSecurityTrustHtml, etc.), or justify why you need to do so
- [ ] If displaying markdown, use the `markdown-wrapper` component, which does extra sanitization
- [ ] Do not use cookies, although this may change in the future
- [ ] Run `npm audit` and ensure you are not introducing new vulnerabilities
- [ ] Do due diligence on new 3rd party libraries, checking for CVEs
- [ ] Don't allow user-uploaded images to be served from the Dockstore domain

PR #1533, opened 2022-05-13
Add blog link below Username on User-pages

eileenzhh:
**Description**
Add blog link (from github user page) below the username in user-pages. 

**Issue**
[DOCK-2165](https://ucsc-cgl.atlassian.net/browse/DOCK-2165)

### DIFF
--- a/src/app/user-page/user-page.component.html
+++ b/src/app/user-page/user-page.component.html
@@ -13,6 +13,9 @@
       <img mat-card-image [src]="user?.avatarUrl" class="mt-0 mx-auto mb-3" alt="Dockstore avatar" />
       <mat-card-content fxLayout="column" fxLayoutAlign="center center">
         <h4 matLine class="accent-1-dark mt-2 mb-0">{{ user?.username }}</h4>
+        <a href="{{ gitHubProfile?.link }}" target="_blank" class="accent-1-dark">
+          <span>{{ gitHubProfile?.link }}</span>
+        </a>
       </mat-card-content>
     </mat-card>
 


eileenzhh:
image

denis-yuen (commented):
Let's do the bio in the card as well in this PR. The SonarCloud warning is worth doing as well, not sure if you need permission to follow the link but let me know

src/app/user-page/user-page.component.html
```diff
@@ -13,6 +13,9 @@
       <img mat-card-image [src]="user?.avatarUrl" class="mt-0 mx-auto mb-3" alt="Dockstore avatar" />
       <mat-card-content fxLayout="column" fxLayoutAlign="center center">
         <h4 matLine class="accent-1-dark mt-2 mb-0">{{ user?.username }}</h4>
+        <a href="{{ gitHubProfile?.link }}" target="_blank" class="accent-1-dark">
```
coverbeck, line 16: You should add `rel="noopener noreferrer"`, like (almost?) all our other external links.

src/app/user-page/user-page.component.html
```diff
@@ -13,6 +13,9 @@
...
+          <span>{{ gitHubProfile?.link }}</span>
```
coverbeck, line 17: It certainly doesn't hurt, but just curious, did you add `<span>` for a reason? Is there some styling that relies on the `<span>`?

coverbeck (commented):
In general, please try to refer to the GitHub issue in your PR and/or commit message, if there is one. See [this thread](https://github.com/dockstore/dockstore-ui2/pull/1529#pullrequestreview-968522037).

src/app/user-page/user-page.component.scss
```diff
@@ -24,6 +24,10 @@ img[alt='Google avatar'] {
   background: $accent-3-dark;
 }
 
+.text-center-align {
```
svonworl, line 27: In general, if you need to create a new style, first check the global `styles.css` to see if something similar already exists.   If not, and your new style might be broadly useful, put it into the global `styles.css`, along with a comment that explains how it is/should be used.

src/app/user-page/user-page.component.html
```diff
@@ -13,6 +13,12 @@
       <img mat-card-image [src]="user?.avatarUrl" class="mt-0 mx-auto mb-3" alt="Dockstore avatar" />
       <mat-card-content fxLayout="column" fxLayoutAlign="center center">
         <h4 matLine class="accent-1-dark mt-2 mb-0">{{ user?.username }}</h4>
+        <mat-card-subtitle *ngIf="gitHubProfile.bio" class="normal-line-height weight-bold accent-1-dark mt-2 mb-0 text-center-align">
+          {{ gitHubProfile.bio }}
+        </mat-card-subtitle>
+        <a rel="noopener noreferrer" href="{{ gitHubProfile?.link }}" target="_blank" class="accent-1-dark">
```
kathy-t, line 19: Right now, it doesn't show a link if you don't have one because it's just an empty `<a>` tag. Can add `*ngIf="gitHubProfile?.link` to only add the `<a>` tag if the link exists

src/app/user-page/user-page.component.html
```diff
@@ -13,6 +13,12 @@
...
+          {{ gitHubProfile?.link }}
```
kathy-t, line 20: If you add the `*ngIf` check, you don't need the question mark here

src/app/user-page/user-page.component.scss
coverbeck: Can you please address this comment? Addressing it might be justifying via a comment why it's proper for it to be here, or actually moving the style, but I just want to make sure that you saw the comment and responded to it (maybe you already had, but there's no indication of it here in the PR).

Thanks!

eileenzhh: Yup I'm sorry I completely missed this comment! I did a simple keyword search in `styles.css`and found that there was no global style existing, I'll add it to the global file after double checking!

denis-yuen (approved):

src/styles.scss
```diff
@@ -1000,6 +1000,13 @@ mat-tab-group.no-pagination {
   color: rgba(0, 0, 0, 0.54);
 }
 
+// Styling for information below the user icon in user-pages
+.mat-card-user-info {
```
coverbeck, line 1004: I think this file should have global styles that might be used in multiple places. This seems sort of specific to the one page. Having a class that encapsulates `text-align: center;` seems like a good candidate, which I think I was Steve's original point.

On the other hand, if you envision this being used in other places, then it seems OK. I think I'll defer to Kathy on this, who has a better sense of our styling.

eileenzhh: Ohh okay that makes sense to me, I'll make the changes :+1: thank you for the suggestion and clarification!

kathy-t (approved):

PR #1537, opened 2022-05-19
Fix 'info' icon not appearing under Activity of user page

eileenzhh:
**Description**
On the new user page, the 'info' icon was not appearing correctly and was appearing as text instead under the activity bar. This was fixed by manually adding material-icon and mat-icon classes to the mat-icon element that should appear for it automatically but does not in this case. Unsure as to why it does not work here and would love to hear ideas/alternative solutions from other devs.

**Issue**
GitHub issue: [#4902](https://github.com/dockstore/dockstore/issues/4902)
Jira: [DOCK-2168](https://ucsc-cgl.atlassian.net/browse/DOCK-2168)

### DIFF
--- a/src/app/home-page/recent-events/recent-events.component.html
+++ b/src/app/home-page/recent-events/recent-events.component.html
@@ -1,6 +1,6 @@
 <app-loading [loading]="loading$ | async">
   <div *ngIf="noEvents$ | async">
-    <mat-card class="alert alert-info" role="alert">
+    <mat-card class="alert alert-info mx-1" role="alert">
       <mat-icon class="mat-icon material-icons">info</mat-icon> No events found. Star entries or organizations to see events for them. Read
       <a [href]="starringDocUrl" target="_blank" rel="noopener noreferrer">Starring Tools and Workflows</a>
       to learn how to star entries.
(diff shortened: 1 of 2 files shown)


eileenzhh:
Before:
image
After: 
image

coverbeck (commented):
> Unsure as to why it does not work here and would love to hear ideas/alternative solutions from other devs.

To clarify, you're saying this PR does not fix the problem, correct?

If not, did you see my comment in the issue: "When I open the [html file](https://github.com/dockstore/dockstore-ui2/blob/ee443beff9b305997b81927ffa107d9f46c6b354/src/app/home-page/recent-events/recent-events.component.html#L4) in WebStorm, I see an error "Component or directive matching mat-icon element is out of the current Angular module's scope", so probably related to that."

If that's really the issue adding this line to recent-events.module.ts may fix it:

```
import { MatIconModule } from '@angular/material/icon';
```

eileenzhh:
> To clarify, you're saying this PR does not fix the problem, correct?
No this PR does fix the problem, it's just that the root of the problem of 'mat-icon' not working was very random and strange to me and this fix was kind of a 'hard-coded' fix. I do think your solution may solve it though and will give it a shot.

kathy-t (approved):

PR #1544, opened 2022-05-26
rename smoke tests

Richard-Hansen:
**Description**
This PR changes the name of the CI smoke tests such that they are a bit more intuitive.

**Issue**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ executors:
 jobs:
   # Specify the webpage url (https://dev.dockstore.net or https://staging.dockstore.org or https://dockstore.org)
   # and stack (dev or staging or prod) and the no auth smoke tests will be run against the corresponding webpage.
-  uptime_monitor_no_auth:
+  no_auth_smoke_tests:
     parameters:
       stack:
         type: string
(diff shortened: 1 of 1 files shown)


chmreid (approved):
🎉

.circleci/config.yml
```diff
@@ -413,38 +413,38 @@ workflows:
               only:
                 - develop
     jobs:
-      - uptime_monitor_no_auth:
-          name: "uptime_monitor_no_auth_dev"
+      - no_auth_smoke_tests:
+          name: "no_auth_smoke_tests_dev"
```
denis-yuen, line 417: if this is in quotes, does it work without `_`s?

denis-yuen (approved):

.circleci/config.yml
Richard-Hansen: It would still work. The `name` key here overrides the default test name in CircleCI, so if we remove the `_`s, we'll have camelCase amongst snake_case, which might look a bit odd.

No strong feelings, I can change them if we want.

denis-yuen: I was more curious about actual spaces, no biggie

PR #1553, opened 2022-06-03
Don't request categories for unpublished entries (dockstore/dockstore#4821)

amarjandu:
**Description**
Removes fetching of categories for non-published entries. Attempting to request categories for a 
non-published workflow will currently result in a 400.

Screen Shot 2022-06-08 at 2 19 40 PM

**Issue**
https://github.com/dockstore/dockstore/issues/4821

### DIFF
--- a/src/app/categories/state/entry-categories.service.ts
+++ b/src/app/categories/state/entry-categories.service.ts
@@ -26,7 +26,7 @@ export class EntryCategoriesService {
   constructor(
     private entryCategoriesStore: EntryCategoriesStore,
     private entryCategoriesQuery: EntryCategoriesQuery,
-    private entriesService: EntriesService,
+    private entriesService: EntriesService
   ) {
     /**
      * An observable list of the categories for a given entry.
@@ -42,14 +42,13 @@ export class EntryCategoriesService {
   /**
    * Updates the list of categories for the specified entry.
    */
-  updateEntryCategories(entryId: number) {
+  updateEntryCategories(entryId: number, published: boolean) {
     this.entryCategoriesStore.setLoading(true);
     this.entryCategoriesStore.setError(false);
     this.entryCategoriesStore.remove();
     this.currentSubscription?.unsubscribe();
-    this.currentSubscription = this.entriesService
-      .entryCategories(entryId)
-      .subscribe(
+    if (published) {
+      this.currentSubscription = this.entriesService.entryCategories(entryId).subscribe(
         (categories: Array<Category>) => {
           this.entryCategoriesStore.setLoading(false);
           this.entryCategoriesStore.setError(false);
@@ -60,5 +59,8 @@ export class EntryCategoriesService {
           this.entryCategoriesStore.setError(true);
         }
       );
+    } else {
+      this.entryCategoriesStore.set(null);
+    }
   }
 }
--- a/src/app/categories/state/entry-categories.service.ts
+++ b/src/app/categories/state/entry-categories.service.ts
@@ -41,14 +41,17 @@ export class EntryCategoriesService {
 
   /**
    * Updates the list of categories for the specified entry.
+   * @param {number} entryId - The ID for the associated entry.
+   * @param {boolean} published - The published status for the entry, non-published entries cannot belong to a category.
    */
   updateEntryCategories(entryId: number, published: boolean) {
-    this.entryCategoriesStore.setLoading(true);
-    this.entryCategoriesStore.setError(false);
     this.entryCategoriesStore.remove();
     this.currentSubscription?.unsubscribe();
     this.currentSubscription = null;
+    this.entryCategoriesStore.setError(false);
+    this.entryCategoriesStore.setLoading(false);
     if (published) {
+      this.entryCategoriesStore.setLoading(true);
       this.currentSubscription = this.entriesService.entryCategories(entryId).subscribe(
         (categories: Array<Category>) => {
           this.entryCategoriesStore.setLoading(false);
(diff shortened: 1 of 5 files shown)


svonworl:
You fixed my bug, thanks!

src/app/categories/state/entry-categories.service.ts
```diff
@@ -42,14 +42,13 @@ export class EntryCategoriesService {
   /**
    * Updates the list of categories for the specified entry.
    */
-  updateEntryCategories(entryId: number) {
+  updateEntryCategories(entryId: number, published: boolean) {
     this.entryCategoriesStore.setLoading(true);
     this.entryCategoriesStore.setError(false);
     this.entryCategoriesStore.remove();
     this.currentSubscription?.unsubscribe();
```
svonworl, line 49: In this code, `currentSubscription` isn't always set to a new value each time through this function, it may remain set to the old value if `published` is false.  So, it'd probably be good to either:
* confirm via testing that `unsubscribe` can be called multiple times on a given subscription, or
* set `currentSubscription` to `null` after the `unsubscribe`.

src/app/categories/state/entry-categories.service.ts
```diff
@@ -60,5 +59,8 @@ export class EntryCategoriesService {
           this.entryCategoriesStore.setError(true);
         }
       );
+    } else {
+      this.entryCategoriesStore.set(null);
```
coverbeck, line 63: In the code above, if an entry doesn't have any categories, this method is called with an empty array. I think you should do the same here. The signature for the method is: `set(entities: SetEntities<EntityType>...` where EntityType is the generic, and `SetEntities` is `export declare type SetEntities<Entity> = Entity[] | Entities<Entity> | HashMap<Entity>`, which is a long-winded way of saying the method shouldn't be passed a `null`.

Digression: We mean to turn on null checking some day so the compiler would flag this, but there are currently too many exceptions, so we can't. But we want to avoid introducing new code passing null unless the signature explicitly allows it.

src/app/categories/state/entry-categories.service.ts
```diff
@@ -42,14 +42,13 @@ export class EntryCategoriesService {
   /**
    * Updates the list of categories for the specified entry.
    */
-  updateEntryCategories(entryId: number) {
+  updateEntryCategories(entryId: number, published: boolean) {
```
coverbeck, line 45: This is works, and can stay if you want, but I would be more inclined to do either:

1. Allow entryId to be null , e.g., `entryId: number | null`
2. Just pass the entry, e.g., `entry: Workflow | Tool` 

Subjective, but it's one less parameter to pass around.

src/app/categories/state/entry-categories.service.ts
svonworl, line 63: The `remove()` above will remove everything from the categories store, this line may not be necessary.

src/app/categories/state/entry-categories.service.ts
amarjandu: I opted to set the currentSubscription to null after the unsubscribe.

src/app/categories/state/entry-categories.service.ts
amarjandu: Confirmed that this suggestion works.

amarjandu: This section was removed per Steve's suggestion.

src/app/categories/state/entry-categories.service.ts
amarjandu: I'm going to keep this as it is, I tried implementing both suggestions and ran into some issues...

Allowing the entryId to be `null` is ok until we have to start adding control flow branches to all the call sites, which seems like a bit of a code smell. It also implicitly expects the developer to add this control flow before the call to the service method, which brings potential bugs in the future if someone does not do that.

I liked the idea of passing the entry around, but the primary issue is that `addEntryToCollection` doesn't have a reference to the type of entry, rather it interfaces with the entryID. I would have liked to use this approach because it abstracts the attribute usage of `is_published` down into the service method, but I don't think its doable at the moment.

svonworl (approved):

denis-yuen (approved):

src/app/categories/state/entry-categories.service.ts
```diff
@@ -42,14 +42,14 @@ export class EntryCategoriesService {
...
     this.entryCategoriesStore.setLoading(true);
     this.entryCategoriesStore.setError(false);
     this.entryCategoriesStore.remove();
     this.currentSubscription?.unsubscribe();
-    this.currentSubscription = this.entriesService
-      .entryCategories(entryId)
-      .subscribe(
+    this.currentSubscription = null;
+    if (published) {
```
coverbeck, line 51: This comment is for line 45/46 -- you should move that `...setLoading(true)` inside this if statement -- loading is never going to get set back to false if `published` is false.

src/app/categories/state/entry-categories.service.ts
```diff
@@ -42,14 +42,14 @@ export class EntryCategoriesService {
   /**
    * Updates the list of categories for the specified entry.
```
coverbeck, line 43: Suggest updating to describe what the published parameter does.

src/app/categories/state/entry-categories.service.ts
svonworl: Good catch.  We should probably also `setLoading(false)` and `setError(false)` around where we unsubscribe, before the `if`.

svonworl: Basically, one way to think about `updateEntryCategories` is that it operates in two stages.  The first stage "cancels" the existing request and resets the store state, and the next stage fires off a new request, if necessary, that will set the store state when it completes...

coverbeck (approved):

PR #1554, opened 2022-06-06
Change recent events to show user-specific events only

eileenzhh:
**Description**
Recent events in user pages as well as home-page dashboard should now show events that are related to either the user being viewed or the logged in user on the dashboard.

**Issue**
JIRA ticket: [SEAB-4314](https://ucsc-cgl.atlassian.net/browse/SEAB-4314)


src/app/home-page/recent-events/recent-events.component.ts
```diff
@@ -31,15 +39,38 @@ export class RecentEventsComponent implements OnInit {
...
+          this.recentEventsService.get(currentUser);
+        },
+        (error: HttpErrorResponse) => {
+          this.alertService.detailedError(error);
+        }
+      );
+    } else {
+      this.userQuery.user$.subscribe((user: User) => {
+        this.recentEventsService.get(user);
+      });
```
eileenzhh, line 66: The if-else statement checks if the username exists, I was also considering adding a boolean input to assess whether it is on the dashboard or a user-page instead

eileenzhh:
Right now there's an `undefined` error that occurs only if I hard refresh in the dashboard, if I click out of the dashboard and back into it without refreshing, the recent events show up as expected.

coverbeck:
> Right now there's an `undefined` error that occurs only if I hard refresh in the dashboard, if I click out of the dashboard and back into it without refreshing, the recent events show up as expected. I'm not sure how to fix this..

Where is the error? Do you have a stack trace?

src/app/home-page/recent-events/recent-events.component.ts
```diff
@@ -31,15 +39,38 @@ export class RecentEventsComponent implements OnInit {
...
+        (currentUser: User) => {
+          this.recentEventsService.get(currentUser);
+        },
+        (error: HttpErrorResponse) => {
+          this.alertService.detailedError(error);
+        }
+      );
+    } else {
+      this.userQuery.user$.subscribe((user: User) => {
+        this.recentEventsService.get(user);
```
coverbeck, line 65: You may need to check that `user` is defined, but I'm not sure. It could be related to the question you asked about seeing an `undefined` error (but just guessing).

eileenzhh:
> > Right now there's an `undefined` error that occurs only if I hard refresh in the dashboard, if I click out of the dashboard and back into it without refreshing, the recent events show up as expected. I'm not sure how to fix this..
> 
> Where is the error? Do you have a stack trace?

image
user is undefined  @coverbeck

src/app/home-page/recent-events/recent-events.component.ts
eileenzhh, line 65: this doesn't seem to get the user on the initial loading of the dashboard page

src/app/home-page/state/recent-events.service.ts
```diff
@@ -1,17 +1,24 @@
...
 import { finalize, tap } from 'rxjs/operators';
 import { RecentEventsStore } from './recent-events.store';
 
 @Injectable({ providedIn: 'root' })
 export class RecentEventsService {
-  constructor(private recentEventsStore: RecentEventsStore, private eventsService: EventsService) {}
+  constructor(
+    private recentEventsStore: RecentEventsStore,
+    private eventsService: EventsService,
+    protected userQuery: UserQuery,
```
coverbeck, line 14: It doesn't look like you're using `userQuery` nor `usersService`.

I tried removing them locally and got some weird compilation errors, but I think you should figure out how to remove them (unless I'm wrong about them being used).

coverbeck: > I'm using `userQuery` and `usersService` to find the user either from the logged in user or from the username respectively.

I may have my IDE code configured incorrectly -- what lines of code are they invoked from?

svonworl:
Here's some overall architecture recommendations:

PR https://github.com/dockstore/dockstore/pull/4901 added a new endpoint `getUserEvents`, which gets the events for a specified user, while maintaining the original `getEvents` endpoint, which gets the events for the logged-in user.   I think it's true that we could someday define `getEvents()` to return a different set of events than `getUserEvents(currentLoggedInUser())`, so we should probably use both endpoints to implement the appropriate functionality in this PR.

Maybe the "recent events" component can take an optional input `userName`.  If `userName` isn't specified, the component does what it did before, and if `userName` is specified, it looks up the id and retrieves the events via a new service method that uses `getUserEvents`?

eileenzhh:
> Here's some overall structural ideas:
> 
> PR [dockstore/dockstore#4901](https://github.com/dockstore/dockstore/pull/4901) added a new endpoint `getUserEvents`, which gets the events for a specified user, while maintaining the original `getEvents` endpoint, which gets the events for the logged-in user. Maybe it's true that we could someday define `getEvents()` to return a different set of events than `getUserEvents(currentLoggedInUser())`, so we should probably use both endpoints to implement the appropriate functionality in this PR. And `getEvents` eliminates the "current user" lookup on the client side...
> 
> Maybe the "recent events" component can take an optional input `userName`. If `userName` isn't specified, the component does what it did before, and if `userName` is specified, it looks up the id and retrieves the events via a new service method that uses `getUserEvents`?

Does that mean if username isn't specified (in the situation that the user is logged in and viewing their own dashboard), we want to show recent events from every user with starred events, exactly the same was as before?

src/app/home-page/state/recent-events.service.ts
eileenzhh: Oh wait nevermind, I was looking at the wrong file. I removed them locally on mine and didn't come across any compilation errors

svonworl:
> Does that mean if username isn't specified (in the situation that the user is logged in and viewing their own dashboard), we want to show recent events from every user that are starred, exactly the same was as before?

That was my impression, but I could be wrong. :)  After re-reading the ticket, if the goal is uncertain, it'd probably be best to clarify by talking with the appropriate personnel (the person who wrote the ticket, a lead, etc).

eileenzhh:
> > Does that mean if username isn't specified (in the situation that the user is logged in and viewing their own dashboard), we want to show recent events from every user that are starred, exactly the same was as before?
> 
> That was my impression, but I could be wrong. :) After re-reading the ticket, if the goal is still uncertain, it'd probably be best to clarify by talking with the appropriate personnel (the person who wrote the ticket, a lead, etc).



> > Does that mean if username isn't specified (in the situation that the user is logged in and viewing their own dashboard), we want to show recent events from every user that are starred, exactly the same was as before?
> 
> That was my impression, but I could be wrong. :) After re-reading the ticket, if the goal is still uncertain, it'd probably be best to clarify by talking with the appropriate personnel (the person who wrote the ticket, a lead, etc).

I think you're right, the dashboard should be using the original endpoint so I changed it back. I think the new `getUserEvents` endpoint does not work correctly since it is showing what the user is interested in (all activity related to their starred organizations and workflows), not the activity that they personally participated in. I created a bug ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-4476

eileenzhh:
> > Does that mean if username isn't specified (in the situation that the user is logged in and viewing their own dashboard), we want to show recent events from every user that are starred, exactly the same was as before?
> 
> That was my impression, but I could be wrong. :) After re-reading the ticket, if the goal is still uncertain, it'd probably be best to clarify by talking with the appropriate personnel (the person who wrote the ticket, a lead, etc).



> > Does that mean if username isn't specified (in the situation that the user is logged in and viewing their own dashboard), we want to show recent events from every user that are starred, exactly the same was as before?
> 
> That was my impression, but I could be wrong. :) After re-reading the ticket, if the goal is still uncertain, it'd probably be best to clarify by talking with the appropriate personnel (the person who wrote the ticket, a lead, etc).

I think you're right, the dashboard should be using the original endpoint so I changed it back. I think the new `getUserEvents` endpoint does not work correctly since it is showing what the user is interested in (all activity related to their starred organizations and workflows), not the activity that they personally participated in. I created a bug ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-4476

src/app/home-page/state/recent-events.service.ts
```diff
@@ -1,24 +1,36 @@
...
+          finalize(() => this.recentEventsStore.setLoading(false)),
+          tap((allStarredEvents) => {
+            this.recentEventsStore.set(allStarredEvents);
+          })
+        )
+        .subscribe();
+    } else {
+      this.eventsService
+        .getEvents('PROFILE')
+        .pipe(
```
coverbeck, line 26: Nit-pick, this code block duplicates lines 16-22.

coverbeck (commented):
When I run this locally, I get several 403 errors, when, for example, I'm not logged in to Dockstore. Is this already tracked in another ticket?

src/app/home-page/state/recent-events.service.ts
eileenzhh: This uses 'getEvents' rather than 'getUserEvents', the difference is getEvents gets it for the user that is logged in, so I don't have to pass on a user which was taken from the activatedRoute param

eileenzhh: However the 403 errors are concerning, I will run on my local and try to see why that's happening.

coverbeck: For the duplication, I was referring to lines 16-22, and lines 26-32 -- they're identical, I think.

Even though you're calling different endpoints, they both return the same response type, so you don't need to repeat that block.

Something like, untested:

```
const events = user ? this.eventsService.getUser(user.id, 'PROFILE') : this.eventsService.getEvents('PROFILE');
events.pipe(
   finalize(...
...
```

kathy-t:
Re: Charles' errors that he encountered, I was also able to reproduce them locally

When I'm logged in, I also get 403 errors when viewing the user page that's not my own:
image
Nothing shows up under the Activity section and the bar continually loads.

Same thing happens when I'm logged out, except it's a 401 error instead:
image

The errors are from the `users` endpoint: `http://localhost:4200/api/users/username/aofarrel` and `http://localhost:4200/api/users/username/aofarrel?include=userProfiles`. I think Richard was previously working on fixing the user profile access: https://ucsc-cgl.atlassian.net/browse/SEAB-4240, but it hasn't been completed yet

eileenzhh:
> 

Is it possible that when you're logged in, your user isn't set to administrator? I get those 403 errors often when I rebuild and my admin status is lost. As for when you are logged in, I'm not quite sure if users are supposed to see other user-pages, 401 errors are what is happening for develop, so it's not specific to this branch :/

src/app/home-page/state/recent-events.service.ts
```diff
@@ -1,24 +1,44 @@
...
-    this.eventsService
-      .getEvents('ALL_STARRED')
-      .pipe(
-        finalize(() => this.recentEventsStore.setLoading(false)),
-        tap((allStarredEvents) => {
-          this.recentEventsStore.set(allStarredEvents);
-        })
-      )
-      .subscribe();
+    // const events = user ? this.eventsService.getUserEvents(user.id, 'PROFILE') : this.eventsService.getEvents('PROFILE');
```
denis-yuen, line 13: probably doesn't make sense to check in commented code 
(unless there is a very specific situation in which is could be uncommented, in which case there should be a `//TODO comment` or similar)

denis-yuen (approved):

src/app/home-page/recent-events/recent-events.component.ts
```diff
@@ -31,15 +39,36 @@ export class RecentEventsComponent implements OnInit {
...
     Event.TypeEnum.UNPUBLISHENTRY,
   ];
-  constructor(private recentEventsQuery: RecentEventsQuery, private recentEventsService: RecentEventsService) {}
+  constructor(
+    private recentEventsQuery: RecentEventsQuery,
+    private recentEventsService: RecentEventsService,
+    private activatedRoute: ActivatedRoute,
+    private usersService: UsersService,
+    private alertService: AlertService,
+    protected userQuery: UserQuery
```
coverbeck, line 48: I don't think this is used; please remove if not.

coverbeck (approved):

kathy-t (approved):

PR #1556, opened 2022-06-14
Get dockstore script from cli repo

wshands:
**Description**
Gets the latest release of the Dockstore CLI from the CLI repo (dockstore/dockstore-cli) and uses that as the location of the dockstore script rather than the Dockstore webservice repo (dockstore/dockstore), when setting up the 'Download CLI' button and printing instructions in the 'Quick Start'.

The supporting webservice PR is https://github.com/dockstore/dockstore/pull/4948

**Issue**
Addresses SEAB-3952


src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -38,20 +39,30 @@ export class DownloadCLIClientComponent implements OnInit {
...
-          .subscribe(
-            (json: any) => {
-              if (json) {
-                this.cwltoolVersion = json.cwltool;
-              }
-            },
-            (err) => {
-              console.log('Unable to retrieve requirements.txt file.');
+        this.metadataService.getCliVersion().subscribe(
+          (json: any) => {
```
coverbeck, line 43: Avoid `any`; it's a string not JSON (but probably should be JSON per Steve's comment in your web service PR).

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -38,20 +39,30 @@ export class DownloadCLIClientComponent implements OnInit {
...
+            if (json) {
+              this.dockstoreCliVersion = json;
             }
-          );
+            this.downloadCli = `https://github.com/dockstore/dockstore-cli/releases/download/${this.dockstoreCliVersion}/dockstore`;
```
coverbeck, line 47: Looking at how you're constructing this, why not instead have the web service endpoint return the complete url (or urls if you also return the unstable)?

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
wshands: Fixed now uses CLIInfo class

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
wshands: now returns complete URL

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -6,6 +6,7 @@ import { Dockstore } from '../../../shared/dockstore.model';
 import { MetadataService } from '../../../shared/swagger';
 import { GA4GHService } from './../../../shared/swagger/api/gA4GH.service';
 import { Metadata } from './../../../shared/swagger/model/metadata';
+import { CLIInfo } from './../../../shared/openapi/model/cLIInfo';
```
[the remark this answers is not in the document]
wshands: CLIInfo will be there after the webservice PR is merged.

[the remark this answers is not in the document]
svonworl: I wonder if there's a way to override the generated name, `cLIInfo` makes me feel funny, like I dRANK too much coffee or sOMEThing...

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -38,20 +39,29 @@ export class DownloadCLIClientComponent implements OnInit {
...
+              .getRunnerDependencies(apiVersion, '3', 'cwltool', 'json')
+              .pipe(finalize(() => this.generateMarkdown()))
+              .subscribe(
+                (json: any) => {
+                  if (json) {
+                    this.cwltoolVersion = json.cwltool;
+                  }
+                },
+                (err) => {
+                  console.log('Unable to retrieve requirements.txt file.');
```
svonworl, line 57: Depending upon what we want to happen when we can't retrieve the necessary information, `ngOnInit` might need a few tweaks.  Do we always generate the markdown, for every error?  Use some rationale defaults in lieu of actual values?  Conditionally render the markdown to omit the parts we don't know?  Or, maybe failure is rare enough in this case that the particular behavior doesn't really matter, as long as it's reasonable?

coverbeck: Good points, and your comment made me notice, we shouldn't be doing `console.log`, we should be surfacing the error.

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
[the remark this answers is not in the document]
wshands: Could rename it, but would require a webservice PR also...

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -34,26 +41,41 @@ export class DownloadCLIClientComponent implements OnInit {
...
-            (err) => {
-              console.log('Unable to retrieve requirements.txt file.');
+        this.metadataService.getCliVersion().subscribe(
+          (clijson: CLIInfo) => {
+            if (clijson) {
+              this.alertService.simpleSuccess();
+              this.downloadCli = clijson.cliLatestDockstoreScriptDownloadUrl;
             }
-          );
+            this.metadataService
```
coverbeck, line 55: Instead of nesting this call, you could use `forkJoin()` --  you don't need any data from the CLIInfo response to make the call to get runner dependencies.

coverbeck (commented):
Torn on the `forkJoin` change -- it probably won't make a huge difference performance-wise, but it is best practice to not do http requests sequentially when you can do them in parallel.

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
wshands: Now using forkJoin

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -34,27 +40,30 @@ export class DownloadCLIClientComponent implements OnInit {
...
-              if (json) {
-                this.cwltoolVersion = json.cwltool;
-              }
-            },
-            (err) => {
-              console.log('Unable to retrieve requirements.txt file.');
-            }
-          );
+
+        const cliInfo$ = this.metadataService.getCliVersion();
```
coverbeck, line 49: Nitpick, we tend to use the `$` only in variable names for Akita.

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -34,27 +40,30 @@ export class DownloadCLIClientComponent implements OnInit {
...
+        const dependencies$ = this.metadataService.getRunnerDependencies(apiVersion, '3', 'cwltool', 'json');
+        // forkJoin returns an array of values, here we map those values to an object
+        forkJoin([cliInfo$, dependencies$]).subscribe(
+          (result) => {
```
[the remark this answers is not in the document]
wshands: done

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
wshands: removed $

coverbeck (approved):

package.json
```diff
@@ -3,10 +3,10 @@
   "version": "2.8.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "1.13.0-alpha.6",
+    "webservice_version": "1.13.0-alpha.7",
     "use_circle": true,
-    "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/7871/workflows/d603dc05-1719-4205-bac8-06f628aad895/jobs/20475/artifacts",
-    "circle_build_id": "20475",
+    "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/7985/workflows/1b5411e1-bd09-40ae-9fb7-845f144fa864/jobs/21164/artifacts",
```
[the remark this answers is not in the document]
wshands: I was unable to push a webservice tag to develop, so had to create a tag from my own branch and then an artifact from that tag. @denis-yuen @coverbeck is this OK or does someone need to create a webservice tag from develop and an artifact from that tag to use for this PR?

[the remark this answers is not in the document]
denis-yuen: I don't think hubflow has much to say on the topic of pre-releases. Our wiki releases page has a little more and suggests a release branch that you don't end up finishing, that's not terribly different from a feature branch.

Did you end up merging `feature/walts-build` into develop and does it have anything important?
If so, you can use the merged build into develop here. 
If not and it won't break when the next developer updates to a newer develop build, then I don't think it matters.

denis-yuen (commented):
comment above

package.json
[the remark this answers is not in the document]
wshands: I merged PR https://github.com/dockstore/dockstore/pull/4948 (which implements the new CLI info API) to webservice develop
I created branch feature/walts-build from webservice develop
I created tag 1.13.0-alpha.7 on feature/walts-build
I created artifact 1.13.0-alpha.7 from that tag
  (so the artifact is really constructed from the equivalent develop branch)
I used that artifact in the current UI PR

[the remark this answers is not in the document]
kathy-t: `use_circle` is set to `true` so it's not using the `1.13.0-alpha.7` artifact right now. I think you have to set it to `false`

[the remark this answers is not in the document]
wshands: Now using circle_ci_source set to artifacts from a CircleCI build on the develop branch

kathy-t (approved):

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
```diff
@@ -34,27 +41,28 @@ export class DownloadCLIClientComponent implements OnInit {
...
+        forkJoin([this.metadataService.getCliVersion(), this.metadataService.getRunnerDependencies(apiVersion, '3', 'cwltool', 'json')])
           .pipe(finalize(() => this.generateMarkdown()))
           .subscribe(
-            (json: any) => {
-              if (json) {
-                this.cwltoolVersion = json.cwltool;
-              }
+            ([cliInfo, dependencies]) => {
+              this.downloadCli = cliInfo.cliLatestDockstoreScriptDownloadUrl;
+              this.cwltoolVersion = JSON.parse(JSON.stringify(dependencies)).cwltool;
```
svonworl, line 56: We call `alertService.start` near the beginning of `ngOnInit`, so  at the end of every path through the subsequent code, we need to call one of alertService methods to signal that the operation is complete.  You've got the calls in both of the error handlers, but you probably need to add a `simpleSuccess` or `detailedSuccess` call to the handler that processes the successfully-retrieved information.
More details here:
https://github.com/dockstore/dockstore-ui2/blob/66af5d00069ed864ffbc19aacdcac2b5337643e2/src/app/shared/alert/state/alert.service.ts#L25

denis-yuen (approved):

src/app/loginComponents/onboarding/downloadcliclient/downloadcliclient.component.ts
wshands: added simple success msg

PR #1557, opened 2022-06-15
Update README for devs

aofarrel:
**Description**
This change is needed because I installed the wrong version of Angular seven times.

**Issue**
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,14 +38,14 @@ wget -qO- https://raw.githubusercontent.com/creationix/nvm/v0.34.0/install.sh | 
 ```
 Close current terminal and open a new one or `source ~/.bashrc`
 ```
-nvm install 10.13.0
+nvm install 16.14.0
 ```
 Optionally, install a global Angular CLI in order to execute `ng` commands without prepending `npx`.   
-Otherwise, prepend `npx` to every command in this README if a global @angular/cli was not installed.
+Otherwise, prepend `npx` to every `ng` command in this README if a global @angular/cli was not installed.
 Before installing, follow https://docs.npmjs.com/getting-started/fixing-npm-permissions#option-two-change-npms-default-directory to fix permissions if needed.
 
 ```
-$npm i -g @angular/cli@1.3.1		
+$npm i -g @angular/cli@13.2.2    
 ```
 
 #### NPM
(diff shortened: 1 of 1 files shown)


aofarrel:
As you may have noticed from the `OS: darwin x64` line, I'm on a Mac, so the Angular output might need to be linuxafied.

denis-yuen (approved):
thanks!

kathy-t (approved):

README.md
```diff
@@ -38,14 +38,14 @@ wget -qO- https://raw.githubusercontent.com/creationix/nvm/v0.34.0/install.sh |
...
+nvm install 16.14.0
 ```
 Optionally, install a global Angular CLI in order to execute `ng` commands without prepending `npx`.   
-Otherwise, prepend `npx` to every command in this README if a global @angular/cli was not installed.
+Otherwise, prepend `npx` to every `ng` command in this README if a global @angular/cli was not installed.
 Before installing, follow https://docs.npmjs.com/getting-started/fixing-npm-permissions#option-two-change-npms-default-directory to fix permissions if needed.
 
 ```
-$npm i -g @angular/cli@1.3.1		
+$npm i -g @angular/cli@13.2.2
```
coverbeck, line 48: We should just remove this entirely, and rely on the Angular installed via package.json.

README.md
```diff
@@ -71,32 +71,43 @@ ghusky > pre-commit (node v10.13.0)
...
-@angular/flex-layout: 2.0.0-beta.9
-@angular/forms: 4.3.6
-@angular/http: 4.3.6
-@angular/platform-browser: 4.3.6
-@angular/platform-browser-dynamic: 4.3.6
-@angular/router: 4.3.6
-@angular/cli: 1.3.1
-@angular/compiler-cli: 4.3.6
-@angular/language-service: 4.3.6
+
```
coverbeck, line 80: I would remove this whole section, starting with line 77

coverbeck (changes requested):
I appreciate this PR -- thanks for taking the time! Sorry for all the feedback on almost every line.

But the whole global Angular thing was before we understood how to do things via package.json and a local Angular. We updated the code but never updated the README.md. If we keep references to the global Angular in here, it will quickly get out of date again, as Angular releases new versions once or twice a year. We should just remove all refs to global Angular.

aofarrel:
@coverbeck 

> But the whole global Angular thing was before we understood how to do things via package.json and a local Angular. 

I'm getting the feeling I am not part of this "we" because what I put in this readme is what I actually did 🙃

It might be best to have someone who actually understands what is going on update this readme...

README.md
```diff
@@ -71,32 +71,43 @@ ghusky > pre-commit (node v10.13.0)
 [feature/2130/prettier b6da3e7c] Test
 ```
 
+To ensure the correct dependencies are installed, run `npm install check-dependencies --save-dev` for good measure.
```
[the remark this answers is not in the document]
aofarrel: `npm ci` does work on my machine, but iirc, I had to also check the dependencies in addition to doing that.

coverbeck:
> @coverbeck
> 
> I'm getting the feeling I am not part of this "we" because what I put in this readme is what I actually did 🙃

I was referring the to the dev team at the time who worked on updating the package.json. No insult intended.

When you run `npm run start`, you are doing `"npx ng serve --proxy-config=proxy.conf.json` which doesn't use the global Angular. So although you installed the global Angular and things then worked, having the global Angular installed or not should be irrelevant.

aofarrel:
@coverbeck has since updated the readme, which I followed for a fresh installation with ease. It's much better than what I have in this PR.

PR #1562, opened 2022-06-27
Dashboard redesign - Layout process and Entries widgets

eileenzhh:
**Description**
This PR only covers a general layout of the new dashboard redesign with only functional behaviour of the workflows/tools/services widgets based on the design in [Zeplin](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6022ab2d42df9f2c57c20982).

It is currently accessible through /newDashboard and uses the [featureflag](https://github.com/dockstore/dockstore-ui2/pull/1572) PR by Charles. 

**Issue**
[SEAB-4511](https://ucsc-cgl.atlassian.net/browse/SEAB-4511)


eileenzhh:
What it looks like after changes: image

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -13,10 +13,148 @@
   ~    See the License for the specific language governing permissions and
   ~    limitations under the License.
   -->
-
 <div fxLayout>
   <app-my-sidebar fxFlexAlign="stretch"></app-my-sidebar>
 
+  <div id="row-holder" fxLayout="column" fxLayoutAlign="start start" class="p-3 w-100">
+    <mat-card class="alert alert-info mx-1 w-100" role="alert"> Major Release: </mat-card>
```
coverbeck, line 20: I see it in the mock, but I'm not sure what the intent is there. Is it to display the latest news item, or is it to figure out the latest major release (which we don't have an API for?). Depending, you may not want to hard-code `Major Release:` here.

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -13,10 +13,148 @@
...
+    <div id="entry-holder" fxLayout="row" fxLayoutGap="10px" class="w-100 mb-4">
```
coverbeck, line 21: Prefer rem to px, I think, if that part of the guide is still correct.

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -13,10 +13,148 @@
...
+        <mat-card-content class="p-3">
+          <mat-form-field appearance="outline" class="search-form-field w-100">
+            <mat-label>Search your Workflows...</mat-label>
+            <input matInput type="text" data-cy="basic-search" />
+          </mat-form-field>
+          <span class="weight-bold">Recent Activity</span>
+        </mat-card-content>
+      </mat-card>
+
+      <mat-card class="mat-card-grow">
```
coverbeck, line 51: A lot of this duplicates the previous block and gets duplicated again below. You may want to consider creating an Angular component to avoid duplication. Plus it would make it easier if we ever add a new entity type.

src/app/home-page/home-logged-in/home-logged-in.component.html
denis-yuen: I think this was a placeholder for what because our current "news and events"
So I think "news and events" can replace this

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -13,10 +13,148 @@
...
 
+  <div id="row-holder" fxLayout="column" fxLayoutAlign="start start" class="p-3 w-100">
+    <mat-card class="alert alert-info mx-1 w-100" role="alert"> Major Release: </mat-card>
+    <div id="entry-holder" fxLayout="row" fxLayoutGap="10px" class="w-100 mb-4">
+      <mat-card class="mat-card-grow">
+        <div id="header-bar" fxLayout="row" fxLayoutGap="5px" class="mat-card-header-separate">
+          <div id="workflow-number" fxLayout="row">
+            <mat-card-header fxLayout="row">
+              <mat-card-title>Workflows</mat-card-title>
+              <span class="bubble">0</span>
```
kathy-t, line 27: Will this use an API to get a user's number of workflows instead of being hard-coded as `0`?

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -13,10 +13,148 @@
...
+            </mat-card-header>
+          </div>
+          <button
+            mat-button
+            class="private-btn small-btn-structure ml-3 mb-3"
+            id="registerWorkflowButton"
+            matTooltip="Register Workflow"
+            matTooltipPosition="after"
+          >
+            <img src="../../../assets/svg/icons-plus.svg" alt="Register a tool" class="pr-2 pb-1" />
```
kathy-t, line 37: Typo for `alt` attribute: `Register a tool` -> `Register a workflow`

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -13,10 +13,148 @@
...
+            Register a Workflow
+          </button>
+        </div>
+        <mat-divider></mat-divider>
+        <mat-card-content class="p-3">
+          <mat-form-field appearance="outline" class="search-form-field w-100">
+            <mat-label>Search your Workflows...</mat-label>
```
kathy-t, line 44: In the case of `0` workflows (or tools/services), the [zeplin mockup](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6022ab2d42df9f2c57c20982) doesn't show the search bar and recent activity in the component. It also shows a "You have not registered any ___" message. (Not sure if this will be added later)

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -13,10 +13,148 @@
...
+            <mat-card-header fxLayout="row">
+              <mat-card-title>Organizations</mat-card-title>
+              <span class="bubble">3</span>
+            </mat-card-header>
+          </div>
+          <button
+            mat-button
+            class="private-btn small-btn-structure ml-3 mb-3"
+            id="registerServiceButton"
+            matTooltip="Register Service"
```
kathy-t, line 135: Typo for `id` and `matToolTip`, should be for an organization

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -13,10 +13,148 @@
...
+            matTooltipPosition="after"
+          >
+            <img src="../../../assets/svg/icons-plus.svg" alt="Register a Service" class="pr-2 pb-1" />
```
kathy-t, line 138: Typo in `alt` attribute, should be for an organization

src/app/home-page/home-logged-in/home-logged-in.component.html
eileenzhh: You're right, this will be added later! I just didn't have a condition to work with yet :)

src/app/home-page/home-logged-in/home-logged-in.component.html
eileenzhh: Yes it will! This is just a placeholder for the visual effect

denis-yuen (commented):
commenting till build passes

coverbeck:
I'm running it locally, but I don't see any of my tools nor services appearing. Is that just not done yet? If not, could you please clarify what's supposed to be working and what's not, so we don't review things not ready for review.

src/app/home-page/widget/entry-boxes/entry-boxes.component.html
```diff
@@ -0,0 +1,63 @@
...
+      </mat-form-field>
+      <span class="weight-bold">Recent Activity</span>
+      <div *ngFor="let entry of listOfEntries" fxLayout="row" class="list-items w-100">
+        <a class="no-wrap" [routerLink]="allEntriesLink + entry.path">
+          {{ entry.prettyPath }}
+        </a>
+        <div class="date">{{ entry.lastUpdateDate | date: 'MMM d, yyyy' }}</div>
+      </div>
+      <a class="weight-bold" [routerLink]="allEntriesLink">
+        All {{ entryType }}<mat-icon class="mt-1">keyboard_double_arrow_right</mat-icon>
```
coverbeck, line 36: This shows up as `All workflow` for me (missing the "s").

src/app/home-page/home-page.component.html
```diff
@@ -4,6 +4,11 @@
 <ngx-json-ld [json]="websiteSchema"></ngx-json-ld>
 
 <ng-template #loggedInHomePage>
+  <div *ngIf="Dockstore.FEATURES.enableNewDashboard; else oldDashboard">
```
coverbeck, line 7: I was thinking you would have this in home-logged-in.component.html, and then that component would display either the old or new dashboard. Or you introduce an intermediate component.

src/app/app.routing.ts
```diff
@@ -35,6 +36,12 @@ export const CLIENT_ROUTER_PROVIDERS = [AuthGuard];
 
 const APP_ROUTES: Routes = [
   { path: '', component: HomeComponent, pathMatch: 'full', data: { title: 'Dockstore' } },
+  {
```
coverbeck, line 39: If you do the other change I suggested, you should be able to get rid of this entirely. Then you toggle the new dashboard with `http://localhost:4200?newDashboard` question mark, not slash.

denis-yuen:
> What it looks like after changes: image

Note, after the widgets are hooked up, you'll want to note that the "empty" widgets look different from this with the icons and "You have not registered any X" wording.

denis-yuen (commented):
rough layout looks good from the screenshot, will need to deal with build issues though

src/app/home-page/widget/entry-boxes/entry-boxes.component.html
eileenzhh: oh yes!

eileenzhh:
> > What it looks like after changes: image
> 
> Note, after the widgets are hooked up, you'll want to note that the "empty" widgets look different from this with the icons and "You have not registered any X" wording.

Hmm I believe I should have the "you have not registered any X" already implemented :thinking:

coverbeck:
Looking at it just from running it realizing now that the data is fake or non-existent:

* If you type  for example `p` in the workflows search box, the number of workflows increases.
* When there are no entries, the text for a box says, for example, "Learn about Tool" and "Learn about Service" -- should be "tools" and "services". Also, I don't think you should capitalize the first letter (tools not Tools).
* It's odd there's an Organizations panel in the dashboard, but not an Organizations item in the left-panel (you probably just haven't gotten to it).

Overall, I think it looks really good, but I'm not sure how much deeper to go into it.

src/app/home-page/home-page.component.html
```diff
@@ -6,4 +6,5 @@
 <ng-template #loggedInHomePage>
   <app-home-logged-in></app-home-logged-in>
 </ng-template>
+<ng-template #oldDashboard> </ng-template>
```
coverbeck, line 9: I don't think you need this any more.

src/app/home-page/widget/entry-boxes/entry-boxes.component.ts
```diff
@@ -0,0 +1,96 @@
...
+import { MyEntriesQuery } from 'app/shared/state/my-entries.query';
+import { SessionQuery } from 'app/shared/session/session.query';
+import { SessionService } from 'app/shared/session/session.service';
+import { MyWorkflowsService } from 'app/myworkflows/myworkflows.service';
+import { RegisterToolService } from 'app/container/register-tool/register-tool.service';
+import { Base } from 'app/shared/base';
+import { Dockstore } from 'app/shared/dockstore.model';
+
+@Component({
+  selector: 'app-entry-boxes',
```
coverbeck, line 17: Technically, this should be `app-entry-box`, shouldn't it? You have one instance of these per entry type?

You'd have to rename the files too, technically; not sure that it's worth it.

src/app/home-page/new-dashboard/new-dashboard.component.html
```diff
@@ -0,0 +1,55 @@
...
+            </mat-card-header>
+          </div>
+        </div>
+        <mat-divider></mat-divider>
+        <mat-card-content class="p-3">
+          <span class="weight-bold">Recent Activity</span>
+        </mat-card-content>
+      </mat-card>
+      <mat-card class="fx-grow">
+        <div id="header-bar" fxLayout="row" fxLayoutGap="5px" class="mat-card-header-separate">
```
coverbeck, line 26: Duplicate id; should be avoided, although not fatal: https://softwareengineering.stackexchange.com/questions/127178/two-html-elements-with-same-id-attribute-how-bad-is-it-really

src/app/home-page/new-dashboard/new-dashboard.component.html
```diff
@@ -0,0 +1,55 @@
...
+          <div id="workflow-number" fxLayout="row">
```
coverbeck, line 27: Duplicate id

src/app/home-page/new-dashboard/new-dashboard.component.html
```diff
@@ -0,0 +1,55 @@
...
+            <mat-card-header fxLayout="row">
+              <mat-card-title>Organizations</mat-card-title>
+              <span class="bubble">3</span>
+            </mat-card-header>
+          </div>
+          <button
+            mat-button
+            class="private-btn small-btn-structure ml-3 mb-3"
+            id="registerServiceButton"
+            matTooltip="Register Service"
```
coverbeck, line 37: Organization

src/app/home-page/new-dashboard/new-dashboard.component.html
```diff
@@ -0,0 +1,55 @@
...
+            matTooltipPosition="after"
+          >
+            <img src="../../../assets/svg/icons-plus.svg" alt="Register a Service" class="pr-2 pb-1" />
```
coverbeck, line 40: Organization

src/app/home-page/new-dashboard/new-dashboard.component.ts
```diff
@@ -0,0 +1,86 @@
...
+})
+export class NewDashboardComponent extends Base implements OnInit {
+  public user$: Observable<User>;
+  user: User;
+  EntryType: EntryType;
+  public listOfTools: Array<EntryUpdateTime> = [];
+  public listOfWorkflows: Array<EntryUpdateTime> = [];
+  public listOfServices: Array<EntryUpdateTime> = [];
+  workflows: Array<Workflow>;
+  public workflowsObject$: Observable<Array<OrgWorkflowObject<Workflow>>>;
```
coverbeck, line 33: Unused, I think

src/app/home-page/new-dashboard/new-dashboard.component.ts
```diff
@@ -0,0 +1,86 @@
...
+  public toolsObject$: Observable<Array<OrgToolObject<DockstoreTool>>>;
```
coverbeck, line 34: Unused, I think

src/app/home-page/new-dashboard/new-dashboard.component.ts
```diff
@@ -0,0 +1,86 @@
...
+          .subscribe(() => {
+            this.alertService.clearEverything();
+          });
+      } else {
+        this.dialog.closeAll();
+      }
+    });
+  }
+
+  protected getMyEntries() {
```
coverbeck, line 70: You may want to push this down into the entry-boxes. Actually, it looks like you already started down that route.

src/app/home-page/widget/entry-boxes/entry-boxes.component.ts
```diff
@@ -0,0 +1,96 @@
...
+      this.allEntriesLink = '/my-tools/';
+    } else if (this.entryType === 'Service') {
+      this.helpLink = Dockstore.DOCUMENTATION_URL + '/getting-started/getting-started-with-services.html';
+      this.allEntriesLink = '/my-services/';
+    }
+  }
+
+  getMyEntries() {
+    this.usersService
+      .getUserEntries(10, this.filterText)
```
coverbeck, line 65: It would be nice if we could just fetch entries of a given type, but that will require an API change.

src/app/home-page/widget/entry-boxes/entry-boxes.component.ts
```diff
@@ -0,0 +1,96 @@
...
+import { Dockstore } from 'app/shared/dockstore.model';
+
+@Component({
+  selector: 'app-entry-boxes',
+  templateUrl: './entry-boxes.component.html',
+  styleUrls: ['./entry-boxes.component.scss'],
+})
+export class EntryBoxesComponent extends Base implements OnInit {
+  Dockstore = Dockstore;
+  @Input() entryType: string;
```
coverbeck, line 23: `@Input() entryType: EntryUpdateTime.EntryTypeEnum` would be more better.

coverbeck (commented):
Started going through in detail, not sure if you want all of this, so stopping.

Overall, I think the layout looks really good! And we should try to iron this out, although time might be a little tight.

src/app/home-page/new-dashboard/new-dashboard.component.html
eileenzhh: this will be changed and renamed when the starred/organization widgets are going to be made in the new ticket https://ucsc-cgl.atlassian.net/browse/SEAB-4623

src/app/home-page/widget/entry-boxes/entry-boxes.component.ts
eileenzhh: Yeah that was a consideration but would need some backend assistance, it would be a good idea to consider for future changes/improvements

src/app/home-page/widget/entry-boxes/entry-boxes.component.ts
eileenzhh: I'm currently just manually passing a string from the 'new-dashboard.component.html' into the 'entry-box' so it knows which one it is rendering

src/app/home-page/new-dashboard/new-dashboard.component.ts
eileenzhh: Yup I just forgot to delete this function in this file, thanks for catching that!

src/app/home-page/widget/entry-boxes/entry-boxes.component.ts
coverbeck: Ugh, I can't believe I wrote `more better`. :( At any rate, I think you can still declare the type as `EntryUpdateTime.EntryTypeEnum`, and you can pass a string from the HTML, and the compiler would be able to verify that the string is a legit value (although we may have template validation turned off).

eileenzhh: haha oh nice I'll do that :+1:

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  ) {
+    super();
+    this.user = this.userQuery.getValue().user;
+  }
+
+  ngOnInit(): void {
+    this.getMyEntries();
+    if (this.entryType) {
+      this.entryTypeLowerCase = this.entryType.toLowerCase();
+      this.entryTypeCapitalize = this.entryTypeLowerCase[0].toUpperCase() + this.entryTypeLowerCase.substring(1);
```
eileenzhh, line 52: By changing the `@Input() entryType` from the string, I did have to do this clunky changing back into a capitalized format, is there a more efficient way? I tried to use `text-transform` but that made the UI look wonky unless more styles are added which may be a bit more work than this option. The capitalized version does make specific elements look nicer so ideally I didn't want to use the lowercase version.

eileenzhh: Note that it is really only used in 3 places (Header title, Register a ___ button, and All ___ link at the bottom)

coverbeck: There is an Angular [title case pipe](https://angular.io/api/common/TitleCasePipe) you can use, either in the .ts file or in the HTML.

Sorry, I know it's sometimes a bit of a hassle to have to use a type instead of just free text, but that's how one is generally supposed to use typed languages like TypeScript and Java, to let the compiler catch any errors.

src/app/myworkflows/myworkflows.service.ts
```diff
@@ -98,7 +98,7 @@ export class MyWorkflowsService extends MyEntriesService<Workflow, OrgWorkflowOb
    * @param {number} id  The user ID
    * @memberof MyBioWorkflowsService
    */
-  getMyBioWorkflows(id: number): void {
+  getMyBioWorkflows(id: number, showEntry: boolean): void {
```
kathy-t, line 101: Does this function (and `getMyEntries`) need to change? From what I can tell, it's not used in your other files

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
+
+  getMyEntries() {
+    this.usersService
+      .getUserEntries(10, this.filterText)
+      .pipe(debounceTime(500), takeUntil(this.ngUnsubscribe))
+      .subscribe((myEntries: Array<EntryUpdateTime>) => {
+        myEntries.forEach((entry: EntryUpdateTime) => {
+          if (this.entryType === 'WORKFLOW' && entry.entryType === 'WORKFLOW') {
```
kathy-t, line 73: `entry.entryType` is an enum, so should use enum to compare instead of a string (ex: `entry.entryType === EntryType.Workflow`)

kathy-t (commented):
Looking good! I have some comments based on what I experienced when running it locally:

This box says I have 6 workflows, but I have more than that, so this count is inaccurate. Also there's alot of white space near the bottom of the box, which could display more workflows.
image

If I search for a nonsense word and there's no entry that matches it, the loading bar never stops. Also the "All tools" should be on the next line.
image

The highlighted search box should have "Search your Workflows/Tools/Service" like you did for organizations:
image

The services entry box is missing the `PREVIEW` label.

src/app/home-page/widget/entry-box/entry-box.component.ts
eileenzhh: Oh yeah duh :facepalm: , I'll fix it!

src/app/myworkflows/myworkflows.service.ts
eileenzhh: Oops this was a change from a while back from when I was still using this function to get the list, good catch!

src/app/home-page/new-dashboard/new-dashboard.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    super();
+  }
+
+  ngOnInit() {
+    this.userQuery.user$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((user) => {
+      this.user = user;
+    });
+    this.registerToolService.isModalShown.pipe(takeUntil(this.ngUnsubscribe)).subscribe((isModalShown: boolean) => {
+      if (isModalShown) {
+        const dialogRef = this.dialog.open(RegisterToolComponent, { width: '600px' });
```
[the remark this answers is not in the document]
eileenzhh: this was taken directly from `my-tool.component.ts` line 117, so should I change it there too?

src/app/home-page/widget/entry-box/entry-box.component.ts
eileenzhh: Ohh that is something I didn't know but is useful to know, it also makes this a LOT cleaner and simpler. Having it as a type instead of a string makes sense to me especially to prevent errors, thank you! :clap:

cypress/integration/group1/dashboard.ts
```diff
@@ -0,0 +1,60 @@
...
+import { goToTab, isActiveTab, resetDB, setTokenUserViewPort } from '../../support/commands';
+
+describe('Dockstore dashboard', () => {
+  resetDB();
+  setTokenUserViewPort();
+
+  const cwlDescriptorType = 'CWL';
+  const wdlDescriptorType = 'WDL';
+  const nextflowDescriptorType = 'Nextflow';
+  //   it('can open register modals from the homepage', () => {
```
denis-yuen, line 27: Probably doesn't make sense to check-in commented code without a comment explaining under what circumstance it would be uncommented

denis-yuen (approved):

cypress/integration/group1/dashboard.ts
eileenzhh: Oh yeah I commented it out to see if that was the test that was flaking, I'll just remove it since it fails >50% of the time.

denis-yuen (approved):
mis-click

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,121 @@
...
+      this.helpLink = Dockstore.DOCUMENTATION_URL + '/getting-started/dockstore-tools.html';
+      this.allEntriesLink = '/my-tools/';
+    } else if (this.entryType === EntryUpdateTime.EntryTypeEnum.SERVICE) {
+      this.helpLink = Dockstore.DOCUMENTATION_URL + '/getting-started/getting-started-with-services.html';
+      this.allEntriesLink = '/my-services/';
+    }
+  }
+
+  getMyEntries() {
+    this.usersService
```
coverbeck, line 70: This is partially my fault for asking you to componentize, but because there are 3 components, we have the exact same 3 calls being made to start off. We should probably add a `type` query parameter to the endpoint so you don't have to do all the work of filtering client-side. Can be a followup.

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,121 @@
...
+      .getUserEntries(null, this.filterText)
+      .pipe(debounceTime(750), takeUntil(this.ngUnsubscribe))
+      .subscribe((myEntries: Array<EntryUpdateTime>) => {
+        this.listOfEntries = [];
+        myEntries.forEach((entry: EntryUpdateTime) => {
```
coverbeck, line 75: There's a lot of repeated code inside the block; you could use a filter to simplify it, or put the repeated code in a method.

```
myEntries.filter(entry => <condition>).forEach( entry => {
  this.totalEntries...
}
```

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,121 @@
...
+          if (this.entryType === entry.entryType && entry.entryType === EntryUpdateTime.EntryTypeEnum.WORKFLOW) {
+            if (this.firstLoad) {
```
coverbeck, line 77: I could see this both ways -- should the badge show the total number of entries, or the total number of entries you have filtered? I'd lean towards the latter and getting rid of this test.

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  getMyEntries() {
+    this.usersService
+      .getUserEntries(null, this.filterText)
+      .pipe(debounceTime(750), takeUntil(this.ngUnsubscribe))
+      .subscribe((myEntries: Array<EntryUpdateTime>) => {
+        this.listOfEntries = [];
+        myEntries.forEach((entry: EntryUpdateTime) => {
+          if (this.entryType === entry.entryType && entry.entryType === EntryUpdateTime.EntryTypeEnum.WORKFLOW) {
```
coverbeck, line 76: Do you need the the second part of this if condition? Can't you just do `if (this.entryType === entry.entryType)`? It would simplify the code a lot, and you wouldn't even need the filter I suggest above (although I have a personal preference for using the filter).

Looking at it again, I guess tools complicate the issue, but I think you could have a special case for tools.

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,121 @@
...
+            if (this.firstLoad) {
+              this.totalEntries += 1;
+            }
+            if (this.listOfEntries.length < 7) {
+              this.listOfEntries.push(entry);
+            }
+          }
+        });
+      });
+    this.isLoading = false;
```
coverbeck, line 103: This needs to be done as part of the subscription, e.g., in a `finalize`. As it is, I see have No Tools Registered briefly displayed because this is done too early.

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  ngOnInit(): void {
+    this.getMyEntries();
+
+    if (this.entryType) {
+      this.entryTypeLowerCase = this.entryType.toLowerCase();
+    }
+
+    //Get the links for the specified entryType
+    if (this.entryType === EntryUpdateTime.EntryTypeEnum.WORKFLOW) {
```
coverbeck, line 57: Too much work for this PR, but I wonder if we shouldn't have sub components to avoid all these if/else conditions, e.g., a workflow-box-component, a tool-box-component, etc.

coverbeck (commented):
Ran this locally and it's looking good!

src/app/home-page/widget/entry-box/entry-box.component.ts
eileenzhh: strongly support adding a `type` query :raised_hands:

src/app/home-page/new-dashboard/new-dashboard.component.html
```diff
@@ -0,0 +1,55 @@
...
+              <mat-card-title>Organizations</mat-card-title>
+              <span class="bubble">3</span>
+            </mat-card-header>
+          </div>
+          <button
+            mat-button
+            class="private-btn small-btn-structure ml-3 mb-3"
+            id="registerOrganizationButton"
+            matTooltip="Register Organization"
+            matTooltipPosition="after"
```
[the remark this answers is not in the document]
eileenzhh: Yup, implementing in other PR

src/app/home-page/widget/entry-box/entry-box.component.ts
eileenzhh: I guess it could be separated into separate components for sure, I'm not against that change later on and it shouldn't be a messy change either.

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -14,73 +14,79 @@
...
       <div
         fxLayout="column"
         fxLayoutAlign="start start"
         fxLayoutGap="20px"
         fxFlex="1 2 100%"
-        class="px-4 side-column"
+        class="px-4 side-column left-side-column"
         fxFlexAlign="stretch"
       >
-        <app-requests *ngIf="user$ | async" class="w-100"></app-requests>
```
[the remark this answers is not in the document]
eileenzhh: Nope, not sure how this happened but I'll revert it!

src/app/home-page/widget/entry-box/entry-box.component.ts
eileenzhh: Oh yes a filter would work very well, I was definitely stuck in the headspace of neverending `if` statements for a while :dizzy_face:

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,101 @@
...
+      .subscribe((myEntries: Array<EntryUpdateTime>) => {
+        this.listOfEntries = [];
+        myEntries
+          .filter(
+            (entry) =>
+              entry.entryType === this.entryType ||
+              (this.entryType === EntryUpdateTime.EntryTypeEnum.TOOL && entry.entryType === EntryUpdateTime.EntryTypeEnum.APPTOOL)
+          )
+          .forEach((entry: EntryUpdateTime) => {
+            if (this.listOfEntries.length < 7) {
```
coverbeck, line 76: To followup, in addition to simplification, you're also going through the loop n times, e.g., 1,000 times if there are 1,000 items in the array, even though you only need the first 7. For the amount of data we're currently dealing with, it probably won't matter, performance-wise, but just as a best practice...

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,101 @@
...
+        debounceTime(750),
+        takeUntil(this.ngUnsubscribe)
+      )
+      .subscribe((myEntries: Array<EntryUpdateTime>) => {
+        this.listOfEntries = [];
+        myEntries
+          .filter(
+            (entry) =>
+              entry.entryType === this.entryType ||
+              (this.entryType === EntryUpdateTime.EntryTypeEnum.TOOL && entry.entryType === EntryUpdateTime.EntryTypeEnum.APPTOOL)
```
coverbeck, line 73: This won't work if `this.entryType` is `EntryType.EntryType.APPTOOL`, but maybe we don't need to worry about it.

src/app/home-page/new-dashboard/new-dashboard.component.ts
```diff
@@ -0,0 +1,48 @@
...
+import { AlertService } from 'app/shared/alert/state/alert.service';
+
+@Component({
+  selector: 'app-new-dashboard',
+  templateUrl: './new-dashboard.component.html',
+  styleUrls: ['./new-dashboard.component.scss'],
+})
+export class NewDashboardComponent extends Base implements OnInit {
+  public user$: Observable<User>;
+  user: User;
```
coverbeck, line 19: It doesn't look like you're using `user` nor `$user`. And if you don't need them, you probably don't need `userQuery` neither, not the block on lines 31-33.

src/app/home-page/new-dashboard/new-dashboard.component.ts
```diff
@@ -0,0 +1,48 @@
...
+    private alertService: AlertService
+  ) {
+    super();
+  }
+
+  ngOnInit() {
+    this.userQuery.user$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((user) => {
+      this.user = user;
+    });
+    this.registerToolService.isModalShown.pipe(takeUntil(this.ngUnsubscribe)).subscribe((isModalShown: boolean) => {
```
coverbeck, line 34: I'd never noticed this before, but it's odd... entry-box.component.ts sets isModalShown to true, then this component opens the dialog.

Without revisiting the whole thing, which I'm not quite grokking (I know it was like this), it seems like this code should be in entry-box.component.ts, because that's the one thats isModalShown to true.

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -14,10 +14,10 @@
   ~    limitations under the License.
   -->
 
-<div fxLayout>
+<div fxLayout *ngIf="!Dockstore.FEATURES.enableNewDashboard; else newDashboard">
   <app-my-sidebar fxFlexAlign="stretch"></app-my-sidebar>
 
-  <div id="column-holder" fxLayout="row" fxLayout.lt-md="column" fxLayoutAlign="start start" fxLayoutGap="10px">
+  <div id="column-holder" fxLayout="row" fxLayout.lt-md="column" fxLayoutAlign="start start" fxLayoutGap="10px" fxFlex="1 2 100%">
```
coverbeck, line 20: You removed all your other changes to the old dashboard, but this still remains. Just making sure it's intentional? Maybe it's required because of the new div this is child of?

src/app/home-page/widget/entry-box/entry-box.component.scss
```diff
@@ -0,0 +1,46 @@
+@use '@angular/material' as mat;
+@import '/src/materialColorScheme.scss';
+
+.entry-boxes {
```
coverbeck, line 4: Not sure if we should be having all this component-specific CSS. @kathy-t will know better than I.

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -0,0 +1,101 @@
...
+import { Dockstore } from 'app/shared/dockstore.model';
+
+@Component({
+  selector: 'app-entry-box',
+  templateUrl: './entry-box.component.html',
+  styleUrls: ['./entry-box.component.scss'],
+})
+export class EntryBoxComponent extends Base implements OnInit {
+  Dockstore = Dockstore;
+  @Input() entryType: EntryUpdateTime.EntryTypeEnum;
```
coverbeck, line 18: Suggest you change this to `@Input() entryType: typeof EntryUpdateTime.EntryTypeEnum.TOOL | typeof EntryUpdateTime.EntryTypeEnum.SERVICE | typeof EntryUpdateTime.EntryTypeEnum.WORKFLOW;`, because your code below assumes it's never set to an APPTOOL.

src/app/home-page/widget/entry-box/entry-box.component.ts
coverbeck: Is there a ticket for that? If not, I'll create one.

src/app/home-page/widget/entry-box/entry-box.component.ts
eileenzhh: Hmm that is true I originally did it to count the number of total entries but there should be a better endpoint for that.

src/app/home-page/widget/entry-box/entry-box.component.ts
eileenzhh: I don't think `this.entryType` will ever be `APPTOOL` because based off of the three boxes and calling this component, it can only be `workflow` `tool` or `service`, I'm just grouping the APPTOOLs into the same tools box. I think this makes sense?

src/app/home-page/new-dashboard/new-dashboard.component.ts
eileenzhh: Oh yes

src/app/home-page/new-dashboard/new-dashboard.component.ts
eileenzhh: I initially put it in the entry-box.component.ts file but I ran into some issues that resulted in me needing to put it here (to be honest i don't remember what the exact errors were but I can try it again to check it out)

src/app/home-page/home-logged-in/home-logged-in.component.html
eileenzhh: Yeah without it, it caused everything to be shifted towards the left and did not fill the screen, I thought that it would behave as before if I just added the `*ngIf` into the original div instead of making a new parent div, but the problem still persisted so I had to add the new `flex` definitions.

src/app/home-page/widget/entry-box/entry-box.component.ts
eileenzhh: https://ucsc-cgl.atlassian.net/browse/SEAB-4699 I just created a ticket for it, I'm not sure what labels are appropriate to add to it and if the definition is clear enough, could you review it?

coverbeck (approved):
There are tickets for the issues we've discussed, and this is behind a flag, so this looks good. Looking forward to seeing it in dev.

kathy-t (approved):

src/app/home-page/widget/entry-box/entry-box.component.scss
kathy-t: Hmm, it seems component specific enough because it's a lot of layout styling

PR #1563, opened 2022-06-28
Fix disabled member edit/delete buttons do not look disabled

eileenzhh:
**Description**
As an admin of an organization, when looking at the members page, their own member card's edit and delete buttons are disabled however they do not appear as disabled. Since the icons are fixed in colour, I think it makes sense to not show the buttons at all in this case.

**Issue**
GitHub Issue [4963](https://github.com/dockstore/dockstore/issues/4963)
JIRA ticket: [DOCK-2188](https://ucsc-cgl.atlassian.net/browse/DOCK-2188)

### DIFF
--- a/src/app/organizations/organization-members/organization-members.component.html
+++ b/src/app/organizations/organization-members/organization-members.component.html
@@ -49,7 +49,13 @@
             </div>
           </div>
           <div fxLayoutAlign="center" *ngIf="canEditMembership$ | async">
-            <mat-card-content fxLayout="row" fxLayoutGap="35px" class="m-0" fxLayoutAlign="space-evenly">
+            <mat-card-content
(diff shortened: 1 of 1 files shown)


eileenzhh:
Logged in as bethsheets: 
image

eileenzhh:
A consideration was adding a bubble text indicating that this member card was their own like "Myself" or something so that the area appears less empty... thoughts?

denis-yuen:
> A consideration was adding a bubble text indicating that this member card was their own like "Myself" or something so that the area appears less empty... thoughts?

I think both work for me, I figure they know who they are so they should be able to put two and two together if the icons are missing

denis-yuen (approved):

coverbeck (approved):

PR #1572, opened 2022-07-19
A feature flag approach

coverbeck:
**Description**
One approach to the feature flag.

* User manually enters https://dev.dockstore.net?newDashboard when initially navigating to the site
* New dashboard would stay active until user reloads the page via the browser without that query param.
* Eileen changes src/app/home-page/home-logged-in/home-logged-in.component.html to have `*ngIf`, e.g.,
```
<div *ngIf="Dockstore.FEATURES.enableNewDashboard">
<!-- New Dashboard html or component -->
</div>
<div *ngIf="!Dockstore.FEATURES.enableNewDashboard">
<!-- Old dashboard html or component -->
</div>
```
Anybody who knows the query param would be able to access.

Other approaches are discussed in the [issue](https://ucsc-cgl.atlassian.net/browse/SEAB-4597), e.g., we could do something fancier with the routes so that if the flag is enabled, the route is for the new dashboard component, otherwise it's for the old dashboard component. Or we can add "guards" to only allow certain users to see the new dashboard, etc.

**Issue**
[SEAB-4597](https://ucsc-cgl.atlassian.net/browse/SEAB-4597)

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -46,12 +46,14 @@ import { CustomHeaderInterceptor } from './interceptors/custom-header.intercepto
 import { WorkflowVersionsInterceptor } from './interceptors/workflow-versions.interceptor';
 import { LoginComponent } from './login/login.component';
 import { LoginService } from './login/login.service';
+import { AccountSidebarModule } from './loginComponents/accounts/account-sidebar/account-sidebar.module';
 import { AccountsComponent } from './loginComponents/accounts/accounts.component';
 import { ControlsComponent } from './loginComponents/accounts/controls/controls.component';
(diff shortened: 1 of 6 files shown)


denis-yuen:
> Anybody who knows the query param would be able to access.

I think this is fine, presumably we wouldn't document this anywhere on public documentation. Additionally, there's nothing secret on the dashboard, the webservice is doing the real handling of permissions

denis-yuen (approved):

kathy-t (approved):

coverbeck:
This has been merged into #1562, which will end up getting merged to develop, so closing this PR and deleting the branch.

PR #1573, opened 2022-07-20
Get logged in user during canActivate to check authentication

kathy-t:
**Description**
Companion PRs:
- compose_setup: https://github.com/dockstore/compose_setup/pull/230
- webservice: https://github.com/dockstore/dockstore/pull/5031

I recommend reviewing the compose_setup PR first, then this PR, then the webservice PR for things to make the most sense.

This PR checks if the user is authenticated by calling the get logged in user endpoint before navigating to authenticated pages. In a multiple webservice scenario where a user might have two browsers open, if they delete their account in browser 1, then browser 2 will still think they're authenticated until the page is refreshed because the token store isn't cleared in browser 2. I believe this is what happened during my experiment in the ticket description. Adding this check to the AuthGuard ensures that a deleted user gets logged out when they try to navigate to different authenticated pages in browser 2.

Also added auth guards to some pages that should have it.

How I tested this locally:
- To imitate the scenario where the user still has access on a second instance:
  - In the webservice, don't invalidate the user tokens when the user is deleted
  - In the UI, don't automatically log a user out when they delete their account so the user can still navigate pages
- Logged in as myself, delete my account.
- Navigate to My Workflows.
- Get redirected to the logged out page.
- In the console, see that the get logged in endpoint returned a 401 unauthorized or a 404 user not found. You'll get a 401 if you wait 10 seconds (the time it takes for the cache to expire) or you'll get a 404 if you navigate to an authorized page before 10 seconds.

**Issue**
[SEAB-3895](https://ucsc-cgl.atlassian.net/browse/SEAB-3895)

### DIFF
--- a/src/app/app.routing.ts
+++ b/src/app/app.routing.ts
@@ -129,7 +129,7 @@ const APP_ROUTES: Routes = [
     path: 'starred',
     component: StarredEntriesComponent,
     canActivate: [AuthGuard],
-    data: { title: 'Dockstore | Starred Tools & Workflows' },
+    data: { title: 'Dockstore | Starred Tools, Workflows & Organizations' },
   },
   { path: 'maintenance', component: MaintenanceComponent, data: { title: 'Dockstore | Maintenance' } },
   { path: 'funding', component: FundingComponent, data: { title: 'Dockstore | Funding' } },
--- a/src/app/app.routing.ts
+++ b/src/app/app.routing.ts
@@ -40,6 +40,7 @@ const APP_ROUTES: Routes = [
     component: HomeLoggedInComponent,
     pathMatch: 'full',
     data: { title: 'Dockstore | Dashboard' },
+    canActivate: [AuthGuard],
   },
   {
     path: 'docs',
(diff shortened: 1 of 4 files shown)


src/app/app.routing.ts
```diff
@@ -124,7 +125,12 @@ const APP_ROUTES: Routes = [
   { path: 'onboarding', component: OnboardingComponent, canActivate: [AuthGuard], data: { title: 'Dockstore | Onboarding' } },
   { path: 'accounts', component: AccountsComponent, canActivate: [AuthGuard], data: { title: 'Dockstore | Accounts' } },
   { path: 'auth/:provider', component: AuthComponent },
-  { path: 'starred', component: StarredEntriesComponent, data: { title: 'Dockstore | Starred Tools & Workflows' } },
+  {
+    path: 'starred',
+    component: StarredEntriesComponent,
+    canActivate: [AuthGuard],
+    data: { title: 'Dockstore | Starred Tools & Workflows' },
```
coverbeck, line 132: It was like this, but I just noticed in reviewing this PR, that the page also shows starred organizations, so text is wrong.

src/app/shared/auth.guard.ts
```diff
@@ -17,15 +17,35 @@
...
+    private logoutService: LogoutService,
+    private usersService: UsersService
+  ) {}
   canActivate(next: ActivatedRouteSnapshot, state: RouterStateSnapshot) {
-    if (this.auth.isAuthenticated()) {
-      return true;
-    }
-    this.router.navigate(['/login']);
-    return false;
+    return this.usersService.getUser().pipe(
```
coverbeck, line 34: Just noting that this is going to block the auth pages load until the API request completes, which is ~90-100ms when I tried it just now against prod. Which isn't a lot, but it adds up.

OTOH, it's the right thing to do, and we could come up with some optimization later.

coverbeck (approved):

david4096 (approved):

denis-yuen (approved):

PR #1574, opened 2022-07-21
add version date and use date for old dates

denis-yuen:
**Description**
* Display date when repo as a whole is modified in addition to when the display version was modified
* Don't think we have a ticket for this, but during UI design process was remarked that past a certain amount of time, probable better to display date rather than "time ago"

CircleCI changes/observations
- looks like the issue I ran into was that the tests broke in a way that pages didn't come up, cypress didn't timeout, so CircleCI would run into a ten minute limit (so no output to debug with)
- extended timeout (to 30 min) results in Cypress eventually giving up, timing out, and properly giving out debug info
  -  debug mode does this too, but results in way too much to read
- turns out I just needed to add some null checks (looks like some old/stub workflows are missing selected versions)
- going to leave in 30 min timeouts, this may help with nightly auth tests which currently die with the same symptom (but probably not the same cause)

**Issue**
dockstore/dockstore#4636
dockstore/dockstore#4640


denis-yuen:
picture
Screenshot from 2022-07-22 11-14-24
Screenshot from 2022-07-22 11-14-27
Screenshot from 2022-07-22 11-14-37
s

denis-yuen:
note for self:
-  instance size seems to be ruled out (upgrading to medium+ and large doesn't make a difference)
- upgrading cypress is inconclusive (different failures)
- re-running older develop builds pass (with same dependencies)
- a lot of nightly auth tests are dying in the same way

david4096 (approved):

kathy-t (approved):

src/app/workflow/workflow.component.ts
```diff
@@ -228,7 +230,10 @@ export class WorkflowComponent extends Entry implements AfterViewInit, OnInit {
     this.resetWorkflowEditData();
     // messy prototype for a carousel https://developers.google.com/search/docs/guides/mark-up-listings
     // will need to be aggregated with a summary page
-    this.schema = this.bioschemaService.getWorkflowSchema(this.workflow, this.selectedVersion);
+    if (this.selectedVersion) {
+      this.schema = this.bioschemaService.getWorkflowSchema(this.workflow, this.selectedVersion);
+      this.versionAgoMessage = this.dateService.getAgoMessage((this.selectedVersion as WorkflowVersion).last_modified);
```
coverbeck, line 235: Cool, I didn't know or had forgot you can use `as` like that in TypeScript.

src/app/workflow/workflow.component.html
```diff
@@ -80,11 +80,21 @@ <h3 id="workflow-path" data-cy="workflowTitle" fxLayout="row" fxLayoutAlign="sta
         </mat-form-field>
       </div>
       <p class="update">
-        <span [matTooltip]="(extendedWorkflow$ | async)?.last_modified_date | date: 'medium'"
-          >Last Modified: {{ (extendedWorkflow$ | async)?.agoMessage || 'n/a' }}</span
-        >
+        <span *ngIf="selectedVersion !== null">
+          <span *ngIf="'TAG' === selectedVersion.referenceType" [matTooltip]="selectedVersion.last_modified | date: 'medium'"
```
coverbeck, line 84: There is an enum, ReferenceTypeEnum, you could use instead of `TAG`.

coverbeck (approved):

src/app/workflow/launch-third-party/launch-third-party.component.html
```diff
@@ -13,10 +13,6 @@
     <div class="container-source-repos">
       <div class="container-launch-with">
         <div *ngIf="workflow?.descriptorType === WorkflowModel.DescriptorTypeEnum.WDL">
-          <div *ngIf="(hasFileImports$ | async) || (hasHttpImports$ | async)" class="import-warning">
```
coverbeck, line 16: Just noticed this is targetted for 1.12.x.; should this PR be for hotfix? Or this part extracted for another PR?

denis-yuen: should be 1.13 for the tickets and this PR, did I miss something?

coverbeck: I meant the issue dockstore/dockstore#5028, which this block addresses, has 1.12.x as the fix version.

src/app/workflow/workflow.component.html
denis-yuen: Running into 
https://stackoverflow.com/questions/35835984/how-to-use-a-typescript-enum-value-in-an-angular2-ngswitch-statement
I can declare a reference in `workflow.component.ts` but it doesn't seem like a very good way of doing it

src/app/workflow/launch-third-party/launch-third-party.component.html
denis-yuen: Was copied out for https://github.com/dockstore/dockstore-ui2/pull/1575/files

src/app/workflow/workflow.component.html
coverbeck: Looks like we do it already: https://github.com/dockstore/dockstore-ui2/blob/21b3959c8f21202a8be4e3cb500a16028d2b4b7c/src/app/workflow/workflow.component.ts#L95

But I don't feel super strongly about this.

denis-yuen: I'm ok with it, if its established practice :grin:

aofarrel:
@kathy-t @denis-yuen What is the intended and actual behavior for this scenario?
1. Someone make a .dockstore.yml workflow
2. On April 20th, as per Santa Cruz tradition, an undergraduate makes some changes and tags a commit that fills the workflow's output with botanical references as 4.2.0
3. On April 21th, the PI finds out and politely requests that the 4.2.0 tagged commit be replaced by something more professional, and because git allows rewriting history and tags are not immutable, this is done such that the botanical version of tag 4.2.0 no longer exists and is replaced by a new 4.2.0 tag

Which "tag created" date shows up on Dockstore? April 20th or April 21st?

denis-yuen:
I suspect that the second tag will be rejected since it will look like a duplicate for the first tag (since we don't support/track deletions on github) I haven't tested it though

This could be viewed as a variant of https://github.com/dockstore/dockstore/issues/2800

PR #1576, opened 2022-07-26
Update the links in the dashboard

eileenzhh:
**Description**
After the dashboard URL was changed to '/dashboard', the 'Quick start', 'Accounts', and 'Starred' links brought you to the wrong links ('dashboard/...'). This PR updates the links to the working addresses.

**Issue**
[SEAB-4594](https://ucsc-cgl.atlassian.net/browse/SEAB-4594)

### DIFF
--- a/src/app/home-page/home-logged-in/home-logged-in.component.html
+++ b/src/app/home-page/home-logged-in/home-logged-in.component.html
@@ -64,9 +64,9 @@
         <div class="w-100">
           <h3>Links</h3>
           <ul>
-            <li><a routerLink="quick-start">Quickstart</a></li>
-            <li><a routerLink="accounts">My Account</a></li>
-            <li><a routerLink="starred">My Starred</a></li>
+            <li><a [routerLink]="['../quick-start']">Quickstart</a></li>
+            <li><a [routerLink]="['../accounts']">My Account</a></li>
+            <li><a [routerLink]="['../starred']">My Starred</a></li>
             <li>
               <a href="https://discuss.dockstore.org/" target="_blank" rel="noopener noreferrer">Forum <mat-icon>open_in_new</mat-icon></a>
             </li>
--- a/src/app/home-page/home-logged-in/home-logged-in.component.html
+++ b/src/app/home-page/home-logged-in/home-logged-in.component.html
@@ -64,9 +64,9 @@
         <div class="w-100">
           <h3>Links</h3>
           <ul>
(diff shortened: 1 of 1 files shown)


src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -64,9 +64,9 @@ <h3>Recent Activity</h3>
         <div class="w-100">
           <h3>Links</h3>
           <ul>
-            <li><a routerLink="quick-start">Quickstart</a></li>
-            <li><a routerLink="accounts">My Account</a></li>
-            <li><a routerLink="starred">My Starred</a></li>
+            <li><a [routerLink]="['../quick-start']">Quickstart</a></li>
```
coverbeck, line 67: I think it would be better to just have, for example, `['/quick-start']`, without the `..`; that way if this component gets moved to a different location (unlikely, but still), the links will continue to work.

On the other hand, you could argue that the links will always be siblings of this page, but I feel it's more flexible to remove the `..`.

eileenzhh: oh yup I agree :+1: better to be more flexible

denis-yuen (approved):

kathy-t (approved):

src/app/home-page/home-logged-in/home-logged-in.component.html
```diff
@@ -64,9 +64,9 @@ <h3>Recent Activity</h3>
         <div class="w-100">
           <h3>Links</h3>
           <ul>
-            <li><a routerLink="quick-start">Quickstart</a></li>
-            <li><a routerLink="accounts">My Account</a></li>
-            <li><a routerLink="starred">My Starred</a></li>
+            <li><a [routerLink]="['/quick-start']">Quickstart</a></li>
+            <li><a [routerLink]="['/accounts']">My Account</a></li>
+            <li><a [routerLink]="['/starred']">My Starred</a></li>
```
svonworl, line 69: No need to change, but for future reference, I think you can just assign the property to a string literal like so:
`<a routerLink="/quick-start">`

svonworl (approved):

src/app/home-page/home-logged-in/home-logged-in.component.html
eileenzhh: Ooh okay got it! :+1:

PR #1577, opened 2022-07-28
SEAB-3792: Fix search self-DOS

svonworl:
**Description**

**UPDATE: The PR has been changed to implement Change 2 only (explained below), Change 1 has been removed.**
 
This PR implements a fix for [the currently-known variety] of the search "self-DOS" bug.  This is a somewhat hacky fix that feels a bit gross to me, but is probably an ok call from a practical POV.

This self-DOS occurs when the browser back (or forward) button is pressed and there are at least two search URLs at the top of the browser URL stack.  If the current page is a search with a text term and facet selected, and the destination page is also a search, a loop is triggered where the search code appears to repeatedly, rapidly, and endlessly switch between two different search states, triggering a rapid stream of ES queries.

The exact mechanics of the bug were difficult to determine - the search code includes some subtle circular data flows that are susceptible to this type of problem.  And, even if I figured it out, which could take days, I wasn't sure if I could easily fix it, and felt that there was a good chance that I'd introduce other bugs in the process...

So, the solution was to try to avoid the situation which can trigger the issue (two search URLs at the top of the browser url stack) and take evasive action in the cases we could not.

There are two changes:

1. The search URL rewrite code is modified to _replace_ the entry at the top of the browser stack, rather than _add_ another entry.  This prevents search URLs from piling up, and has the added benefit that when the user presses the back button, they go back to the page they moved to search from, which feels more appropriate (to me) than bumping back one step in the search state.

2. If we detect a situation where the forward/back button takes us to another search (as can happen after we're on the search page and and we click on a link to the search page [such as in the menu bar]), we force the search component to reinitialize, causing the destination search to run in newly-initialized component, nipping the self-DOS in the bud.

Do note that there may be other undiscovered varieties of the self-DOS lurking...

**Issue**
SEAB-3792


src/app/search/search.component.ts
```diff
@@ -201,6 +204,24 @@ export class SearchComponent implements OnInit, OnDestroy {
     this.hasFacetAutoCompleteTerms$ = this.searchQuery.hasFacetAutoCompleteTerms$;
   }
 
+  /**
+   * Handle the popstate events generated by browser back/forward button presses.
+   * If the destination page is also a search page, silently navigate to a nonexistent
+   * url, then back to the destination page again, causing this component to
+   * reinitiatize, which terminates the previous search and loads the destination page
```
coverbeck, line 211: Typo in `reinitiatize`

coverbeck (approved):
I find the new back behavior unintuitive; I tried with Amazon as a comparison, and if you select a facet, then click the back button, you go back to the page with the facet unselected. Not a perfect comparison, but still...

But this is exponentially better than self-DOSing ourselves. :) Maybe in the future we can revisit the page; as you said the interactions are complex.

svonworl:
> 



> I find the new back behavior unintuitive; I tried with Amazon as a comparison, and if you select a facet, then click the back button, you go back to the page with the facet unselected. Not a perfect comparison, but still...
> 
> But this is exponentially better than self-DOSing ourselves. :) Maybe in the future we can revisit the page; as you said the interactions are complex.

Change #1 (on search facet/text change, replacing the top URL on the browser stack with the new URL rather than adding) is optional, we can remove it if the team deems it objectionable.

src/app/search/search.component.ts
svonworl: Doing my part to expedite the evolution of the English language.

denis-yuen:
> Change #1 (on search facet/text change, replacing the top URL on the browser stack with the new URL rather than adding) is optional, we can remove it and keep the previous back behavior if the team deems it objectionable.

I think I agree, I would probably find that unexpected as well.  That said, does it reduce the chance of a self-dos or support change 2 in some way?

denis-yuen (approved):
comment above

svonworl:
> > Change #1 (on search facet/text change, replacing the top URL on the browser stack with the new URL rather than adding) is optional, we can remove it and keep the previous back behavior if the team deems it objectionable.
> 
> I think I agree, I would probably find that unexpected as well. That said, does it reduce the chance of a self-dos or support change 2 in some way?

Change #1 (replacing the url on top of the browser stack on search rewrite) largely prevents the situation that leads to the self-DOS (two search urls on top of the browser stack), but Change #2 seems to work well enough that it can stand on its own.  I will remove Change #1.

PR #1583, opened 2022-08-03
Show all GitHub orgs you have access to, even if there are no entries

coverbeck:
**Description**
Call `/users/registries/github.com/organizations` endpoint instead of `/users/github/organizations` and use it. Create empty `OrgWorkflowObject`s for orgs without repos.

The new endpoint is currently slow, but it's the right thing to do if we want to show the logs for all GitHub orgs.

I will also create a web service PR that:

1. Optimizes the endpoint -- I think I see a way
2. Only returns logs for repos that you have permissions to
3. Deletes the now unused endpoint.

Apologies for the first review I requested; I was initially going to break this into 2 PRs, but then realized it was easier to to repurpose the code making the call to the unused endpoint.

***Small Other Changes***
* Changed a constructor to protected in abstract class because WebStorm gave  a warning
* Did a `forkJoin` because the way we were doing the success message wasn't quite right.

**Issue**
dockstore/dockstore#4845

### DIFF
--- a/src/app/mytools/my-tool/my-tool.component.ts
+++ b/src/app/mytools/my-tool/my-tool.component.ts
@@ -29,20 +29,21 @@ import { RegisterToolComponent } from '../../container/register-tool/register-to
 import { RegisterToolService } from '../../container/register-tool/register-tool.service';
 import { Tool } from '../../container/register-tool/tool';
 import { AccountsService } from '../../loginComponents/accounts/external/accounts.service';
+import { OrgWorkflowObject } from '../../myworkflows/my-workflow/my-workflow.component';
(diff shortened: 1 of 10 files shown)


svonworl (approved):

denis-yuen (approved):

kathy-t (approved):

coverbeck:
Here is a screenshot, although it may be hard to tell what's going on -- I have no workflows for DockstoreTestUser2, but it's still showing up.

Screen Shot 2022-08-04 at 5 00 37 PM

svonworl (approved):

denis-yuen (approved):

kathy-t (approved):

PR #1585, opened 2022-08-05
what if no email

denis-yuen:
**Description**
UI counterpart to https://github.com/dockstore/dockstore/pull/5062
Most of this is a red herring, the stargazers page seems to only really be using avatarurls anyway (due to lazy loading of user profiles)

**Issue**
https://github.com/dockstore/dockstore/pull/5062

### DIFF
--- a/src/app/shared/user/user.service.ts
+++ b/src/app/shared/user/user.service.ts
@@ -120,15 +120,11 @@ export class UserService {
     this.getExtendedUserData();
   }
 
-  gravatarUrl(email: string | null, defaultImg: string | null): string | null {
-    if (email) {
-      return this.gravatarService.gravatarUrlForEmail(email, defaultImg);
+  gravatarUrl(defaultImg: string | null): string | null {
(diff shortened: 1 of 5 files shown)


src/app/shared/user/user.service.ts
```diff
@@ -120,15 +120,11 @@ export class UserService {
     this.getExtendedUserData();
   }
 
-  gravatarUrl(email: string | null, defaultImg: string | null): string | null {
-    if (email) {
-      return this.gravatarService.gravatarUrlForEmail(email, defaultImg);
```
coverbeck, line 125: I can't remember why we did it here instead of just using the gravatarUrl directly from GitHub/Google. I wonder if there were CORS issues, but that doesn't sound right. Maybe we weren't capturing that info in the DB at the time? At any rate, this seems right.

coverbeck (approved):

david4096 (approved):

svonworl (approved):

PR #1578, opened 2022-07-28
Feature/dock 1810/galaxy launch button test part 1

fhembroff:
**Description**
This PR adds some `data-cy` attributes required for a second PR I'll make for this ticket. This PR is required as my tests will not run on `localhost:4200` in circleCI due to this line,
https://github.com/dockstore/dockstore-ui2/blob/e88cc80d0fc2dc2b15a126e265202e4eae16348d/cypress/integration/smokeTests/sharedTests/basic-enduser.ts#L173

So, what I will do is merge this PR into develop, then use the circleCI nightly test setup (slightly modified) to run my proposed test for this ticket to verify it is working as intended and then make another PR.

**Issue**
[DOCK-1810](https://ucsc-cgl.atlassian.net/browse/DOCK-1810)
https://github.com/dockstore/dockstore/issues/4278

Please make sure that you've checked the following before submitting your pull request. Thanks!

- [x] Check that your code compiles by running `npm run build`
- [x] If this is the first time you're submitting a PR or even if you just need a refresher, consider reviewing our [style guide](https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide#pr-checklist)
- [x] Do not bypass Angular sanitization (bypassSecurityTrustHtml, etc.), or justify why you need to do so
- [x] If displaying markdown, use the `markdown-wrapper` component, which does extra sanitization
- [x] Do not use cookies, although this may change in the future
- [x] Run `npm audit` and ensure you are not introducing new vulnerabilities
- [x] Do due diligence on new 3rd party libraries, checking for CVEs
- [x] Don't allow user-uploaded images to be served from the Dockstore domain

PR #1586, opened 2022-08-06
Galaxy Launch Button Test

fhembroff:
**Description**
This PR adds testing for the galaxy launch button on this [web page](https://dockstore.org/workflows/github.com/dockstore-testing/galaxy-workflow-dockstore-example-2:0.9.6?tab=info). The change opens up the menu and selects each button to ensure they have an accurate URL.

These changes do not execute under the `everything` task in CircleCI and instead only run in the job `no_auth_smoke_tests_*`, for this reason,
https://github.com/dockstore/dockstore-ui2/blob/f72a583436091ad33322978da8f2e27f1870c5cd/cypress/integration/smokeTests/sharedTests/basic-enduser.ts#L172
 I have tested these tests in the job `no_auth_smoke_tests_dev`, results found [here](https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/8260/workflows/a406f5b4-78ab-4730-bb2d-40dfeac68d76/jobs/31319).

Currently, the test fails for all `no_auth_smoke_tests` except for `no_auth_smoke_tests_dev` as this test requires the use of this [PR](https://github.com/dockstore/dockstore-ui2/pull/1578), this is why the galaxy tests are located in `devOnly/basic-enduser.ts` (which means that `devOnly/basic-enduser.ts` will only run in the `no_auth_smoke_tests_dev` job). I have kept most of the formatting from `sharedTests/basic-enduser.ts` so that `devOnly/basic-enduser.ts` can be easily merged with `sharedTests/basic-enduser.ts`  in the future.

**Issue**
[DOCK-1810](https://ucsc-cgl.atlassian.net/browse/DOCK-1810)
https://github.com/dockstore/dockstore/issues/4278


denis-yuen (commented):
> Currently, the test fails for all `no_auth_smoke_tests` except for `no_auth_smoke_tests_dev` as this test requires the use of this [PR](https://github.com/dockstore/dockstore-ui2/pull/1578), which I believe is only available in develop.

If this only works in dev, might be worth putting it into a `devOnly` directory. 
These are blank right now, I'm assuming because all current smoke tests work in all environments. But if this one won't work, may make sense to limit it to one environment. 

Example https://github.com/dockstore/dockstore-ui2/blob/develop/package.json#L33

fhembroff:
> If this only works in dev, might be worth putting it into a `devOnly` directory. These are blank right now, I'm assuming because all current smoke tests work in all environments. But if this one won't work, may make sense to limit it to one environment.
> 
> Example https://github.com/dockstore/dockstore-ui2/blob/develop/package.json#L33

For this change would I put all of `basic-enduser.ts` into the `devOnly` directory or would I just add the parts that I added in this PR?

denis-yuen:
Just the parts that test the galaxy launch buttons

cypress/integration/smokeTests/devOnly/basic-enduser.ts
```diff
@@ -0,0 +1,85 @@
...
+    cy.contains('.mat-tab-label', 'DAG').click();
+    cy.url().should('contain', '?tab=dag');
+    cy.get('[data-cy=dag-holder]').children().should('have.length.of.at.least', 1);
+  });
+
+  if (type === 'Galaxy') {
+    it('test that galaxy button exists', () => {
+      cy.get('[data-cy=galaxyLaunchWith] button').should('exist');
+      cy.get('[data-cy=galaxyLaunchWith] button').click();
+      cy.get('[data-cy=multiCloudLaunchOption]').each(($el, index) => {
```
coverbeck, line 71: Could you check that at least one of the non-"specify your own url" radiobuttons exists? Or maybe check that at least the usegalaxy.org radiobutton choice is present? While those radiobuttons are API driven, if they're not present, it wouldn't be good.

cypress/integration/smokeTests/devOnly/basic-enduser.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  it('DAG tab works', () => {
+    /// New material have to click twice
+    cy.contains('.mat-tab-label', 'DAG').click();
+    cy.contains('.mat-tab-label', 'DAG').click();
+    cy.url().should('contain', '?tab=dag');
+    cy.get('[data-cy=dag-holder]').children().should('have.length.of.at.least', 1);
+  });
+
+  if (type === 'Galaxy') {
```
coverbeck, line 67: Please create a followup ticket to move this test out of dev-only after 1.13 is deployed (which means you dependent PR will be in all regions), specifying where it should go.

coverbeck (commented):
Could you create a followup ticket

fhembroff:
@denis-yuen I have implemented your suggestion and updated the PR comment.

cypress/integration/smokeTests/devOnly/basic-enduser.ts
fhembroff: Ticket created [SEAB-4686](https://ucsc-cgl.atlassian.net/browse/SEAB-4686)

cypress/integration/smokeTests/devOnly/basic-enduser.ts
fhembroff: Implemented.

It should now be testing that their is at least one non-"specify your own url" and that one of them contains `usegalaxy.org`

cypress/integration/smokeTests/devOnly/basic-enduser.ts
denis-yuen: Might also document the sections that will be moved to `./sharedTests/basic-enduser.ts` after 1.13 is deployed
I think there is some code here which will end up being duplicate/redundant

denis-yuen (approved):

cypress/integration/smokeTests/devOnly/basic-enduser.ts
fhembroff: I have added some comments explaining which code will have to be moved to `../sharedTests/basic-enduser.ts`.

coverbeck (approved):

kathy-t (approved):

PR #1589, opened 2022-08-10
add more search boxes

denis-yuen:
**Description**
Adds mores search boxes. 

Efficiency issue, should be more familiar with change detection than I am, does it make sense that `GetFacetSearchUpdatePipe` is being called for all facets even if only one facet is being typed-in/changed?

**Issue**
https://github.com/dockstore/dockstore/issues/2062

### DIFF
--- a/src/app/search/facet-search/facet-search.pipe.ts
+++ b/src/app/search/facet-search/facet-search.pipe.ts
@@ -14,7 +14,11 @@ import { Pipe, PipeTransform } from '@angular/core';
  */
 export class GetFacetSearchResultsPipe implements PipeTransform {
   transform(items: Array<any>, searchText: string, facet: string): Array<any> {
-    if (!items || !searchText || facet !== 'author') {
+    if (
+      !items ||
+      !searchText ||
+      (facet !== 'author' && facet !== 'organization' && facet !== 'labels.value.keyword' && facet !== 'namespace')
+    ) {
       return items;
     }
     const value = searchText.toLowerCase();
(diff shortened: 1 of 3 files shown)


coverbeck (approved):
Threw out a suggestion, but it could certainly wait as another ticket.

> Efficiency issue, should be more familiar with change detection than I am, does it make sense that GetFacetSearchUpdatePipe is being called for all facets even if only one facet is being typed-in/changed?

Tried it in the browser and it seems OK; it's only getting invoked once per facet click.

denis-yuen:
> Tried it in the browser and it seems OK; it's only getting invoked once per facet click.

Hmmm. When I tried it, it seemed to be invoked once per facet with a search box every time I typed or even moved the mouse over a specific search box.

svonworl:
FUNCTION CALLS IN THE TEMPLATES!  WUT?!!?!

Just kidding, they look reasonably light-and-airy, it's fine.

src/app/search/facet-search/facet-search.pipe.ts
```diff
@@ -14,7 +14,15 @@ import { Pipe, PipeTransform } from '@angular/core';
...
   transform(items: Array<any>, searchText: string, facet: string): Array<any> {
-    if (!items || !searchText || facet !== 'author') {
+    if (
+      !items ||
+      !searchText ||
+      (facet !== 'author' &&
+        facet !== 'organization' &&
+        facet !== 'labels.value.keyword' &&
+        facet !== 'namespace' &&
+        facet !== 'categories.name.keyword')
```
svonworl, line 24: Maybe we could encapsulate this list of ES fields in a single function/data structure which we could use throughout.

svonworl (approved):

kathy-t (approved):

src/app/search/facet-search/facet-search.pipe.ts
denis-yuen: Adding this to https://github.com/dockstore/dockstore/issues/5072

PR #1591, opened 2022-08-12
DOCK-2012: Fix the search "quick click" bug (and some other problems)

svonworl:
**Description**
This PR addresses three bugs in the UI2 search interface:
1. The "quick click" bug, which occurred when toggling rapidly between the workflows and tools results.  The search code was updating the `checkboxMap` object, and the template was accessing it while it was not fully-formed.
2. A bug with the back button, which would not allow the user to move from more than one level down in the search to back past the root search page. 
3. An inconsistent setting of `searchMode`.  It should be `file`, rather than `File` (I think).

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2012
#4619

### DIFF
--- a/src/app/search/state/search.service.ts
+++ b/src/app/search/state/search.service.ts
@@ -359,8 +359,8 @@ export class SearchService {
     return [url, httpParams.toString()];
   }
 
-  handleLink(linkArray: Array<string>) {
-    this.router.navigateByUrl('search?' + linkArray[1]);
+  handleLink(linkArray: Array<string>, currentLocation: string) {
+    this.router.navigateByUrl('search?' + linkArray[1], currentLocation == '/search' ? { replaceUrl: true } : {});
     this.setShortUrl(linkArray[0] + '?' + linkArray[1]);
   }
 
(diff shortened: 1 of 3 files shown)


denis-yuen (approved):

src/app/search/state/search.service.ts
```diff
@@ -644,9 +644,9 @@ export class SearchService {
    */
   saveCurrentTabAndClear(index: number) {
     if (index === SearchService.WORKFLOWS_TAB_INDEX) {
-      this.router.navigateByUrl('search?entryType=workflows&searchMode=Files');
+      this.router.navigateByUrl('search?entryType=workflows&searchMode=files');
```
coverbeck, line 647: I looked and it's still not clear to me what searchMode actually does. It looks like it can be `files` or `description` (and I digress, but the type [here](https://github.com/dockstore/dockstore-ui2/blob/23d6e519f214dfaec4b9476a0ece63129a96a04a/src/app/shared/models/AdvancedSearchObject.ts#L22) should so indicate, although the type there wouldn't have fixed this anyway).

No action required on this comment.

coverbeck (approved):

kathy-t (approved):

PR #1594, opened 2022-08-16
Stop never-ending indeterminate progress bar

coverbeck:
**Description**
Stop never-ending indeterminate progress bar.

As long as there is an uncleared item, created by the
alertService.start, the progress bar will keep going.

**Issue**
#3975

### DIFF
--- a/src/app/workflow/permissions/permissions.component.ts
+++ b/src/app/workflow/permissions/permissions.component.ts
@@ -10,8 +10,8 @@ import { Dockstore } from '../../shared/dockstore.model';
 import { TokenSource } from '../../shared/enum/token-source.enum';
 import { TokenQuery } from '../../shared/state/token.query';
 import { Permission, Workflow, WorkflowsService } from '../../shared/swagger';
-
 import RoleEnum = Permission.RoleEnum;
+
 @Component({
   selector: 'app-permissions',
   templateUrl: './permissions.component.html',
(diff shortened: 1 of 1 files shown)


src/app/workflow/permissions/permissions.component.ts
```diff
@@ -76,6 +76,7 @@ export class PermissionsComponent implements OnInit {
         .addWorkflowPermission(this.workflow.full_workflow_path, { email: value, role: permission }, this.entryType === EntryType.Service)
         .subscribe(
           (userPermissions: Permission[]) => {
+            this.alertService.detailedSuccess('Permissions updated');
```
svonworl, line 79: Should we add similar calls to the alert service in the `remove()` method, also?

denis-yuen (commented):
build issue seems relevant (error comes from application code)

src/app/workflow/permissions/permissions.component.ts
coverbeck: Added

coverbeck:
> build issue seems relevant
> 
> > TypeError: The following error originated from your application code, not from Cypress.
> 
> > Cannot read properties of undefined (reading 'nativeElement')
> 
> > When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

Seems flaky, alas; it was marked so in CircleCI. Rebased and now it's passing.

src/app/workflow/permissions/permissions.component.ts
```diff
@@ -76,6 +78,7 @@ export class PermissionsComponent implements OnInit {
...
```
denis-yuen, line 81: out of a silly desire for consistency, adding permissions results in "Permissions updated" but removing permissions actually says that permissions were removed :wink:

denis-yuen (approved):

src/app/workflow/permissions/permissions.component.ts
coverbeck: I like consistency. :)

denis-yuen (approved):

svonworl (approved):

PR #1595, opened 2022-08-17
Create the Starred events box in the new-dashboard page

eileenzhh:
**Description**
Last widget box in the entire new dashboard, follows [design](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6022ab2d42df9f2c57c20982)

**Issue**
[SEAB-4649](https://ucsc-cgl.atlassian.net/browse/SEAB-4649)


eileenzhh:
image

denis-yuen (commented):
screenshots looking good, note build error

src/app/home-page/widget/starred-box/starred-box.component.scss
```diff
@@ -0,0 +1,28 @@
+@use '@angular/material' as mat;
+@import '/src/materialColorScheme.scss';
+
+.list-events {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  display: -webkit-box;
+  -webkit-line-clamp: 2; /* number of lines to show */
```
kathy-t, line 8: There's a global style that truncates lines: https://github.com/dockstore/dockstore-ui2/blob/develop/src/styles.scss#L1189. In this case, you would do something like `truncate-text-2` to only show 2 lines

src/app/home-page/widget/starred-box/starred-box.component.scss
```diff
@@ -0,0 +1,28 @@
...
+  border: solid 1px mat.get-color-from-palette($dockstore-app-gray, 4);
+  margin: 0.8rem;
+}
+
+img[alt='Org avatar'] {
+  width: 6rem;
+}
+
+img[alt='star'] {
+  width: 2rem;
```
kathy-t, line 26: Would this global style work for that image? https://github.com/dockstore/dockstore-ui2/blob/3df74a1d5c2764a09cb15d5f475e2be8300ad5b7/src/styles.scss#L395

src/app/home-page/widget/starred-box/starred-box.component.html
```diff
@@ -0,0 +1,188 @@
...
+        <div fxLayout="row" fxLayoutAlign="start center">
+          <img src="{{ event?.initiatorUser?.avatarUrl }}" alt="User avatar" />
+          <div [ngSwitch]="event.type">
+            <div *ngSwitchCase="EventType.ADDVERSIONTOENTRY" class="my-3">
+              <div class="list-events">
+                <strong>{{ event.initiatorUser?.username }} </strong> created the {{ event.version?.referenceType | lowercase }}
+                <strong>{{ event.version?.name }}</strong> in {{ event?.tool || event?.apptool ? 'tool' : 'workflow' }}
+                <a
+                  [routerLink]="
+                    event?.apptool
```
coverbeck, line 47: I think this would be better with a pipe, and line 54 as well. Both because it would be easier to read and test, but also because the logic is duplicated below. But it can be a followup ticket.

src/app/home-page/widget/starred-box/starred-box.component.ts
```diff
@@ -0,0 +1,75 @@
...
+    });
+    this.usersService.getStarredWorkflows().subscribe((starredWorkflows) => {
+      this.totalStarredWorkflows = starredWorkflows.length;
+    });
+    this.usersService.getStarredOrganizations().subscribe((starredOrganizations) => {
+      this.totalStarredOrganizations = starredOrganizations.length;
+    });
+  }
+
+  getMyEvents() {
```
coverbeck, line 53: Looks like it can be private.

FYI, on my requests to make this an the other field private: It may seem trivial and not worth it, and it won't change the way your code runs. I guess it is trivial, but adding `private` makes it easier to ready the code.

If I see `private getMyEvents()`, I know the method is only called from with this class. But because it is public, I had to double-check to see if you were invoking it from the HTML -- because if you were calling it from somewhere else, you would need to reinitialize `this.events` (`this.events` would keep growing).

src/app/home-page/widget/starred-box/starred-box.component.ts
```diff
@@ -0,0 +1,75 @@
...
+export class StarredBoxComponent extends Base implements OnInit {
+  Dockstore = Dockstore;
+  totalStarredWorkflows: number = 0;
+  totalStarredTools: number = 0;
+  totalStarredServices: number = 0;
+  totalStarredOrganizations: number = 0;
+  events: Array<Event> = [];
+  public isLoading = true;
+  EventType = Event.TypeEnum;
+  readonly supportedEventTypes = [
```
coverbeck, line 23: Can be private

src/app/home-page/widget/starred-box/starred-box.component.ts
```diff
@@ -0,0 +1,75 @@
...
+  getMyEvents() {
+    this.eventsService
+      .getEvents('ALL_STARRED')
+      .pipe(
+        finalize(() => (this.isLoading = false)),
+        takeUntil(this.ngUnsubscribe)
+      )
+      .subscribe(
+        (events) => {
+          events
```
coverbeck, line 62: I think we chatted about this on another PR, and you can leave it, but you could write this as this, which to me at least, is more readable (but others' opinions may vary):

```
this.events = events.filter((event) => this.supportedEventTypes.includes(event.type))
   .slice(0,4);
```

Also, I'm not sure if the 4 should be a 3 or a 4. :)

src/app/home-page/widget/starred-box/starred-box.component.html
eileenzhh: I will add this to the list of followup tickets I make before leaving :D

src/app/home-page/widget/starred-box/starred-box.component.html
```diff
@@ -0,0 +1,188 @@
+<mat-card class="h-100">
+  <div id="header-bar" fxLayout="row" fxLayoutAlign="space-between">
+    <mat-card-header fxLayout="row">
+      <mat-card-title>Starred</mat-card-title>
+    </mat-card-header>
+    <div id="total-starred" fxLayout="row" fxLayoutGap="1rem" class="size-small pb-3">
+      <div [ngPlural]="totalStarredWorkflows" id="starred-workflows" fxLayout="row" fxLayoutAlign="start center">
+        <ng-template ngPluralCase="=0"> <img src="../assets/svg/star-off.svg" alt="star" class="mr-1" /> 0 Workflows </ng-template>
```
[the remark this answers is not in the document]
coverbeck: I don't think you need this line at all -- line 10 should handle the 0 workflows case as well. Same comment for tools and workflows.

coverbeck (approved):
Would be nice to remove the redundant ngPluralCase, but not a blocker; please add it to the list of outstanding items if not.

kathy-t (approved):

eileenzhh:
> Would be nice to remove the redundant ngPluralCase, but not a blocker; please add it to the list of outstanding items if not.

The '0' case is unique for it's empty star icon vs the filled star for the other options, but I agree it is making the code a lot clunkier as it is repeated. It may do better with a for loop through different entrytypes, but I will make a ticket for it.

coverbeck:
> > Would be nice to remove the redundant ngPluralCase, but not a blocker; please add it to the list of outstanding items if not.
> 
> The '0' case is unique for it's empty star icon vs the filled star for the other options, but I agree it is making the code a lot clunkier as it is repeated. It may do better with a for loop through different entrytypes, but I will make a ticket for it.

Ah, I see. I was only looking at the text.

src/app/home-page/widget/starred-box/starred-box.component.html
```diff
@@ -0,0 +1,192 @@
...
+                  >{{ event.collection?.name }}
+                </a>
+                in organization
+                <a [routerLink]="'/organizations/' + event.organization?.name">{{ event.organization?.displayName }}</a>
+              </div>
+              <div class="subtitle">{{ event.dbCreateDate | date: 'medium' }}</div>
+            </div>
+            <div *ngSwitchCase="EventType.ADDTOCOLLECTION" class="my-3">
+              <div class="truncate-text-2">
+                <strong>{{ event.initiatorUser.username }} </strong> added the {{ event?.tool || event?.apptool ? 'tool' : 'workflow' }}
```
svonworl, line 78: In general, whenever a tag will change the formatting of the enclosed content, it's usually a good idea to avoid whitespace between the content and the opening and closing tags, so that the formatting doesn't spill out into the whitespace.  For example:
```
<strong>xyz</strong>
```
rather than
```
<strong>  xyz  </strong>
```

eileenzhh: ohh okay noted, i'll change it

PR #1596, opened 2022-08-17
looks like an import error

denis-yuen:
**Description**
Not very exciting, seems to have broken registering quay.io tools
Might be bad merge conflict, not sure why this gets past compiler before

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4706


fhembroff (approved):
Just ran this PR locally and the issue seems to be resolved. Thanks for fixing this.

svonworl:
Should we write a test to make sure it doesn't regress?

svonworl:
> Should we write a test to make sure it doesn't regress?

Oh, sorry, just noticed the ticket says that a test is failing because of this.  If that's so, no need to write another test...

svonworl (approved):

PR #1598, opened 2022-08-18
add a custom 404 page instead of redirecting

denis-yuen:
**Description**
Guess we're grown up enough now to need a specific 404 page. 
Added some possible reasons for landing on it. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1974


src/app/pagenotfound/pagenotfound.component.html
```diff
@@ -0,0 +1,21 @@
+<div class="overflow-hidden">
+  <div fxLayout="column" fxLayoutAlign="space-around" fxLayoutGap="4rem" class="mb-5 w-100">
+    <div class="about-dockstore">
+      <div fxLayout="row" fxLayout.lt-md="column" LayoutAlign="space-between center" fxLayoutGap="1rem" class="container">
+        <div fxLayout="column" fxLayoutAlign="space-between start" fxLayoutGap="2rem">
+          <h1>Page not found</h1>
+          <h3>
+            Don't panic. You seem to have taken a wrong turn. In some cases, this may mean that you have bookmarked a tool or workflow that
+            has been unpublished by it's author. Alternatively, maybe this was a page that you accessed while logged-in but you haven't
```
coverbeck, line 9: its

src/app/pagenotfound/pagenotfound.component.scss
```diff
@@ -0,0 +1,44 @@
+@use '@angular/material' as mat;
+@import '/src/materialColorScheme.scss';
+
+.about-dockstore {
```
coverbeck, line 4: Slightly wary of all this component-specific styling, but it is a unique page unlike any other, so... I always defer to Kathy on this; I should get more up to speed.

coverbeck (approved):

src/app/pagenotfound/pagenotfound.component.html
```diff
@@ -0,0 +1,21 @@
+<div class="overflow-hidden">
+  <div fxLayout="column" fxLayoutAlign="space-around" fxLayoutGap="4rem" class="mb-5 w-100">
+    <div class="about-dockstore">
+      <div fxLayout="row" fxLayout.lt-md="column" LayoutAlign="space-between center" fxLayoutGap="1rem" class="container">
+        <div fxLayout="column" fxLayoutAlign="space-between start" fxLayoutGap="2rem">
+          <h1>Page not found</h1>
+          <h3>
```
svonworl, line 7: Unsure about this use of `h3`... would this block of text be more appropriate as a styled `p`?

denis-yuen: This parallels the "about" page that I stole the styling from, 
I think it looks ok, but have no real preference.

david4096 (approved):

src/app/pagenotfound/pagenotfound.component.scss
```diff
@@ -0,0 +1,44 @@
...
+  font-weight: 500;
+}
+
+h2 {
+  font-weight: inherit;
+  font-size: 2.5rem;
+}
+
+hr {
+  width: 100%;
```
kathy-t, line 30: Can use `w-100` instead of creating this class: https://github.com/dockstore/dockstore-ui2/blob/develop/src/bootstrap4.scss#L33

src/app/pagenotfound/pagenotfound.component.scss
```diff
@@ -0,0 +1,44 @@
...
+@import '/src/materialColorScheme.scss';
+
+.about-dockstore {
+  margin-top: 6rem;
+  background-image: url(../../assets/images/about-page/city.png);
+  background-repeat: repeat-x;
+  background-position-y: bottom;
+}
+
+.truck-img {
```
kathy-t, line 11: Seems like these are duplicated from the about page scss? If so, I would be inclined to move them to the global sheet with an explanation of what these are for (text page with truck background, for example)

kathy-t (commented):
Maybe just me, but I find it a little hard to read with the words overlapping with the buildings, especially the black font. I think it would be easier to read if all the text was in the white area. No need to change if others don't feel the same
image

denis-yuen:
Not sure if there is a modern equivalent to the multiple `<br>`s
Screenshot from 2022-08-23 10-44-36

coverbeck (approved):

aofarrel:
@denis-yuen Late to the party, but it appears that a standard browser zoom-in will not cause the text to overflow onto the image, which is great! ✅ Thanks for doing this.

PR #1599, opened 2022-08-18
Fix second revert not working in hosted workflows

coverbeck:
**Description**
The problem was that a deep copy was made initially, so the first
revert worked, but the first revert used a shallow copy, so the
original data ended up being modified.

Did a little extra cleanup, removing some `any` types, and specifying
the correct imports (openapi vs swagger).

Tried to create a integration test, but ACE Editor and Cypress is not
a happy combination; seems safe/obvious enough, and added a unit test.

**Issue**
dockstore/dockstore#3800


svonworl (approved):

denis-yuen (approved):
FYI, issue link not very linky

kathy-t (approved):

PR #1601, opened 2022-08-19
Added in test for hiding version

fhembroff:
**Description**
This PR adds a test for hiding and un-hiding workflow versions. This PR is incomplete and regrettably I won't be able to see it through, so I am leaving some comments.

**Issue**
[DOCK-2202](https://ucsc-cgl.atlassian.net/browse/DOCK-2202)
https://github.com/dockstore/dockstore/issues/5017

### DIFF
--- a/cypress/integration/smokeTests/sharedTests/auth-tests.ts
+++ b/cypress/integration/smokeTests/sharedTests/auth-tests.ts
@@ -256,6 +256,26 @@ function testWorkflow(registry: string, repo: string, name: string) {
       cy.contains('button', 'Restub').click();
       cy.contains('button', 'Publish').should('be.disabled');
     });
+    it('hide and un-hide a version', () => {
+      goToTab('Versions');
+      // hide
+      cy.get('[data-cy=refreshButton]').click();
+      cy.get('[data-cy=versionRow]').last().contains('button', 'Actions').should('be.visible').click();
+      cy.contains('button', 'Edit').click();
+      cy.contains('div', 'Hidden:').within(() => {
+        cy.get('[name=checkbox]').click();
+      });
+      cy.contains('button', 'Save Changes').click();
+      cy.get('[data-cy=hidden]').should('have.length', 1);
+      // un-hide
+      cy.get('[data-cy=versionRow]').last().contains('button', 'Actions').should('be.visible').click();
+      cy.contains('button', 'Edit').click();
+      cy.contains('div', 'Hidden:').within(() => {
+        cy.get('[name=checkbox]').click();
+      });
+      cy.contains('button', 'Save Changes').click();
+      cy.get('[data-cy=hidden]').should('not.exist');
+    });
   });
 }
 
@@ -300,7 +320,7 @@ function testCollection(org: string, collection: string, registry: string, repo:
     deleteTool();
   });
 }
-
+//
 testCollection(collectionTuple[0], collectionTuple[1], toolTuple[0], toolTuple[1], toolTuple[2]);
 testTool(toolTuple[0], toolTuple[1], toolTuple[2]);
 testWorkflow(workflowTuple[0], workflowTuple[1], workflowTuple[2]);
--- a/cypress/integration/smokeTests/sharedTests/auth-tests.ts
+++ b/cypress/integration/smokeTests/sharedTests/auth-tests.ts
@@ -170,6 +170,27 @@ function registerToolOnDockstore(repo: string, name: string) {
   });
 }
 
+function toggleHiddenToolVersion() {
+  cy.contains('button', 'Actions').should('be.visible');
+  cy.contains('td', 'Actions').first().click();
+  cy.contains('button', 'Edit').click();
+  cy.contains('div', 'Hidden:').within(() => {
+    cy.get('[name=checkbox]').click();
+  });
+  cy.contains('button', 'Save Changes').click();
+}
+
+function toggleHiddenWorkflowVersion() {
+  cy.get('[data-cy=refreshButton]').click();
+  cy.get('[data-cy=versionRow]').last().contains('button', 'Actions').should('be.visible').click();
+  cy.contains('button', 'Edit').click();
+  // TODO: Use [data-cy=hiddenCheck] -- do after 1.14 deployed
+  cy.contains('div', 'Hidden:').within(() => {
+    cy.get('[name=checkbox]').click();
+  });
+  cy.contains('button', 'Save Changes').click();
+}
+
 function testTool(registry: string, repo: string, name: string) {
   describe('Register, publish, unpublish, and delete a tool', () => {
     registerQuayTool(repo, name);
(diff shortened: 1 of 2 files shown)


cypress/integration/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -250,6 +250,26 @@ function testWorkflow(registry: string, repo: string, name: string) {
       cy.contains('button', 'Restub').click();
       cy.contains('button', 'Publish').should('be.disabled');
     });
+    it('hide and un-hide a version', () => {
```
fhembroff, line 253: It would be great to create a function for these hiding and un-hiding as the workflow test and the tool test are near identical. However, when I tried moving this into a function I ran into problems.

cypress/integration/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -250,6 +250,26 @@ function testWorkflow(registry: string, repo: string, name: string) {
...
+      cy.contains('button', 'Save Changes').click();
+      cy.get('[data-cy=hidden]').should('have.length', 1);
+      // un-hide
+      cy.get('[data-cy=versionRow]').last().contains('button', 'Actions').should('be.visible').click();
+      cy.contains('button', 'Edit').click();
+      cy.contains('div', 'Hidden:').within(() => {
+        cy.get('[name=checkbox]').click();
+      });
+      cy.contains('button', 'Save Changes').click();
+      cy.get('[data-cy=hidden]').should('not.exist');
```
fhembroff, line 271: The `data-cy` tag in the hidden box in the workflow options is `hidden`, whereas in the tool options it is `hiddenCheck`. I'm thinking if these were the same it would make it easier to write a function to do the test (and make sense?).

fhembroff: The box I am referring to when I say "hidden box" is the one highlighted in this image,
Screenshot from 2022-08-19 17-22-52

cypress/integration/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -250,6 +250,26 @@ function testWorkflow(registry: string, repo: string, name: string) {
...
+      cy.get('[data-cy=refreshButton]').click();
+      cy.get('[data-cy=versionRow]').last().contains('button', 'Actions').should('be.visible').click();
+      cy.contains('button', 'Edit').click();
+      cy.contains('div', 'Hidden:').within(() => {
+        cy.get('[name=checkbox]').click();
+      });
+      cy.contains('button', 'Save Changes').click();
+      cy.get('[data-cy=hidden]').should('have.length', 1);
+      // un-hide
+      cy.get('[data-cy=versionRow]').last().contains('button', 'Actions').should('be.visible').click();
```
fhembroff, line 265: This is required as the first option that would be selected is the default version (which cannot be hidden). It is also for this version that I didn't  just use `cy.contains('button', 'Actions')` (as it only grabs the first match). 

It would be good to come up with a way to match the first non-default option (as the default could change, breaking this test).

fhembroff: The tool test just uses `cy.contains('button', 'Actions') ` as the first occurrence is not the default. It would be good to correct this as well.

fhembroff:
This https://github.com/dockstore/dockstore-ui2/blob/d5a2584eda86571730d48bb75fb9b7a7ea7adc67/cypress/integration/smokeTests/sharedTests/auth-tests.ts#L183 only tests hiding, it does not test un-hiding.

cypress/integration/smokeTests/sharedTests/auth-tests.ts
coverbeck: They seem quite a bit different, unfortunately; going to leave the same for now.

coverbeck:
> This
> 
> https://github.com/dockstore/dockstore-ui2/blob/d5a2584eda86571730d48bb75fb9b7a7ea7adc67/cypress/integration/smokeTests/sharedTests/auth-tests.ts#L183
> only tests hiding, it does not test un-hiding. This should be corrected.

Updated to  test un-hiding.

cypress/integration/smokeTests/sharedTests/auth-tests.ts
coverbeck: Punting on this for now.

cypress/integration/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -251,6 +257,25 @@ function testWorkflow(registry: string, repo: string, name: string) {
       cy.contains('button', 'Restub').click();
       cy.contains('button', 'Publish').should('be.disabled');
     });
+    it('hide and un-hide a version', () => {
+      function toggleHidden() {
```
svonworl, line 261: Nitpicky: would be great if both versions of `toggleHidden` were defined at the same lexical level.

cypress/integration/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -251,6 +257,25 @@ function testWorkflow(registry: string, repo: string, name: string) {
...
+        });
+        cy.contains('button', 'Save Changes').click();
+      }
+      goToTab('Versions');
+      // hide
+      toggleHidden();
+      cy.get('[data-cy=hidden]').should('have.length', 1);
+      // un-hide
+      toggleHidden();
+      cy.get('[data-cy=hidden]').should('not.exist');
```
svonworl, line 277: Flynn was suggesting that we set the same value of `data-cy` (`hiddenCheck` is probably good) for both the tools and workflows pages, which sounds like a good idea.

cypress/integration/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -182,17 +182,23 @@ function testTool(registry: string, repo: string, name: string) {
...
       cy.contains('button', 'Edit').click();
       cy.contains('div', 'Hidden:').within(() => {
         cy.get('[name=checkbox]').click();
       });
       cy.contains('button', 'Save Changes').click();
+    }
+    registerQuayTool(repo, name);
+    it('hide a version', () => {
+      goToTab('Versions');
+      toggleHidden();
```
svonworl, line 197: Since it's easy, we could add a test for the precondition (the hidden check shouldn't exist) before the first toggle (here and in the workflow version below), thus foreclosing some possible avenues of later confusion if things go wrong...

cypress/integration/smokeTests/sharedTests/auth-tests.ts
coverbeck: I had added the `data-cy=hiddenCheck`, but I don't want to use it here yet because it will not be deployed to staging/prod. I'll create a followup ticket.

cypress/integration/smokeTests/sharedTests/auth-tests.ts
coverbeck: Done.

cypress/integration/smokeTests/sharedTests/auth-tests.ts
coverbeck: Looked at it briefly, it's not as trivial as it seems to do this. I'd like to get this in to automate a little bit more of the 1.13 release, so going to punt.

denis-yuen (approved):
check checklist items

svonworl (approved):

kathy-t (approved):

PR #1602, opened 2022-08-19
Cancel workflow edit now reverts

coverbeck:
**Description**
Cancel workflow edit now reverts

It was sort of confusing, code-wise. See [this comment](https://github.com/dockstore/dockstore/issues/4592#issuecomment-1221073554) for details.

TLDR; There was a workflow object in InfoTabService that was being reverted, but the object wasn't being used by the component, so no reversion happened in the UI. So I removed the object entirely from the service to avoid confusion, and I do the revert in the component.

**Issue**
dockstore/dockstore#4592

### DIFF
--- a/cypress/integration/group2/myworkflows.ts
+++ b/cypress/integration/group2/myworkflows.ts
@@ -13,7 +13,6 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-import { contains } from 'cypress/types/jquery';
 import { Repository } from '../../../src/app/shared/openapi/model/repository';
 import { goToTab, isActiveTab, resetDB, setTokenUserViewPort } from '../../support/commands';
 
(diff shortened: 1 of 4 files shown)


svonworl (approved):

cypress/integration/group2/myworkflows.ts
```diff
@@ -108,24 +107,37 @@ describe('Dockstore my workflows', () => {
       cy.contains('1 – 2 of 2');
       cy.contains('Close').click();
     });
-    it('Should contain the extended properties and be able to edit the info tab', () => {
+    it.only('Should contain the extended properties and be able to edit the info tab', () => {
```
kathy-t, line 110: Should remove `.only` so other tests aren't ignored

coverbeck: Ugh, thanks for catching that.

kathy-t (approved):

denis-yuen (approved):

PR #1603, opened 2022-08-23
Show more generic message when updating workflow

coverbeck:
**Description**

When updating a workflow, we can't do it with one API request, we need to do it at least 2; one request for hidden and/or workflow path, then 1 or 2 more requests for adding/removing test parameter files. We were displaying the updating test parameter file success message last, which is what the user saw. Also, we displayed the updating test parameter files message, even if no request was made to update the test parameters.

The user shouldn't care about how many requests we are making underneath the hood, so now it's just a generic "Saving workflow version succeeded" message. I could have broken it apart, e.g., "Hiding version and adding one test parameter file", but it didn't seem worth the effort.

Also, if a refresh happens, a refresh message will be the last thing to display, but, oh well - I think you'd want to know that.

Also:

* Got rid of an `any`
* Changed the dialog title from `Edit Workfow` to `Edit Workflow Version`, because that's what it is doing
* Noticed a compiler error in WebStorm for the `trackByIndex` function (it doesn't exist), so just removed `trackBy`.

**Issue**
dockstore/dockstore#5014

### DIFF
--- a/src/app/workflow/version-modal/version-modal.service.ts
+++ b/src/app/workflow/version-modal/version-modal.service.ts
@@ -20,11 +20,12 @@ import { concatMap } from 'rxjs/operators';
 import { AlertService } from '../../shared/alert/state/alert.service';
 import { RefreshService } from '../../shared/refresh.service';
 import { WorkflowQuery } from '../../shared/state/workflow.query';
+import { WorkflowService } from '../../shared/state/workflow.service';
+import { AppTool, BioWorkflow, DockstoreTool, Service, Workflow } from '../../shared/swagger';
 import { WorkflowsService } from '../../shared/swagger/api/workflows.service';
 import { SourceFile } from '../../shared/swagger/model/sourceFile';
 import { WorkflowVersion } from '../../shared/swagger/model/workflowVersion';
-import { WorkflowService } from '../../shared/state/workflow.service';
-import { AppTool, BioWorkflow, Service } from '../../shared/swagger';
+import ModeEnum = DockstoreTool.ModeEnum;
 
 @Injectable()
 export class VersionModalService {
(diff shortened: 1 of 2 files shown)


src/app/workflow/version-modal/version-modal.service.ts
```diff
@@ -64,28 +65,28 @@ export class VersionModalService {
...
+          this.alertService.detailedSuccess();
+          this.matDialog.closeAll();
+        } else {
           this.modifyTestParameterFiles(workflowVersion, originalTestParameterFilePaths, newTestParameterFiles).subscribe(
-            (success) => {
+            (files) => {
               // Checks if there was a test parameter file added/removed
-              // The this.modifyTestParameterFiles function returns a {} observable if nothing was done, this is a way of checking for it
-              if (!(Object.keys(success).length === 0 && success.constructor === Object)) {
+              if (files !== null) {
```
svonworl, line 89: If I'm reading the code right, in all cases, even when `modifyTestParameterFiles` returns the value of `observableOf(null)`, `files` is an `Observable` object, so this conditional will always evaluate to true.  But that's ok, I propose we always refresh the version after we try to update it, because the overhead is low, and we'll avoid the case where a bug prevents us from seeing the updated version.

coverbeck: `files` is `SourceFile[] | null` because we're in the subscribe. "Proof" of a sorts is this WebStorm screenshot:

Screen Shot 2022-08-23 at 3 56 30 PM

Plus I had run it in the debugger as well, and was able to verify that.

I think the idea is that just toggling `hidden` shouldn't refresh the whole workflow/version, as that's sort of an expensive operation.

svonworl: Yes, you are totally correct on all counts.  Disregard.

svonworl (approved):

denis-yuen (approved):

kathy-t (approved):

PR #1604, opened 2022-08-24
Add https to custom multi-cloud launch URLs without a protocol

kathy-t:
**Description**
This PR:
- Adds `https://` to URLs that don't specify a protocol so `usegalaxy.ca` becomes `https://usegalaxy.ca`.
- Validates that the URL uses https. If it doesn't, the URL is considered invalid and a warning message is displayed.
- Removes the `href` attribute for the `Launch` button if the URL is invalid so the user isn't directed to that URL.
- Performs a truthy check on `url` so we only add new user cloud instances to the DB if the url was created successfully. Without this check, we could get an `undefined` error for invalid URLs.

**Issue**
[SEAB-3965](https://ucsc-cgl.atlassian.net/browse/SEAB-3965)

### DIFF
--- a/src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.spec.ts
+++ b/src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.spec.ts
@@ -33,7 +33,7 @@ describe('MultiCloudLaunchComponent', () => {
   it('Should construct url with https protocol', () => {
     expect(component.constructUrlWithHttpsProtocol('usegalaxy.ca')).toBe('https://usegalaxy.ca');
     expect(component.constructUrlWithHttpsProtocol('subdomain.usegalaxy.ca')).toBe('https://subdomain.usegalaxy.ca');
-    expect(() => component.constructUrlWithHttpsProtocol('http://foobar.com')).toThrow();
+    expect(() => component.constructUrlWithHttpsProtocol('http://foobar.com')).toThrow(); // NOSONAR suppress insecure http protocol analysis
     expect(() => component.constructUrlWithHttpsProtocol('foobar://foobar.com')).toThrow();
   });
 });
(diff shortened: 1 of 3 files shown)


src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.spec.ts
```diff
@@ -29,4 +29,11 @@ describe('MultiCloudLaunchComponent', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  it('Should construct url with https protocol', () => {
+    expect(component.constructUrlWithHttpsProtocol('usegalaxy.ca')).toBe('https://usegalaxy.ca');
+    expect(component.constructUrlWithHttpsProtocol('subdomain.usegalaxy.ca')).toBe('https://subdomain.usegalaxy.ca');
+    expect(() => component.constructUrlWithHttpsProtocol('http://foobar.com')).toThrow();
```
denis-yuen, line 36: Amused that this triggers a security alert (alert can be ignored of course for other reviewers).
May be possible to put in a comment to have sonarcloud ignore this line

denis-yuen (approved):

src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.ts
```diff
@@ -141,32 +142,52 @@ export class MultiCloudLaunchComponent extends Base implements OnInit {
...
+          },
+          (error: HttpErrorResponse) => {
+            // It's okay for the save to the db to fail for now. At this step, we only care about what URL's users
+            // are trying to use. Probably most of these failures will be the for violating the uniqueness constraint
+            console.log(error.message);
+          }
+        );
+      } else {
+        // Remove link to prevent navigating to invalid URL
+        document.getElementById('multiCloudLaunchButton').removeAttribute('href');
```
coverbeck, line 167: I didn't try running your branch and don't 100% follow the code flow, but does this ever need to get added back? What if the user subsequently specifies a good/valid url, if that makes sense?

coverbeck (approved):

svonworl (approved):

src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.ts
kathy-t: I don't think we need to manually add it back. `this.launchWith.url` is binded to `href` [here](https://github.com/dockstore/dockstore-ui2/blob/08b5b9f7ddf4398a728b6b454ea0390c96bc37b8/src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.html#L62) so if a user specifies a different URL (valid or invalid) then the `href` property is updated to the new URL

PR #1605, opened 2022-08-24
Update smoke tests

coverbeck:
**Description**
Get smoke tests working.

1. On prod and staging, we would sometimes run into the search self-DOS. This would cause the test to fail and our WAF rate limit to be hit. Removed the test that led to that by creating staging and prod-specific files that comment out the test that causes the self-DOS. Created a followup ticket, [SEAB-4767](https://ucsc-cgl.atlassian.net/browse/SEAB-4767), to restore the tests once 1.13 is deployed to staging/prod.
2. In looking at this, noticed we were installing chrome and chromedriver twice (they're installed in `setup_nightly_tests`, so removed.
3. Another issue was introduced by the GitHub app logs fix -- before, when you deleted the user's last tool, the `There are currently no tools registered under this account` message would display, now it doesn't. Created an issue to follow up on that, dockstore/dockstore#5105.

Smoke tests passed although the staging auth one is still flaky :(.

**Issue**
dockstore/dockstore#4588


cypress/integration/smokeTests/prodOnly/search.ts
```diff
@@ -0,0 +1,64 @@
...
+      cy.url().should('not.include', '/search?descriptorType=WDL');
+      cy.contains('the Language is WDL').should('not.exist');
+      cy.go('forward');
+      cy.url().should('include', '/search?descriptorType=WDL');
+      cy.contains('the Language is WDL').should('exist');
+    });
+    it('should be able to use basic search box and have suggestions', () => {
+      cy.get('[data-cy=basic-search]').type('dockstore_i{enter}');
+      cy.contains(' Sorry, no matches found for dockstore_i');
+      cy.contains(/Do[ ]you[ ]mean:[ ].+/);
```
svonworl, line 34: The code robot has a valid beef about these, I think.  It says we should replace the character class with the actual character, for example:
```
cy.contains(/Do you mean: .+/);
```

svonworl (approved):

cypress/integration/smokeTests/prodOnly/search.ts
coverbeck: It's not clear from my PR description, but I took "cypress/integration/smokeTests/**sharedTests**/search.ts", copied it to "cypress/integration/smokeTests/**stagingOnly**/search.ts" and "cypress/integration/smokeTests/**prodOnly**/search.ts", then commented out lines 43-55 in the copied versions. Lines 43-55 are what are causing the self-DOSsing.

This is the pattern for running dev/staging/prod specific tests.

Once 1.13 is deployed I will delete this file and the `stagingOnly` copy as well.

TLDR; it's a copy of existing code, and the copy is going to get deleted, so I don't think it's worth it for this PR. But I'll create a followup ticket for the "core" test.

coverbeck: On further reflection, I think I'll pass on creating a ticket. The next person who edits the core file can deal with it then, but it's working for now. Feel free to push back.

denis-yuen (approved):

kathy-t (approved):

PR #1606, opened 2022-08-27
Show user-friendly Add Tool/Workflow button

coverbeck:
**Description**

The problem was that we now display your GitHub orgs even if you have no tools/workflows registered (so you can see the app logs). Now check also if there is a least one registered entry in one of the GitHub orgs.

Also restored commented out smoke test that this bug had broken.

**Issue**
dockstore/dockstore#5105

### DIFF
--- a/cypress/integration/group3/mytools.ts
+++ b/cypress/integration/group3/mytools.ts
@@ -16,23 +16,6 @@
 import { DockstoreTool } from '../../../src/app/shared/openapi';
 import { goToTab, resetDB, setTokenUserViewPort, setTokenUserViewPortCurator, typeInInput } from '../../support/commands';
 
-describe('Should handle no tools correctly', () => {
-  resetDB();
-  setTokenUserViewPortCurator(); // Curator has no tools
-  beforeEach(() => {
-    cy.server();
-    cy.route({
-      method: 'GET',
-      url: /github.com\/organizations/,
-      response: ['dockstore'],
-    });
-  });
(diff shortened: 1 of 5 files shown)


kathy-t (approved):

cypress/integration/group3/mytools.ts
```diff
@@ -14,7 +14,24 @@
...
+    cy.server();
+    cy.route({
+      method: 'GET',
+      url: /github.com\/organizations/,
+      response: ['dockstore'],
+    });
+  });
+  it('My tools should prompt to register a tool', () => {
+    cy.visit('/my-tools');
+    cy.contains('Add Tool');
```
svonworl, line 32: Apologies in advance for being borderline nitpicky, and I vaguely recall having this convo before, but if not:
We use 'Register Workflow' in one place, and 'Add Tool' here.  For the sake of consistency, maybe we pick one verb?

svonworl (approved):

denis-yuen (approved):

cypress/integration/group3/mytools.ts
coverbeck: I noticed the difference for the first time in doing this PR, debated whether to change it, and had decided to leave it, just to keep the PR focused. But looking again, the mocks for the new screen use "Register", so I'll go with that.

coverbeck: On further reflection, I'm going to leave it. It has a small ripple effect where I don't want risk introducing something.

PR #1608, opened 2022-09-01
Fix flaky group test 1

coverbeck:
**Description**
The Twitter "placeholder" div doesn't exist in the new dashboard, the
lack of which would frequently cause a native element exception and a
Cypress failure.

Ideally there would be another component that's a container for the old
and new dashboards, instead of the way it is now, where this is both
the old dashboard and a container for the new dashboard. But I may want
to bring this into the release branch and we're eventually getting rid
of the old dashboard anyway, so I went with a more minimal approach.

**Issue**
Repurposed https://ucsc-cgl.atlassian.net/browse/SEAB-659

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -443,53 +443,53 @@ workflows:
             - lint_license_unit_test_coverage
           context:
             - dockerhub
-#      - integration_test_2:
-#          filters:
-#            tags:
-#              only: /.*/
-#          requires:
-#            - lint_license_unit_test_coverage
-#          context:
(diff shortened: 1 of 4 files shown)


coverbeck:
Going to do this one instead of #1610, as in addition to upgrading the artifacts, if fixes the flaky failing integration test there.

The accessibility test is failing because it's trying to compare against a develop build, but that fails because of the expired artifacts.

kathy-t (approved):

svonworl (approved):

PR #1617, opened 2022-09-21
DOCK-2114: Fix Duplicate Link to Entry Discuss Topic

svonworl:
**Description**
Sometimes, the UI includes duplicate links to the discuss topic (post) from a workflow page.  The source of the issue seems to be this code:
https://github.com/dockstore/dockstore-ui2/blob/4d77cfdea42ecd7d696b0b32eddccadde743f2eb/src/app/workflow/workflow.component.ts#L174

In some situations, it appears that the handler, which calls `discourseHelper` to add the link, is being invoked more than once.

On staging, I can consistently trigger the problem by going to the "private" page for a published hosted workflow, then clicking "View public page".  On the public page, observe the duplicate discuss links at bottom.

This PR adds a check to prevent the discuss `script` tag from being added multiple times.  The assumption here is that the duplicate links always have the same `topicId`, which, based on my understanding of how the code works, I believe to be true.  If not, this PR won't work right.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2114
dockstore/dockstore#4812

### DIFF
--- a/src/app/shared/entry.ts
+++ b/src/app/shared/entry.ts
@@ -461,6 +461,9 @@ export abstract class Entry implements OnDestroy {
    * @param topicId The ID of the topic on discourse
    */
   discourseHelper(topicId: number): void {
+    if (document.getElementById('discourse-embed-script') !== null) {
+      return;
+    }
     const element = document.getElementById('discourse-embed-frame');
     if (element !== null) {
       element.remove();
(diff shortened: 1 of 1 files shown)


coverbeck (commented):
We can only verify this works in staging, right? Because staging and prod are the only envs hooked up to Discourse?

What we could do is update staging with the UI from your branch, give it a try, then revert staging to match prod again.

Looking at the code, I'm not convinced this will work, and I'm concerned that it might end up showing old links (although I think it's unlikely). Before it was removing the parent element, now it's checking for a child element of that parent element and not doing anything if the child is present.

I could be wrong, so instead of speculating, maybe we should give your branch a shot in staging.

PR #1618, opened 2022-09-21
DOCK-2114: Fix Duplicate Links to Entry Discuss Topic

svonworl:
**Description**
Sometimes, the UI includes duplicate links to the discuss topic (post) from a workflow page. The source of the issue seems to be this code:
https://github.com/dockstore/dockstore-ui2/blob/4d77cfdea42ecd7d696b0b32eddccadde743f2eb/src/app/workflow/workflow.component.ts#L174

In some situations, it appears that the handler, which calls `discourseHelper` to add the link, is being invoked more than once for a given view of a workflow.  Alas, the invoked discourse embed script adds a link, rather than replacing the existing link if it exists, resulting in duplicates.
 
There is code in `discourseHelper` that attempts to remove the previous link.  However, multiple scripts may be in flight at once, and at removal time, a script from a previous call may not have generated its link yet (or even be completely downloaded), resulting in duplicates. 

On staging, I can consistently trigger the problem by going to the "private" page for a published hosted workflow, then clicking "View public page". On the public page, observe the duplicate discuss links at bottom.

One approach fiddles with the observable/akita code to prevent multiple the above-mentioned handler from being invoked multiple times.  But that feels dangerous, so I took a different approach.

This PR "deduplicates" the links by adding logic to `discourseHelper` to ensure that consecutive calls with the same `topicId` only fire the discourse embed script a single time, on the first call.  The underlying mechanism is to store the passed `topicId`, compare it with the next value, and do nothing if they match.  The element that the `topicId` is stored in is part of the workflow component, so it will disappear when the view is regenerated, causing the embed script to be always be invoked on the next call in that case, too.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2114
https://github.com/dockstore/dockstore/issues/4812

### DIFF
--- a/src/app/shared/entry.ts
+++ b/src/app/shared/entry.ts
@@ -461,12 +461,15 @@ export abstract class Entry implements OnDestroy {
    * @param topicId The ID of the topic on discourse
    */
   discourseHelper(topicId: number): void {
-    // If the #discourse-comments element does not exist, or if the
-    // previous topicId is the same as the requested topic id, no dothing.
+    // If the #discourse-comments element does not exist, or if the previous topicId
+    // (from the previous call) is the same as the requested topicId, no dothing.
     const comments = document.getElementById('discourse-comments');
     if (!comments || (<any>comments).topicId === topicId) {
+      console.log('no action');
       return;
     }
+    // Store the topicId so we can check it the next time this function is called.
+    // The stored topicId will persist until the view is regenerated.
     (<any>comments).topicId = topicId;
 
     const element = document.getElementById('discourse-embed-frame');
--- a/src/app/shared/entry.ts
+++ b/src/app/shared/entry.ts
@@ -461,9 +461,14 @@ export abstract class Entry implements OnDestroy {
    * @param topicId The ID of the topic on discourse
    */
   discourseHelper(topicId: number): void {
-    if (document.getElementById('discourse-embed-script') !== null) {
+    // If the #discourse-comments element does not exist, or if the
+    // previous topicId is the same as the requested topic id, no dothing.
+    const comments = document.getElementById('discourse-comments');
+    if (!comments || (<any>comments).topicId === topicId) {
       return;
     }
+    (<any>comments).topicId = topicId;
+
     const element = document.getElementById('discourse-embed-frame');
     if (element !== null) {
       element.remove();
@@ -474,7 +479,6 @@ export abstract class Entry implements OnDestroy {
     };
     (function () {
       const d = document.createElement('script');
-      d.id = 'discourse-embed-script';
       d.type = 'text/javascript';
       d.async = true;
       d.src = (<any>window).DiscourseEmbed.discourseUrl + '/javascripts/embed.js';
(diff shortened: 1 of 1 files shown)


src/app/shared/entry.ts
```diff
@@ -461,6 +461,16 @@ export abstract class Entry implements OnDestroy {
    * @param topicId The ID of the topic on discourse
    */
   discourseHelper(topicId: number): void {
+    // If the #discourse-comments element does not exist, or if the previous topicId
+    // (from the previous call) is the same as the requested topicId, no dothing.
+    const comments = document.getElementById('discourse-comments');
+    if (!comments || (<any>comments).topicId === topicId) {
```
coverbeck, line 467: The more TypeScript-y way would be to use `setAttribute` and `getAttribute`. It's our goal one day to get rid of `any` and to avoid introducing new usages of it in the meantime. Although we use it so much, I don't know if we ever will.

If you do change it, note that `setAttribute` converts the value to a string :(

src/app/shared/entry.ts
```diff
@@ -461,6 +461,16 @@ export abstract class Entry implements OnDestroy {
    * @param topicId The ID of the topic on discourse
    */
   discourseHelper(topicId: number): void {
+    // If the #discourse-comments element does not exist, or if the previous topicId
```
coverbeck, line 464: > If the #discourse-comments element 

Suggest something like "If the #discourse-comments placeholder element..." -- I had to go look at the code to understand this, and was initially confused why we you are checking for existence of an element I assumed that the Discourse JavaScript creates. And then I realized it's defined in our HTML.

coverbeck (approved):

src/app/shared/entry.ts
svonworl: I added a sentence to the front of this comment that explains the purpose and provenance of the #discourse-comments element.

src/app/shared/entry.ts
svonworl: I understand the desire to get of `any`, but in this case, I literally want to read/write a property from/to the raw Javascript object, rather than set an attribute.

src/app/shared/entry.ts
```diff
@@ -461,6 +461,18 @@ export abstract class Entry implements OnDestroy {
    * @param topicId The ID of the topic on discourse
    */
   discourseHelper(topicId: number): void {
+    // The #discourse-comments element, included in the entry component's html template,
+    // marks the location where the discourse embed script will add the discourse link.
+    // If the #discourse-comments element does not exist, or if the previous topicId
+    // (from the previous call) is the same as the requested topicId, no dothing.
```
kathy-t, line 467: Typo: `no dothing` :slightly_smiling_face:

kathy-t (approved):

src/app/shared/entry.ts
svonworl: That's my trial balloon of "Olde English" comments, hackernews sez they're gonna be the next big thing...

PR #1620, opened 2022-09-23
DOCK-2061: Fix Weird Label Editing Behavior

svonworl:
**Description**
This PR changes the entry label editing interface so that its behavior is more natural.  Both issues mentioned in the ticket are fixed.

Part of problem stemmed from the `matChipInputAddOnBlur` = `true` setting, which caused a new chip (label) to be added when the chip list lost focus.  This setting, when there were no labels, for reasons that I was not able to determine, prevented the "save"/"cancel" button handler from being fired when the button was clicked.  It would also trigger unexpected behavior, like adding a partially-typed label when, for example, you had to click on another window to look up how to spell the rest of it.

I also fixed some code that was likely a bug - `undefined` should probably not be quoted in that `if` statement.

Please try this out, because if we hotfix, it's important to user test as much as possible.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2061
https://github.com/dockstore/dockstore/issues/4719

### DIFF
--- a/src/app/workflow/workflow.component.ts
+++ b/src/app/workflow/workflow.component.ts
@@ -485,6 +485,7 @@ export class WorkflowComponent extends Entry implements AfterViewInit, OnInit {
     if (input) {
       input.value = '';
     }
+    this.labelFormControl.setValue(null);
   }
 
   removeLabel(label: any): void {
--- a/src/app/workflow/workflow.component.ts
+++ b/src/app/workflow/workflow.component.ts
@@ -413,15 +413,21 @@ export class WorkflowComponent extends Entry implements AfterViewInit, OnInit {
       this.labelsEditMode = true;
       return;
     }
+    const value = this.labelFormControl.value;
+    if ((value || '').trim()) {
+      this.workflowEditData.labels.push(value.trim());
+    }
     // the edit object should be recreated
-    if (this.workflowEditData.labels !== 'undefined') {
+    if (this.workflowEditData.labels !== undefined) {
       this.setWorkflowLabels();
     }
+    this.labelFormControl.setValue(null);
   }
 
   cancelLabelChanges(): void {
     this.workflowEditData.labels = this.dockstoreService.getLabelStrings(this.workflow.labels);
     this.labelsEditMode = false;
+    this.labelFormControl.setValue(null);
   }
 
   // TODO: Move most of this function to the service, sadly 'this.labelsEditMode' makes it more difficult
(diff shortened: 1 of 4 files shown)


src/app/workflow/workflow.component.ts
```diff
@@ -413,15 +413,21 @@ export class WorkflowComponent extends Entry implements AfterViewInit, OnInit {
       this.labelsEditMode = true;
       return;
     }
+    const value = this.labelFormControl.value;
+    if ((value || '').trim()) {
+      this.workflowEditData.labels.push(value.trim());
+    }
     // the edit object should be recreated
-    if (this.workflowEditData.labels !== 'undefined') {
+    if (this.workflowEditData.labels !== undefined) {
```
coverbeck, line 421: I wonder about the value of this `if` statement in general -- if `this.workflowEditData.labels` really is `undefined` and `value` is truthy, it would have already failed on line 418. But I assume nobody (or at least I don't) knows this code that well, so it's probably safest not to tinker with it.

Heh, also, since the labels always seems to be an array, it "worked" before because an array does not equal the string value `"undefined"`.

coverbeck (approved):
I was able to reproduce the problem on qa; could not reproduce with a local build. Looks good!

kathy-t (commented):
I was able to reproduce the two bugs locally and your PR fixed them! I did notice a different bug, almost similar to Issue 2 in the ticket.

Steps to reproduce:
1. Go to "My Workflows" and click on any workflow with existing labels or add your own labels and click "Save".
2. Click on "Manage Labels". Observe that there is no duplicate label at the end like issue 2.
3. Type a label, like "potato", then press the <Enter> key.
4. Click on "Save" and observe that there are two "potato" labels while the save is occurring. Once the save is done, there's only one "potato".

Here's a video of the bug on my local build:

https://user-images.githubusercontent.com/25287123/194625035-a8579db4-229a-4dcb-937d-686b2dd6575d.mp4

svonworl:
> I was able to reproduce the two bugs locally and your PR fixed them! I did notice a different bug, almost similar to Issue 2 in the ticket.
> 
> Steps to reproduce:
> 
> 1. Go to "My Workflows" and click on any workflow with existing labels or add your own labels and click "Save".
> 2. Click on "Manage Labels". Observe that there is no duplicate label at the end like issue 2.
> 3. Type a label, like "potato", then press the `Enter` key on your keyboard.
> 4. Click on "Save" and observe that there are two "potato" labels while the save is occurring. Once the save is done, there's only one "potato", as expected.
> 
> Here's a video of the bug on my local build (if the video doesn't load, try Chrome)
> 
>  label-editing.mp4

Good catch!  Should be fixed in latest commit.

kathy-t (approved):

PR #1621, opened 2022-09-29
fixed github app link on new dashboard

hyunnaye:
**Description**
Fixes the github app link when registering  workflows, tools, and services on the new dashboard.

The changes specifically fixes the issue where the 'Manage Dockstore Installations on GitHub' button would redirect to the homepage instead of to github. This issue was caused due to the new dashboard not having an EntryType set in its session.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2239
https://github.com/dockstore/dockstore/issues/5123

### DIFF
--- a/cypress/integration/group1/dashboard.ts
+++ b/cypress/integration/group1/dashboard.ts
@@ -14,6 +14,7 @@
  *  limitations under the License.
  */
 import { resetDB, setTokenUserViewPort } from '../../support/commands';
+import { Dockstore } from '../../../src/app/shared/dockstore.model';
 
 describe('Dockstore dashboard', () => {
   resetDB();
@@ -42,5 +43,15 @@ describe('Dockstore dashboard', () => {
       .contains('Learn more about services')
       .should('have.attr', 'href')
       .and('include', 'getting-started-with-services');
+  });
+  it('Registering new tool through Github redirects correctly', () => {
+    cy.visit('/dashboard?newDashboard');
+    cy.contains('Tools');
+    cy.get('#registerToolButton').should('be.visible').click();
+    cy.get('#3-register-workflow-option').should('be.visible').click();
+    cy.contains('button', 'Next').should('be.visible').click();
+    cy.contains('a', 'Manage Dockstore installations on GitHub')
+      .should('have.attr', 'href')
+      .and('include', 'https://github.com/apps/dockstore-testing-application');
   });
 });
--- a/cypress/integration/group1/dashboard.ts
+++ b/cypress/integration/group1/dashboard.ts
@@ -14,7 +14,6 @@
  *  limitations under the License.
  */
 import { resetDB, setTokenUserViewPort } from '../../support/commands';
-import { Dockstore } from '../../../src/app/shared/dockstore.model';
 
 describe('Dockstore dashboard', () => {
   resetDB();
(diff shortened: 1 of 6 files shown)


svonworl:
Often, in our PRs, we include a short description of what was fixed and/or how the changes work.  If you could please add a bit of that to this PR, it would help me to better understand it.  Thanks!

test/web.yml
```diff
@@ -25,8 +25,6 @@ orcidClientSecret: <fill me in>
 gitHubAppPrivateKeyFile: <fill me in>
 gitHubAppId: <fill me in>
 
-orcidClientID: <fill me in>
-orcidClientSecret: <fill me in>
```
kathy-t, line 29: Why was this removed?

cypress/integration/group1/dashboard.ts
```diff
@@ -14,6 +14,7 @@
  *  limitations under the License.
  */
 import { resetDB, setTokenUserViewPort } from '../../support/commands';
+import { Dockstore } from '../../../src/app/shared/dockstore.model';
```
kathy-t, line 17: This is unused, can be removed

cypress/integration/group1/dashboard.ts
```diff
@@ -43,4 +44,14 @@ describe('Dockstore dashboard', () => {
       .should('have.attr', 'href')
       .and('include', 'getting-started-with-services');
   });
+  it('Registering new tool through Github redirects correctly', () => {
```
kathy-t, line 47: Might be worth having a similar test for workflows and services too

kathy-t (commented):
Thanks for fixing this for workflows and services too!

test/web.yml
hyunnaye: oops looks like it was removed by accident. Added it back now.

cypress/integration/group1/dashboard.ts
hyunnaye: Added one for workflows. The button to register services seems to link straight to the github link so I could not add a test for it as cypress does not support different domain redirects

cypress/integration/group1/dashboard.ts
```diff
@@ -43,4 +43,24 @@ describe('Dockstore dashboard', () => {
...
+    cy.visit('/dashboard?newDashboard');
+    cy.contains('Tools');
+    cy.get('#registerToolButton').should('be.visible').click();
+    cy.get('#3-register-workflow-option').should('be.visible').click();
+    cy.contains('button', 'Next').should('be.visible').click();
+    cy.contains('a', 'Manage Dockstore installations on GitHub')
+      .should('have.attr', 'href')
+      .and('include', 'https://github.com/apps/dockstore-testing-application');
+  });
+  it('Registering new workflow through Github redirects correctly', () => {
```
coverbeck, line 56: The blocks starting at line 47 and 57 are almost identical. You could create a function that takes a parameter(s) for what's different and avoid the duplication.

cypress/integration/group1/dashboard.ts
```diff
@@ -43,4 +43,24 @@ describe('Dockstore dashboard', () => {
       .should('have.attr', 'href')
       .and('include', 'getting-started-with-services');
   });
+  it('Registering new tool through Github redirects correctly', () => {
+    cy.visit('/dashboard?newDashboard');
+    cy.contains('Tools');
+    cy.get('#registerToolButton').should('be.visible').click();
+    cy.get('#3-register-workflow-option').should('be.visible').click();
```
coverbeck, line 50: 1. You're copying an unfortunate existing pattern we have everywhere, but ideally, you should be using `data-cy` instead of `id` to select in Cypress. See [Cypress best practices](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements). While we don't need to go back and fix everywhere where we're using `id`, we should avoid introducing new usages of `id` as a Cypress selector, if possible.
2. The `#3` is fragile (if we change the order), although the tests will catch it. If you switch to `data-cy` you may be able to come up with a better solution.

coverbeck:
Nice find, by the way. I know tracking down these code paths with the observables can be tricky.

cypress/integration/group1/dashboard.ts
hyunnaye: changed from using `id` to `data-cy`

coverbeck (approved):

kathy-t (approved):

PR #1627, opened 2022-10-11
Remove overrides to avoid self-DOSsing

coverbeck:
**Description**
Now that self-DOSsing bug has been fixed, we no longer need the test overrides that prevented the triggering of it.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4767


svonworl:
Should we target 1.13.x?

coverbeck:
> Should we target 1.13.x?

Not branch-wise, if that's what you mean. Smoke tests, a bit confusingly, are always run off the develop branch.

cypress/integration/smokeTests/prodOnly/search.ts
```diff
@@ -1,64 +0,0 @@
-/**
- * Remove this file after 1.13 deployed and update package.json "test-staging-no-auth"
- * and "test-prod-no-auth" to restore the sharedTests version of this file.
```
kathy-t, line 3: Do you need to update the package.json?

svonworl (approved):

cypress/integration/smokeTests/prodOnly/search.ts
coverbeck: Good catch, all I needed to do was read my own comment.

PR #1635, opened 2022-10-25
Upgrade to Angular 14

coverbeck:
**Description**
Upgrades to Angular 14 from Angular 13. I mainly just ran the migration tools Angular provides. Ran into 3 main issues:

1. Spun up dockstore/dockstore#5218 out of this. Json-ld is working, but it's a hassle.
2. One odd thing with Cypress integration tests -- the Angular tooltips in the Advanced Search was partially over an item being clicked on, and Cypress would fail. The tooltips' positions were the same as in Angular 13, but for some reason Cypress is only now having issues. I'm guessing the styling may have changed, even though they look the same. As a fix, I change the position of the tooltip.
3. Denis had to generate licenses because my machine and Docker containers don't match what we're using on CircleCI -- dockstore/dockstore#4979 -- not new, but a pain.

**Review Instructions**
Do some basic sanity testing -- click around a few pages and ensure things are working OK. We're mainly going to have to rely on the tests catching things.

**Issue**
[SEAB-4851](https://ucsc-cgl.atlassian.net/browse/SEAB-4851)


tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "moduleResolution": "node",
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
-    "target": "es2018",
+    "target": "es2020",
```
svonworl, line 12: Are we pretty confident that everyone who visits Dockstore will have a browser that supports `es2020`?  Related: https://kangax.github.io/compat-table/es2016plus/

svonworl (approved):

tsconfig.json
coverbeck: > Are we pretty confident that everyone who visits Dockstore will have a browser that supports `es2020`? Related: https://kangax.github.io/compat-table/es2016plus/

Hmm, nice catch, I hadn't even noticed that.

Looking into it, that is the Angular 14 default (their migration tools updated it without me noticing). Looking at your link, all desktop browsers support it, except Safari doesn't support Function.prototype.toString(), which is in es2019. It's not a feature we rely on (gets the source code of a function).

The only one that scares me is Samsung mobile browsers not support optional chaining... So I lean towards dropping to 2019. Thoughts by anybody else?

denis-yuen: > The only one that scares me is Samsung mobile browsers not support optional chaining... So I lean towards dropping to 2019. Thoughts by anybody else?

FWIW, in the last 30 days, according to Google Analytics, mobile users are 4.18% of users. 
Alternatively by browser, "Samsung Internet" is .09% of all possible Browser & OS combinations. 
That said, some folks may be blocking GA. And of course, this could be a chicken and the egg problem due to issues like https://github.com/dockstore/dockstore/issues/5144

coverbeck: Looking more closely at Analytics, we're only getting 14 and 18 of Samsung Internet -- in the table, version 14 does support optional chaining, although not with the spread operator after.

I'm inclined to leave it at es2020. Last chance for anybody to push back.

kathy-t (approved):

PR #1637, opened 2022-11-01
SEAB-4839 Reset dockstoretestuser4 resources via API

hyunnaye:
**Description**
Created a script which will reset the test user (dockstoretest4) resources to ensure the test user only has workflows in unpublished state and has no registered tools. 

More information on this issue: https://docs.google.com/document/d/17-dJk8SJ4ZrTp0tFucp_Y8fWESDUeUYIgmOfDLVcH4g/edit?usp=sharing

**Review Instructions**
Verify the script calls the correct APIs to unpublish workflows and delete registered tools. Verify that the config file correctly calls this script during auth tests and the auth tests still passes.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4839


.circleci/config.yml
```diff
@@ -552,7 +556,7 @@ workflows:
     jobs:
       - auth_smoke_tests:
           name: "auth_smoke_tests_dev"
```
coverbeck, line 558: Can you please update `dev` to `qa`? Look like I missed both of these in my PR (also please update line 532).

scripts/reset-dockstoretestuser4-resources.sh
```diff
@@ -0,0 +1,66 @@
...
+fi
+
+
+USER=$(curl -X 'GET' \
+  "${URL}/api/users/username/DockstoreTestUser4" \
+  -H 'accept: application/json')
+
+USER_ID=$(echo "$USER" | jq -r ".id")
+
+while IFS=$'\n' read -r line
```
coverbeck, line 26: Not a BASH expert nor do I know its best practices that well, but I would still suggest adding comments and/or creating functions, e.g., something like `getUserPublishedWorkflows`.

scripts/reset-dockstoretestuser4-resources.sh
```diff
@@ -0,0 +1,77 @@
...
+
+
+function getUserPublishedWorkflows {
+  while IFS=$'\n' read -r line
+  do
+    PUBLISHED_WORKFLOWS_IDS+=("$line")
+  done < <(curl -X 'GET' \
+    "${URL}/api/users/${USER_ID}/workflows/published" \
+    -H 'accept: application/json' \
+    -H "Authorization: Bearer ${TOKEN}" | jq .[].id)
```
svonworl, line 35: Thanks for the hard work on this!

If you wanted to make this block a little tighter, you could simply assign the result of the command substitution to the variable, ala:
```
IDS=$(curl -X GET http://<hostname>/users/<id>/workflows/published -H 'accept: application/json' -H 'Authorization: Bearer <token>' | jq .[].id)
```
Bash drops the newlines and `IDS` is set to something like this:
```
90210 5150 777

File: scripts/reset-dockstoretestuser4-resources.sh [hunk 4]
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+
```
svonworl, line 2: A lot of our scripts use some combination of:
```
set -o errexit
set -o pipefail
set -o nounset
set -o xtrace
```
If you find that any of that behavior might be desirable, please add as appropriate.

scripts/reset-dockstoretestuser4-resources.sh
```diff
@@ -0,0 +1,77 @@
...
+      -H 'Content-Type: application/json' \
+      -H "Authorization: Bearer ${TOKEN}" \
+      -d '{
+      "publish": false
+    }'
+  done
+
+  for TOOL_ID in "${REGISTERED_TOOLS_IDS[@]}"
+  do
+    curl -X 'DELETE' \
```
kathy-t, line 65: Are all registered tools guaranteed to be unpublished? If not, should check that they're unpublished before deleting them because you can't delete published tools

scripts/reset-dockstoretestuser4-resources.sh
```diff
@@ -0,0 +1,77 @@
...
+if [ "$STACK" == "qa" ]
+then
+  URL="https://qa.dockstore.org"
+  TOKEN=${CYPRESS_QA_TOKEN}
+elif [ "$STACK" == "staging" ]
+then
+  URL="https://staging.dockstore.org"
+  TOKEN=${CYPRESS_STAGING_TOKEN}
+else
+  URL="https://dockstore.org"
```
denis-yuen, line 14: this is a little undesirable to have prod as a default, could imagine someone adding a new environment in .circleci/config.yml and forgetting to modify this or mis-spelling something and hitting prod 
It's a test user, so its not a big deal but would prefer to have this just fail on unknown to make it obvious

scripts/reset-dockstoretestuser4-resources.sh
denis-yuen: > Thanks for the hard work on this!

Seconding this

scripts/reset-dockstoretestuser4-resources.sh
denis-yuen: Agreed on most, but will need to avoid `set -o xtrace` though due to how the token is used in the curl calls below. 
https://circleci.com/docs/env-vars/#secrets-masking has more details

denis-yuen (commented):
FYI, for other reviewers, google docs link is protected and requires login (which is good)

scripts/reset-dockstoretestuser4-resources.sh
hyunnaye: I added echo commands instead of `set -o xtrace` to document the steps. Should I add more details?

scripts/reset-dockstoretestuser4-resources.sh
hyunnaye: thanks for this! I wasn't aware that you couldn't delete published tools. I edited my script to call to unpublish first then delete them

scripts/reset-dockstoretestuser4-resources.sh
denis-yuen: No, I think that is fine. There are ways of passing environment variables around without having them show up on the command-line. For example a pgpass file with postgress. I think that may be a good follow-up ticket to create now and can tackle if we run into issues or need to improve this script.

denis-yuen (approved):

coverbeck (approved):

svonworl (approved):

kathy-t (approved):

PR #1638, opened 2022-11-01
DOCK-2240 Changed colors of each entry widget bubbles on new dashboard

hyunnaye:
**Description**
Fixed some styling differences between the implemented entry widgets for the new dashboard on dev and the [zeplin mockup](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6022ab2d42df9f2c57c20982).

The number of entries label in each entry widget should have a color specific to the entry type (green for workflows, blue for tools, orange for services).

Before fix, they were all grey:
Screenshot from 2022-11-01 15-09-44

After fix:
Screenshot from 2022-11-01 15-06-49

**Review Instructions**
Verify the implemented color match the zeplin mockup and that it correctly changes each bubble color.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2240

### DIFF
--- a/src/app/home-page/widget/entry-box/entry-box.component.html
+++ b/src/app/home-page/widget/entry-box/entry-box.component.html
@@ -3,9 +3,9 @@
     <div id="workflow-number" fxLayout="row">
       <mat-card-header fxLayout="row">
         <mat-card-title>{{ entryTypeLowerCase | titlecase }}s</mat-card-title>
-        <span class="bubble" id="bubble-green" *ngIf="entryTypeLowerCase === 'workflow'">{{ totalEntries }}</span>
-        <span class="bubble" id="bubble-blue" *ngIf="entryTypeLowerCase === 'tool'">{{ totalEntries }}</span>
-        <span class="bubble" id="bubble-orange" *ngIf="entryTypeLowerCase === 'service'">{{ totalEntries }}</span>
+        <span class="bubble workflow-background" *ngIf="entryType === entryTypeEnum.WORKFLOW">{{ totalEntries }}</span>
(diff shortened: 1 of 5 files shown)


src/app/home-page/widget/entry-box/entry-box.component.html
```diff
@@ -3,7 +3,9 @@
     <div id="workflow-number" fxLayout="row">
       <mat-card-header fxLayout="row">
         <mat-card-title>{{ entryTypeLowerCase | titlecase }}s</mat-card-title>
-        <span class="bubble">{{ totalEntries }}</span>
+        <span class="bubble" id="bubble-green" *ngIf="entryTypeLowerCase === 'workflow'">{{ totalEntries }}</span>
```
coverbeck, line 6: I suggest doing the styling based on a class, not the id. It just feels better :) Plus there's a chance you could end up with duplicate ids; there are several instances of the entry-box.component -- each entry-box.component should have different a entry type, and hence a unique id, but it's not enforced and you never know if another instance might be instantiated.

Also, I suggest changing `entryTypeLowerCase === 'workflow'` to (untested) `entryType === entryTypeEnum.WORKFLOW` (you'll need to add a property in the TS file for this to work) -- relies on types the compiler can catch instead of spelling a string correctly.

hyunnaye: Done! For the ids, they were replaced by the preexisting global styles in `styles.scss` following Kathy's comment.

coverbeck (approved):

kathy-t (approved):

svonworl (approved):

denis-yuen (approved):

PR #1645, opened 2022-11-11
DOCK-2199: Added domain onto the copy block on Copy build info button

hyunnaye:
**Description**
This PR will now copy the domain when the Copy build info button is clicked. 

Before the fix, the Copy build info button copies all the build versions, but doesn't copy the domain, so you have to manually add the host/domain to that section of the issue.

**Review Instructions**
Verify that the Copy build info button down at the footer now copies the domain as well.

**Issue**
https://github.com/dockstore/dockstore/issues/5015
https://ucsc-cgl.atlassian.net/browse/DOCK-2199

### DIFF
--- a/src/app/footer/footer.component.ts
+++ b/src/app/footer/footer.component.ts
@@ -31,6 +31,7 @@ import { versions } from './versions';
   styleUrls: ['./footer.component.scss'],
 })
 export class FooterComponent extends Base implements OnInit {
+  domain: string;
   version: string;
   tag: string;
   public prod = true;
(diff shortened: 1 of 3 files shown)


src/app/footer/footer.component.ts
```diff
@@ -78,6 +80,7 @@ export class FooterComponent extends Base implements OnInit {
               this.version = metadatum;
             }
             this.content = this.footerService.versionsToMarkdown(
+              this.domain,
```
coverbeck, line 83: You can use `DOCKSTORE.HOSTNAME`. It already does the calculation, and it will just return `https://dockstore.org`, whereas `window.location.href` will include the current path based on where you are, e.g., if you are on the orgs page, `https://dockstore.org/organizations`, which we probably don't want.

coverbeck (approved):
In your PR description, could you please also reference the GitHub issue (as opposed to/in addition to the JIRA issue)? That way if somebody looks at the GitHub issue, they'll see a link to the PR.

denis-yuen (approved):

PR #1652, opened 2022-11-29
update and add warning for Linux

denis-yuen:
**Description**
Add warning against the use of Docker Desktop with Dockstore CLI

**Review Instructions**
See that the warning appears in the quick-start

**Issue**
https://github.com/dockstore/dockstore/issues/5135

### DIFF
--- a/cypress/integration/group2/organizations.ts
+++ b/cypress/integration/group2/organizations.ts
@@ -17,7 +17,7 @@ import { approvePotatoMembership, approvePotatoOrganization, resetDB, setTokenUs
 import { TokenUser } from '../../../src/app/shared/swagger';
 import { TokenSource } from '../../../src/app/shared/enum/token-source.enum';
 
-const imageURL = 'https://fakeUrl.com/potato.png';
+const imageURL = 'https://superduperfakepotatourl.com/potato.png';
 describe('Dockstore Organizations', () => {
   resetDB();
   setTokenUserViewPort();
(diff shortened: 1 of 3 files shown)


denis-yuen:
Screenshot from 2022-11-29 18-09-07

denis-yuen:
Think build failure is a false positive

aofarrel:
Do we want to continue telling Mac users to install Docker Desktop? That is what I use without issue, but there's two orange flags here: 
* I'm wondering if some of the Mac permissions weirdness interacts with version of Docker the Mac user has installed. I don't have the bandwidth to investigate this myself.
* [Our docs lead Mac users to Docker Engine instructions but the quick start takes them to a Docker Desktop installer](https://github.com/dockstore/dockstore-documentation/pull/230/files#r1036407213)

denis-yuen:
Hmmm, added an additional comment in the documentation PR, but from https://github.com/dockstore/dockstore/issues/5135 I think I would recommend against Docker Desktop for Linux. I have little experience with Mac, but my understanding is Mac users do not have a choice?

coverbeck:
> Hmmm, added an additional comment in the documentation PR, but from [dockstore/dockstore#5135](https://github.com/dockstore/dockstore/issues/5135) I think I would recommend against Docker Desktop for Linux. I have little experience with Mac, but my understanding is Mac users do not have a choice?

It looks like there are some unofficial workarounds for not installing Docker Desktop on a Mac, but they're not official. So yes, you basically have to install Desktop.

Also, FWIW, my understanding is that the Docker Engine on Mac requires a VM as well... This link is old, but presumably still applicable: https://dev.to/ericnograles/why-is-docker-on-macos-so-much-worse-than-linux-flh

coverbeck (approved):

PR #1657, opened 2022-11-30
Remove ngx-lite/json-ld dependency

coverbeck:
**Description**
We're removing the dependency because they haven't issued newer releases to keep up with new Angular releases.

Copy the relevant part of their code and commit it to our repo, with attribution.

In testing it, I noticed that the json-ld was not showing up on the
home page. I thought I'd broken it, but it turns out that component
is not ever used (json-ld is not showing up in prod either, without my
changes). I deleted the component and moved the json-ld to the
actual home page component.

**Review Instructions**
1. Go to the home page, not logged into Dockstore
2. Open Browser developer tools
3. Verify there are 2 instances of a script tag, with the attribute `type=application/ld+json`

**Issue**
dockstore/dockstore#5218


coverbeck:
Justification for bypassing Angular sanitization, from checklist, FYI @david4096:

1. We were effectively invoking the same code before, just via an NPM module
2. The code has a test for an injection attack.

denis-yuen (approved):

kathy-t (approved):

src/app/shared/json-ld/json-ld.component.spec.ts
```diff
@@ -0,0 +1,112 @@
+/*
+ * MIT License
```
svonworl, line 2: Whether or not to address this is up to you:
I'm a little bit concerned that, absent our context, it's not clear what these two license blocks mean, in terms of exactly what and how we're licensing.  Maybe we could use the comment below the imports as a starting point, and say something like:

```
The following code was adapted from https://github.com/coryrylan/ngx-lite, 
as that repo and its corresponding NPM releases have not kept up with the latest Angular releases.
The original code was licensed under the included MIT license, and we sublicense the derived code
under the included Apache license.  The included MIT license does not apply to other source files,
except where noted.
```

coverbeck: That's better said and more explicit than what I have... Good suggestion.

svonworl (approved):

PR #1673, opened 2023-01-04
Add language version search facet and language version column in versions tab

kathy-t:
**Description**

**Review Instructions**

**Issue**

**Security**

### DIFF
--- a/src/app/container/versions/versions.component.html
+++ b/src/app/container/versions/versions.component.html
@@ -94,6 +94,23 @@
         {{ version.last_built | date: 'yyyy-MM-dd HH:mm' }}
       </td>
     </ng-container>
+    <ng-container matColumnDef="descriptor_type_versions">
+      <th
+        scope="col"
+        mat-header-cell
+        *matHeaderCellDef
+        mat-sort-header
+        matTooltip="The language versions used by the version's the descriptor file(s)."
+      >
+        Language Versions
+      </th>
+      <td mat-cell *matCellDef="let version">
+        <!-- Only display descriptor type versions if there's one descriptor type otherwise we can't tell which version belongs to which type without looking at the source files -->
+        <span *ngIf="tool?.descriptorType.length === 1">{{
+          tool.descriptorType[0] | descriptorLanguageVersions: version.descriptorTypeVersions
+        }}</span>
+      </td>
+    </ng-container>
     <ng-container matColumnDef="valid">
       <th scope="col" mat-header-cell *matHeaderCellDef mat-sort-header matTooltip="">Valid</th>
       <td mat-cell *matCellDef="let version">
--- a/src/app/workflow/versions/versions.component.html
+++ b/src/app/workflow/versions/versions.component.html
@@ -89,6 +89,21 @@
       </td>
     </ng-container>
 
+    <ng-container matColumnDef="descriptor_type_versions">
+      <th
+        scope="col"
+        mat-header-cell
+        *matHeaderCellDef
+        mat-sort-header
+        matTooltip="The language versions used by the version's the descriptor file(s)."
+      >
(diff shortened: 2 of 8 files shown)


coverbeck:
I haven't looked at the code yet, nor tried running it, but isn't this facet is going to behave a bit differently than others? In your example, a single workflow has both WDL 1.1 and 1.0 versions. Assuming your use case/screenshots in the description, if you select both facets, you will see one workflow in the search results, correct? It might be confusing, because you might expect to see 2 items visible.

Do we have any other facet that acts like this?

It's not necessarily wrong, I'm just trying to understand at this point, and I don't really have an alternative if it's so.

src/app/container/versions/versions.component.html
```diff
@@ -94,6 +94,23 @@
         {{ version.last_built | date: 'yyyy-MM-dd HH:mm' }}
       </td>
     </ng-container>
+    <ng-container matColumnDef="descriptor_type_versions">
+      <th
+        scope="col"
+        mat-header-cell
+        *matHeaderCellDef
+        mat-sort-header
+        matTooltip="The language versions used by the version's the descriptor file(s)."
```
coverbeck, line 103: Wording seems off -- remove `the`?

src/app/workflow/versions/versions.component.html
```diff
@@ -89,6 +89,21 @@
       </td>
     </ng-container>
 
+    <ng-container matColumnDef="descriptor_type_versions">
+      <th
+        scope="col"
+        mat-header-cell
+        *matHeaderCellDef
+        mat-sort-header
+        matTooltip="The language versions used by the version's the descriptor file(s)."
```
coverbeck, line 98: Same wording issue as other comment.

coverbeck (approved):

kathy-t:
> In your example, a single workflow has both WDL 1.1 and 1.0 versions. Assuming your use case/screenshots in the description, if you select both facets, you will see one workflow in the search results, correct? It might be confusing, because you might expect to see 2 items visible.
> 
> Do we have any other facet that acts like this?

I believe the Verified Platforms facet behaves in a similar manner. According to its tooltip, it _Indicates which platform a tool or workflow (at least one version) was successfully run on._

For example, this [workflow](https://dockstore.org/workflows/github.com/DataBiosphere/topmed-workflows/UM_variant_caller_wdl:1.32.0?tab=versions) has two versions with different Verified Platforms, Dockstore CLI and Cromwell. If you select both Verified Platform facets, you would only see one workflow in the search results. This is an [example search query](https://dockstore.org/search?verified_platforms.keyword=Dockstore%20CLI&verified_platforms.keyword=Cromwell&entryType=workflows&search=UM_variant_caller_wdl) for that workflow.

denis-yuen (approved):

svonworl (approved):

coverbeck:
> > Do we have any other facet that acts like this?
> 
> I believe the Verified Platforms facet behaves in a similar manner. According to its tooltip, it _Indicates which platform a tool or workflow (at least one version) was successfully run on._
> 
> For example, this [workflow](https://dockstore.org/workflows/github.com/DataBiosphere/topmed-workflows/UM_variant_caller_wdl:1.32.0?tab=versions) has two versions with different Verified Platforms, Dockstore CLI and Cromwell. If you select both Verified Platform facets, you would only see one workflow in the search results. This is an [example search query](https://dockstore.org/search?verified_platforms.keyword=Dockstore%20CLI&verified_platforms.keyword=Cromwell&entryType=workflows&search=UM_variant_caller_wdl) for that workflow.

Thanks for clarifying! I did not know about those facets. Also, in hindsight, the use case I cited of mixed versions is rarely going to come up in any case.

PR #1680, opened 2023-01-12
404 Redirects for Remaining Pages - Dock 1170

KimberleyChong:
**Description**

Current 404 pages are displayed with the Angular Wildcard Route ('**'), which redirects _undefined_ URLs to the `PageNotFoundComponent`, however this does not handle pages displaying Organizations, Collections, Workflows, and Tools because every URL that is `https://dockstore.org/organizations/{ORG-NAME}` (for example) is accepted as a _defined_ URL and the component is loaded using the URL param whether or not that organization, tool, workflow, or collection exists -> resulting in a black page.  For some non-existing tools/workflows we handle the 404 by redirecting to the homepage, so those were also updated to redirect to the 404 page.

### Solution:
I tried a few methods (ex. comparing the desired org name to a full list of org names) but found that the best/most reliable method is when requesting to load the organization, tool, workflow, or collection, check if the HTTP response was a 404 and redirect to `page-not-found`.

**Previous:**
http://localhost:4200/organizations/kim
image

**With Redirect:**
image

### Examples of fixed URL formats:
**Non-existing Organizations:**
Non-existing organization: http://localhost:4200/organizations/my-org
Non-existing organization & non-existing collection: http://localhost:4200/organizations/my-org/collections/collection-name
Existing organization & non-existing collection: http://localhost:4200/organizations/iwc/collections/kim

**Non-existing workflows:**
http://localhost:4200/workflows/1234
http://localhost:4200/workflows/github.com/kim-workflows
http://localhost:4200/workflows/github.com/iwc-workflows/kim

**Non-existing tool:**
http://localhost:4200/containers/dog
http://localhost:4200/containers/quay.io/kim-org
http://localhost:4200/containers/quay.io/pancancer/kim

I did some troubleshooting to find these URLs and that should cover all of the remaining/necessary 404 redirects, but if anyone can find any that's missing let me know :)

**Issue**
[DOCK-1170
](https://ucsc-cgl.atlassian.net/browse/DOCK-1170?atlOrigin=eyJpIjoiM2FmYTdlZjRmZTI1NDY2ZWJjZTM5YjRmZmRkM2E4YzYiLCJwIjoiaiJ9)
**Security**


src/app/organizations/state/collections.service.ts
```diff
@@ -101,8 +101,11 @@ export class CollectionsService {
           this.collectionsStore.setActive(collection.id);
           this.organizationService.updateOrganizationFromID(collection.organizationID);
         },
-        () => {
+        (collection: Collection) => {
```
coverbeck, line 104: I think you can just have `(error: HttpErrorResponse)`, then you don't need do the casting on line 106. https://angular.io/guide/http#error-handling. It's weird, I'd never looked at it before, but the IDE/compiler doesn't seem to be able to infer that.

That said, the doc I linked to says it returns an error object, so I think it's safe to declare it as an ErrorResponseObject. It's definitely not going to be a `Collection` though, so I wouldn't declare it as that.

src/app/organizations/state/organization.service.ts
```diff
@@ -78,8 +81,11 @@ export class OrganizationService {
           this.updateOrganization(organization);
           this.organizationMembersService.updateCanModify(organization.id);
         },
-        () => {
+        (organization: Organization) => {
```
coverbeck, line 84: Same comment as other

src/app/organizations/state/collections.service.ts
KimberleyChong: Ah that makes a lot of sense, thank you

src/app/organizations/state/collections.service.ts
```diff
@@ -162,51 +165,49 @@ export class CollectionsService {
...
+          this.router.navigate(['/organizations', organizationName]);
         }
-      );
+      },
+      (error: HttpErrorResponse) => {
+        this.collectionsStore.setLoading(false);
+        this.collectionsStore.setError(true);
+        this.alertService.detailedError(error);
+      }
+    );
```
svonworl, line 211: As far as I can tell, the above changes are probably due to a code reformat by our automatic tool.  Did anything functional change in `deleteCollection`?

svonworl (approved):

src/app/organizations/state/collections.service.ts
KimberleyChong: Yeah everything below my change on line 108 was automatic reformatting (makes the changes look so confusing on git).  No changes to `deleteCollection`

denis-yuen (approved):

src/app/container/container.component.ts
```diff
@@ -255,8 +255,10 @@ export class ContainerComponent extends Entry implements AfterViewInit, OnInit {
             this.updateUrl(this.tool.tool_path, 'my-tools', 'containers');
           }
         },
-        () => {
-          this.router.navigate(['../']);
+        (error) => {
+          if (error.status == 404) {
```
kathy-t, line 259: Should use `===`

src/app/organizations/state/collections.service.ts
```diff
@@ -101,8 +101,11 @@ export class CollectionsService {
           this.collectionsStore.setActive(collection.id);
           this.organizationService.updateOrganizationFromID(collection.organizationID);
         },
-        () => {
+        (error: HttpErrorResponse) => {
           this.collectionsStore.setError(true);
+          if (error.status == 404) {
```
kathy-t, line 106: Should use `===`

src/app/organizations/state/organization.service.ts
```diff
@@ -78,8 +81,11 @@ export class OrganizationService {
           this.updateOrganization(organization);
           this.organizationMembersService.updateCanModify(organization.id);
         },
-        () => {
+        (error: HttpHeaderResponse) => {
           this.organizationStore.setError(true);
+          if (error.status == 404) {
```
kathy-t, line 86: Should use `===`

KimberleyChong:
Please note that the `accessibility_test_base` will be failing until develop has this [commit](https://github.com/dockstore/dockstore-ui2/pull/1680/commits/a8c62c1d3953f93deca0b6073371b7b96ab5e9af) with the updated circleci build id (I have included it in this pr so the test should be resolved after this is merged)

kathy-t (approved):

coverbeck (approved):

david4096 (approved):

PR #1685, opened 2023-01-19
DOCK-2290: updated navbar dropdown

hyunnaye:
**Description**
Updated dropdown on navbar to look like the zeplin mockups.

UI after changes:
Screenshot from 2023-01-19 16-52-26
Screenshot from 2023-01-19 16-52-34

Zeplin mockups
https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6022acb6632937281e6935ec. 

**Review Instructions**
Verify that the new changes represent the mockups accurately and the integration tests still pass.

**Issue**
https://github.com/dockstore/dockstore/issues/5261

**Security**

### DIFF
--- a/cypress/e2e/immutableDatabaseTests/dropdown.ts
+++ b/cypress/e2e/immutableDatabaseTests/dropdown.ts
@@ -384,7 +384,7 @@ describe('Dropdown test', () => {
       });
       goToAccountsOnboarding();
       everythingOk();
-      cy.get('[data-cy=dropdown-main').should('be.visible').click();
+      cy.get('[data-cy=dropdown-main]').should('be.visible').click();
       cy.get('[data-cy=dropdown-logout-button]').should('be.visible').click();
       cy.get('[data-cy=header]').should('contain', 'Logged Out');
     });
--- a/cypress/e2e/immutableDatabaseTests/dropdown.ts
+++ b/cypress/e2e/immutableDatabaseTests/dropdown.ts
@@ -384,6 +384,8 @@ describe('Dropdown test', () => {
       });
       goToAccountsOnboarding();
       everythingOk();
+      cy.get('[data-cy=dropdown-logout-button]').should('be.visible').click();
+      cy.get('[data-cy=header]').should('contain', 'Logged Out');
     });
     // TODO: this part of the wizard has been reworked
     // it('Go through steps', () => {
@@ -429,9 +431,5 @@ describe('Dropdown test', () => {
     //         .get('#finish_step')
     //         .click()
     // });
-    it('Should be able to log out through the dropdown button', () => {
-      cy.get('[data-cy=dropdown-logout-button]').should('be.visible').click();
-      cy.get('[data-cy=header]').should('contain', 'Logged Out');
-    });
   });
 });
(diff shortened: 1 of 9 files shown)


src/app/navbar/navbar.component.scss
```diff
@@ -79,3 +79,29 @@ mat-menu {
   color: mat.get-color-from-palette($kim-accent1-success, 4);
   transform: scale(1.25);
 }
+
+#dropdown-main {
+  padding: 14px 14px;
```
coverbeck, line 84: Aren't we supposed to avoid pixels and use rems instead? It's in the [opinionated style guide](https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide), although at one point I thought Gary mentioned it may not be necessary any more. @kathy-t may know?

If rems aren't required, can you please update the style guide, so I don't keep making comments like this. If they are required, please update here.

hyunnaye: I was thinking about that too but wanted to make it the same padding as the notification bell and My Dockstore buttons which had 14px as padding. https://github.com/dockstore/dockstore-ui2/blob/e5d682ce2906cd65b99dc5be0b41861404244f87/src/app/navbar/navbar.component.scss#L21

If we should use rem insread, i'll change it for those as well or consistency!

kathy-t: Hmm, I generally follow the style guide and use rem. I don't really see a downside to using REM because it makes the design more accessible

svonworl (approved):

kathy-t (approved):

denis-yuen (approved):

cypress/e2e/immutableDatabaseTests/dropdown.ts
```diff
@@ -431,4 +430,10 @@ describe('Dropdown test', () => {
     //         .click()
     // });
   });
+  describe('Log out', () => {
+    it('Should be able to log out through the dropdown button', () => {
+      cy.get('[data-cy=dropdown-logout-button]').should('be.visible').click();
```
coverbeck, line 435: This isn't going to work with my upcoming Cypress 12 PR :( . You had no way of knowing that, and we have this pattern all over -- I've been changing them.

I can fix it after you merge, or if you want to get it ready in advance...

In Cypress 12, different tests are isolated by default. This test currently assumes you are

1. On a page, carried over from the previous test
2. Logged in

In Cypress 12
1. You won't be on a page because you need to explicitly navigate to one
2. You won't be logged in (local storage, with the token, is wiped between tests).

There are two possible fixes:

1. Get rid of lines 433 and 434 (and their closing braces), so that the previous page and logged in token "carry over".
2. Keep lines 433 and 434, but do a `cy.visit` and the call to set the user token.

I will probably just do 1 because it's easier; if you feel strongly and want to keep the log out as a separately named test, then could you please do 2.

coverbeck (approved):
Approved, but with an option if you don't want me to change your code shortly after it's merged. :)

cypress/e2e/immutableDatabaseTests/dropdown.ts
hyunnaye: I'll change it!

PR #1694, opened 2023-01-27
DOCK-2213: edit source code to link to primary descriptor

hyunnaye:
**Description**
This PR changes the Source Code link to point to the primary descriptor file instead of pointing to the root of the repo. Also, some workflows had very long paths so the text of the link was changed to only display the file name.

**Review Instructions**
Verify the link is edited correctly

**Issue**
https://github.com/dockstore/dockstore/issues/5044

**Security**

### DIFF
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -45,9 +45,9 @@
             <a
               id="sourceRepository"
               data-cy="sourceRepository"
-              [href]="workflow?.providerUrl | versionProviderUrl: (isPublic ? selectedVersion?.name : '')"
+              [href]="workflow?.providerUrl | versionProviderUrl: (isPublic ? selectedVersion?.name : '') + '/' + sourceCodeFile"
             >
-              {{ workflow?.providerUrl | urlDeconstruct: (isPublic ? selectedVersion?.name : '') }}
+              {{ sourceCodeFile }}
             </a>
           </li>
         </span>
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -47,7 +47,7 @@
               data-cy="sourceRepository"
               [href]="workflow?.providerUrl | versionProviderUrl: (isPublic ? selectedVersion?.name : '') + '/' + sourceCodeFile"
             >
-              {{ sourceCodeFile }}
+              {{ workflow?.providerUrl | versionProviderUrl: (isPublic ? selectedVersion?.name : '') + '/' + sourceCodeFile }}
             </a>
           </li>
         </span>
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -39,7 +39,7 @@
     <mat-divider></mat-divider>
     <mat-card-content class="p-3">
       <ul class="list-unstyled container-info" *ngIf="workflow">
-        <span *ngIf="workflow?.mode !== WorkflowType.ModeEnum.HOSTED">
+        <span *ngIf="workflow?.mode !== WorkflowType.ModeEnum.HOSTED" class="truncate-text-1">
           <li *ngIf="workflow?.provider && workflow?.providerUrl">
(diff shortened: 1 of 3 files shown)


src/app/workflow/info-tab/info-tab.component.ts
```diff
@@ -150,6 +151,7 @@ export class InfoTabComponent extends EntryTab implements OnInit, OnChanges {
       .subscribe((entryType) => (this.displayTextForButton = this.infoTabService.getTRSId(this.workflow, entryType)));
     this.infoTabService.forumUrlEditing$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((editing) => (this.forumUrlEditing = editing));
     this.infoTabService.topicEditing$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((topicEditing) => (this.topicEditing = topicEditing));
+    this.sourceCodeFile = this.selectedVersion.workflow_path.slice(1); //removes first slash to prevent from displaying on text
```
coverbeck, line 154: There are several checks for `this.selectedVersion` existing or not, it seems like you should add the check. Also, the other accesses to `selectedVersion` seem to be in `ngOnChanges`; I suggest moving this change there -- if the version changes, `sourceCodeFile` won't get updated.

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -45,9 +45,9 @@
             <a
               id="sourceRepository"
               data-cy="sourceRepository"
-              [href]="workflow?.providerUrl | versionProviderUrl: (isPublic ? selectedVersion?.name : '')"
+              [href]="workflow?.providerUrl | versionProviderUrl: (isPublic ? selectedVersion?.name : '') + '/' + sourceCodeFile"
             >
-              {{ workflow?.providerUrl | urlDeconstruct: (isPublic ? selectedVersion?.name : '') }}
+              {{ sourceCodeFile }}
```
kathy-t, line 50: I wonder if we should keep the text as before because looking at it locally, the file name looks out of place next to `Source Code`.

image

> Also, some workflows had very long paths so the text of the link was changed to only display the file name.

Did it not wrap properly with the long paths?

denis-yuen: > Did it not wrap properly with the long paths?

think it wrapped but looked silly (i.e. wrapped more than once with some examples)

hyunnaye: It wraps but it looked a bit weird when the link got too long. Here was an example that I encountered
image

kathy-t: Perhaps a compromise is to truncate the text if it's more than one line? Looks something like this:

image

I used the the `truncate-text-1` class from [here](https://github.com/dockstore/dockstore-ui2/blob/7dd11b7dd5d46c56ad7e6016030b4fc023a1d338/src/styles.scss#L1201)

denis-yuen: (personally ok either way as long as the mouse-over URL is good)

denis-yuen (approved):

denis-yuen:
(with pending comments above)

svonworl (approved):

src/app/workflow/info-tab/info-tab.component.ts
```diff
@@ -119,6 +120,7 @@ export class InfoTabComponent extends EntryTab implements OnInit, OnChanges {
         this.selectedVersion.workflow_path,
         this.sessionQuery.getValue().entryType
       );
+      this.sourceCodeFile = this.selectedVersion.workflow_path.slice(1); //removes first slash to prevent from displaying on text
```
coverbeck, line 123: I just queried the workflowversions table in  recent-ish snapshot of prod data  for workflowpath not starting with a `/`, and about 1% of them don't -- this would end up truncating the first character in those cases. So I think you need either need to check here for a leading slash, or just take the whole value here and change the display logic in the HTML.

I ran this query:

```
select count(*) from workflowversion where workflowpath not like '/%';
```

coverbeck (changes requested):

kathy-t (approved):

coverbeck (approved):

PR #1698, opened 2023-02-01
Add the Open Data facet

coverbeck:
**Description**
Surface the Open Data facet. 

The facets column is getting really long, I wonder if it should start out collapsed by default? But that's for another issue/discussion.

Screen Shot 2023-01-31 at 5 17 04 PM

Screen Shot 2023-01-31 at 5 12 09 PM

**Review Instructions**
Verify that the open data facets are present as in the screenshots, and that clicking a search result when the `Has public test data` checkbox is selected, takes you to a workflow with an open data test param file.

**Issue**
dockstore/dockstore#5136

**Security**

### DIFF
--- a/src/app/search/map-friendly-values.pipe.ts
+++ b/src/app/search/map-friendly-values.pipe.ts
@@ -68,6 +68,13 @@ export class MapFriendlyValuesPipe implements PipeTransform {
       ]),
     ],
     [
+      'openData',
+      new Map([
+        ['1', 'has public test data'],
+        ['0', 'no public test data'],
+      ]),
+    ],
+    [
       'registry',
       new Map([
         ['QUAY_IO', 'Quay.io'],
@@ -110,13 +117,6 @@ export class MapFriendlyValuesPipe implements PipeTransform {
         ['PUSH', 'Push'],
         ['DELETE', 'Delete'],
         ['INSTALL', 'Install'],
-      ]),
-    ],
-    [
-      'openData',
-      new Map([
-        ['0', 'No public test data'],
-        ['1', 'Has public test data'],
       ]),
     ],
   ]);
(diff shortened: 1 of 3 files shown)


coverbeck:
Failing test is the flaky one, I think.

src/app/search/map-friendly-values.pipe.ts
```diff
@@ -112,6 +112,13 @@ export class MapFriendlyValuesPipe implements PipeTransform {
         ['INSTALL', 'Install'],
       ]),
     ],
+    [
+      'openData',
+      new Map([
+        ['0', 'No public test data'],
+        ['1', 'Has public test data'],
```
kathy-t, line 119: Minor, but should keep the casing the same as the other facets. `Verified` and `Has Checker Workflows` have lower case facets

kathy-t (approved):

KimberleyChong (approved):

denis-yuen (approved):

KimberleyChong:
> The facets column is getting really long, I wonder if it should start out collapsed by default? But that's for another issue/discussion.

That's a good point, maybe we can do something like in the [mocks](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/5fc690e874e0c807afeb299b) where about half of them are closed and the most priority ones are kept open.

coverbeck:
> > The facets column is getting really long, I wonder if it should start out collapsed by default? But that's for another issue/discussion.
> 
> That's a good point, maybe we can do something like in the [mocks](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/5fc690e874e0c807afeb299b) where about half of them are closed and the most priority ones are kept open.

Nice find on the mocks, created a followup ticket, dockstore/dockstore#5341

svonworl (approved):

PR #1707, opened 2023-02-10
Use Friendly Name for Galaxy Workflows - SEAB-5244

KimberleyChong:
**Description**
Use friendly name for Galaxy workflows in the `starredentries.component`.

Taken straight from `search-workflow-table.component`:
https://github.com/dockstore/dockstore-ui2/blob/507d493626e698b0eb7d8f38646bc74d2c0e365e/src/app/search/search-workflow-table/search-workflow-table.component.html#L42

Also updated the unit test for the `DescriptorLanguageService` provider and `HttpClientTestingModule` import.

**Before:**
image

**After:**
image

**Review Instructions**

**Issue**
[SEAB-5244
](https://ucsc-cgl.atlassian.net/browse/SEAB-5244)
**Security**

### DIFF
--- a/src/app/shared/entry/descriptor-language.pipe.ts
+++ b/src/app/shared/entry/descriptor-language.pipe.ts
@@ -1,0 +1,24 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { DescriptorLanguageService } from 'app/shared/entry/descriptor-language.service';
+import { Workflow } from '../../shared/swagger';
+
+@Pipe({
+  name: 'descriptorLanguage',
+})
+export class DescriptorLanguagePipe implements PipeTransform {
+  /**
+   * When displaying the descriptor type typically the workflow descriptor type string is used
+   * However for Galaxy this is 'gxformat2', so change it to be Galaxy.
+   * The shortFriendlyName is 'Galaxy' so use that.
+   * @param descriptorType
+   * @returns {string}
+   */
+  constructor(private descriptorLanguageService: DescriptorLanguageService) {}
+  transform(descriptorType: string): string {
+    if (descriptorType === 'gxformat2') {
+      return this.descriptorLanguageService.workflowDescriptorTypeEnumToShortFriendlyName(Workflow.DescriptorTypeEnum.Gxformat2);
+    } else {
+      return descriptorType.toUpperCase();
+    }
+  }
+}
--- a/src/app/shared/entry/descriptor-language.pipe.ts
+++ b/src/app/shared/entry/descriptor-language.pipe.ts
@@ -7,18 +7,12 @@ import { Workflow } from '../../shared/swagger';
 })
 export class DescriptorLanguagePipe implements PipeTransform {
   /**
-   * When displaying the descriptor type typically the workflow descriptor type string is used
-   * However for Galaxy this is 'gxformat2', so change it to be Galaxy.
-   * The shortFriendlyName is 'Galaxy' so use that.
+   * Transforms DescriptorTypeEnum to their friendly names
    * @param descriptorType
    * @returns {string}
    */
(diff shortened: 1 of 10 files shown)


coverbeck (approved):
You could make this a pipe and avoid the duplication, but it's barely worth it.

svonworl (approved):

kathy-t (approved):

KimberleyChong:
> You could make this a pipe and avoid the duplication, but it's barely worth it.

Mmm, I'll actually look into that

denis-yuen (approved):

KimberleyChong:
> You could make this a pipe and avoid the duplication, but it's barely worth it.

Created new `descriptor-language` pipe and replaced instances where the previous method was being used

coverbeck (approved):
Thanks for adding the pipe!

src/app/shared/entry/descriptor-language.pipe.ts
```diff
@@ -0,0 +1,24 @@
...
+   * The shortFriendlyName is 'Galaxy' so use that.
+   * @param descriptorType
+   * @returns {string}
+   */
+  constructor(private descriptorLanguageService: DescriptorLanguageService) {}
+  transform(descriptorType: string): string {
+    if (descriptorType === 'gxformat2') {
+      return this.descriptorLanguageService.workflowDescriptorTypeEnumToShortFriendlyName(Workflow.DescriptorTypeEnum.Gxformat2);
+    } else {
+      return descriptorType.toUpperCase();
```
svonworl, line 21: Do we uppercase everything else because it's mostly acronyms (CWL, WDL, NFL)?   I wonder if we should use the "short" friendly name for all cases?  Or, create another field for the "abbreviated" friendly name in `ExtendedDescriptorLanguageBean` and use that?

svonworl (approved):

src/app/shared/entry/descriptor-language.pipe.ts
```diff
@@ -0,0 +1,24 @@
...
+  /**
+   * When displaying the descriptor type typically the workflow descriptor type string is used
+   * However for Galaxy this is 'gxformat2', so change it to be Galaxy.
+   * The shortFriendlyName is 'Galaxy' so use that.
+   * @param descriptorType
+   * @returns {string}
+   */
+  constructor(private descriptorLanguageService: DescriptorLanguageService) {}
+  transform(descriptorType: string): string {
+    if (descriptorType === 'gxformat2') {
```
kathy-t, line 18: I think you can use `Workflow.DescriptorTypeEnum.Gxformat2` here too

kathy-t (approved):

src/app/shared/entry/descriptor-language.pipe.ts
KimberleyChong: Yeah everything else should be acronyms, I think the only additional difference would be using `Nextflow` instead of `NFL`. (and `SMK` to `Snakemake`) @denis-yuen any opinion?
image
image

denis-yuen: Nextflow has approved NFL as an acronym, no idea about Snakemake. 
But from and end-user POV, Nextflow and Snakemake are probably better than the acronyms for those two languages, so:
`CWL, WDL, Galaxy, Nextflow, Snakemake` could work given that there's plenty of space on this screen. (There is less on say, the search page)

denis-yuen (approved):
With discussion above

src/app/shared/entry/descriptor-language.pipe.ts
KimberleyChong: Nextflow is only _slightly_ bigger than Galaxy on the search page, but I agree readability is better than the acronyms... I'll switch them all to friendly names.

image

PR #1713, opened 2023-02-17
Dashboard UI Fixes - SEAB 4712/SEAB 5198/SEAB 4713

KimberleyChong:
**Description**
Mix of Dashboard related UI fixes so let's dive in:

**My-Sidebar:**

Dividing line was missing the purple highlighting when visiting the dashboard page.  The `routerLink` needed to be updated, and cleaning up I removed the `[routerLinkActiveOptions]="{ exact: true }"` since that was needed originally when the dashboard button would link to the homepage (ie. routerLink="" ).
| Before | After |
|:---:|:---:|
| image | image | 

Added the "User Guide" link to the sidebar:
| Before | After |
|:---:|:---:|
| image | image | 

**Navbar:**

"My Dockstore" button was using the wrong icon, should be the smaller light purple one:

| Before | After | Mocks
|:---:|:---:|:---:|
| image | image | image

Small updates to the new dropdown, bolded text, new "Log out" icon, colour changes and some additional padding.

| Before | After | Mocks
|:---:|:---:|:---:|
| image | image | image

Looking at it now, there is no line in the mocks, so this `mat-divider` should be removed:
https://github.com/dockstore/dockstore-ui2/blob/385cce574745a7d079138a907900750be6f06650/src/app/navbar/navbar.component.html#L105

I'll wait for any additional feedback before pushing anything.

**Organization Box:**

Pending request/invitation icons should be greyed out if equal to zero.
| Before | After |
|:---:|:---:|
| image | image |

Noticed pending requests/invitations weren't updating so I changed them to `observables`, updating as expected now.
https://github.com/dockstore/dockstore-ui2/blob/385cce574745a7d079138a907900750be6f06650/src/app/home-page/widget/organization-box/organization-box.component.ts#L23-L24
https://github.com/dockstore/dockstore-ui2/blob/385cce574745a7d079138a907900750be6f06650/src/app/home-page/widget/organization-box/organization-box.component.ts#L42-L43
image

Empty card was updated with the circle org icon:
| Before | After |
|:---:|:---:|
| image | image | 

Added the search bar (similar implementation to `entry-box.component`), changed "Recent Activity" to "My Organizations", and switched from using the organization `Name` to the `DisplayName`.

| Before | After |
|:---:|:---:|
| image | image |

Changes to the Org box still aren't complete, we have a ticket to add a new API that returns a user's organization events (SEAB-5266) and a follow-up ticket to use that API and display a user's organization events following the mocks (SEAB-5267).

Last change, the pending requests/invitations links were updated to go to the "Requests" tab of the account page:
`routerLink="/accounts" [queryParams]="{ tab: 'requests' }"`

Same for the link on this organization pop-up which previously linked to the default tab on the accounts page:
image

Finally, just an overview of the current state of the Dashboard:
image

**Review Instructions**

**Issue**
[SEAB-4712](https://ucsc-cgl.atlassian.net/browse/SEAB-4712) 
[SEAB-5198](https://ucsc-cgl.atlassian.net/browse/SEAB-5198)
[SEAB-4713](https://ucsc-cgl.atlassian.net/browse/SEAB-4713)

**Security**

[SEAB-4712]: https://ucsc-cgl.atlassian.net/browse/SEAB-4712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/home-page/widget/organization-box/organization-box.component.ts
```diff
@@ -38,45 +39,22 @@ export class OrganizationBoxComponent extends Base implements OnInit {
...
-          myOrgs.forEach((org: OrganizationUpdateTime) => {
-            if (this.listOfOrganizations.length < 5) {
-              this.listOfOrganizations.push(org);
-            }
-            if (this.firstCall) {
-              this.totalOrgs += 1;
-            }
+          myOrgs.forEach(() => {
+            this.listOfOrganizations = myOrgs.slice(0, 7);
+            this.totalOrgs = myOrgs.length;
```
coverbeck, line 57: This is going to be the total of filtered organizations you belong to, not the total organizations you belong to. So if belong to at least one org, but type in a filter that doesn't match any of your orgs, you're going to get the message `You are not in any organizations`, which seems wrong.

src/app/home-page/widget/organization-box/organization-box.component.html
```diff
@@ -42,22 +42,31 @@
...
           <div class="date-display">{{ org.lastUpdateDate | date: 'MMM d, yyyy' }}</div>
         </div>
       </div>
       <ng-template #noOrganizations>
         <div *ngIf="!isLoading" fxLayout="column" class="h-100" fxLayoutAlign="center center">
-          <img src="../assets/svg/sub-nav/organization.svg" alt="my organizations" class="widget-box-type-logo mb-4" />
+          <img src="../../../../assets/svg/organization-circle.svg" alt="my organizations" class="widget-box-type-logo mb-4" />
           <span>You are not in any organizations </span>
           <div class="size-small text-align-center">
             <span>Use the </span>
```
coverbeck, line 72: For line 67/76, I can't comment there, and it was already like this, but my IDE is flagging the id `requestOrganizationButton` as a duplicate id. Looks like the id isn't referenced anywhere, so I think you can just remove both ids.

src/app/home-page/widget/organization-box/organization-box.component.ts
```diff
@@ -38,45 +39,22 @@ export class OrganizationBoxComponent extends Base implements OnInit {
...
         (myOrgs: Array<OrganizationUpdateTime>) => {
-          this.listOfOrganizations = [];
-          myOrgs.forEach((org: OrganizationUpdateTime) => {
-            if (this.listOfOrganizations.length < 5) {
-              this.listOfOrganizations.push(org);
-            }
-            if (this.firstCall) {
-              this.totalOrgs += 1;
-            }
+          myOrgs.forEach(() => {
```
coverbeck, line 55: You don't need the `forEach()` at all, the body currently just repeats the same operation each time.

src/app/home-page/widget/organization-box/organization-box.component.ts
```diff
@@ -38,45 +39,22 @@ export class OrganizationBoxComponent extends Base implements OnInit {
...
+            this.listOfOrganizations = myOrgs.slice(0, 7);
+            this.totalOrgs = myOrgs.length;
           });
         },
         (error: HttpErrorResponse) => {
```
coverbeck, line 60: It was like this, but you could set `this.listOfOrganizations = []` in the error block to handle this case:

1. User's organizations are successfully fetched
2. User types in some text and there's an error fetching the organizations, e.g., the network connection is lost, web service fails, etc.
3. We display the snackbar message, but the organizations from step are still displayed.

You could argue it's more user-friendly to keep the old orgs displayed, and maybe it is. On the other hand, our error messages aren't very prominent and the user might be surprised to see orgs that don't match the filter. So your call :)

src/app/home-page/widget/organization-box/organization-box.component.ts
KimberleyChong: Yeah no I agree the `forEach()` is redundant, I was following the other dashboard boxes for consistency and couldn't understand why the `forEach()` was there...:
https://github.com/dockstore/dockstore-ui2/blob/b7ddce0c20b7c4991750cfccd039dbe225a91d74/src/app/home-page/widget/entry-box/entry-box.component.ts#L93-L103

I couldn't explain why at the time but when I removed the `forEach()` the search box would go blank when no results returned (From your previous comment I see its because `this.totalOrgs` is set to zero!!):

https://user-images.githubusercontent.com/97123241/219804286-6c4d4cd1-7ee7-4703-a0a4-8d59fa8a4395.mp4

However, with the `forEach()` this is avoided somehow... now I see that it's because the `forEach()` never runs (because `myOrgs` is null) and so the previous state of `this.listOfOrganizations` and `this.totalOrgs` is never reassigned.

https://user-images.githubusercontent.com/97123241/219804487-cd28d349-e7eb-4860-a597-197285c5cc44.mp4

I'll remove the `forEach()`, then addressing your previous comment, assigning `this.totalOrgs` correctly will prevent the card from going to `You are not in any organizations`.

KimberleyChong:
Updates to the `organization-box.component` following Charles' feedback: `totalOrgs` remains constant,  removed `forEach()`, added `this.listOfOrganizations = [];` in the error block, and added a "No matching search results." text when returned result is empty.

Quick demo of the new behaviour:


https://user-images.githubusercontent.com/97123241/220383586-f7864e86-949e-4bd8-a348-7d10616a465a.mp4

Also made a followup ticket SEAB-5268 to update how search is handled in the `entry-box.component` to match.

src/app/home-page/widget/organization-box/organization-box.component.html
```diff
@@ -52,6 +51,7 @@
           />
         </mat-form-field>
         <h5 class="weight-bold mt-0">My Organizations</h5>
+        <div *ngIf="listOfOrganizations.length === 0" class="text-align-center date-display mt-5">No matching search results.</div>
```
coverbeck, line 54: I think his is going to "flash"; where when the page first loads, it will display this message, the request will complete, then this will go away. It would be nice if it didn't.

coverbeck (approved):

src/app/home-page/widget/organization-box/organization-box.component.ts
```diff
@@ -19,10 +20,11 @@ import { AlertService } from 'app/shared/alert/state/alert.service';
...
   Dockstore = Dockstore;
   listOfOrganizations: Array<OrganizationUpdateTime> = [];
-  firstCall = true;
+  pendingRequests$: Observable<Array<OrganizationUser>>;
+  pendingInvites$: Observable<Array<OrganizationUser>>;
+  filterText: string;
   totalOrgs: number = 0;
-  pendingRequests: Array<OrganizationUser> = [];
-  pendingInvites: Array<OrganizationUser> = [];
+  firstCall: Boolean = true;
```
coverbeck, line 27: https://stackoverflow.com/questions/64443288/what-is-the-difference-between-boolean-and-boolean-in-typescript -- you should use `boolean` instead.

KimberleyChong: Ah neat, didn't know that thanks

src/app/home-page/widget/organization-box/organization-box.component.html
KimberleyChong: Adding `&& !isLoading` to the `*ngIf` which avoids the problem, good detail thanks. Making a note to update the other `entry-boxes`

src/app/my-sidebar/my-sidebar.component.html
```diff
@@ -72,9 +65,12 @@
   </button>
   <hr routerLink="/accounts" routerLinkActive="active" />
 
+  <!-- User Guide -->
```
denis-yuen, line 68: Think docs.dockstore.org may be more of a user guide, the quick start is really just a quick-start. Since we have docs up in the header, think this should say "Quickstart"

denis-yuen (approved):
> I'll wait for any additional feedback before pushing anything.

 looking good so far

> Finally, just an overview of the current state of the Dashboard:
 
Looks good, and thanks for catching all the details!

svonworl (approved):

kathy-t (approved):

KimberleyChong:
Last change: "User Guide" updated to "Quickstart":

image

PR #1714, opened 2023-02-17
Collapse Search Panels - Dock 2317

KimberleyChong:
**Description**
Following the [mocks](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/5fc690e874e0c807afeb299b), the search sidebar has a new default state in which about half of the panels are collapsed.  A nice little change which avoids that long scroll to the bottom of the page.  As Charles suggested in the [ticket](https://ucsc-cgl.atlassian.net/browse/DOCK-2317), the state of the panels is saved to local storage, so upon page reload or navigation, the panels that have been collapsed or expanded by the user is remembered.

The tools search looks identical, collapsed panels follow the tools search [mocks](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/5fc7ceb4fd8d340926d62d6d):
| Workflows | Tools |
|:---:|:---:|
| image | image |

**Before:**
Very simple, all panels begin expanded, no state is saved, if you collapsed any panels on page refresh they will be expanded again.

https://user-images.githubusercontent.com/97123241/219798551-ed3d54a0-fb80-4efb-8644-75e4e2d711a9.mp4

**After:**
Half of the panels begin collapsed, when a user collapses or expands any panels it's state is remembered upon refresh and navigation. I also set the `reset` button to clear local storage of the panel's state and reset them back to their default (half collapsed) state.

https://user-images.githubusercontent.com/97123241/219798880-0ebbca57-99e7-41c7-9cc2-3cd94925eb5e.mp4

Also created a few spin-off tickets to update some of the search UI: SEAB-5258, SEAB-5257

**Review Instructions**
Check out the search panels, ensure some are collapsed by default and state is saved.

**Issue**
DOCK-2317

**Security**

### DIFF
--- a/src/app/search/search.component.html
+++ b/src/app/search/search.component.html
@@ -74,7 +74,6 @@
         ></app-basic-search>
         <mat-expansion-panel
           *ngFor="let key of getKeys(orderedBuckets)"
-          #bucket
           [expanded]="expandedPanels.get(key)"
           (afterExpand)="updateExpandedPanels(key, true)"
           (afterCollapse)="updateExpandedPanels(key, false)"
--- a/src/app/search/search.component.html
+++ b/src/app/search/search.component.html
@@ -58,7 +58,7 @@
             fxFlex="1 0 auto"
             class="m-1 private-btn mat-accent small-btn-structure"
             mat-button
-            (click)="resetFilters()"
+            (click)="resetFilters(); resetExpansionPanels()"
             type="button"
           >
             <img src="../assets/svg/reset.svg" alt="Reset filters" class="pb-1 pr-1" />
@@ -72,7 +72,13 @@
           (changedDebounced)="handleChangedDebounced($event)"
           (submitted)="handleSubmitted($event)"
         ></app-basic-search>
-        <mat-expansion-panel *ngFor="let key of getKeys(orderedBuckets)" #bucket expanded>
+        <mat-expansion-panel
+          *ngFor="let key of getKeys(orderedBuckets)"
+          #bucket
+          [expanded]="expandedPanels.get(key)"
+          (afterExpand)="updateExpandedPanels(key, true)"
+          (afterCollapse)="updateExpandedPanels(key, false)"
+        >
           <mat-expansion-panel-header>
             {{ friendlyNames.get(key) }}
             <img
--- a/src/app/search/state/search.service.ts
+++ b/src/app/search/state/search.service.ts
@@ -586,7 +586,7 @@ export class SearchService {
    */
   initializeExpandedPanels() {
     if (localStorage.getItem(this.expandedPanelsStorageKey)) {
-      return new Map<string, Boolean>(JSON.parse(localStorage.getItem(this.expandedPanelsStorageKey)));
+      return new Map<string, boolean>(JSON.parse(localStorage.getItem(this.expandedPanelsStorageKey)));
     } else {
       return new Map([
         ['descriptorType', true],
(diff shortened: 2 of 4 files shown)


src/app/search/search.component.html
```diff
@@ -72,7 +72,13 @@
           (changedDebounced)="handleChangedDebounced($event)"
           (submitted)="handleSubmitted($event)"
         ></app-basic-search>
-        <mat-expansion-panel *ngFor="let key of getKeys(orderedBuckets)" #bucket expanded>
+        <mat-expansion-panel
+          *ngFor="let key of getKeys(orderedBuckets)"
+          #bucket
```
coverbeck, line 77: What is `#bucket` for? It was there before, but just curious.

src/app/search/state/search.service.ts
```diff
@@ -580,6 +581,35 @@ export class SearchService {
     ]);
   }
 
+  /**
+   * Initialize expanded panels to default state or restore previous state from local storage
+   */
+  initializeExpandedPanels() {
+    if (localStorage.getItem(this.expandedPanelsStorageKey)) {
+      return new Map<string, Boolean>(JSON.parse(localStorage.getItem(this.expandedPanelsStorageKey)));
```
coverbeck, line 589: Use boolean, if possible; see my comment in your other PR.

src/app/search/state/search.service.ts
```diff
@@ -580,6 +581,35 @@ export class SearchService {
...
+    } else {
+      return new Map([
```
coverbeck, line 591: Looking at search.service.ts, we now have already had  4 maps where we initialize panels/buckets, now we have 5. It seems like they could be all combined into one data structure, but that would probably be a pretty massive change, so I don't know if it's worth it.

coverbeck (commented):
Generally looks great. Since it's a draft PR, I just put my question out there about a bigger refactoring, but not a blocker. Also, what will happen in the following scenario?

1. In 1.14, a user A's expanded panels are saved to local storage
2. We add a new facet in 1.15
3. User A goes to search page for the first time in 1.15; Dockstore loads from local storage, which doesn't have the 1.15 facet.

I didn't trace through the code -- will it handle this gracefully?

KimberleyChong:
> I didn't trace through the code -- will it handle this gracefully?

Yeah I believe it should, if we look through local storage and see a panel that maps to `true` then we add the `expanded` attribute to that panel's html so it begins expanded.  If we add a new panel in the future where there is no mapping to local storage, the `expanded` attribute won't be added and the panel will just begin closed.

src/app/search/search.component.html
KimberleyChong: Good point, I have no idea... can't find any reference to it and removing it doesn't seem to do anything.  Looks like it was introduced [here](https://github.com/dockstore/dockstore-ui2/pull/48/files#diff-4b285e04baef6b0ed18c0c0501ca161ba52f82e37927b3f22689840438778c26R30) in 2017. I'm gonna remove it

src/app/search/state/search.service.ts
KimberleyChong: Yeah, big change but we can probably make a ticket if this is something we want to do

svonworl (approved):

kathy-t (approved):

denis-yuen (approved):

src/app/search/state/search.service.ts
coverbeck: On further reflection, it doesn't have to be that big of a change. We could have one data structure, and derive all the maps from it, and not have to change any of the HTML. Please do create a ticket for it.

coverbeck (approved):

KimberleyChong:
> Please do create a ticket for it.

Follow-up ticket: SEAB-5271

PR #1715, opened 2023-02-21
Fix Broken Smoke Test for Staging/Prod  - SEAB-5269

KimberleyChong:
**Description**
Previously updated the smoke tests in this [pr](https://github.com/dockstore/dockstore-ui2/pull/1707/files#diff-0e83ac8adb043831f00e87fd557c3cbe8a82257c2ca3d5772e2a31d01dbc77feR110) to look for `Nextflow` instead of `NFL` since we switched to the friendly names, however the change in smoke test does not accommodate for staging/develop where these changes from develop are not present.  Updated the test to handle both cases of `NFL` and `Nextflow`, tested locally with `NFL` and `Nextflow` and appears to be passing well.

Added an additional note [here](https://github.com/dockstore/dockstore/issues/5372#issuecomment-1439116907) to remove the check for "NFL" after the 1.14 release.

**Review Instructions**

**Issue**
SEAB-5269

**Security**


coverbeck (approved):
Thanks for addressing this so quickly.

svonworl (approved):

kathy-t (approved):

denis-yuen (approved):

PR #1719, opened 2023-02-28
Update Entry Box Search - SEAB-5268

KimberleyChong:
**Description**
Updates search on `entry-box.component` to handle empty search results.

**Before:**

https://user-images.githubusercontent.com/97123241/221948529-f5616d21-16aa-4ffd-a5ec-3bf73806f855.mp4

**After:**

https://user-images.githubusercontent.com/97123241/221948545-acaa1221-f85d-48ae-94ce-1a3b33036e21.mp4

**Review Instructions**
Empty search results should display message.

**Issue**
SEAB-5268

**Security**

### DIFF
--- a/src/app/home-page/widget/entry-box/entry-box.component.html
+++ b/src/app/home-page/widget/entry-box/entry-box.component.html
@@ -58,8 +58,8 @@
           </a>
           <div class="date-display">{{ entry.lastUpdateDate | date: 'MMM d, yyyy' }}</div>
         </div>
-        <div *ngIf="listOfEntries.length === 0" class="mt-2">No matching {{ entryTypeLowerCase }}s</div>
-        <div class="weight-bold py-2">
+        <div *ngIf="listOfEntries.length === 0" class="text-align-center date-display mt-5">No matching {{ entryTypeLowerCase }}s.</div>
+        <div *ngIf="listOfEntries.length > 0" class="weight-bold py-2">
           <a [routerLink]="allEntriesLink" data-cy="all-entries-btn">
             All {{ entryTypeLowerCase | titlecase }}s<mat-icon class="mat-icon-sm">keyboard_double_arrow_right</mat-icon>
           </a>
--- a/src/app/home-page/widget/entry-box/entry-box.component.ts
+++ b/src/app/home-page/widget/entry-box/entry-box.component.ts
@@ -48,6 +48,7 @@ export class EntryBoxComponent extends Base implements OnInit {
   public isLoading = true;
   userEntries$: Observable<EntryUpdateTime[]>;
   entryTypeParam: any;
+  firstCall: boolean = true;
 
   constructor(
     private registerToolService: RegisterToolService,
@@ -91,10 +92,11 @@ export class EntryBoxComponent extends Base implements OnInit {
       )
       .subscribe(
         (myEntries: Array<EntryUpdateTime>) => {
-          myEntries.forEach(() => {
-            this.listOfEntries = myEntries.slice(0, 7);
+          this.listOfEntries = myEntries.slice(0, 7);
+          if (this.firstCall) {
             this.totalEntries = myEntries.length;
-          });
+            this.firstCall = false;
+          }
         },
         (error: HttpErrorResponse) => {
           this.alertService.detailedError(error);


src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -91,10 +92,11 @@ export class EntryBoxComponent extends Base implements OnInit {
       )
       .subscribe(
         (myEntries: Array<EntryUpdateTime>) => {
-          myEntries.forEach(() => {
-            this.listOfEntries = myEntries.slice(0, 7);
+          this.listOfEntries = myEntries.slice(0, 7);
+          if (this.firstCall) {
```
coverbeck, line 96: Because the requests are asynchronous, there's no guarantee you'll get the response from the first request first. It's extremely unlikely in this scenario, given the likely lag between the page loading and a character being typed, but still possible. I know there was code like this before, but in hindsight it wasn't quite right.

Example:

1. Page loads, request to get entries without filter starts
2. Network is or becomes super slow, on cellular
3. User types in a filter string, a request to get (filtered) entries starts
4. Network switched over to high speed wifi, response for 3 comes back before response for 1.

One way to solve it would be to get the HttpResponse, like [here](https://github.com/dockstore/dockstore-ui2/blob/develop/src/app/workflows/list/published-workflows.datasource.ts#L61), then check to see if the filterText was set in the request. There might be other ways.

KimberleyChong:
> One way to solve it would be to get the HttpResponse, like [here](https://github.com/dockstore/dockstore-ui2/blob/develop/src/app/workflows/list/published-workflows.datasource.ts#L61), then check to see if the filterText was set in the request.

Spoke with Charles about this, updated to recieve the HttpResponse and put in a check for the `filter` param in the request url to ensure it is unset or empty before assigning `this.totalEntries`. Went with using `if (!url.searchParams.get('filter'))` because it handles requests without the filter param (after page load, returns null) and with an empty filter param (i.e. `?filter=` for when doing an empty search, returns empty string)

https://github.com/dockstore/dockstore-ui2/blob/1d9f6a58b4bf2f63ef6f878952f53faaaa3ded48/src/app/home-page/widget/entry-box/entry-box.component.ts#L90-L101

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -82,19 +80,20 @@ export class EntryBoxComponent extends Base implements OnInit {
   }
 
   getMyEntries() {
-    this.userEntries$ = this.usersService.getUserEntries(null, this.filterText, this.entryTypeParam);
```
coverbeck, line 85: I'm glad you got rid of this; I was going to comment on it before, but didn't want to get too nit-picky when I was already commenting on an unlikely timing edge case. :) But yeah, we aren't using it, so this is good.

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -82,19 +80,20 @@ export class EntryBoxComponent extends Base implements OnInit {
...
       .subscribe(
-        (myEntries: Array<EntryUpdateTime>) => {
-          myEntries.forEach(() => {
-            this.listOfEntries = myEntries.slice(0, 7);
-            this.totalEntries = myEntries.length;
-          });
+        (myEntries: HttpResponse<EntryUpdateTime[]>) => {
+          this.listOfEntries = myEntries.body.slice(0, 7);
+          const url = new URL(myEntries.url);
+          if (!url.searchParams.get('filter')) {
```
coverbeck, line 94: Suggestion: you could have a comment explaining what's this for; I wonder if it's only obvious to us because we've been discussing it. Or maybe it's obvious to all.

src/app/home-page/widget/organization-box/organization-box.component.ts
```diff
@@ -46,17 +45,17 @@ export class OrganizationBoxComponent extends Base implements OnInit {
...
       .subscribe(
-        (myOrgs: Array<OrganizationUpdateTime>) => {
-          this.listOfOrganizations = myOrgs.slice(0, 7);
-          if (this.firstCall) {
-            this.totalOrgs = myOrgs.length;
-            this.firstCall = false;
+        (myOrgs: HttpResponse<OrganizationUpdateTime[]>) => {
+          this.listOfOrganizations = myOrgs.body.slice(0, 7);
+          const url = new URL(myOrgs.url);
+          if (!url.searchParams.get('filter')) {
```
coverbeck, line 57: Same suggestion on comment.

coverbeck (approved):

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -82,19 +80,22 @@ export class EntryBoxComponent extends Base implements OnInit {
...
       )
       .subscribe(
-        (myEntries: Array<EntryUpdateTime>) => {
-          myEntries.forEach(() => {
-            this.listOfEntries = myEntries.slice(0, 7);
-            this.totalEntries = myEntries.length;
-          });
+        (myEntries: HttpResponse<EntryUpdateTime[]>) => {
+          this.listOfEntries = myEntries.body.slice(0, 7);
+          // Update total entries only when no search filter applied (i.e. non-filtered total)
```
kathy-t, line 93: Why are we only updating the total entries when there's no search filter? Wouldn't it be helpful if we displayed the number of search results?

src/app/home-page/widget/organization-box/organization-box.component.ts
```diff
@@ -46,17 +45,19 @@ export class OrganizationBoxComponent extends Base implements OnInit {
...
       )
       .subscribe(
-        (myOrgs: Array<OrganizationUpdateTime>) => {
-          this.listOfOrganizations = myOrgs.slice(0, 7);
-          if (this.firstCall) {
-            this.totalOrgs = myOrgs.length;
-            this.firstCall = false;
+        (myOrgs: HttpResponse<OrganizationUpdateTime[]>) => {
+          this.listOfOrganizations = myOrgs.body.slice(0, 7);
+          // Update total orgs only when no search filter applied (i.e. non-filtered total)
```
kathy-t, line 56: Same comment

src/app/home-page/widget/entry-box/entry-box.component.ts
KimberleyChong: This is true and really up to preference, https://github.com/dockstore/dockstore-ui2/pull/1713#discussion_r1110367559 is why we originally decided to keep `totalOrgs`/ `listOfEntries` static.

But in turn, the UI displays `listOfEntries`, so now that it's static, it doesn't change... whether we want the UI to display the the number of search results instead is up for debate @coverbeck @denis-yuen 

I can change `this.listOfEntries = myEntries.body.slice(0, 7);` to `this.listOfEntries = myEntries.body;` then display `this.listOfEntries.length` in the bubble as the new total and use the slice pipe in the `*ngFor` to display the first 7 entries.

Just checked the mocks [here](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6022abd3355a9d26154a8d95) and the search bar is handled a bit differently, with the results being in a dropdown and the total displayed being unchanged. Whether we want to go with that option is also up for discussion.

image

coverbeck: > This is true and really up to preference, https://github.com/dockstore/dockstore-ui2/pull/1713#discussion_r1110367559 is why we originally decided to keep totalOrgs/ listOfEntries static.

My comment was pointing out that you could end up displaying a confusing "You are in no organizations" when the user does belong to at least one org. I took no stance there on which way to solve it. :) 

But I do lean towards displaying the total count.

Since we're discussing it, the other confusing thing that the mocks don't cover is what portion of the filtered results you are seeing, i.e., if the search finds 10 results, we only display 7 and don't tell you there are more. In this case, I like what we do in the search page, with the "x more", although I don't think in this case we need to make it an expandable link. Plus it's not in the mocks.

Screen Shot 2023-03-03 at 1 21 47 PM

In summary:

* I lean towards displaying the total
* I'd like it if we showed how many the filter is matching; if it's more than 7, show how many matches aren't being shown, otherwise just show the matches.

Which I guess is saying I'd like to see both :).

But I don't feel super strongly, and could be easily convinced either way.

kathy-t: > Just checked the mocks [here](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6022abd3355a9d26154a8d95) and the search bar is handled a bit differently, with the results being in a dropdown and the total displayed being unchanged. Whether we want to go with that option is also up for discussion.

I like it :slightly_smiling_face: feels like the best of both worlds, you can see the total number of entries, and also all of the search results in the drop down. I think currently, you don't see all search results

KimberleyChong: Those are all really great points, I guess I also agree, the mocks are probably the best of both worlds

KimberleyChong: Just playing around, going with the dropdown idea, any preference on introducing a scroll bar to limit the number of search results shown or display the total results list no matter the length?

We can also style the scroll to be less of an eye sore

image

denis-yuen: I think the mock-scroll-bar is more of an auto-complete than a way of showing results (i.e. would probably skip the dates). Otherwise, appreciate the conversation so far and don't have too much to add.

KimberleyChong:
**Changes:** Added drop-down for search and included a message to indicate the number of search results we are displaying of the total matching results (limited to 5, less than 5 will hide the message).

**With message:**
image

**Without message:**
image

**Demo**

https://user-images.githubusercontent.com/97123241/223242932-47c12925-f23d-43b0-bf27-871f160f2d57.mp4

KimberleyChong:
On a related note, did not add the drop-down to the organization box, we have a [ticket](https://ucsc-cgl.atlassian.net/browse/SEAB-5267) to display events following the mocks which will be implemented soon, but I think it would make more sense to  display search results in the card rather than a drop-down because there is more information to display than just an entry name:

image

The mocks do not show how search is handled here, but we can always discuss how we will handle this when I pick up that ticket.

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -82,19 +81,24 @@ export class EntryBoxComponent extends Base implements OnInit {
...
       .subscribe(
-        (myEntries: Array<EntryUpdateTime>) => {
-          myEntries.forEach(() => {
-            this.listOfEntries = myEntries.slice(0, 7);
-            this.totalEntries = myEntries.length;
-          });
+        (myEntries: HttpResponse<EntryUpdateTime[]>) => {
+          const url = new URL(myEntries.url);
+          if (url.searchParams.get('filter')) {
+            this.listOfResults = myEntries.body;
```
svonworl, line 95: If there's lots of filtered results, do we want to `slice` them also, to limit the number?

src/app/home-page/widget/entry-box/entry-box.component.html
```diff
@@ -43,22 +43,29 @@
...
+          <mat-label>Search your {{ entryTypeLowerCase | titlecase }}s...</mat-label>
+          <input matInput data-cy="search-field" [(ngModel)]="filterText" (keyup)="onTextChange($event)" [matAutocomplete]="auto" />
+          <button *ngIf="filterText" mat-button matSuffix mat-icon-button (click)="clearSearch()">
+            <mat-icon>cancel</mat-icon>
+          </button>
         </mat-form-field>
+        <mat-autocomplete #auto="matAutocomplete">
+          <span>
+            <a *ngFor="let result of listOfResults | slice: 0:5" [routerLink]="allEntriesLink + result.path" class="no-wrap no-underline">
+              <mat-option>{{ result.prettyPath }}</mat-option>
```
coverbeck, line 55: I wonder if it's worth showing the organization here?

Trying it locally, if I type an organization name, e.g., `coverbeck` or `dockstore`, two GitHub orgs I belong to, I see several matching results in the UI, but I don't see `coverbeck` or `dockstore` displayed. It was a little confusing -- why is it matching something that doesn't look like it matches?

In the screenshot, all the workflows belong to the org `coverbeck`, which is why they are displayed.

Screen Shot 2023-03-07 at 2 20 56 PM

On the other hand, maybe most users don't belong to multiple repos and typically won't search by org name. And there's a reason the repetitive org name is removed from the mocks :). Still... I don't know, I'll leave it for you to ponder. :)

coverbeck (commented):
I left a similar comment on Nayeon's PR, even though she's using a different component than you. Maybe there's something with my computer. :) But, I'm not able to act on a search item by navigating to it with the keyboard:

1. Type "helloworld"
2. Press down arrow to select the first result, doesn't select it.
3. I have to click on the first result.

Expectation:

I should be able to down arrow to my choice. See Google search as an example.

src/app/home-page/widget/entry-box/entry-box.component.ts
KimberleyChong: Good idea since we only display a max of 5 results anyway, will just add in a `totalResults` varaible so we can still display the number of entries matched

KimberleyChong:
> I should be able to down arrow to my choice. See Google search as an example.

Thanks for pointing that out, had no idea!  Issue was _every_ keyboard event triggered `getMyEntries()`, putting in a check to ignore arrow keys and replacing the `routerlink` with a navigation function so it works both on `click` and `enter`.

src/app/home-page/widget/entry-box/entry-box.component.html
denis-yuen: Well, the full path with "github.com" is obviously redundant in this case. I do see the argument for displaying organisation.  In an ideal world, I could see it displaying the organization only when you have conflicts (i.e. if you were searching `md5sum` and you needed the organization to distinguish between `coverbeck/md5sum` and `dockstore/md5sum`. 

Right now, I think that's probably over-engineering things, so I'm inclined to open the ticket (on how to display search results, with the ideas here and above), see if there are user requests, and only improve this on user request.  

The bigger question right now is, if you have multiple workflows with the same name, how does it display in the results below the search box once you press enter?

KimberleyChong: I suppose we could remove the github.com part and just display `coverbeck/md5sum` and `dockstore/md5sum` part for all entries since that's also how we display entries on search:
image

> The bigger question right now is, if you have multiple workflows with the same name, how does it display in the results below the search box once you press enter?

Not sure I understand this, clicking on an entry in the search results (or pressing enter, wait for changes to be pushed) will navigate you to your workflow or tool page. If they have the same name I'm guessing they would both display in the drop-down but with the same name appearing twice. Like so?
image

coverbeck: Agreed, it's fine to wait and see, and yes, I was thinking we could also drop the `github.com`.

My use case is pretty extreme, I suspect I belong to a lot more orgs than the average Dockstore user, and maybe this will be moot most of the time.

KimberleyChong:
**Update:** Allow arrow keys for navigation


https://user-images.githubusercontent.com/97123241/223862253-8094316e-c9ae-4008-b8dc-3b486ff4aac9.mp4

Put in check to ignore arrow key events:
https://github.com/dockstore/dockstore-ui2/blob/440c988c90e2d40367000b1140f799693fa438e7/src/app/home-page/widget/entry-box/entry-box.component.ts#L50
https://github.com/dockstore/dockstore-ui2/blob/440c988c90e2d40367000b1140f799693fa438e7/src/app/home-page/widget/entry-box/entry-box.component.ts#L112-L118

Then I removed the `routerlinks` (because `enter` on a link doesn't work) and replaced it with a navigation function that fires on the `optionSelected` event. Now it handles both `enter` and `click`:
https://github.com/dockstore/dockstore-ui2/blob/440c988c90e2d40367000b1140f799693fa438e7/src/app/home-page/widget/entry-box/entry-box.component.html#L52
https://github.com/dockstore/dockstore-ui2/blob/440c988c90e2d40367000b1140f799693fa438e7/src/app/home-page/widget/entry-box/entry-box.component.ts#L120-L122

Also minimized the size of `listOfResults` and added in `totalResults` to keep track of the number of matching entries.
https://github.com/dockstore/dockstore-ui2/blob/440c988c90e2d40367000b1140f799693fa438e7/src/app/home-page/widget/entry-box/entry-box.component.ts#L97-L98

kathy-t (commented):
Looks good!

It seems like the search results don't get updated when you delete characters. When I delete all characters, there are still search results in the drop down:

https://user-images.githubusercontent.com/25287123/224048424-efe0bd59-403f-451d-a173-745fa235beb2.mp4  



I also noticed some console errors on the new dashboard related to the organization box while trying it out locally. Possibly related to this PR?

image

KimberleyChong:
> I also noticed some console errors on the new dashboard related to the organization box while trying it out locally. Possibly related to this PR?

Nope, actually I've seen those there for a while, I've been meaning to write it down somewhere but I'm planning on clearing them up  when I go to work on the org box again (SEAB-5267) since the implementation is going to be majorly changed.  I'll leave a note on the ticket

src/app/home-page/widget/entry-box/entry-box.component.ts
```diff
@@ -80,19 +83,25 @@ export class EntryBoxComponent extends Base implements OnInit {
...
-        (myEntries: Array<EntryUpdateTime>) => {
-          myEntries.forEach(() => {
-            this.listOfEntries = myEntries.slice(0, 7);
-            this.totalEntries = myEntries.length;
-          });
+        (myEntries: HttpResponse<EntryUpdateTime[]>) => {
+          const url = new URL(myEntries.url);
+          if (url.searchParams.get('filter')) {
+            this.listOfResults = myEntries.body.slice(0, 5);
+            this.totalResults = myEntries.body.length;
```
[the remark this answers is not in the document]
kathy-t: Oops, I see it now :grin:

src/app/home-page/widget/entry-box/entry-box.component.html
```diff
@@ -43,22 +43,25 @@
...
-            (keyup)="onTextChange($event)"
-            placeholder="Search your {{ entryTypeLowerCase }}s..."
-          />
+          <mat-label>Search your {{ entryTypeLowerCase | titlecase }}s...</mat-label>
+          <input matInput data-cy="search-field" [(ngModel)]="filterText" (keyup)="onTextChange($event)" [matAutocomplete]="auto" />
+          <button *ngIf="filterText" mat-button matSuffix mat-icon-button (click)="clearSearch()" type="button">
+            <mat-icon>cancel</mat-icon>
+          </button>
         </mat-form-field>
+        <mat-autocomplete #auto="matAutocomplete" (optionSelected)="navigateToEntry(value)">
```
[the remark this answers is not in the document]
KimberleyChong: 😯 Thank you!!!

Two problems: `navigateToEntry` was always receiving `undefined` because I wasn't referencing the `mat-option` value correctly. Updated to:`(optionSelected)="navigateToEntry($event.option.value)"`. Secondly, need to navigate using the full `path` not `prettyPath`: `[value]="result.path"` and `this.router.navigate([this.allEntriesLink + path]);`

src/app/home-page/widget/entry-box/entry-box.component.html
```diff
@@ -43,22 +43,25 @@
...
+          </div>
+        </mat-autocomplete>
         <h5 class="weight-bold my-0">Recent Activity</h5>
         <div *ngFor="let entry of listOfEntries" fxLayout="row" class="widget-list-items space-between w-100 py-2">
           <a class="no-wrap" [routerLink]="allEntriesLink + entry.path">
             {{ entry.prettyPath }}
           </a>
           <div class="date-display">{{ entry.lastUpdateDate | date: 'MMM d, yyyy' }}</div>
         </div>
-        <div *ngIf="listOfEntries.length === 0" class="mt-2">No matching {{ entryTypeLowerCase }}s</div>
```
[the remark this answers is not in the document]
KimberleyChong: Sure, we can display like this:

image

svonworl (approved):

src/app/home-page/widget/entry-box/entry-box.component.html
```diff
@@ -43,22 +43,26 @@
...
+          <mat-label>Search your {{ entryTypeLowerCase | titlecase }}s...</mat-label>
+          <input matInput data-cy="search-field" [(ngModel)]="filterText" (keyup)="onTextChange($event)" [matAutocomplete]="auto" />
+          <button *ngIf="filterText" mat-button matSuffix mat-icon-button (click)="clearSearch()" type="button">
+            <mat-icon>cancel</mat-icon>
+          </button>
         </mat-form-field>
+        <mat-autocomplete #auto="matAutocomplete" (optionSelected)="navigateToEntry($event.option.value)">
+          <mat-option *ngFor="let result of listOfResults" [value]="result.path" class="no-wrap">{{ result.prettyPath }}</mat-option>
+          <div *ngIf="totalResults > 5" class="date-display no-wrap p-2 px-4">
+            Showing 5 of {{ totalResults }} matching {{ entryTypeLowerCase }}s
```
denis-yuen, line 55: Can the 5 here be a variable shared with `this.listOfResults = myEntries.body.slice(0, 5);` (different file) below? 
Seems like the kind of thing where someone may change this and forget to change the other one

denis-yuen (approved):
Tiny change if simple and/or going back-in for some other reason. 
Thanks for the video and addressing all the other feedback!

kathy-t (approved):

PR #1720, opened 2023-03-02
DOCK-1525: reordering version dropdown in my workflow

hyunnaye:
**Description**
This PR adds a search feature onto the version dropdown in My Workflows and reorders the dropdown to alphabetical order.
Library used: https://www.npmjs.com/package/ngx-mat-select-search 

Before change:
Screenshot from 2023-03-03 13-47-40

After change (search bar and alphabetical sorting):
Screenshot from 2023-03-08 16-19-50
Screenshot from 2023-03-08 16-19-58
Screenshot from 2023-03-08 16-20-22

**Review Instructions**
Review that the newly created integration test pass and the search feature correctly filters through the different versions of a workflow.

**Issue**
https://github.com/dockstore/dockstore/issues/3722

**Security**


svonworl:
Hi, Nayeon, thanks for the work on this.  Could you please add a few comments at key spots in the code, it will help me to follow it.

THIRD-PARTY-LICENSES.csv
```diff
@@ -805,6 +805,7 @@
 "make-fetch-happen@10.2.1","ISC","https://github.com/npm/make-fetch-happen"
 "make-fetch-happen@5.0.2","ISC","https://github.com/zkat/make-fetch-happen"
 "marked@4.2.3","MIT","https://github.com/markedjs/marked"
+"mat-select-search@1.2.21","Custom: https://github.com/shafi-sahal/MatSelectSearch",""
```
denis-yuen, line 808: [Was a big thread ](https://github.com/angular/components/issues/5697), didn't fully read through. 
Did you consider https://github.com/bithost-gmbh/ngx-mat-select-search which seems to be more popular?
(this is a discussion, I know little about the two aside from seeing both in the thread and this one seems to have more commits/stars)

denis-yuen (commented):
question

THIRD-PARTY-LICENSES.csv
hyunnaye: I tried ngx-mat-select-search but I couldn't get it to compile due to various module errors like `ERROR in ./node_modules/ngx-mat-select-search/fesm2020/ngx-mat-select-search.mjs 842:108-124 export 'MatIconButton' (imported as 'i4') was not found in '@angular/material/button' (possible exports: MatAnchor, MatButton, MatButtonModule)`

denis-yuen: Did you try version 5? 
https://github.com/bithost-gmbh/ngx-mat-select-search#compatibility

hyunnaye: no, i'll try that out right now!

hyunnaye: changed to ngx-mat-select-search as it seemed to resolve Charles' issues above

hyunnaye:
> Looks cool!
> 
> Some observations about using the control, that you might be constrained by what customization the control allows:
> 
> 1. You have to click on the selection. For example, type `develop`, assuming there is a match; it only shows `develop`, but you then have to click on it. Could be just me, I try to use the mouse as little as possible.
> 2. If you type a matching version, then press tab (no click), you end up in a state where it doesn't display anything. Screenshot attached.
> 3. You could argue this both ways, but I don't like how it "remembers" your last search. I type develop, select it, now I want to look for master; it's still showing what I typed for `develop` beforehand.
> 
> Screen Shot 2023-03-07 at 11 24 05 AM

Thank you for this! I switched over to ngx-mat-select-search following Denis' comment and it seems to resolve this issue. I will rerequest review after verifying all tests pass in my latest commit

denis-yuen (approved):
Hmmm, so if I understand correctly, the build issues was the bad cache in the end?

hyunnaye:
> Hmmm, so if I understand correctly, the build issues was the bad cache in the end?

Yea, I changed the version on package.json and package-lock.json and that seems to fix it.. Very odd

denis-yuen:
> > Hmmm, so if I understand correctly, the build issues was the bad cache in the end?
> 
> Yea, I changed the version on package.json and package-lock.json and that seems to fix it.. Very odd

I think what happened is the first time the cache was generated, it was messed up and every subsequent build it pulled the messed up cache until you busted the cache with the change

src/app/workflow/workflow.component.ts
```diff
@@ -302,11 +314,38 @@ export class WorkflowComponent extends Entry implements AfterViewInit, OnInit {
...
+        });
+        this.filteredVersions.next(this.workflowVersionAlphabetical.slice());
+        this.workflowVersionsCtrl.setValue(this.selectedVersion);
       }
       this.setUpWorkflow(workflow);
     });
   }
 
+  protected setInitialValue() {
+    this.filteredVersions.pipe(takeUntil(this.ngUnsubscribe)).subscribe(() => {
```
coverbeck, line 328: I don't think you need this. Line 230 doesn't use the response from this. I find it a tad confusing, because you have a method called `setInitialValue`, but you're creating a subscription that's going to be active until the component goes away.

In a quick test locally, it seems unnecessary, but I didn't extensively test it.

coverbeck (commented):
I tried running it locally, and it looks and acts a lot nicer than the previous iteration. I did hve one question though that I left inline.

svonworl (approved):

src/app/workflow/workflow.component.ts
hyunnaye: Removed

kathy-t (approved):

src/app/workflow/workflow.component.ts
coverbeck: > Removed

Just to confirm, was it OK to also remove the `this.singleSelect.compareWith = (a: Tag | WorkflowVersion, b: Tag | WorkflowVersion) => a && b && a.id === b.id;`? It was potentially getting executed many times before with the subscription, which was too much, but now it's never getting executed.

coverbeck (approved):
I left a comment with one more question, but assuming that's resolved, approved.

src/app/workflow/workflow.component.ts
hyunnaye: Yes, it was okay to remove. I suspect our code seem to work without it since we already have the initial value set through `this.selectedVersion`

PR #1722, opened 2023-03-03
Updated Blue Banner Style

KimberleyChong:
**Description**
Updates the blue `alert-info` banner style following the zeplin mock-ups and creates a new palette for `Accent 2` as defined [here](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/5fc699d068c46f17551f745b).
New style has a blue border, light blue background and no box-shadow.
https://github.com/dockstore/dockstore-ui2/blob/cc245466e179fa7c0e80446f9f232fb26896854e/src/styles.scss#L1331-L1337
Zeplin mocks: [Dashboard](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60229885ad1715098caf03e5), [Starred](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/602c3030c7ca5718b6e37c1c)

**Dashboard:**
**Before:**
image
**After:**
image

**Starred:**
**Before:**
image

**After:**
image
In this case, following the mocks, I removed the warning icon.

**Zeplin:**
image

**General:**
For areas where we leave the icons, Zeplin didn't have any examples so I picked a matching blue from the new `Accent 2` colour palette:

**Before:**
image

**After:**
image

**Review Instructions**

**Issue**
SEAB-5201

**Security**

### DIFF
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -30,14 +30,14 @@
       (tool?.mode === ModeEnum.AUTODETECTQUAYTAGSAUTOMATEDBUILDS || tool?.mode === ModeEnum.AUTODETECTQUAYTAGSWITHMIXED) && !tool?.gitUrl
     "
   >
-    <mat-card class="alert alert-info" role="alert">
+    <mat-card class="alert alert-info mat-elevation-z" role="alert">
       <mat-icon>warning</mat-icon> This Quay.io tool does not have an associated Git repository. Please view the Quay.io repository and add
       a build trigger for a valid Git repository. Once the build has been added, refresh the tool.
     </mat-card>
   </div>
 
   <div class="row m-1" *ngIf="tool?.mode === ModeEnum.HOSTED && tool?.workflowVersions?.length === 0">
-    <mat-card class="alert alert-info" role="alert">
+    <mat-card class="alert alert-info mat-elevation-z" role="alert">
       <mat-icon>warning</mat-icon> Your manually added tool does not have any tags. Go to the versions tab to add one.
     </mat-card>
   </div>
@@ -443,7 +443,7 @@
 </div>
 <ng-template #noVersions>
   <div class="p-3">
-    <mat-card class="alert alert-info p-3" role="alert">
+    <mat-card class="alert alert-info mat-elevation-z p-3" role="alert">
       <mat-icon>info</mat-icon>
       <span *ngIf="tool?.mode === ModeEnum.HOSTED"> To see versions, please add a new version in the Files tab. </span>
       <span *ngIf="tool?.mode === ModeEnum.MANUALIMAGEPATH"> To see versions, click the Add Tag button below. </span>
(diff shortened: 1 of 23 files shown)


coverbeck (approved):

src/app/container/container.component.html
```diff
@@ -30,14 +30,14 @@
       (tool?.mode === ModeEnum.AUTODETECTQUAYTAGSAUTOMATEDBUILDS || tool?.mode === ModeEnum.AUTODETECTQUAYTAGSWITHMIXED) && !tool?.gitUrl
     "
   >
-    <mat-card class="alert alert-info" role="alert">
+    <mat-card class="alert alert-info mat-elevation-z" role="alert">
```
kathy-t, line 33: If this `mat-elevation-z` is for all `alert-info` mat-cards, then it should be added to the `alert-info` class so that future uses of `alert-info` will automatically have this style. Can investigate using the `mat-elevation` [mixin](https://v6.material.angular.io/guide/elevation#mixins) or just using `box-shadow: none`

KimberleyChong: Looked into `mat-elevation` mixin, few caveats:
- can only do elevation from 0 - 24 (not z) so its not possible to have no shadow with this method
- it applies the `box-shadow` style to the component, but without the class `mat-elevation-z` it gets overridden by the general `mat-card` shadow style (see below)
- doesn't seem to allow `!important`  but either way style doesn't work and the `!importants` would clash

I thought about `box-shadow: none !important;` when implementing but we have a problem with conflicting overrides in shadow styles by the general `mat-card` shadow style again (see below), so it doesn't work either

**General mat-card shadow style:**
https://github.com/dockstore/dockstore-ui2/blob/0007b986539ab3883bf72b9239d165e64038a725/src/styles.scss#L1315-L1318

So the only method/ probably most robust method that I found is to just include the `mat-elevation-z` class at every instance and avoid the overrides completely (`alert-warning` and `alert-danger` also follow the same method from when I updated them last)

denis-yuen (approved):
After resolving comment above

src/app/container/container.component.html
kathy-t: > can only do elevation from 0 - 24 (not z) so its not possible to have no shadow with this method

I think `mat-elevation-z` is the `0` value (no elevation)

Thanks for looking into it. It doesn't seem possible without a bit of refactoring. I think it would still be good to document that all banners should have `mat-elevation-z` so that it has a flat appearance. Perhaps you can add something to https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide about the banners?

KimberleyChong: > I think mat-elevation-z is the 0 value (no elevation)

Yeah I thought that too, but `mat-elevation(0)` adds a shadow style so I don't think they correspond in this case? Which is interesting...
| image | image |
|:--:|:--:|

>  Perhaps you can add something to https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide about the banners?

Makes sense 👍️ I'll add a note in the style guide

KimberleyChong: Note added!

kathy-t (approved):

PR #1723, opened 2023-03-04
Indicate open on versions table

coverbeck:
**Description**
Adds an Open column to the versions table, and does corresponding renaming from Open Data to Open on the facets page.

***Background on the renaming***
* Initially, we were considering an open data version to be a workflow with a test parameter file with publicly accessible test data
* Now, we are considering an open version to be a version with publicly accessible test data, **OR**, a version that does not require any file inputs. The key point being that in both cases you don't need any additional permissions to run the version.

The versions table is getting crowded, although I think it might just be acceptable. Suggestions welcome.

Before:
Screen Shot 2023-03-03 at 4 04 24 PM

After:
Screen Shot 2023-03-03 at 3 45 15 PM

***Facets***
Before
Screen Shot 2023-03-03 at 4 02 40 PM

After
Screen Shot 2023-03-03 at 4 01 46 PM

**Review Instructions**
Verify that the Open column appears on the versions table in both the public and My workflows views. Verify that the facet looks good on the search page.

We only update public accessible when updating/creating a new version (it's not retroactive), so you'll need to push new versions to test this.

**Issue**
dockstore/dockstore#5340

**Security**

- [x] If this PR is for a user-facing feature, create and link a documentation ticket for this feature (usually in the same milestone as the linked issue). Style points if you create a documentation PR directly and link that instead. dockstore/dockstore#5402

### DIFF
--- a/src/app/search/map-friendly-values.pipe.ts
+++ b/src/app/search/map-friendly-values.pipe.ts
@@ -70,8 +70,8 @@ export class MapFriendlyValuesPipe implements PipeTransform {
     [
       'openData',
       new Map([
-        ['1', "doesn't need public test data"],
-        ['0', 'needs public test data'],
+        ['1', "doesn't need more data"],
+        ['0', 'needs more data'],
       ]),
     ],
     [
--- a/src/app/search/map-friendly-values.pipe.ts
+++ b/src/app/search/map-friendly-values.pipe.ts
@@ -70,8 +70,8 @@ export class MapFriendlyValuesPipe implements PipeTransform {
     [
       'openData',
       new Map([
-        ['1', 'has public test data'],
-        ['0', 'no public test data'],
+        ['1', "doesn't need public test data"],
+        ['0', 'needs public test data'],
       ]),
     ],
     [
--- a/src/app/search/map-friendly-values.pipe.ts
+++ b/src/app/search/map-friendly-values.pipe.ts
@@ -70,8 +70,8 @@ export class MapFriendlyValuesPipe implements PipeTransform {
     [
       'openData',
       new Map([
-        ['1', "doesn't need more data"],
-        ['0', 'needs more data'],
+        ['1', 'open data'],
+        ['0', 'requires restricted data'],
       ]),
     ],
     [
--- a/src/app/search/state/search.service.ts
+++ b/src/app/search/state/search.service.ts
@@ -534,7 +534,7 @@ export class SearchService {
       ['verified_platforms.keyword', 'Verified Platforms'],
       ['categories.name.keyword', 'Category'],
       ['descriptor_type_versions.keyword', 'Language Versions'],
-      ['openData', 'Open Data'],
+      ['openData', 'Open'],
     ]);
   }
 
@@ -555,7 +555,7 @@ export class SearchService {
         'descriptor_type_versions.keyword',
         'Indicates that the tool or workflow contains at least one version that is written with the workflow language version',
       ],
-      ['openData', 'Indicates whether an entry has a test parameter file with all publicly accessible data.'],
+      ['openData', 'Indicates whether an entry can be run with a provided test parameter file and no additional access permissions.'],
     ]);
   }
 
--- a/src/app/search/state/search.service.ts
+++ b/src/app/search/state/search.service.ts
@@ -534,7 +534,7 @@ export class SearchService {
       ['verified_platforms.keyword', 'Verified Platforms'],
       ['categories.name.keyword', 'Category'],
       ['descriptor_type_versions.keyword', 'Language Versions'],
-      ['openData', 'Open'],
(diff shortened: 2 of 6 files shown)


denis-yuen:
> The versions table is getting crowded, although I think it might just be acceptable. Suggestions welcome.

Think it may be ok, at least without adding additional columns. 
It seems like you could go two rows for the header since the "verified platforms" header is much bigger than its contents. The language versions column could also be a bit less redundant but with more work. (all different ticket though)

src/app/search/state/search.service.ts
```diff
@@ -555,7 +555,7 @@ export class SearchService {
         'descriptor_type_versions.keyword',
         'Indicates that the tool or workflow contains at least one version that is written with the workflow language version',
       ],
-      ['openData', 'Indicates whether an entry has a test parameter file with all publicly accessible data.'],
+      ['openData', 'Indicates whether an entry can be run with a provided test parameter file and no additional access permissions.'],
```
denis-yuen, line 558: Technically, this is "can be run with (a provided test parameter file with no additional access permissions) or (no additional access permissions" which is clearly terrible. 
How about "can be run with no additional access permissions, potentially with a test parameter file referencing open data"?

src/app/search/state/search.service.ts
```diff
@@ -534,7 +534,7 @@ export class SearchService {
       ['verified_platforms.keyword', 'Verified Platforms'],
       ['categories.name.keyword', 'Category'],
       ['descriptor_type_versions.keyword', 'Language Versions'],
-      ['openData', 'Open Data'],
+      ['openData', 'Open'],
```
denis-yuen, line 537: I think you're being overly careful here, but as in the SEAB ticket, I think it's fine to consider this "Open Data" due to expectations in this field

src/app/search/map-friendly-values.pipe.ts
```diff
@@ -70,8 +70,8 @@ export class MapFriendlyValuesPipe implements PipeTransform {
     [
       'openData',
       new Map([
-        ['1', 'has public test data'],
-        ['0', 'no public test data'],
+        ['1', "doesn't need more data"],
+        ['0', 'needs more data'],
```
denis-yuen, line 74: Not really a fan of the contraction in the facet. 
How about "requires non-open data", "open data"

src/app/search/state/search.service.ts
KimberleyChong: Yeah I prefer "Open Data" to "Open", more clear to me at least. But if it's changed in the facet I think it would make sense for it to match with its header in the versions table... as Denis mentioned we may need two rows in a future ticket

src/app/search/map-friendly-values.pipe.ts
KimberleyChong: I think it looks nice/ is clear:
image

src/app/workflow/versions/versions.component.html
```diff
@@ -134,6 +134,28 @@
...
+        scope="col"
+        mat-header-cell
+        *matHeaderCellDef
+        mat-sort-header
+        matTooltip="The version has a test parameter file with open data or requires no input files."
+      >
+        Open
+        <a
+          class="ds-green"
+          [href]="Dockstore.DOCUMENTATION_URL + '/faq.html#what-is-an-open-tool-or-workflow'"
```
kathy-t, line 148: Does this exist?

src/app/search/map-friendly-values.pipe.ts
kathy-t: Instead of "non-open data", maybe "restricted data" - you used it in a tooltip I think

src/app/workflow/versions/versions.component.html
coverbeck: > Does this exist?

Not yet, but I have a ticket for it dockstore/dockstore#5402

coverbeck:
> > The versions table is getting crowded, although I think it might just be acceptable. Suggestions welcome.
> 
> Think it may be ok, at least without adding additional columns. It seems like you could go two rows for the header since the "verified platforms" header is much bigger than its contents. The language versions column could also be a bit less redundant but with more work. (all different ticket though)

dockstore/dockstore#5412

svonworl (approved):

denis-yuen:
> * [ ]  If this PR is for a user-facing feature, create and link a documentation ticket for this feature (usually in the same milestone as the linked issue). Style points if you create a documentation PR directly and link that instead.
> *

Remember to create and link this, we're going to need to explain where we are (i.e. parameter files with only http/https linnks with no auth) vs where we want to go (s3 and gs links that are requester pays)

denis-yuen (approved):

KimberleyChong (approved):

kathy-t (approved):

PR #1726, opened 2023-03-07
SEAB-5122: added notebook tab on my starred page

hyunnaye:
**Description**

**Review Instructions**

**Issue**

**Security**

### DIFF
--- a/src/app/starredentries/starredentries.component.html
+++ b/src/app/starredentries/starredentries.component.html
@@ -35,7 +35,8 @@
         [ngClass]="{
           'workflow-tab-group': selected.value === 0,
           'tool-tab-group': selected.value === 1,
-          'purple-tab-group': selected.value === 2
+          'notebook-tab-group': selected.value === 2,
+          'purple-tab-group': selected.value === 3
         }"
       >
         <mat-tab>
@@ -83,12 +84,42 @@
           </div>
         </mat-tab>
 
-        <ng-template #workflowSummary let-workflow="workflow" let-isAppTool="isAppTool">
+        <mat-tab>
+          <ng-template mat-tab-label>
+            <div fxLayoutGap="0.5rem">
+              <img class="notebook-icon" src="../../../assets/svg/sub-nav/notebooks.svg" alt="notebook logo" />
+              <span>Notebooks</span>
+              <span class="tab-display">{{ starredNotebooks?.length }}</span>
+            </div>
+          </ng-template>
+          <div>
+            <mat-card class="alert alert-info starred-header" role="alert" *ngIf="starredNotebooks?.length === 0">
+              <mat-icon>warning</mat-icon>
+              You have no starred notebooks.
+            </mat-card>
+            <div fxLayout="column" fxLayoutGap="1rem" class="mb-2 mx-1">
+              <mat-card fxFlex *ngFor="let notebook of starredNotebooks; let last = last" class="m-1 pb-3">
+                <ng-container
+                  *ngTemplateOutlet="workflowSummary; context: { workflow: notebook, isAppTool: false, isNotebook: true }"
+                ></ng-container>
+              </mat-card>
+            </div>
+          </div>
+        </mat-tab>
+
+        <ng-template #workflowSummary let-workflow="workflow" let-isAppTool="isAppTool" let-isNotebook="isNotebook">
           <div fxLayout="row" fxLayoutGap="10px" fxLayoutAlign="space-between">
             <a [routerLink]="(isAppTool ? entryType.AppTool : entryType.BioWorkflow) | routerLink: workflow" class="no-underline">
               <mat-card-header>
                 <mat-card-title fxLayoutAlign="space-between" fxLayoutGap="1rem">
                   <img
+                    *ngIf="isNotebook"
+                    src="../../../assets/svg/sub-nav/notebooks.svg"
+                    alt="{{ isAppTool ? 'tool' : 'workflow' }} logo"
+                    class="notebook-icon"
+                  />
+                  <img
+                    *ngIf="!isNotebook"
                     src="../../../assets/svg/sub-nav/{{ isAppTool ? 'tool' : 'workflow' }}.svg"
                     alt="{{ isAppTool ? 'tool' : 'workflow' }} logo"
                   />
--- a/src/app/starredentries/starredentries.component.scss
+++ b/src/app/starredentries/starredentries.component.scss
@@ -58,6 +58,11 @@
   text-decoration: none;
   transition: 0.4s color;
 }
+.notebook-icon {
+  height: 1em;
+  align-self: center;
+  justify-content: center;
+}
 @media screen and (max-width: 750px) {
   .columns {
     column-gap: 0px;
--- a/src/app/starredentries/starredentries.component.ts
+++ b/src/app/starredentries/starredentries.component.ts
@@ -14,6 +14,7 @@ import { OrgLogoService } from '../shared/org-logo.service';
 import { EntryType } from '../shared/enum/entry-type';
 import { ActivatedRoute, Params } from '@angular/router';
 import { Location } from '@angular/common';
+import { Notebook } from '../shared/openapi';
 
 @Component({
   selector: 'app-starredentries',
@@ -24,6 +25,7 @@ export class StarredEntriesComponent extends Base implements OnInit {
   starredTools: Array<ExtendedDockstoreTool> | null;
   starredWorkflows: Array<ExtendedWorkflow> | null;
   starredServices: Array<Entry> | null;
+  starredNotebooks: Array<ExtendedWorkflow> | null;
   starredOrganizations: Array<Organization> | null;
   user: any;
   starGazersClicked = false;
@@ -66,6 +68,14 @@ export class StarredEntriesComponent extends Base implements OnInit {
     this.usersService.getStarredWorkflows().subscribe((starredWorkflow) => {
       this.starredWorkflows = <ExtendedWorkflow[]>starredWorkflow.filter((entry: Workflow) => entry.is_published);
       this.starredWorkflows.forEach((workflow) => {
+        this.providerService.setUpProvider(workflow);
+        // This doesn't work because there's no workflowVersions
+        // workflow.versionVerified = this.dockstoreService.getVersionVerified(workflow.workflowVersions);
+      });
+    });
+    this.usersService.getStarredNotebooks().subscribe((starredNotebook) => {
+      this.starredNotebooks = <ExtendedWorkflow[]>starredNotebook.filter((entry: Workflow) => entry.is_published);
+      this.starredNotebooks.forEach((workflow) => {
         this.providerService.setUpProvider(workflow);
         // This doesn't work because there's no workflowVersions
         // workflow.versionVerified = this.dockstoreService.getVersionVerified(workflow.workflowVersions);


hyunnaye:
Will need https://github.com/dockstore/dockstore/pull/5406 to be merged for this PR to pass tests

src/app/starredentries/starredentries.component.html
```diff
@@ -81,12 +82,42 @@
...
+
+        <ng-template #workflowSummary let-workflow="workflow" let-isAppTool="isAppTool" let-isNotebook="isNotebook">
           <div fxLayout="row" fxLayoutGap="10px" fxLayoutAlign="space-between">
             <a [routerLink]="(isAppTool ? entryType.AppTool : entryType.BioWorkflow) | routerLink: workflow" class="no-underline">
               <mat-card-header>
                 <mat-card-title fxLayoutAlign="space-between" fxLayoutGap="1rem">
                   <img
+                    *ngIf="isNotebook"
+                    src="../../../assets/svg/sub-nav/notebooks.svg"
+                    alt="{{ isAppTool ? 'tool' : 'workflow' }} logo"
```
coverbeck, line 116: You can hard-code `"notebook logo"` here.

src/app/starredentries/starredentries.component.scss
```diff
@@ -58,6 +58,11 @@
   text-decoration: none;
   transition: 0.4s color;
 }
+.notebook-icon {
```
coverbeck, line 61: Why does this icon need special styling? Can't it use the same styling as the other icons on the page? Maybe it does, I didn't look at it closely, but it seems like it shouldn't.

src/app/starredentries/starredentries.component.html
```diff
@@ -81,12 +82,42 @@
...
+              <mat-card fxFlex *ngFor="let notebook of starredNotebooks; let last = last" class="m-1 pb-3">
+                <ng-container
+                  *ngTemplateOutlet="workflowSummary; context: { workflow: notebook, isAppTool: false, isNotebook: true }"
+                ></ng-container>
+              </mat-card>
+            </div>
+          </div>
+        </mat-tab>
+
+        <ng-template #workflowSummary let-workflow="workflow" let-isAppTool="isAppTool" let-isNotebook="isNotebook">
```
svonworl, line 108: Recommend that we pass the proper `entryType` into the template and dispense with the `isAppTool` and `isNotebook`.

src/app/starredentries/starredentries.component.html
svonworl: The Entry responses from the webservice now contain an `entryTypeMetadata` property, which has sub-properties `term` and `termPlural` that contain the singular and plural words used to refer to that kind of entry.  So, if we make our image names consistent (either singular or plural, depending on context), we could coalesce all the cases into a single passage, like:
```
<img
  src="../../../assets/svg/sub-nav/{{ workflow.entryTypeMetadata.termPlural }}.svg"
  alt="{{ workflow.entryTypeMetadata.term }} logo"
/>
```
We might even put this in a pipe someday:
```
{{ workflow | subNavImg }}
```

Sometime (not now), we should refactor the `routerLink` pipe similarly.  Now that the workflow object contains its type and other necessary information, the entryType argument is superfluous.

kathy-t:
> Will need [dockstore/dockstore#5406](https://github.com/dockstore/dockstore/pull/5406) to be merged for this PR to pass tests

You can update the `package.json` [here](https://github.com/dockstore/dockstore-ui2/blob/fa45f93c742f44335c10c43542e713b495d897bc/package.json#L8-L9) to use the artifacts from your webservice branch if you want to run the tests before your webservice PR is merged. If you do that, remember to switch it back to use develop artifacts before you merge this UI PR.

src/app/starredentries/starredentries.component.html
```diff
@@ -81,12 +82,42 @@
...
+        <mat-tab>
+          <ng-template mat-tab-label>
+            <div fxLayoutGap="0.5rem">
+              <img class="notebook-icon" src="../../../assets/svg/sub-nav/notebooks.svg" alt="notebook logo" />
+              <span>Notebooks</span>
+              <span class="tab-display" data-cy="starred-notebooks-count">{{ starredNotebooks?.length }}</span>
+            </div>
+          </ng-template>
+          <div>
+            <mat-card class="alert alert-info starred-header" role="alert" *ngIf="starredNotebooks?.length === 0">
```
kathy-t, line 94: This banner needs to also have `mat-elevation-z` added to the class so that it matches the banner style of the redesign. (Last point of the UI design section https://github.com/dockstore/dockstore/wiki/Dockstore-Frontend-Opinionated-Style-Guide#ui-design)

src/app/starredentries/starredentries.component.html
```diff
@@ -81,12 +82,42 @@
...
+        </mat-tab>
+
+        <ng-template #workflowSummary let-workflow="workflow" let-isAppTool="isAppTool" let-isNotebook="isNotebook">
           <div fxLayout="row" fxLayoutGap="10px" fxLayoutAlign="space-between">
             <a [routerLink]="(isAppTool ? entryType.AppTool : entryType.BioWorkflow) | routerLink: workflow" class="no-underline">
               <mat-card-header>
                 <mat-card-title fxLayoutAlign="space-between" fxLayoutGap="1rem">
                   <img
+                    *ngIf="isNotebook"
+                    src="../../../assets/svg/sub-nav/notebooks.svg"
```
kathy-t, line 115: If you changed the name of this `notebooks.svg` to singular `notebook.svg` and you followed Steve's suggestion of using `entryType` and getting rid of `isAppTool` and `isNotebook` in this [comment](https://github.com/dockstore/dockstore-ui2/pull/1726/files#r1138091424), I think that you can get rid of the new `img` block and modify the existing one to get the image `src` and `alt` from `entryType` because [EntryType](https://github.com/dockstore/dockstore-ui2/blob/fa45f93c742f44335c10c43542e713b495d897bc/src/app/shared/enum/entry-type.ts#L11) is already the singular word.

Something like:
```
<img
  src="../../../assets/svg/sub-nav/{{ entryType }}.svg"
  alt="{{ entryType }} logo"
/>

File: src/app/starredentries/starredentries.component.html [hunk 6]
```diff
@@ -85,41 +87,39 @@
...
-                    src="../../../assets/svg/sub-nav/notebooks.svg"
-                    alt="{{ isAppTool ? 'tool' : 'workflow' }} logo"
-                    class="notebook-icon"
-                  />
-                  <img
-                    *ngIf="!isNotebook"
-                    src="../../../assets/svg/sub-nav/{{ isAppTool ? 'tool' : 'workflow' }}.svg"
-                    alt="{{ isAppTool ? 'tool' : 'workflow' }} logo"
+                    class="site-icons-small entryType-icons"
+                    src="../../../assets/svg/sub-nav/{{ entryType }}.svg"
```
kathy-t, line 121: I didn't try this locally, but I don't think this will work for `EntryType.AppTool` because it should use `tool.svg`, not `appTool.svg`. You'll need a conditional to check for `AppTool` and set it to `tool`

src/app/starredentries/starredentries.component.html
```diff
@@ -73,22 +76,52 @@
...
+              <mat-card fxFlex *ngFor="let notebook of starredNotebooks; let last = last" class="m-1 pb-3">
+                <ng-container
+                  *ngTemplateOutlet="workflowSummary; context: { workflow: notebook, entryType: entryType.Notebook, isAppTool: false }"
+                ></ng-container>
+              </mat-card>
+            </div>
+          </div>
+        </mat-tab>
+
+        <ng-template #workflowSummary let-workflow="workflow" let-entryType="entryType" let-isAppTool="isAppTool">
```
coverbeck, line 116: Now that you're passing the `EntryType` (good!), you should be able to get rid of `isAppTool`. It's redundant and possibly contradictory (what if somebody passed EntryType.Notebook and isAppTool true)?

hyunnaye: I removed that before but I realized I needed it to add the conditional for https://github.com/dockstore/dockstore-ui2/blob/f34e653c63fd99e15c957561c584d3acb71d8577/src/app/starredentries/starredentries.component.html#L123

coverbeck: Can you do `entryType === entryType.AppTool ? 'tool' : entryType`?

hyunnaye: No, it comes with an `Unresolved variable AppTool ` warning on intellij. Since the variable on line 116 is named to entryType, it looks like its checking `appTool === entryType.AppTool` and resulting to false. One possible fix is to rename the variable to something like `let-type="entryType"`. That way we can do `type === entryType.AppTool ? 'tool' : type`. Do you think that would be better than `isAppTool`?

coverbeck: Yes, I think it's better than using `isAppTool` -- you already have all the info you need in the entryType parameter.

Renaming makes sense, I can see how the compiler would get confused between entryType defined in the .ts file vs entryType defined in the template. In fact, I wonder if it's actually using the entryType you think it's using on line 124 (maybe it is, maybe it's not). :)

hyunnaye: Changed

coverbeck (approved):

src/app/starredentries/starredentries.component.html
```diff
@@ -73,22 +76,50 @@
...
-            <a [routerLink]="(isAppTool ? entryType.AppTool : entryType.BioWorkflow) | routerLink: workflow" class="no-underline">
+            <a [routerLink]="type | routerLink: workflow" class="no-underline">
               <mat-card-header>
                 <mat-card-title fxLayoutAlign="space-between" fxLayoutGap="1rem">
                   <img
-                    src="../../../assets/svg/sub-nav/{{ isAppTool ? 'tool' : 'workflow' }}.svg"
-                    alt="{{ isAppTool ? 'tool' : 'workflow' }} logo"
+                    class="site-icons-small entryType-icons"
+                    src="../../../assets/svg/sub-nav/{{ type === entryType.AppTool ? 'tool' : type }}.svg"
+                    alt="{{ type }} logo"
```
kathy-t, line 122: Should do `{{ type === entryType.AppTool ? 'tool' : type }}` for the `alt` tag too

kathy-t (approved):

src/app/starredentries/starredentries.component.ts
```diff
@@ -71,6 +72,14 @@ export class StarredEntriesComponent extends Base implements OnInit {
         // workflow.versionVerified = this.dockstoreService.getVersionVerified(workflow.workflowVersions);
       });
     });
+    this.usersService.getStarredNotebooks().subscribe((starredNotebook) => {
+      this.starredNotebooks = <ExtendedWorkflow[]>starredNotebook.filter((entry: Workflow) => entry.is_published);
+      this.starredNotebooks.forEach((workflow) => {
+        this.providerService.setUpProvider(workflow);
+        // This doesn't work because there's no workflowVersions
+        // workflow.versionVerified = this.dockstoreService.getVersionVerified(workflow.workflowVersions);
```
denis-yuen, line 80: was copied but comment needs some updating

denis-yuen (approved):

svonworl (approved):

src/app/starredentries/starredentries.component.ts
hyunnaye: comment removed

coverbeck (approved):

PR #1730, opened 2023-03-08
Position tooltip to the left in Actions menu dropdown

kathy-t:
**Description**
By default, the tooltip is positioned below the button. In the Actions menu, this can block other buttons. This PR sets the position of the tooltip to the left of the buttons in the Actions menu dropdown so that none of the tooltips are preventing users from pressing buttons under them. 

Before:

https://user-images.githubusercontent.com/25287123/223879359-5618a084-553b-44a4-aba0-58c7de28fc70.mp4

After:

https://user-images.githubusercontent.com/25287123/223879374-acfaf7b4-0a6b-46e7-b3a0-33bece08ab6d.mp4 

  
**Review Instructions**
On QA,
- Go to My Workflows and click on a workflow with more than one valid version (so that there's the option to snapshot). If the version is a tag, even better because the Actions menu dropdown for the version will have options to Snapshot, Request DOI, and Export to ORCID
- Click on the Versions tab and click the Actions dropdown for the valid version
- Hover over each button (at minimum, ensure there's the Snapshot button in order to verify this fix) and verify that the tooltip is positioned to the left of the button and each tooltip does not block the user from pressing buttons underneath it

**Issue**
https://github.com/dockstore/dockstore/issues/5258
[DOCK-2288](https://ucsc-cgl.atlassian.net/browse/DOCK-2288)

**Security**

[DOCK-2288]: https://ucsc-cgl.atlassian.net/browse/DOCK-2288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/workflow/view/view.component.html
+++ b/src/app/workflow/view/view.component.html
@@ -51,6 +51,7 @@
       (click)="snapshotVersion()"
       data-cy="dockstore-snapshot"
       matTooltip="Snapshotting a version makes it read-only."
+      matTooltipPosition="left"
     >
       Snapshot
     </button>
@@ -63,6 +64,7 @@
       data-cy="dockstore-request-doi-button"
       *ngIf="version.referenceType !== 'BRANCH' && !version.doiURL && workflow?.is_published"
       matTooltip="Request a Digital Object Identifier (DOI) for this version."
+      matTooltipPosition="left"
     >
       Request DOI
     </button>
@@ -80,6 +82,7 @@
         workflow?.is_published
       "
       matTooltip="Export this version to ORCID."
+      matTooltipPosition="left"
     >
       Export to ORCID
     </button>
@@ -91,7 +94,10 @@
     >
       Set as Default Version
     </button>
-    <div matTooltip="Refresh an individual version. Disabled for GitHub App workflows, hosted workflows, and frozen versions.">
+    <div
+      matTooltip="Refresh an individual version. Disabled for GitHub App workflows, hosted workflows, and frozen versions."
+      matTooltipPosition="left"
+    >
       <button
         mat-menu-item
         color="accent"


KimberleyChong (approved):

denis-yuen (approved):

hyunnaye (approved):

coverbeck (approved):
This works by moving the tooltip out of the way, which is fine, but just curious, did you look at the underlying issue, that the tooltip doesn't disappear when the mouse is over it? It seems to do so in prod. 
 
I think the difference is the newer Angular we (I) upgraded to, is behaving differently as per angular/components#25173, and there's probably not much we can do about it. As Denis said, we've noticed it in other sites, include GitHub itself, where things don't seem to get out of the way like they used to.

kathy-t:
> This works by moving the tooltip out of the way, which is fine, but just curious, did you look at the underlying issue, that the tooltip doesn't disappear when the mouse is over it? It seems to do so in prod.
> 
> I think the difference is the newer Angular we (I) upgraded to, is behaving differently as per [angular/components#25173](https://github.com/angular/components/issues/25173), and there's probably not much we can do about it. As Denis said, we've noticed it in other sites, include GitHub itself, where things don't seem to get out of the way like they used to.

Yeah, it's the new Angular version. The [commit](https://github.com/angular/components/commit/0dfc4909883af35d89c11d8bfcd73263578cd1cd) that introduced this behaviour made it into Angular 14 (can see version 14 tags in the commit). The change is intentional and addresses an accessibility bug, tracked by this [ticket](https://github.com/angular/components/issues/4942).

I think there's a way to override it, but I lean against doing that since it would not comply with accessibility standards.

PR #1731, opened 2023-03-13
Add Discover Existing Tools Button - DOCK-2269

KimberleyChong:
**Description**
Essentially copies the `Discover Existing Dockstore Workflows` button and adds it to the my-tools page as the `Discover Existing Dockstore Tools`  button to add existing GitHub App Tools from your organization.

image
image
image

**Wording preference:** Any preference to update the text on the `Discover Existing Dockstore Workflows` button (and new tools button) to match the mocks or keep it as is? The tooltip has the updated text but the button itself is simplified:
| Current | Mocks |
|:---:|:---:|
| image | image |

**Review Instructions**
Check that button exists and succeeds when adding existing tools.

**Issue**
DOCK-2269 [GitHub Issue](https://github.com/dockstore/dockstore/issues/5195)

**Security**

### DIFF
--- a/src/app/shared/user/user.service.ts
+++ b/src/app/shared/user/user.service.ts
@@ -46,14 +46,12 @@ export class UserService {
   }
 
   /**
-   * Add Workflows or GitHub App Tools from a user's organization
+   * Add Workflows and GitHub App Tools from a user's organization
    *
    * @param userId
-   * @param isWorkflow  Display 'workflows' or 'tools' in the alert message
    */
-  addUserToWorkflows(userId: number, isWorkflow: boolean): void {
-    // Should use 'tools' when adding app tools from myTools page and 'workflows' when adding from myWorkflows page
-    this.alertService.start('Adding user to existing ' + (isWorkflow ? 'workflows' : 'tools') + ' on Dockstore');
+  addUserToWorkflows(userId: number): void {
+    this.alertService.start('Adding user to existing workflows and tools on Dockstore');
     this.usersService.addUserToDockstoreWorkflows(userId).subscribe(
       (workflows: Array<Workflow>) => {
         this.alertService.detailedSuccess();
--- a/src/app/shared/user/user.service.ts
+++ b/src/app/shared/user/user.service.ts
@@ -45,8 +45,15 @@ export class UserService {
     });
   }
 
-  addUserToWorkflows(userId: number): void {
-    this.alertService.start('Adding user to existing workflows on Dockstore');
+  /**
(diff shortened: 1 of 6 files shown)


src/app/shared/user/user.service.ts
```diff
@@ -45,8 +45,15 @@ export class UserService {
...
-    this.alertService.start('Adding user to existing workflows on Dockstore');
+  /**
+   * Add Workflows or GitHub App Tools from a user's organization
+   *
+   * @param userId
+   * @param isWorkflow  Display 'workflows' or 'tools' in the alert message
+   */
+  addUserToWorkflows(userId: number, isWorkflow: boolean): void {
+    // Should use 'tools' when adding app tools from myTools page and 'workflows' when adding from myWorkflows page
+    this.alertService.start('Adding user to existing ' + (isWorkflow ? 'workflows' : 'tools') + ' on Dockstore');
```
coverbeck, line 56: Two contradictory suggestions:

1. instead of passing a boolean, pass the value or enum to display, perhaps using entry type metadata from #1732. For example, if we ever add this feature for notebooks or some other entity, we don't have to update this code, we just pass in the new display value/enum as a parameter.
2. The web service endpoint discovers both app tools and workflows, so maybe we shouldn't have a type-specific message anyway. Maybe it should just say "Adding user to existing entries on Dockstore", or "Adding user to existing tools and workflows on Dockstore". If we do that, do we also update the menu choices to say "Discover Existing Dockstore Entries"? 

My vote is:

* Keep the menu choice labeling the same.
* Change the success message to "Adding user to existing workflows and tools on Dockstore".

svonworl:
Looks good!

It's a preexisting issue, but would it make sense to fix the uneven vertical empty space in the "More" menu?  When there's a single one-line item, the space at top is narrower than at bottom.  I think it's because the menu has a `min-height` of 64.

src/app/shared/user/user.service.ts
KimberleyChong: > My vote is:
> 
> * Keep the menu choice labeling the same.
> * Change the success message to "Adding user to existing workflows and tools on Dockstore".

Mm I see, I agree.  I'll keep the labels the same and change the success message to both tools and workflows

KimberleyChong:
**Update:**

Remove `min-height` requirement that was resulting in extra padding in the menu dropdown:

**Before:**
image

**After:**
image

Change success message to apply for both workflows and tools:

image

coverbeck (approved):

svonworl (approved):

kathy-t (approved):

PR #1733, opened 2023-03-16
Fix missing unpublished workflows on refresh

kathy-t:
**Description**
The unpublished tab was empty if an unpublished workflow was selected and the browser was refreshed because the tab group was quickly selecting the published tab then the unpublished tab. This seems to be a problem with the animation when the tab content changes quickly (found this issue https://github.com/angular/components/issues/25236). 

It selected the published tab first because the `workflowId` observable that the `selectTab` pipe was waiting for was not ready yet so the pipe selected the published tab according to its logic. Then when `workflowId` was ready, the pipe selected the unpublished tab, resulting in the empty tab. This PR adds a condition to the mat-tab-group so that the tabs are only rendered if the `workflowId` is ready. This ensures that the correct tab is always selected on the first render. While I was in there, I did some minor simplifications of the select tab pipe's logic.

I also fixed an `undefined` console error I noticed in the My Tools page when I tried to reproduce this bug for GitHub app tools.

**Review Instructions**
On QA:
- Go to My Workflows
  - Select an unpublished workflow
  - Refresh the page
  - The unpublished tab that your workflow belongs to should not be empty
- Go to My Tools
  - Select an unpublished GitHub App tool
  - Refresh the page
  - The unpublished tab that your workflow belongs to should not be empty
  - Check that the tabs of the other expansion panels look okay (they're not empty)

**Issue**
dockstore/dockstore#4882
[DOCK-2158](https://ucsc-cgl.atlassian.net/browse/DOCK-2158)

**Security**


kathy-t:
Not sure what's up with CircleCI. It's failing at the build step because the `get-circleci-artifact-url.sh` script fails, but it doesn't fail all the time. Checking locally, this curl sometimes returns a 404: `curl -I https://circleci.com/api/v2/project/gh/dockstore/dockstore/31327/artifacts -H 'Accept: application/json'`.

Will try re-running later, but requesting reviews for now.

coverbeck (approved):
Pending passing tests

KimberleyChong (approved):

src/app/shared/state/workflow.query.ts
```diff
@@ -26,10 +26,12 @@ export class WorkflowQuery extends QueryEntity<WorkflowState, Service | BioWorkf
...
   );
   public extendedDescriptorLanguageBean$: Observable<ExtendedDescriptorLanguageBean> = this.workflow$.pipe(
-    map((workflow) => this.descriptorLanguageService.workflowDescriptorTypeEnumToExtendedDescriptorLanguageBean(workflow.descriptorType))
+    map((workflow) =>
+      workflow ? this.descriptorLanguageService.workflowDescriptorTypeEnumToExtendedDescriptorLanguageBean(workflow.descriptorType) : null
+    )
   );
   public launchSupport$: Observable<boolean> = this.extendedDescriptorLanguageBean$.pipe(
-    map((extendedDescriptorLanguage) => extendedDescriptorLanguage.workflowLaunchSupport)
+    map((extendedDescriptorLanguage) => (extendedDescriptorLanguage ? extendedDescriptorLanguage.workflowLaunchSupport : null))
```
svonworl, line 34: No need to change, but for future reference, if we're ok with this expression possibly evaluating to `undefined`, we could the "optional chaining" operator here:
```
extendedDescriptorLanguage?.workflowLaunchSupport
```
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

svonworl (approved):

PR #1734, opened 2023-03-17
Fix CircleCI artifacts script

kathy-t:
**Description**
UI builds have been failing because of the CircleCI artifacts script. This PR fixes the script by adding an empty `Circle-Token` header to the `curl` that the script makes to get the artifacts. 

The [docs](https://circleci.com/docs/api-developers-guide/#authentication-and-authorization) say that the API uses token-based authentication to manage access to the API server. Maybe they recently changed something so now the header is required :thinking:. The `getArtifacts` example [here](https://circleci.com/docs/api-developers-guide/#download-artifacts) also has the `Circle-Token` header. I just made the value an empty string and I think it works because the dockstore/dockstore project is public.

**EDIT:** accessibility_test_base runs the develop version of the script which doesn't have the Circle-Token header so it fails. It should pass once this PR merges into develop.

**Review Instructions**
Builds should pass

**Issue**
n/a

**Security**

### DIFF
--- a/scripts/get-circleci-artifact-url.sh
+++ b/scripts/get-circleci-artifact-url.sh
@@ -14,6 +14,6 @@ fi
 
 CIRCLECI_BUILD_ID=${1}
 ARTIFACT_NAME=${2}
-ARTIFACT_URL=$(curl "https://circleci.com/api/v2/project/gh/dockstore/dockstore/${CIRCLECI_BUILD_ID}/artifacts" -H "Accept: application/json" | jq -r --arg ARTIFACT_NAME "${ARTIFACT_NAME}" '[.items[] | select( .path | contains($ARTIFACT_NAME) ) | .url] | first ')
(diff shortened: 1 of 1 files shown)


kathy-t:
> On tests passing,
> 
> However, test failure seems related
> 
> ```
> ++ jq -r --arg ARTIFACT_NAME swagger.yaml '[.items[] | select( .path | contains($ARTIFACT_NAME) ) | .url] | first '
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100    26  100    26    0     0   1444      0 --:--:-- --:--:-- --:--:--  1444
> jq: error (at <stdin>:0): Cannot iterate over null (null)
> + ARTIFACT_URL=
> + SWAGGER_PATH=
> 
> Exited with code exit status 5
> 
> CircleCI received exit code 5
> ```

`accessibility_test_base` runs the develop version of the script which doesn't have the `Circle-Token` header so it fails. It should pass once this PR merges into develop.

denis-yuen (approved):

KimberleyChong (approved):

coverbeck (approved):
Thanks for figuring it out.

svonworl (approved):
Yay, good job!

PR #1735, opened 2023-03-17
Refactor Recent Events & Add to Org Box- SEAB-5267

KimberleyChong:
**Description**
Adds recent events to the `organization-box` (SEAB-5267), refactors the `recent-events.component` so that it can be  used to display recent events for the `organization-box.component`, `starred-box.component` and `user-page.component`, and adds the `recent-events.pipe` to assist in cleaning up the `recent-events` template (SEAB-4056 SEAB-4715 SEAB-4714).

**Org-box & Starred-box**
image
**User Pages**
image

Note that the [search bar](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6022ab2d42df9f2c57c20982) in the organization box was removed as it doesn't make much sense to implement (at least right now), the mocks don't show how search is handled and not sure how much use users would get out of it given that not many belong to an organization. The "All Organizations" link at the bottom of the card is also missing as the `my-organizations` page does not exist.

This pr also cleans up two console errors that would occur due to the use of `ng-plural` in the `org-box`.

### recent-events.component
Now takes an `eventType` parameter to display organization events for the `org-box` or starred events for the `starred-box`. If no input is provided and it identifies a username in the url (i.e. on user page) then it will retrieve the user's events.
https://github.com/dockstore/dockstore-ui2/blob/44df1f0e8432d2d7320dfe6abb432c62d26bbed2/src/app/home-page/recent-events/recent-events.component.ts#L28

The `recent-events.component` now covers `organization events`, `starred events`, and `entry events` covering these event types:

| CREATEORG | MODIFYORG | APPROVEORG | REJECTORG |
| :---| :---| :---| :---| 
| ADDUSERTOORG | APPROVEORGINVITE | CREATECOLLECTION | MODIFYCOLLECTION |
| DELETECOLLECTION | REMOVEFROMCOLLECTION | ADDTOCOLLECTION | ADDVERSIONTOENTRY | 
| PUBLISHENTRY | UNPUBLISHENTRY |

Notable changes to UI:
- The username and user avatar now link to that user's user page, this wasn't a requirement, but after interacting with the component I thought it made sense to have that functionality
- Events that display an org image where that organization does not have an org logo will display the org icon
image
- Entry events (`ADDVERSIONTOENTRY`, `PUBLISHENTRY`, `UNPUBLISHENTRY`) will display the associated entry type icon
- Org icons and entry icons will link to the associated organization or entry unless the event was `UNPUBLISHENTRY` in which the entry would no longer exist
[...]

PR #1744, opened 2023-03-29
Update Featured Content Styles - SEAB-4628

KimberleyChong:
**Description**
A follow-up to the changes to `feat-content.html`.

Updates the featured content styles and adds Featured Content to the new Dashboard.

**Before:**
image

**After:**
image

**Overview of the Dashboard and Featured Content (at the bottom):**
Group 5

- I felt like News & Updates was missing something, so I added a count bubble so it matches other widgets on the Dashboard, I think it fits nicely

<p align="center">

</p>

- The twitter widget has some built-in padding around it so I added some around the `news-box` so their heights look more similar

**Review Instructions**
Featured Content appears on the new dashboard and the new styles look correct on both the homepage and dashboard.

**Issue**
SEAB-4628

**Security**

### DIFF
--- a/src/app/home-page/widget/news-box/news-box.component.html
+++ b/src/app/home-page/widget/news-box/news-box.component.html
@@ -2,11 +2,11 @@
   <div fxLayout="column" fxLayoutAlign="start stretch" class="h-100">
     <mat-card-header>
       <mat-card-title>News & Updates</mat-card-title>
-      <span class="bubble">4</span>
+      <span class="bubble">{{ newsUpdates.newsEntriesCount }}</span>
     </mat-card-header>
     <mat-divider></mat-divider>
     <mat-card-content class="h-100 mx-3 mt-3 mb-0">
-      <app-news-and-updates></app-news-and-updates>
+      <app-news-and-updates #newsUpdates></app-news-and-updates>
     </mat-card-content>
   </div>
 </mat-card>
--- a/src/app/home-page/widget/news-box/news-box.component.html
+++ b/src/app/home-page/widget/news-box/news-box.component.html
@@ -2,6 +2,7 @@
   <div fxLayout="column" fxLayoutAlign="start stretch" class="h-100">
     <mat-card-header>
       <mat-card-title>News & Updates</mat-card-title>
+      <span class="bubble">4</span>
     </mat-card-header>
     <mat-divider></mat-divider>
     <mat-card-content class="h-100 mx-3 mt-3 mb-0">
--- a/src/app/home-page/new-dashboard/new-dashboard.component.html
+++ b/src/app/home-page/new-dashboard/new-dashboard.component.html
@@ -15,10 +15,16 @@
         <app-starred-box fxFlex="50" fxFlex.lt-md="100" class="updates-holder"></app-starred-box>
       </div>
       <div fxFlex fxLayout fxLayout.lt-md="column" fxLayoutGap="2.5rem" class="w-100">
-        <app-news-box fxFlex.md="65" fxFlex.gt-md="65" fxFlex.lt-md="100"></app-news-box>
+        <div fxFlex.md="65" fxFlex.gt-md="65" fxFlex.lt-md="100" class="py-4">
+          <app-news-box></app-news-box>
+        </div>
         <div fxFlex.md="35" fxFlex.gt-md="35" fxFlex.lt-md="100">
           <div class="my-3 twitter" #twitter></div>
         </div>
+      </div>
+      <div class="w-100">
+        <div class="mat-card-title text-align-center">Featured Content</div>
+        <app-featured-content></app-featured-content>
       </div>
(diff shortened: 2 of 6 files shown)


src/app/home-page/widget/news-box/news-box.component.html
```diff
@@ -2,6 +2,7 @@
   <div fxLayout="column" fxLayoutAlign="start stretch" class="h-100">
     <mat-card-header>
       <mat-card-title>News & Updates</mat-card-title>
+      <span class="bubble">4</span>
```
svonworl, line 5: Is it ever possible to have a number of items which is not 4?

coverbeck (approved):

svonworl (commented):
There's a couple of unrelated bugs in the "Starred" box that probably need fixing:

1. The org icons are broken.
2. Events about categories (corresponding to hidden org 'Dockstore') are leaking.  We should probably filter them out on the webservice side.

If you could please write the necessary tickets, that would be great, thanks!

KimberleyChong:
> > 2. Events about categories (corresponding to hidden org 'Dockstore') are leaking.  We should probably filter them out on the webservice side.
> 
> Hmmm, is this occurring because the hidden org was starred or because your user is a part of it?

Mm referring to something like this? Occurs when starring [this](https://qa.dockstore.org/workflows/github.com/iwc-workflows/sars-cov-2-ont-artic-variant-calling/COVID-19-ARTIC-ONT:main?tab=info) workflow. I've actually wondered about the Dockstore organization because it shows up as `page not found` when you try to visit it

image

KimberleyChong:
> 1. The org icons are broken.

Do they break on qa/ any specific case? I just merged the [updated](https://github.com/dockstore/dockstore-ui2/pull/1735) revent activities which should handle no org logo and broken org logos 🤔

denis-yuen:
> Mm referring to something like this? Occurs when starring [this](https://qa.dockstore.org/workflows/github.com/iwc-workflows/sars-cov-2-ont-artic-variant-calling/COVID-19-ARTIC-ONT:main?tab=info) workflow. I've actually wondered about the Dockstore organization because it shows up as `page not found` when you try to visit it

Hmmm, ah I see. We basically implemented categories using a special organization and collections that are normally hidden from non-curators. It's a little weird that they are leaking  but I guess I  was asking two questions  for both you and @svonworl 

1) Why are they leaking?  
2) Should they be leaking? 

For 1) it seems like you're seeing events related to organizations because you starred a workflow which was previously added to a category. Are you a curator or admin? 

For 2), theoretically I could see a version of Dockstore where we get landing pages https://ucsc-cgl.atlassian.net/browse/SEAB-3631 and end-users can see events relating to categories. Given that we lost team lighthouse, we're not near there yet so agreeing with @svonworl 

For 1.14, we need a ticket for hiding events about categories from regular end-users (not admins and curators)

Screenshot from 2023-04-03 10-32-32

src/app/home-page/widget/news-box/news-box.component.html
denis-yuen: I think this is matching https://github.com/dockstore/extra-content/blob/develop/news.html 
And I do think this may be a bit unwieldy if we need to modify both places when adding/removing news. 

If there's not a simple way to count the number of news items, we may need to skip this (even if it does look better) for usability.

denis-yuen (commented):
Comment

KimberleyChong:
> For 1) it seems like you're seeing events related to organizations because you starred a workflow which was previously added to a category. Are you a curator or admin?

Just checked and the leaking occurs both locally as an admin and on QA as a regular user 👍️

Here's another scenario from the user pages: https://qa.dockstore.org/users/akhadilk

denis-yuen:
> Just checked and the leaking occurs both locally as an admin and on QA as a regular user +1

Yeah, follow-up ticket will all this info then for the web service

src/app/home-page/widget/news-box/news-box.component.html
KimberleyChong: > Is it ever possible to have a number of items which is not 4?

Mm yes, I was thinking about that and it's a good point, I assumed we always maintained four at a time...but that may not always be valid.

I'm adding in a check to count the number of `news-entry` classes in the HTML string retrieved, which will indicate the number of news articles.

Ex. with two articles returned:
image

KimberleyChong:
**Updates**: Changes to accommodate the removal of the nested anchors in `feat-content.html`, also changes to display the number of news items dynamically.  In `news-and-updates` the contents of `OnInit` was moved to the `Constructor` to satisfy dependencies, since its parent `news-box` is displaying `newsEntriesCount` (provided by the child, `news-and-updates`)

denis-yuen (approved):

kathy-t (approved):

svonworl (approved):

PR #1748, opened 2023-04-03
Clean Up Sonarcloud Quality Gate Issues

KimberleyChong:
**Description**
An attempt to clean up some Sonarcloud Quality Gate issues:
<p align=center>

</p>

**Main issues addressed:**
- Adding `alt` tags to images
- Removing un-used imports
- Simplifying multiple imports from the same source
- Switching to `string` from `String`
- Switching to 'boolean' from 'Boolean'
- Removing unexpected empty `Constructor`s

**Additional:**
- When you visit https://qa.dockstore.org/docs you are redirected to https://docs.dockstore.org/en/stable/.html instead of https://docs.dockstore.org/en/stable/, removed `+ .html` to fix redirect:
https://github.com/dockstore/dockstore-ui2/blob/429bd41bede34cfda0118280ee1c4f935648308a/src/app/docs/docs.component.ts#L69
- Removed a chunk of old css in `starredentries.component.scss` to style elements/classes that no longer exist (checked history to verify)
- Removed `src/organizations.scss`, file has no references made to it and history states this file was brought over from the previous `UI1` repo in 2017

**Review Instructions**
All tests should be passing.  There should be less Sonarcloud Quality Gate issues than in the current screenshot.

**Issue**
SEAB-5338

**Security**


src/app/organizations/organization-members/organization-members.component.html
```diff
@@ -38,17 +38,20 @@
...
-                  *ngIf="!organizationUser.user?.avatarUrl"
-                  mat-card-avatar
-                  class="header-image"
-                  src="https://via.placeholder.com/300x300?text=Avatar+Not+Found"
-                />
+                <ng-template #noAvatar>
+                  <img
+                    mat-card-avatar
+                    class="header-image"
+                    src="https://via.placeholder.com/300x300?text=Avatar+Not+Found"
```
denis-yuen, line 51: Was like this before, but https://placeholder.com/ doesn't seem like an obvious source. 
https://www.reddit.com/r/web_design/comments/10lqgja/placeholdercom_is_no_more/ looks like it got bought up which means the image could be replaced unexpectedly. 

Can you find an alternative and/or look for other sources? (also see https://via.placeholder.com/150 elsewhere)

coverbeck (approved):
Thanks for doing the cleanup!

kathy-t (approved):

svonworl (approved):

KimberleyChong:
> A few more placeholders, for example
> https://github.com/dockstore/dockstore-ui2/blob/SEAB-5338/sonarcloud-quality-gate-bugs/cypress/e2e/group2/organizations.ts#L89

Looks like remaining places are using to test for regex, still updated to replace final instances of `placeholder.com` with `gravatar`

PR #1751, opened 2023-04-08
Prevent snapshotting if no author set.

coverbeck:
**Description**
Force the author(s) to be set before snapshotting, so that we can set the author field when creating the DOI. The author field is required when creating a DOI; we currently get around that by adding "Author unknown", which is not ideal; this will prevent that going forward.

The initial idea was to prevent publishing a DOI if the author isn't set, but since you can't add an author after you snapshot, I moved the check up earlier.

***Notes***
* We currently don't prevent this in the web service; that can be handled that in another PR. It looks like there's other enforcement the UI does that is not done by the web service; the ability to snapshot branches is only blocked in the UI.
* There are existing snapshots that slipped through without an author; there's not much we can do about that, since the versions are immutable.

***Alternative***
At the DB level, we don't freeze the authors for a version. It's not currently supported by the API, but in theory we could provide some way to add authors after snapshotting. Or maybe we should be freezing the authors for a frozen version.

**Review Instructions**
1. Try snapshotting a workflow version that does not have the author set; you should be unable to, and you should get the red box that is in the attached screenshot.
3. Try snapshotting a workflow version that does have the author set, you should be able to.

**Issue**
dockstore/dockstore#5437
https://ucsc-cgl.atlassian.net/browse/DOCK-2358

**Screenshot**
This comes up if the authors are not set after you click on the Snapshot menu choice.

Screen Shot 2023-04-07 at 4 40 04 PM

**Security**


denis-yuen:
> * It looks like there's other enforcement the UI does that is not done by the web service; the ability to snapshot branches is only blocked in the UI.

Could probably use a ticket if it doesn't exist already

> At the DB level, we don't freeze the authors for a version. It's not currently supported by the API, but in theory we could provide some way to add authors after snapshotting. Or maybe we should be freezing the authors for a frozen version.

Hmmm, worth a thought either way. The good thing is if we recommend orcid authors more strongly, people can still edit their profiles that way frozen or not frozen. On the other hand, tinkering with authors doesn't change reproducibility, which makes me think it should be changeable like labels.

test/authors_db_dump.sql
```diff
@@ -0,0 +1,38 @@
...
+SET xmloption = content;
+SET client_min_messages = warning;
+SET row_security = off;
+
+--
+-- Data for Name: author; Type: TABLE DATA; Schema: public; Owner: dockstore
+--
+
+COPY public.author (id, affiliation, dbcreatedate, dbupdatedate, email, name, role, versionid) FROM stdin;
+1	\N	\N	\N	Muhammad.Lee@oicr.on.ca	Muhammed Lee	\N	13
```
denis-yuen, line 24: Interesting that the warning in codacy seems only half formed

denis-yuen (approved):

src/app/workflow/snapshot-exporter-modal/snaphot-exporter-modal.component.html
```diff
@@ -139,47 +144,65 @@ <h4 mat-dialog-title>
...
+        Before snapshotting, you should ensure your version has its author(s) set. The authors cannot be added after you have snapshotted
+        the version. You can either add authors via
+        <a
+          href="{{
+            Dockstore.DOCUMENTATION_URL
+          }}/assets/templates/tools/tools.html?highlight=authors#filled-out-example-of-multiple-tools-in-the-same-repository"
+          target="_blank"
+          rel="noopener noreferrer"
+          >the .dockstore.yml</a
+        >, which is preferred if you have ORCID ids for the authors, or by adding them
```
kathy-t, line 187: ORCID wants us to call it `ORCID iDs` instead of `ids` https://info.orcid.org/brand-guidelines/

kathy-t (approved):

KimberleyChong (approved):

PR #1753, opened 2023-04-11
DOCK-2352: Fixed path for external files when using custom readme

hyunnaye:
**Description**
This PR fixes the `baseUrl` of the readme for `ngx-markdown`. Before the fix, when using a custom readme, the base path was still set to the root of the repo which resulted in 404 errors when redirecting to external files in the readme. This PR will set the `baseUrl` to the directory of the custom readme file.

For workflow: https://qa.dockstore.org/workflows/github.com/denis-yuen/myco/myco_cleaned:develop?tab=info
Before, the input.md link would be https://github.com/denis-yuen/myco/blob/develop/inputs.md leading to a 404 error. This PR changes it to https://github.com/denis-yuen/myco/blob/develop/doc/inputs.md 

Before, images would lead to a 404 error on the console. ([Workflow](https://qa.dockstore.org/workflows/github.com/denis-yuen/myco/myco_sra:main?tab=info))
Screenshot from 2023-04-12 13-13-58

It now links to https://github.com/denis-yuen/myco/blob/split-further/doc/doc_making_resources/myco_sra_flowchart_1.png thus removing the 404 errors on console. The images still do not load as they are linked to its GitHub link instead of the raw image, we have created a [ticket](https://github.com/dockstore/dockstore/issues/5459) to implement our wrapper to load our images correctly.
Screenshot from 2023-04-12 13-19-08

**Review Instructions**
Review that the basepath is now correctly set.

**Issue**
https://github.com/dockstore/dockstore/issues/5423

**Security**

### DIFF
--- a/src/app/shared/entry/base-url.pipe.spec.ts
+++ b/src/app/shared/entry/base-url.pipe.spec.ts
@@ -5,7 +5,10 @@ describe('Pipe: BaseUrl', () => {
     const pipe = new BaseUrlPipe();
     expect(pipe).toBeTruthy();
     expect(pipe.transform('https://github.com/foobar/hello_world', 'master')).toBe('https://github.com/foobar/hello_world/blob/master/');
-    expect(pipe.transform('https://gitlab.com/foobar/hello_world', 'develop', 'dir/readme.md')).toBe(
+    expect(pipe.transform('https://github.com/foobar/hello_world', 'master', null)).toBe(
+      'https://github.com/foobar/hello_world/blob/master/'
+    );
+    expect(pipe.transform('https://gitlab.com/foobar/hello_world', 'develop', '/dir/readme.md')).toBe(
       'https://gitlab.com/foobar/hello_world/blob/develop/dir/'
     );
     expect(pipe.transform('https://bitbucket.org/foobar/hello_world', '1.0')).toBe('https://bitbucket.org/foobar/hello_world/src/1.0/');
--- a/src/app/shared/entry/base-url.pipe.spec.ts
+++ b/src/app/shared/entry/base-url.pipe.spec.ts
@@ -4,14 +4,10 @@ describe('Pipe: BaseUrl', () => {
   it('create an instance and return base url', () => {
     const pipe = new BaseUrlPipe();
     expect(pipe).toBeTruthy();
-    expect(pipe.transform('https://github.com/foobar/hello_world', 'master', null)).toBe(
-      'https://github.com/foobar/hello_world/blob/master/'
-    );
+    expect(pipe.transform('https://github.com/foobar/hello_world', 'master')).toBe('https://github.com/foobar/hello_world/blob/master/');
     expect(pipe.transform('https://gitlab.com/foobar/hello_world', 'develop', 'dir/readme.md')).toBe(
       'https://gitlab.com/foobar/hello_world/blob/develop/dir/'
     );
-    expect(pipe.transform('https://bitbucket.org/foobar/hello_world', '1.0', null)).toBe(
-      'https://bitbucket.org/foobar/hello_world/src/1.0/'
-    );
+    expect(pipe.transform('https://bitbucket.org/foobar/hello_world', '1.0')).toBe('https://bitbucket.org/foobar/hello_world/src/1.0/');
   });
 });
--- a/src/app/shared/entry/base-url.pipe.spec.ts
+++ b/src/app/shared/entry/base-url.pipe.spec.ts
@@ -4,8 +4,14 @@ describe('Pipe: BaseUrl', () => {
   it('create an instance and return base url', () => {
     const pipe = new BaseUrlPipe();
     expect(pipe).toBeTruthy();
-    expect(pipe.transform('https://github.com/foobar/hello_world', 'master')).toBe('https://github.com/foobar/hello_world/blob/master/');
-    expect(pipe.transform('https://gitlab.com/foobar/hello_world', 'develop')).toBe('https://gitlab.com/foobar/hello_world/blob/develop/');
-    expect(pipe.transform('https://bitbucket.org/foobar/hello_world', '1.0')).toBe('https://bitbucket.org/foobar/hello_world/src/1.0/');
+    expect(pipe.transform('https://github.com/foobar/hello_world', 'master', null)).toBe(
+      'https://github.com/foobar/hello_world/blob/master/'
+    );
+    expect(pipe.transform('https://gitlab.com/foobar/hello_world', 'develop', 'dir/readme.md')).toBe(
+      'https://gitlab.com/foobar/hello_world/blob/develop/dir/'
+    );
+    expect(pipe.transform('https://bitbucket.org/foobar/hello_world', '1.0', null)).toBe(
+      'https://bitbucket.org/foobar/hello_world/src/1.0/'
+    );
   });
 });
--- a/src/app/shared/entry/base-url.pipe.ts
+++ b/src/app/shared/entry/base-url.pipe.ts
@@ -9,6 +9,7 @@ export class BaseUrlPipe implements PipeTransform {
    * Constructs a base url for relative links in markdown README files.
    * @param providerUrl
    * @param versionName
+   * @param readMePath path for desired base url is different from the root of the repo. Ex. readMePath defined in .dockstore.yml
    * @returns {string} base url to be used by ngx-markdown
    */
   transform(providerUrl: string, versionName: string, readMePath: string): string {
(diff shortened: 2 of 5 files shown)


src/app/workflow/info-tab/info-tab.component.ts
```diff
@@ -132,6 +133,12 @@ export class InfoTabComponent extends EntryTab implements OnInit, OnChanges {
       this.workflowsService.getWorkflowVersionOrcidAuthors(this.workflow.id, this.selectedVersion.id).subscribe((orcidAuthors) => {
         this.authors = [...this.selectedVersion.authors, ...orcidAuthors];
       });
+      if (this.selectedVersion.descriptionSource === 'CUSTOM_README') {
+        const slashIndex = this.selectedVersion.readMePath.slice(0, -1).lastIndexOf('/'); //index of the slash before the readme filename. This allows us to get the path to its parent directory.
```
svonworl, line 137: If I understand correctly, the `slice(0, -1)` removes the last character from the string, why do we need to do that?

src/app/workflow/info-tab/info-tab.component.ts
```diff
@@ -132,6 +133,12 @@ export class InfoTabComponent extends EntryTab implements OnInit, OnChanges {
...
+        this.readMeBasePath = this.selectedVersion.name + this.selectedVersion.readMePath.substring(0, slashIndex);
```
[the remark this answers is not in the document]
hyunnaye: The path always starts with a slash since relative paths get converted to absolute from this PR https://github.com/dockstore/dockstore/pull/5110

coverbeck (approved):

svonworl (approved):

src/app/workflow/info-tab/info-tab.component.ts
```diff
@@ -132,6 +133,12 @@ export class InfoTabComponent extends EntryTab implements OnInit, OnChanges {
       this.workflowsService.getWorkflowVersionOrcidAuthors(this.workflow.id, this.selectedVersion.id).subscribe((orcidAuthors) => {
         this.authors = [...this.selectedVersion.authors, ...orcidAuthors];
       });
+      if (this.selectedVersion.descriptionSource === 'CUSTOM_README') {
```
kathy-t, line 136: I think it would be better to move this logic into the `BaseUrlPipe` so that if the pipe is used else where in the future, we're aware that the custom readme path can affect the base url. Maybe add a `customReadMePath` parameter to the pipe - this way, we can also create some tests for the custom readme path in https://github.com/dockstore/dockstore-ui2/blob/develop/src/app/shared/entry/base-url.pipe.spec.ts

src/app/workflow/info-tab/info-tab.component.ts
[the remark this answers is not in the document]
kathy-t: Should add as a comment that the readme path always starts with a slash because it's not immediately obvious

src/app/workflow/info-tab/info-tab.component.ts
svonworl: A wrinkle is that notebooks can also contain markdown, which can reference images using relative URLs.  The base URL for a notebook's markdown is probably similar to that of a custom `README` (the repo info + the directory portion of the notebook path), so if we can account for that in our pipe mods, that would be great.

src/app/shared/entry/base-url.pipe.spec.ts
```diff
@@ -5,7 +5,9 @@ describe('Pipe: BaseUrl', () => {
     const pipe = new BaseUrlPipe();
     expect(pipe).toBeTruthy();
     expect(pipe.transform('https://github.com/foobar/hello_world', 'master')).toBe('https://github.com/foobar/hello_world/blob/master/');
-    expect(pipe.transform('https://gitlab.com/foobar/hello_world', 'develop')).toBe('https://gitlab.com/foobar/hello_world/blob/develop/');
+    expect(pipe.transform('https://gitlab.com/foobar/hello_world', 'develop', 'dir/readme.md')).toBe(
```
kathy-t, line 8: Based on this other [comment thread](https://github.com/dockstore/dockstore-ui2/pull/1753#discussion_r1164818203), I thought the readme path always started with a slash?

kathy-t, line 8: I would also add a test case for when the readme path is `null`

src/app/shared/entry/base-url.pipe.ts
```diff
@@ -4,21 +4,31 @@ import { Pipe, PipeTransform } from '@angular/core';
   name: 'baseUrl',
 })
 export class BaseUrlPipe implements PipeTransform {
+  private markdownBasePath: string;
```
kathy-t, line 7: I don't think this needs to be a private field because it's only used inside `transform`. It can be a variable inside that function instead

src/app/shared/entry/base-url.pipe.ts
```diff
@@ -4,21 +4,31 @@ import { Pipe, PipeTransform } from '@angular/core';
...
+    if (customPath) {
+      const slashIndex = customPath.lastIndexOf('/'); //index of the slash before the readme filename. This allows us to get the path to its parent directory.
+      this.markdownBasePath = customPath.substring(0, slashIndex) + '/'; // the last slash is needed by ngx-markdown
+    } else {
+      this.markdownBasePath = '';
+    }
+
     if (providerUrl.startsWith('https://github.com/') || providerUrl.startsWith('https://gitlab.com/')) {
-      return providerUrl + '/blob/' + versionName + '/'; // the last slash is needed by ngx-markdown
+      return providerUrl + '/blob/' + versionName + '/' + this.markdownBasePath; // the last slash needed by ngx-markdown is provided by this.markdownBasePath.
```
kathy-t, line 29: The comment doesn't seem quite accurate. `markdownBasePath` can be `''` which doesn't have the last slash. You could remove the last slash assignment on this line and add it to line 25 where you assign `this.markdownBasePath = ''`

src/app/shared/entry/base-url.pipe.ts
```diff
@@ -4,21 +4,31 @@ import { Pipe, PipeTransform } from '@angular/core';
...
     } else if (providerUrl.startsWith('https://bitbucket.org/')) {
-      return providerUrl + '/src/' + versionName + '/'; // the last slash is needed by ngx-markdown
+      return providerUrl + '/src/' + versionName + '/' + this.markdownBasePath; //  the last slash needed by ngx-markdown is provided by this.rmarkdownBasePath.
```
kathy-t, line 31: Similar comment about the last slash

src/app/workflow/info-tab/info-tab.component.ts
hyunnaye: To account for notebook markdown files, I generalized the parameter name in `BaseUrlPipe` to `customPath` instead of `readMePath`

src/app/shared/entry/base-url.pipe.ts
```diff
@@ -4,21 +4,31 @@ import { Pipe, PipeTransform } from '@angular/core';
...
+  private markdownBasePath: string;
   /**
    * Constructs a base url for relative links in markdown README files.
    * @param providerUrl
    * @param versionName
+   * @param customPath If applicable, the relative path of markdown file if its parent directory is not the root of the repo. Ex. custom readMePath defined in .dockstore.yml
    * @returns {string} base url to be used by ngx-markdown
    */
-  transform(providerUrl: string, versionName: string): string {
+  transform(providerUrl: string, versionName: string, customPath?: string): string {
```
kathy-t, line 15: `customPath` can be null, so you could also add the `null` type beside `string`

src/app/shared/entry/base-url.pipe.spec.ts
hyunnaye: Yes, it does. Sorry looks like I didn't add it on this. Thanks for catching that

src/app/shared/entry/base-url.pipe.ts
```diff
@@ -8,17 +8,27 @@ export class BaseUrlPipe implements PipeTransform {
...
+  transform(providerUrl: string, versionName: string, customPath?: string | null): string {
+    let markdownBasePath: string;
     if (!providerUrl || !versionName) {
       console.error('providerUrl or versionName is not truthy');
       return null;
     }
+
+    if (customPath) {
+      const slashIndex = customPath.lastIndexOf('/'); //index of the slash before the readme filename. This allows us to get the path to its parent directory.
+      markdownBasePath = customPath.substring(0, slashIndex) + '/'; // the last slash is needed by ngx-markdown
```
kathy-t, line 23: Can you add a comment that `customPath` always starts with a `/`?

kathy-t (approved):

PR #1755, opened 2023-04-14
Display multiple authors in search

kathy-t:
**Description**
Corresponding webservice PR: https://github.com/dockstore/dockstore/pull/5461

This PR displays multiple authors in the search page. ORCID authors have a link to their ORCID page.

I've removed most usages of the Entry `author` field. The list entries page like this [one](https://qa.dockstore.org/workflows) still use the Entry `author`. Removing that would be a little more involved because it uses the `author` as sorting query parameter to the `allPublished` endpoint. I created a ticket https://github.com/dockstore/dockstore/pull/5461 to remove this usage of the `author` field.

Misc changes while I was in the area:
- Added a tooltip for the topics on the search page so that users can still read the full topic even if they're truncated
- Some facet names were camel case, so I added spaces

Screenshots:
Screenshot 2023-04-14 at 17-29-04 Dockstore Search

Screenshot 2023-04-14 at 17-51-57 Dockstore Search

**Review Instructions**
- Create a GitHub app workflow with a .dockstore.yml that **doesn't** contain authors. The descriptor of the workflow should contain an author.
- Publish the workflow
- Verify that the search page displays the descriptor author for this workflow
- Add two or more authors to the .dockstore.yml (we want multiple authors)
- Verify that the search page displays the .dockstore.yml authors for this workflow. There should be a search facet for each author.

**Issue**
[SEAB-3629](https://ucsc-cgl.atlassian.net/browse/SEAB-3629)

**Security**

[SEAB-3629]: https://ucsc-cgl.atlassian.net/browse/SEAB-3629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/shared/bioschema.service.ts
```diff
@@ -80,9 +81,6 @@ export class BioschemaService {
         '@type': 'Organization',
         name: tool.namespace,
       };
-      if (tool.email) {
```
kathy-t, line 83: I removed this and the corresponding portion for workflows below because an entry's email is the email of the author

svonworl (approved):

coverbeck (approved):

PR #1756, opened 2023-04-17
My Notebooks Page - SEAB-5114

KimberleyChong:
**Description**
Adds the `my-notebooks` page (SEAB-5114), adds the GitHub apps registration interface for notebooks (SEAB-5119), and updates and hides notebooks in collections (SEAB-5386).

~~Opening this PR early just to get some feedback earlier rather than later, but when the release branch is made I'll switch the destination branch from `develop` to `release`.~~

**Quick Overview:**

https://user-images.githubusercontent.com/97123241/232588061-53ecb4a0-927f-4260-84f0-396448847d9b.mp4

What my-notebooks looks like, I just copied the existing notebook in the db and assigned it to myself for the purpose of having a notebook entry:

image

I created the `register-github-app-modal.component` which closely resembles the register a workflow github apps option:

image

I also reused the same registration screen for services since previously clicking "Add a service" would open GitHub with no intermediate step.

image

- Renamed `SharedWorkflowServicesModule` to `SharedWorkflowServicesNotebooksModule` since it includes all three now
- I removed `MyServicesService` and `MyBioWorkflowsService` components, as they both contained identical functions differing in entry type, and altered `selectEntry` in `MyWorkflowsService` to handle workflows, services and notebooks without calling different `entryType` specific functions.

Also added the warning banner to the collections page along with the feature flag to hide notebooks in collections. The entry link here also needed to be updated to properly navigate to a notebook entry.

image

While doing that I decided to re-export the notebook icon, to a smaller size, I noticed the tools and workflows icons have a much smaller "intrinsic size", so I re-exported it to match so we don't need to keep using styles to resize it smaller.

**Review Instructions**
Using the feature flag, my notebooks should appear in the my-sidebar and you should be able to view the my-notebooks page. Everything should appear as the demo video.

**Issue**
SEAB-5114, SEAB-5119, SEAB-5386

**Security**


denis-yuen:
> Opening this PR early just to get some feedback earlier rather than later, but when the release branch is made I'll switch the destination branch from `develop` to `release`.

Think this should stay as `develop` actually, if we're targeting it at 1.15

src/app/workflow/register-workflow-modal/register-github-app-modal/register-github-app-modal.component.ts
```diff
@@ -0,0 +1,58 @@
+/*
+ *    Copyright 2018 OICR
```
coverbeck, line 2: Out of date copied copyright :)

src/app/organizations/collection/collection.component.html
```diff
@@ -101,18 +104,28 @@ <h3 *ngIf="Dockstore.FEATURES.enableNotebooks">Workflows, Notebooks, and Tools</
...
+                "
+                class="pb-3"
+              >
                 <a
                   [routerLink]="[
-                    '/' + (entry.entryType === 'tool' ? 'container' : 'workflow') + 's/',
+                    '/' +
+                      (entry.entryType === entryType.Tool
+                        ? 'container'
+                        : entry.entryType === entryType.Workflow
```
coverbeck, line 118: IDE error on `entryType.Workflow` -- it doesn't exist.

coverbeck, line 118: IDE error on `entryType.Workflow` -- it doesn't exist. Actually, we have a pipe for this src/app/entry/router-link.pipe.ts

coverbeck (commented):
> Think this should stay as develop actually, if we're targeting it at 1.15

I'm unclear, since we haven't cut the release branches, will this part of 1.14 then (behind the feature flag)? I think it's good if we do, but can go either way.

src/app/workflow/register-workflow-modal/register-github-app-modal/register-github-app-modal.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    @Inject(MAT_DIALOG_DATA) public entryType: string,
+    private registerWorkflowModalService: RegisterWorkflowModalService,
+    private alertQuery: AlertQuery,
+    protected sessionQuery: SessionQuery,
+    private userQuery: UserQuery
+  ) {}
+
+  ngOnInit() {
+    this.username$ = this.userQuery.username$;
+    this.usernameChangeRequired$ = this.userQuery.user$.pipe(map((user) => user.usernameChangeRequired));
```
svonworl, line 50: Definitely not a biggie, but in the interest of consistency, we could make `isNameChangeRequired` an observable property of `userQuery`, and remove the dup code from this and the other modal.

svonworl (approved):
Good job, great work!

denis-yuen:
> I'm unclear, since we haven't cut the release branches, will this part of 1.14 then (behind the feature flag)? I think it's good if we do, but can go either way.

Think we just have to keep this open until later in the week when we start cutting release branches

src/app/myworkflows/myworkflows.service.ts
```diff
@@ -91,10 +91,31 @@ export class MyWorkflowsService extends MyEntriesService<Workflow, OrgWorkflowOb
...
+        })
+      )
+      .subscribe(
+        (notebooks: Array<Workflow>) => {
+          this.setupEntries(notebooks, EntryType.Notebook);
           this.alertService.simpleSuccess();
-          this.selectEntry(this.recomputeWhatEntryToSelect([...(services || []), ...[]]), EntryType.Service);
         },
         (error) => {
           this.workflowService.setWorkflows([]);
```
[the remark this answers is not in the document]
KimberleyChong: Yes this is true 👍️, I can call it with an empty array for the error case

src/app/organizations/collection/collection.component.html
KimberleyChong: Oh woah thanks for catching this, supposed to be `BioWorkflow`.  What IDE do you use Charles? That's really useful

KimberleyChong: > Actually, we have a pipe for this src/app/entry/router-link.pipe.ts

Mm previously considered that, unfortunately it takes in a type of `Workflow` to compute the url but here we're dealing with type `CollectionEntry`. But maybe its a good idea to add support for both, I'll update the pipe to take in `entry: Workflow | CollectionEntry` and handle both cases

src/app/organizations/collection/collection.component.html
```diff
@@ -101,18 +104,28 @@ <h3 *ngIf="Dockstore.FEATURES.enableNotebooks">Workflows, Notebooks, and Tools</
             <h3 *ngIf="!Dockstore.FEATURES.enableNotebooks">Workflows and Tools</h3>
 
             <div data-cy="collectionEntry" *ngFor="let entry of collection?.entries">
-              <mat-card class="pb-3">
+              <mat-card
+                *ngIf="
+                  (entry.entryType !== entryType.Notebook && !Dockstore.FEATURES.enableNotebooks) || Dockstore.FEATURES.enableNotebooks
```
kathy-t, line 109: Why do you need this `*ngIf`? If it's not a notebook, then it'll satisfy the condition both when `enableNotebooks` is true and false. If it is a notebook, then `enableNotebooks` should be true, which also satisfies the condition

src/app/organizations/collection/collection.component.html
```diff
@@ -31,6 +31,9 @@
 <app-loading [loading]="(loadingCollection$ | async) || (loadingOrganization$ | async)">
   <div *ngIf="organization$ | async as organization">
     <div class="container mb-5" *ngIf="collection$ | async as collection" fxLayout="column">
+      <div *ngIf="Dockstore.FEATURES.enableNotebooks" class="mt-3">
```
kathy-t, line 34: Looking at the collection page in your video, it's not clear what the banner is referring. Perhaps we should mention the feature in the banner, like Charles suggested during the demo.

src/app/organizations/collection/collection.component.html
```diff
@@ -101,18 +104,28 @@ <h3 *ngIf="Dockstore.FEATURES.enableNotebooks">Workflows, Notebooks, and Tools</
             <h3 *ngIf="!Dockstore.FEATURES.enableNotebooks">Workflows and Tools</h3>
 
             <div data-cy="collectionEntry" *ngFor="let entry of collection?.entries">
-              <mat-card class="pb-3">
+              <mat-card
```
kathy-t, line 107: Brainstorming, does the notebook mat card need the `preview` flag?

src/app/workflow/register-workflow-modal/register-github-app-modal/register-github-app-modal.component.html
```diff
@@ -0,0 +1,33 @@
...
+</p>
+<div fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="2rem" class="mb-2">
+  <a
+    mat-raised-button
+    type="button"
+    class="accent-1-dark-btn mat-elevation-z"
+    [href]="gitHubAppInstallationLink$ | async"
+    [disabled]="usernameChangeRequired$ | async"
+    matTooltip="Manage Dockstore installations on GitHub"
+    ><mat-icon>add</mat-icon> Manage Dockstore installations on GitHub</a
```
kathy-t, line 19: I think this link might have a leading space (I remember seeing something similar on QA)

src/app/myworkflows/my-workflow/my-workflow.component.html
```diff
@@ -152,7 +138,7 @@ <h3>Shared with me</h3>
                   {{ entryType$ | async }}, do one of the following:
                 </p>
                 <ul>
-                  <li *ngIf="entryType === EntryType.Service">
+                  <li *ngIf="entryType === EntryType.Service || entryType === EntryType.Notebook">
                     Press the 'Manage on GitHub' button and follow the GitHub apps web interface to create services from your GitHub
```
kathy-t, line 142: This might be out of date because I don't see a Manage on GitHub button

src/app/workflow/register-workflow-modal/register-github-app-modal/register-github-app-modal.component.html
```diff
@@ -0,0 +1,33 @@
+<div class="mb-3" *ngIf="Dockstore.FEATURES.enableNotebooks && entryType === EntryType.Notebook">
```
kathy-t, line 1: This dialog should have a title, like Register Workflow

src/app/myworkflows/myworkflows.service.ts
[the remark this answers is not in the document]
KimberleyChong: It's there :) just cut off in the diff:

https://github.com/dockstore/dockstore-ui2/blob/12d6858b460398eb511f48cf0827fcbc3f26b51c/src/app/myworkflows/myworkflows.service.ts#L123

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.html
```diff
@@ -63,11 +63,14 @@
...
               </p>
-              <p data-cy="no-published-service-message" *ngIf="pageEntryType === EntryType.Service" class="text-center size-small">
-                No published services.
-              </p>
-              <p data-cy="no-published-workflow-message" *ngIf="pageEntryType === EntryType.BioWorkflow" class="text-center size-small">
-                No published workflows.
+              <p
+                [attr.data-cy]="'no-published-' + pageEntryType + '-message'"
+                *ngIf="
+                  pageEntryType === EntryType.Service || pageEntryType === EntryType.BioWorkflow || pageEntryType === EntryType.Notebook
```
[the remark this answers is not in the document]
KimberleyChong: Yeah, `apptools` isn't included because the format `No published {{ pageEntryType }}s.` doesn't fit well. 

The `pageEntryType` for an appTool is `appTool`, which displays `No published appTools.` instead of `No published tools.` so I decided to keep it in a case of it's own

But great point! I can actually simplify a lot and remove the `*ngIf`:
```
<p [attr.data-cy]="'no-published-' + pageEntryType + '-message'" class="text-center size-small">
  No published {{ pageEntryType === EntryType.Service || pageEntryType === EntryType.BioWorkflow || pageEntryType === EntryType.Notebook ? pageEntryType : 'tool' }}s.
</p>
```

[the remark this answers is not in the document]
svonworl: We also now have the `EntryTypeMetadata` objects, which contain the singular and plural terms that we use to publicly refer to each entry type.  It may not be the right time to do so, but if/when we convert this template to use them, we may not even need the conditional.

src/app/myworkflows/myworkflows.service.ts
```diff
@@ -213,11 +246,30 @@ export class MyWorkflowsService extends MyEntriesService<Workflow, OrgWorkflowOb
...
   }
 
+  /**
+   * Modal that navigates user to GitHub page that allows them to add our GitHub app to their repo
+   * Used by Register Service and Register Notebook buttons
+   *
+   * @param entryType
+   */
+  openRegisterGithubAppModal(entryType: EntryType) {
+    const dialogRef = this.matDialog.open(RegisterGithubAppModalComponent, { width: '600px', data: entryType });
```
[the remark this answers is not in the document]
KimberleyChong: Makes sense, I'll go with `bootstrap4mediumModalSize` and update for both `RegisterWorkflowModal` and `RegisterGithubAppModal` then

src/app/workflow/register-workflow-modal/register-github-app-modal/register-github-app-modal.component.html
KimberleyChong: Thank you!!

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.html
```diff
@@ -94,11 +97,14 @@
...
-              <p data-cy="no-unpublished-workflow-message" *ngIf="pageEntryType === EntryType.BioWorkflow" class="text-center size-small">
-                No unpublished workflows.
+              <p
+                [attr.data-cy]="'no-unpublished-' + pageEntryType + '-message'"
+                *ngIf="
+                  pageEntryType === EntryType.Service || pageEntryType === EntryType.BioWorkflow || pageEntryType === EntryType.Notebook
+                "
+                class="text-center size-small"
+              >
+                No unpublished {{ pageEntryType }}s.
```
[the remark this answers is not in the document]
KimberleyChong: Gotcha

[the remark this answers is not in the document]
KimberleyChong: Should services have it as well then since they're also registered through GitHub apps? I supposed they're excluded for a reason though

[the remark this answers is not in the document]
denis-yuen: Probably never got around to it, probably not a priority still

src/app/organizations/collection/collection.component.html
KimberleyChong: > If it is a notebook, then `enableNotebooks` should be true

Yeah the thing is, this entry being a `notebook` and `enableNotebooks` being set are independent of each other, the issue was without this `*ngIf` entries that were notebooks we being shown regardless of the value of `enableNotebooks` (i.e. all the time)

The conditions here are a bit hard to imagine:

The first part:
`entry.entryType !== entryType.Notebook && !Dockstore.FEATURES.enableNotebooks`

- When the feature flag is **not** enabled, display an entry only if it is **not** a notebook

The second part:
`Dockstore.FEATURES.enableNotebooks`

- When the feature flag **is** enabled, display **all** entries (including notebooks)

That being said, the logic here can be simplified and maintain the same results :) will update to:
`entry.entryType !== entryType.Notebook || Dockstore.FEATURES.enableNotebooks`

image

src/app/organizations/collection/collection.component.html
KimberleyChong: Noted, saw your comment on the original Jira ticket and while doing those updates will add in a `feature` parameter

src/app/myworkflows/my-workflow/my-workflow.component.html
KimberleyChong: I remember thinking it didn't make sense, thanks for verifying. Updating to be as so, which I think makes sense:

```
<li>
  Press the 'Register a {{ entryType$ | async | titlecase }}' button 
  <span *ngIf="entryType === EntryType.Service || entryType === EntryType.Notebook; else manualRegistration">and follow the GitHub apps web interface to create {{ entryType$ | async }}s from your GitHub repositories.</span>
  <ng-template #manualRegistration >to manually add a {{ entryType$ | async }} using the registration wizard.</ng-template>
</li>
```
**For Services/Notebooks:**

Before:
image

After:
image

**For workflows:**

Before:
image

After:
image

Going to update `my-tool.component` to match while here:
Before:
image
After:
image

Changed buttons to say "Register a {{ entryType }}" to match the other registration buttons:
image
image

KimberleyChong:
> > Think this should stay as develop actually, if we're targeting it at 1.15
> 
> I'm unclear, since we haven't cut the release branches, will this part of 1.14 then (behind the feature flag)? I think it's good if we do, but can go either way.

Yep, everything is hidden behind the feature flag, no strong feelings whether it makes the release or not, but I suppose one thing that would be good to have in production from this PR is the feature flag to hide notebooks in collections... otherwise notebooks appear if an organization has added it to their collection without the FF, but also it leads to a 404 if you click on it because the link also gets updated in this PR.

src/app/organizations/collection/collection.component.html
KimberleyChong: Maybe instead of adding it to the card, I could add the preview bubble to the title since we usually add it on the tab/title, similar to my-starred?



image

src/app/workflow/register-workflow-modal/register-github-app-modal/register-github-app-modal.component.html
kathy-t: Yeah, that's the right one. When you hover over the link, you can see an underline for the space, which IMO looks awkward:
image


Does it make a huge difference without the space? If not, I would remove it. If it does, consider adding a right margin to the icon instead of the space. 

Admittedly, this is a very minor thing so your call whether or not you think it's worth doing :slightly_smiling_face:

src/app/organizations/collection/collection.component.html
kathy-t: I think it's a little ambiguous beside the title because it also mentions workflows and tools. Since you're adding the feature name to the feature warning banner, we can probably live without the preview label on this page

KimberleyChong: Got it, makes sense

src/app/workflow/register-workflow-modal/register-github-app-modal/register-github-app-modal.component.ts
KimberleyChong: Makes sense

src/app/workflow/register-workflow-modal/register-github-app-modal/register-github-app-modal.component.html
KimberleyChong: Ah I see, in that case I can add the `no-underline` style which gets rid of the underline on hover which I think looks better for links styled as buttons :)

KimberleyChong:
**Update**: Changes following feedback, main points:
- Updated `routerLinkPipe` to support `collectionEntry` and `DockstoreTool` types to use for routing collections
- added `isUsernameChangeRequired$` as an observable boolean to `userQuery` for consistency/to remove dup subscription code
- Use Bootstrap for modal size, add header to modal and other updates to the `register {{ entryType }}` modals
- Simplify logic, update outdated registration text

denis-yuen:
FYI this is fine to go into develop now from a github perspective

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.html
```diff
@@ -60,17 +60,13 @@
...
-              <p data-cy="no-published-apptool-message" *ngIf="pageEntryType === EntryType.AppTool" class="text-center size-small">
-                No published tools.
-              </p>
-              <p
-                [attr.data-cy]="'no-published-' + pageEntryType + '-message'"
-                *ngIf="
+              <p [attr.data-cy]="'no-published-' + pageEntryType + '-message'" class="text-center size-small">
+                No published
+                {{
                   pageEntryType === EntryType.Service || pageEntryType === EntryType.BioWorkflow || pageEntryType === EntryType.Notebook
```
kathy-t, line 66: This could be simplified to `pageEntryType == EntryType.AppTool ? 'tool' : pageEntryType`

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.html
```diff
@@ -94,25 +90,23 @@
...
-              <p data-cy="no-unpublished-apptool-message" *ngIf="pageEntryType === EntryType.AppTool" class="text-center size-small">
-                No unpublished tools.
-              </p>
-              <p
-                [attr.data-cy]="'no-unpublished-' + pageEntryType + '-message'"
-                *ngIf="
+              <p [attr.data-cy]="'no-unpublished-' + pageEntryType + '-message'" class="text-center size-small">
+                No unpublished
+                {{
                   pageEntryType === EntryType.Service || pageEntryType === EntryType.BioWorkflow || pageEntryType === EntryType.Notebook
```
kathy-t, line 96: Same comment

src/app/entry/router-link.pipe.ts
```diff
@@ -1,25 +1,71 @@
...
-      case EntryType.BioWorkflow:
-        return '/workflows/' + workflow.full_workflow_path;
-      case EntryType.Service:
-        return '/services/' + workflow.full_workflow_path;
-      case EntryType.AppTool:
-        return '/containers/' + workflow.full_workflow_path;
-      case EntryType.Notebook:
-        return '/notebooks/' + workflow.full_workflow_path;
+
+    if ((entry as Workflow).full_workflow_path) {
```
kathy-t, line 22: Steve mentioned that we now have `EntryTypeMetadata` in this [comment](https://github.com/dockstore/dockstore-ui2/pull/1756#discussion_r1171580682) and I think using that here would greatly simplify this function. `EntryTypeMetadata.sitePath` contains the entry type specific portion of the site path (ex: `containers`), so that would remove the need for all the switches.

`EntryTypeMetadata` should be a field in `DockstoreTool | Workflow`, so you can easily access it there. As for `CollectionEntry`, it's not a field in there, but you can use this function in the `EntryTypeMetadataService` to retrieve the `EntryTypeMetadata` based on the collection entry's `entryType`: https://github.com/dockstore/dockstore-ui2/blob/3478bfb2e97481e1b127224bdb06b268ef8c94a1/src/app/entry/type-metadata/entry-type-metadata.service.ts#L30

KimberleyChong:
**Update**:
- refactor `RouterLinkPipe` to use `EntryTypeMetadata` and update `spec.ts`
- add `EntryTypeMetadata` to mock objects in `openapi-mocked-objects.ts`
- create `EntryTypeMetadataStubService` in `service-stubs.ts`

kathy-t (approved):

src/app/organizations/collection/collection.component.html
coverbeck: I use WebStorm, but if you're using IntelliJ, I think you can configure it with the same plugins as WebStorm.

denis-yuen (approved):

PR #1757, opened 2023-04-17
SEAB-5117: Formatted "preview" notebook with "more Angular" style

svonworl:
**Description**
This PR displays a user-friendly formatted non-interactive representation of a Jupyter notebook in the "Preview" tab of its page.  This is a rewrite of #1747 and is intended to be more "Angular"  to address reviewer feedback.   Please see the original PR for background material.

We got to approximately 57% Angular, and we could go further, but IMHO, the Angular equivalents of the remaining HTML syntheses might be harder to understand, buggier, and more difficult to maintain...

This implementation contains four new subcomponents that encapsulate portions of the "preview" notebook as follows:
 
* `NotebookSourceComponent`: highlighted source code of a code cell.
* `NotebookMarkdownComponent`: rendered markdown cell.
* `NotebookStreamOutputComponent`: plain-text "stream" output of a code cell.
* `NotebookMimeBundleOutputComponent`: best item from a mime bundle of the output of a code cell. 

Each generates its contents and then embeds them into its template via a property or `innerHTML`.

All user-supplied HTML (in the markdown and `text/html` mime bundle outputs) is sanitized twice, via the markdown-wrapper's DOMPurify invocation and the Angular sanitizer.

We have disposed of the previous "live" DOM manipulation.  We use MathJax to format embedded TeX math, and because the Angular sanitizer destroys MathJax markup, `NotebookMarkdownComponent` uses `DOMParser` to convert the compiled+sanitized markdown HTML to a DOM representation that is disconnected from the browser context, preventing any naughty contents from being executed.  MathJax is then run on this DOM representation.  We run our own mini-sanitizer on the MathJax output, which deletes everything but text nodes and elements that start with `mjx-`,   Then, we convert to HTML, prepend the Markdown-generated CSS, and display the result via `innerHTML` by bypassing the final Angular sanitization .  We configureMathJax to run with a tight configuration of its `ui/safe` module, which ostensibly prevents anything from being injected into the MathJax-generated CSS.  Creating our own mini-sanitizer for this CSS would be the next logical security improvement, but the current impl is probably pretty strong, and returns are diminishing...

Regarding the previous feedback:

* Instances of `any` are reduced, but not eliminated.  Update: I have added some types to describe the parsed JSON parts of the notebook per Charles' suggestion, the rest of this line item is obsolete.  Our open ticket is to reduce _implicit_ `anys`, not eliminate `any` from use entirely.  https://ucsc-cgl.atlassian.net/browse/DOCK-1473  In this case, `JSON.parse` returns the parsed notebook as type `any`, and we intentionally propagate chunks of it likewise and gracefully fail when the values don't match expectations.  IMHO, it's not worth parsing/validating to a more strongly-typed object for a number of reasons (which I'll not mention here to save space).

* This code triggers some additional IDE feedback about things being "not assignable", which appears to be caused by using Swagger and OpenAPI types as synonyms.  At the Javascript level, they describe the same object with the same fields, so I think "no action" is the best solution.  When we convert the UI to 100% OpenAPI, the problem will go away.

**Review Instructions**
Register some notebooks, pull up their entry pages, click their "Preview" tab, and confirm that they look ok.  You can find some test notebooks in: https://github.com/svonworl/test-notebooks

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5117

**Screenshot**
See https://github.com/dockstore/dockstore-ui2/pull/1747 for screenshots. "Code" tab has been relabeled "Preview".

**Security**
We're displaying lots of user-supplied content, and we use various 3rd party libraries to process it, so we need to be careful to avoid injection vulnerabilities (XSS). 

- (see above) If displaying markdown, use the `markdown-wrapper` component, which does extra sanitization
- (see above) Do due diligence on new 3rd party libraries, checking for CVEs
- (see above) Don't allow user-uploaded images to be served from the Dockstore domain


src/app/notebook/notebook.component.ts
```diff
@@ -0,0 +1,103 @@
...
+    // order.
+    // Assuming that Angular calls a component's `ngOnChanges()` directly
+    // after setting its @Input fields, a pleasant side effect of said
+    // cancellation is that when we handle a response, we know that @Input
+    // fields have the same values as when the corresponding request was
+    // started, which simplifies the code.  For example, the methods
+    // called by the handler to format the notebook can safely use the
+    // value of the `notebook` field to determine the notebook language.
+    this.currentSubscription?.unsubscribe();
+    this.currentSubscription = this.workflowsService
```
coverbeck, line 49: We're not always consistent, but we generally use the AlertService for Ajax requests (and the error).

coverbeck (commented):
Thanks for redoing the PR! I think it was worth it; the code is more digestible, to me at least. Left a few comments.

src/app/notebook/notebook.component.ts
svonworl: Yeah, this code is similar to that behind the Files tab, which doesn't use AlertService either.  Is the alert functionality redundant with the error message that appears where the preview should be?

src/app/notebook/notebook-markdown.component.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (mathjax == undefined) {
+      console.log('mathjax not found');
+      return html;
+    }
+
+    // Convert the input HTML to a DOM representation.
+    // DOMParser creates nodes that are detached from the browser context,
+    // which should prevent any embedded code or handlers from firing.
+    // For example, an `img` tag with an `onerror` handler.
+    const doc = this.domParser.parseFromString(html, 'text/html');
```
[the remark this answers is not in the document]
svonworl: The input to `parseFromString` is the output from the sanitizers, which probably is always well-formed HTML.  Apparently, `parseFromString` always returns a valid DOM, but if there's an error, it's a `parseerror` element.  So, just in case, I will add a check, which displays nothing if the result is a `parseerror` element.

src/app/workflow/workflow.component.html
```diff
@@ -244,12 +244,17 @@ <h3>
...
-                <p>A user-friendly representation of the notebook will appear here, after it is implemented in SEAB-5117.</p>
-                <p>For a Jupyter notebook, it'll be the contents of the code/documentation cells, nicely-formatted.</p>
-                <p>For other kinds of notebooks, it might be something different.</p>
-              </div>
+            <mat-tab id="previewTab" label="Preview">
+              <ng-template matTabContent>
+                <div class="mt-4 ml-4 mr-4">
+                  <app-notebook
+                    *ngIf="workflow && selectedVersion && (extendedWorkflow$ | async)"
+                    [notebook]="workflow"
```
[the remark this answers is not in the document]
coverbeck: 1. I thought we had an issue for this, but we don't, so I created dockstore/dockstore#5471.
2.  Looking at the IDE errors just caught a legit problem [here](https://github.com/dockstore/dockstore-ui2/pull/1756#discussion_r1170582877).
3. I entered the original comment, so I'm biased, but I don't think we should introduce additional errors, even though in the case, it's the same underlying JS object, it won't be a runtime failure. It just make things cleaner for the future.

Another optional way to fix it, instead of changing the import, is to allow both Swagger and OpenAPI for the input in notebook.component.ts.

src/app/notebook/notebook-markdown.component.ts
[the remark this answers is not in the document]
svonworl: Decided to return a resanitized value on parser error.

src/app/notebook/notebook-markdown.component.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+    // Recursively sanitize each child node.
+    // childNodes is "live" so we duplicate it beforehand to avoid problems if we delete a child.
+    for (const child of this.dup(node.childNodes)) {
+      this.sanitizeMathElement(child);
+    }
+  }
+
+  dup(nodes: NodeListOf<Node>): Node[] {
+    // The clumsy syntax below is necessary because in ES5, NodeListOf is not iterable.
```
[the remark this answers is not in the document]
svonworl: Was a compiler error, but I figured out a different solution, using `Array.from`, so `dup` is now gone.

src/app/notebook/notebook.component.ts
```diff
@@ -0,0 +1,103 @@
...
+  templateUrl: './notebook.component.html',
+})
+export class NotebookComponent implements OnChanges {
+  constructor(private workflowsService: WorkflowsService) {}
+  @Input() notebook: Workflow;
+  @Input() version: WorkflowVersion;
+  @Input() baseUrl: string;
+  loading = true;
+  error = false;
+  cells = [];
```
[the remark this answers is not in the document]
svonworl: I added some types to describe the parsed notebook JSON and further reduced the number of `anys`.

src/app/notebook/notebook-helpers.ts
```diff
@@ -0,0 +1,47 @@
+import { MimeBundle } from './notebook-types';
+
+// A list of the mime types we will display, ordered from "best" to "worst".
```
denis-yuen, line 3: (code) elaborate on what best means here, guessing best kind of image or something image like?

denis-yuen: Then again, why no svg love? Maybe this is dictated by the Python notebooks format?

denis-yuen:
> We're displaying lots of user-supplied content, and we use various 3rd party libraries to process it, so we need to be careful to avoid injection vulnerabilities (XSS).

highlighting for @david4096

src/styles.scss
```diff
@@ -1398,6 +1398,168 @@ div.mat-menu-panel {
...
+    .token.italic {
+      font-style: italic;
+    }
+
+    .token.deleted {
+      color: red;
+    }
+  }
+}
+
```
KimberleyChong, line 1562: Really great stuff. Probably missed this, just leaving the same comment as before:

> Good comments here, suggestion: because this is a large chunk of styles all pertaining to notebooks, I think it would be a good idea to create a separate file like `notebooks.scss` and import it into `styles.scss` similar to [featured-content.scss](https://github.com/dockstore/dockstore-ui2/blob/develop/src/featured-content.scss) which also styles external html (just for separation since these styles won't be used elsewhere):
> 
> https://github.com/dockstore/dockstore-ui2/blob/274e8deece1f8fdfbdc3f20b7ea5cb0f8a80aaea/src/styles.scss#L24
> 
> @import is actually going to be [phased out](https://sass-lang.com/documentation/at-rules/import) in a few years so it's recommended to use @use instead. Just tried it locally and styles rendered correctly. Eventually we should switch over our other @imports here as well

src/app/notebook/notebook-helpers.ts
svonworl: I wasn't sure how badly the sanitizers would mangle svg, so I left it off the list.  We could do some experiments and see if the common constructs pass through ok, but recommend that in another later ticket.

angular.json
```diff
@@ -50,7 +50,12 @@
               "node_modules/ace-builds/src-min-noconflict/mode-perl.js",
               "node_modules/ace-builds/src-min-noconflict/theme-idle_fingers.js",
               "node_modules/ace-builds/src-min-noconflict/ext-searchbox.js",
-              "node_modules/bootstrap/dist/js/bootstrap.js"
+              "node_modules/bootstrap/dist/js/bootstrap.js",
+              "src/mathjax-config.js",
```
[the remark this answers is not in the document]
david4096: Can we open a ticket to look at installing mathjax in a more canonical fashion?

src/app/notebook/notebook-helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+export const supportedMimeTypes: string[] = ['image/png', 'image/webp', 'image/jpeg', 'image/gif', 'text/html', 'text/json', 'text/plain'];
```
david4096, line 5: Makes me think if on the webservice side we want to tell people "you are uploading a notebook preview that includes mimetypes that won't render properly. Please refer to a doc or open a ticket if you require this functionality..."

src/app/notebook/notebook-helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+  if (Array.isArray(value)) {
+    return value.join('');
+  }
+  return String(value);
+}
+
+// The below escape() implementation is adapted from mustache.js
+// https://github.com/janl/mustache.js/blob/972fd2b27a036888acfcb60d6119317744fac7ee/mustache.js#L60
+export const escapeCharToEntity = {
```
david4096, line 29: Not my favorite to see we are now manually maintaining this given angular has this feature somewhere but seems simple enough. (ngSanitize?)

src/mathjax-config.js
```diff
@@ -0,0 +1,33 @@
+window.MathJax = {
```
david4096, line 1: Wondering if as an alternative this could be added to the router or ngOnInit?

david4096 (approved):
This is really great! It makes this feature much more exciting from a UX perspective. Steve noted two things to me aside:

1) that since we are showing user generated content there may be possible XSS vulnerabilities. It seems like there are test and mitigations in place for the obvious case. Although it's not blocking, it's not my preference that we maintain any sanitization code if possible.

2) user provided images will be displayed. Arbitrary image files will be flagged by the CSP rules (but not blocked). This is good to keep an eye on, but in general I believe we want this feature in place. It's not ideal in that page loading may be affected by external servers but the alternative to cache them ourselves would be a costly effort

None of my comments are blocking but there may be reason to keep followup tickets to keep an eye on them. I don't think adjusting how mathjax is loaded will have much security implication, just more of a developer experience issue.

src/styles.scss
svonworl: Thanks for the gentle reminder, Kimberley, it had dropped off my radar.  I went with the `@import` syntax just because we use it elsewhere and know it to be problem free, we can convert them all to `@use` in a batch, someday...

src/app/notebook/notebook-helpers.ts
svonworl: Yeah, I wasn't able to find an alternative that gave me more confidence than the known good implementation that I horked (the included mustache.js adaptation).  You'd think it'd be accessible via Angular, but I couldn't find it, and maybe, in general, they want to discourage this kind of thing.

src/app/notebook/notebook-markdown.component.ts
```diff
@@ -0,0 +1,164 @@
...
+    adjusted = this.preprocessLatexMath(adjusted);
+    // Convert the adjusted markdown to HTML using our wrapper.
+    return this.markdownWrapperService.customCompile(adjusted, this.baseUrl);
+  }
+
+  formatMath(html: string): string {
+    // Format embedded math using MathJax: https://docs.mathjax.org/en/latest/web/typeset.html
+    // Find the global MathJax object.
+    const mathjax = (<any>this.document?.defaultView)?.MathJax;
+    if (mathjax == undefined) {
```
kathy-t, line 57: `==` -> `===`

src/app/notebook/notebook-markdown.component.ts
```diff
@@ -0,0 +1,164 @@
...
+      console.log('mathjax not found');
```
kathy-t, line 58: Not sure, do we want to be logging this to the console?

src/app/notebook/notebook-mime-bundle-output.component.ts
```diff
@@ -0,0 +1,52 @@
...
+      return this.markdownWrapperService.customSanitize(data);
+    }
+    if (mimeType?.startsWith('text/')) {
+      return `<pre>${escape(data)}</pre>`;
+    }
+    return undefined;
+  }
+
+  createAttribute(name: string, value: string | number): string {
+    if (value != undefined) {
```
kathy-t, line 46: `!=` -> `!==`

src/app/notebook/notebook-source.component.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  ngOnChanges(): void {
+    this.html = this.highlight(join(this.cell?.source));
+  }
+
+  highlight(code: string): string {
+    const language = (this.language ?? 'python').toLowerCase();
+    const Prism = (<any>this.document?.defaultView)?.Prism;
+    const module = Prism?.languages[language];
+    if (module == undefined) {
```
kathy-t, line 25: `==` -> `===`

src/notebooks.scss
```diff
@@ -0,0 +1,162 @@
...
+  grid-auto-flow: row;
+  padding: 1em 1.5em 1em 1em;
+  border: solid 1px #cfd1d7;
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.1);
+  border-radius: 4px;
+  line-height: 1.5;
+  font-size: 12px;
+
+  h1 {
+    font-size: 24px;
```
kathy-t, line 17: The front end style guide says we should use rem instead of pixels

src/styles.scss
KimberleyChong: Np, just made a ticket to note this

src/app/notebook/notebook-markdown.component.ts
svonworl: `mathjax == undefined`
is a shorthand for
`mathjax === undefined || mathjax === null`
Is it ok in this context, it's certain more readable.

src/notebooks.scss
svonworl: True, although I'm guessing that the intent was related to sizing dialogs and divs, rather than fonts.  In this case, our notebook styles adjust the global styles for some elements, the font sizes of which are also defined in `px`, so it's more likely that our notebook will look ok and exist harmoniously with its surroundings if we define its font sizes in `px`, too.  (Fun fact: about 20% of the `font-size:` in our styles are set with `rem` units.)

svonworl:
I decided to tighten up the FORBID lists in `MarkdownWrapperService`, as another line of defense.  Please take a look and confirm that we never would want any of that stuff to appear on our pages.

src/app/notebook/notebook-markdown.component.ts
svonworl: I changed all instances of `==` and `!=`, because at this point, code clarity is less important than getting this PR merged.  However, we should consider modifying the style guide to allow the shorthand.

svonworl:
Ok, I pushed one final change: we now load MathJax and Prism is a much more conventional fashion, and webpack should now be able to schedule their chunks later.

That is the last of my proposed changes due to feedback, feel free to raise further objections if you think something else really needs to be addressed.

KimberleyChong (approved):
Only nit would just be to address the code smells, but not a necessity ofc :)  (remove the empty constructors and resolve the duplicate css selectors)

kathy-t (approved):

src/app/notebook/notebook-markdown.component.ts
coverbeck: > `mathjax == undefined` is a shorthand for `mathjax === undefined || mathjax === null` Similarly, `!=` Are they ok in this context? They're certainly cleaner and more difficult to break.

Interesting, I'd never known you could do that, although it makes sense now that you've described it. I usually do `if (!mathjax)` for this situation.

src/app/notebook/notebook-mime-bundle-output.component.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  constructor(private markdownWrapperService: MarkdownWrapperService) {}
+
+  ngOnChanges(): void {
+    const mimeBundle = this.output?.data ?? {};
+    const metadataBundle = this.output?.metadata ?? {};
+    this.html = this.createHtmlFromBundles(mimeBundle, metadataBundle);
+  }
+
+  createHtmlFromBundles(mimeBundle: MimeBundle, metadataBundle: OutputMetadata): string {
```
coverbeck, line 22: Should have `string | undefined` as the return type (returns undefined on line 42).

src/app/notebook/notebook-helpers.ts
coverbeck: You're not invoking this from outside the file, so you don't need to export it.

src/app/notebook/notebook-helpers.ts
```diff
@@ -0,0 +1,48 @@
+import { MimeBundle } from './notebook-types';
+
+// A list of the mime types we will display from a mime bundle, ordered from most to least "preferable".
+// To determine which item to display from a mime bundle, we'll select the item with the mime type that appears first in this list.
+export const supportedMimeTypes: string[] = ['image/png', 'image/webp', 'image/jpeg', 'image/gif', 'text/html', 'text/json', 'text/plain'];
+
+export function selectBestFromMimeBundle(mimeBundle: MimeBundle): { mimeType: string; data: string } {
```
coverbeck, line 7: Not blocking for this, but generally, and there are exceptions, but the pattern is generally to create Angular services with corresponding spec files (usually we have the spec files :) ).

coverbeck (approved):
Pleasantly surprised to see the `scripts` issue sorted out! I wasn't expecting that.

denis-yuen (approved):

PR #1758, opened 2023-04-19
SEAB-5318: Created execution tab in UI for metrics

hyunnaye:
**Description**
This PR adds the Execution tab in workflows to show its metrics. This tab can be accessed by the feature flag `metrics`.

When workflow has no metrics data available
Screenshot from 2023-04-26 17-42-21

Available partners can be toggled through the dropdown and if workflow does not have any validations data the bottom table gets hidden
Screenshot from 2023-04-26 17-42-42

If most recent run is valid 
Screenshot from 2023-04-27 09-53-33

If most recent run is not valid 
Screenshot from 2023-04-26 17-43-58

**Review Instructions**

1.  Go to [this workflow](https://qa.dockstore.org/workflows/github.com/dockstore-testing/tooltester-wes-testing/nontrivial:main?tab=info) and add the feature flag `metrics`.
2. Verify that the executions tab is visible with its corresponding metrics data. Also verify that the partners dropdown toggles through the various available metrics data correctly.
3.  Repeat with this [workflow](https://qa.dockstore.org/workflows/github.com/gatk-workflows/seq-format-conversion/BAM-to-Unmapped-BAM:3.0.0?tab=info) but check that the validations table below is hidden (since there are no available validations data)
4. Workflows with no metrics data available should have the banner stating so. [example](https://qa.dockstore.org/workflows/github.com/iwc-workflows/sars-cov-2-ont-artic-variant-calling/COVID-19-ARTIC-ONT:main?tab=info)

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5352

**Security**


src/app/shared/dockstore.model.ts
```diff
@@ -103,5 +103,6 @@ export class Dockstore {
     enableMultiCloudLaunchWithDNAstack: false,
     enableOrcidExport: true,
     enableNotebooks: false,
+    enableMetrics: false,
```
coverbeck, line 106: I thought we were enabling metrics by default? Not in the UI?

src/app/workflow/executions/executions-tab.component.ts
```diff
@@ -0,0 +1,184 @@
...
+        if (metrics) {
+          for (const [partner, metric] of Object.entries(metrics)) {
+            this.setMetricsObject(this.metrics, partner, metric);
+          }
+          this.partners = Array.from(this.metrics.keys());
+          if (this.partners.length > 0) {
+            this.ifMetricsExist = true;
+            this.selectPartner(this.partners[0]);
+          }
+        }
```
coverbeck, line 88: You should have some sort of handling for the error case.

src/app/workflow/executions/executions-tab.component.ts
```diff
@@ -0,0 +1,184 @@
...
+  currentPartner: PartnerEnum;
+  partners: PartnerEnum[];
+  metricsTable: any[];
+  validationsTable: any[];
+  metricsColumns: string[];
+  validationsColumns: string[];
+  totalExecutions: number;
+  successfulExecutions: number;
+  failedExecutions: number;
+  ifMetricsExist: boolean;
```
coverbeck, line 53: Suggest something like `metricsExist`

src/app/workflow/executions/executions-tab.component.ts
```diff
@@ -0,0 +1,184 @@
...
+  selector: 'app-executions-tab',
+  templateUrl: './executions-tab.component.html',
+  styleUrls: ['./executions-tab.component.css'],
+})
+export class ExecutionsTabComponent extends EntryTab implements OnChanges {
+  metrics: Map<PartnerEnum, Metrics>;
+  trsID: string;
+  currentPartner: PartnerEnum;
+  partners: PartnerEnum[];
+  metricsTable: any[];
```
coverbeck, line 46: Can you try and provide types instead of using `any` here and the next line? It's in the style guide to avoid `any`.

src/app/workflow/executions/executions-tab.component.html
```diff
@@ -0,0 +1,111 @@
...
+  </mat-card>
+</div>
+
+<div *ngIf="ifMetricsExist">
+  <div class="metrics-container">
+    <div class="w-25">
+      <div class="partner-dropdown">
+        <span class="metrics-header">Partner: </span>
+        <mat-form-field data-cy="metrics-partner-dropdown">
+          <mat-select [value]="currentPartner" (selectionChange)="matSelectChange($event)">
```
coverbeck, line 13: It would be nice to have a pipe for this; in your screenshot, which is great, it shows `DNA_STACK`, which is our enum, not how the display their name (`DNAstack`). Looks like there is some code (not a pipe, alas) that already does [some of it](https://github.com/dockstore/dockstore-ui2/blob/1cd20066f733c1997ba77675bb59a0bbc2f9ef55/src/app/workflow/launch-third-party/multi-cloud-launch/multi-cloud-launch.component.ts#L81). 

If this is going to be in preview mode (see my other comment/question), can wait for 1.15 with a followup ticket.

src/app/workflow/executions/executions-tab.component.css
```diff
@@ -0,0 +1,48 @@
...
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+.metrics-container {
+  display: flex;
```
kathy-t, line 18: Hmm, I think instead of creating a class, you can achieve similar results using `fxFlex` in the HTML

src/app/workflow/executions/executions-tab.component.css
```diff
@@ -0,0 +1,48 @@
...
+  width: 100%;
+  flex-direction: row;
+}
+.metrics-table-container {
+  justify-content: center;
+  height: 100%;
+  width: 75%;
+  display: flex;
+}
+.partner-dropdown {
```
kathy-t, line 28: Should use `fxFlex` in the HTML. I think you're trying to achieve the same thing as `fxLayout="row"`

src/app/workflow/executions/executions-tab.component.css
```diff
@@ -0,0 +1,48 @@
...
+  flex-direction: row;
+  margin: 2%;
+}
+.metrics-header {
+  font-weight: bold;
+}
+.execution-counts-container {
+  margin: 2%;
```
kathy-t, line 36: I don't think we've used `%` for margin before. Try seeing if one of the margin classes in this file suits your needs https://github.com/dockstore/dockstore-ui2/blob/367606c987bd52766a7c5c8de834098fb3a167dc/src/bootstrap4.scss#L69

(Similar comment for the other classes where you used `%` for margin)

src/app/workflow/executions/executions-tab.component.ts
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  private loadMetricsData(partner: PartnerEnum) {
+    this.metricsTable = [];
+    const metrics = this.metrics.get(partner);
+    if (metrics) {
+      this.metricsColumns = ['metric', 'min', 'avg', 'max'];
+      this.insertToMetricsTable(this.metricsTable, metrics?.cpu, 'CPU');
+      this.insertToMetricsTable(this.metricsTable, metrics?.memory, 'Memory');
+      this.insertToMetricsTable(this.metricsTable, metrics?.executionTime, 'Run Time (s)');
```
kathy-t, line 107: I assume the `(s)` in `Run Time (s)` is to indicate that it's in seconds. I think it might be more clear to display next to the numerical values like `10s`.

Also, memory has a unit as well, GB. You can retrieve the unit using `metrics?.memory?.unit` and `metrics?.executionTime?.unit`

src/app/shared/dockstore.model.ts
hyunnaye: Any thoughts @denis-yuen ? I was following https://ucsc-cgl.atlassian.net/browse/SEAB-5318 where it says to add a feature flag.

denis-yuen: > I thought we were enabling metrics by default? Not in the UI?

Hmmm, the API is always available. I was assuming that we would need more data (i.e. maybe we'd go live in a 1.14.1 or 1.14.2 when we get platform partners on board) before making it widely available in the UI

denis-yuen (approved):

src/app/workflow/executions/executions-tab.component.html
denis-yuen: But agreed, it looks unfinished with the enum value. We probably also need to do something with a display value for the validators (MINIWDL)

src/app/shared/dockstore.model.ts
coverbeck: Sounds good, I was just making sure. :)

src/app/workflow/executions/executions-tab.component.ts
hyunnaye: I updated the screenshots in the PR description to show this change. Thanks for the suggestion!

src/app/workflow/executions/executions-tab.component.html
hyunnaye: Ticket created: https://ucsc-cgl.atlassian.net/browse/SEAB-5413

src/app/workflow/executions/executions-tab.component.ts
```diff
@@ -0,0 +1,218 @@
...
+      if (metrics.executionStatusCount) {
+        this.totalExecutions =
+          metrics.executionStatusCount.numberOfSuccessfulExecutions + metrics.executionStatusCount.numberOfFailedExecutions;
+        this.successfulExecutions = metrics.executionStatusCount.numberOfSuccessfulExecutions;
+        this.failedExecutions = metrics.executionStatusCount.numberOfFailedExecutions;
+      }
+    }
+  }
+
+  private insertToMetricsTable(table: any[], metricsData: CpuMetric | MemoryMetric | ExecutionTimeMetric, metricLabel: string) {
```
coverbeck, line 154: `table` is defined as `any[]`.

src/app/workflow/executions/executions-tab.component.ts
```diff
@@ -0,0 +1,218 @@
...
+import { BioWorkflow, Notebook, Service } from '../../shared/swagger';
+import { CheckerWorkflowService } from '../../shared/state/checker-workflow.service';
+import { CheckerWorkflowQuery } from '../../shared/state/checker-workflow.query';
+import PartnerEnum = CloudInstance.PartnerEnum;
+import { MatSelectChange } from '@angular/material/select';
+import { AlertService } from '../../shared/alert/state/alert.service';
+
+interface MetricsTableObject {
+  metric: string;
+  min: number;
```
coverbeck, line 39: On line 157 you have `min: metricsData?.minimum`, which means the value of `min` can be undefined or null, so you should make this `min?: number`. Same comment for any other applicable fields in this interface and potentially ValidationsTableObject that can be null/undefined.

src/app/workflow/executions/executions-tab.component.ts
```diff
@@ -0,0 +1,218 @@
...
+      if (PartnerEnum[partnerEnum] === partner) {
+        metricsMap.set(PartnerEnum[partnerEnum], metric);
+      }
+    }
+  }
+
+  private resetMetricsData() {
+    this.metricsExist = false;
+    this.metrics = new Map();
+    this.metricsTable = null;
```
coverbeck, line 126: We're wildly inconsistent on this, but you should probably set this and validationsTable to an empty array instead of null. Or change the definition on line 65 to `metricsTable: MetricsTableObject[] | null`.

src/app/workflow/executions/executions-tab.component.ts
```diff
@@ -0,0 +1,218 @@
...
+      max: metricsData?.maximum,
+      unit: metricsData?.unit,
+    });
+  }
+
+  private loadValidationsData(partner: PartnerEnum) {
+    const validations = this.metrics.get(partner).validationStatus?.validatorToolToIsValid;
+    if (validations) {
+      for (const [validatorTool, isValid] of Object.entries(validations)) {
+        this.validationsTable = [
```
coverbeck, line 168: I don't know the data structure of the validations very well, but is this intentional? Because you're in a loop, the second validation, if there is one, is going to overwrite `this.validationsTable`, isn't it?

coverbeck: Belated thought, is the idea that you just want to display the last validation only? If so, it would be less confusing to have no loop and just process the last element.

hyunnaye: Oops, sorry. I meant to push into `this.validationsTable` instead of overwriting. Thanks for catching that!

hyunnaye: Since there may be more than one validator in the table, I changed the "Most recent error message" column to always be present. If a validator tool's most recent run is valid, that column would just be blank for that row.

hyunnaye: Screenshots in description is now updated

kathy-t:
Haven't re-reviewed yet, but I had a thought while looking at this locally. For published workflows, the Executions tab is hidden at first view because there's too many tabs.

image

Wondering if we should move the Executions tab forward, maybe after Files, so it's more noticeable. Probably not a big deal right now because it's accessed through a feature flag so the user probably will look for it, but when it's no longer a feature flag, I think users may overlook it.

denis-yuen:
> Wondering if we should move the Executions tab forward, maybe after Files, so it's more noticeable. Probably not a big deal right now because it's accessed through a feature flag so the user probably will look for it, but when it's no longer a feature flag, I think users may overlook it.

Worth a discussion, there's probably a `ux-design` worthy ticket in here to figure out what to do when there's too many tabs in general. Maybe we need to merge some or combine the data in some way

src/app/workflow/executions/executions-tab.component.ts
```diff
@@ -0,0 +1,210 @@
...
+        'numberOfRuns',
+        'passingRate',
+      ];
+      for (const [validatorTool, isValid] of Object.entries(validations)) {
+        this.validationsTable.push({
+          validatorTool: validatorTool,
+          mostRecentVersion: isValid.mostRecentVersion,
+          isValid: isValid.mostRecentIsValid,
+          successfulValidationVersions: isValid.successfulValidationVersions,
+          mostRecentErrorMessage: isValid?.mostRecentErrorMessage,
```
coverbeck, line 188: I don't think you need the question mark. Or if you do need it, if `isValid` can be be null/undefined, then you need it on all lines between 185 and 191.

coverbeck (approved):

src/app/workflow/executions/executions-tab.component.html
```diff
@@ -0,0 +1,117 @@
...
+            <span *ngIf="element.max">{{ element.max + element.unit }}</span>
+          </td>
+        </ng-container>
+
+        <tr mat-header-row *matHeaderRowDef="metricsColumns"></tr>
+        <tr mat-row *matRowDef="let row; columns: metricsColumns"></tr>
+      </table>
+    </div>
+  </div>
+  <div class="validations-table-container w-100 mt-5">
```
kathy-t, line 71: `validations-table-container` doesn't exist

src/app/workflow/executions/executions-tab.component.html
```diff
@@ -0,0 +1,117 @@
+<div *ngIf="!metricsExist" class="w-100">
+  <mat-card class="alert alert-warning mat-elevation-z" role="alert" data-cy="no-metrics-banner">
+    <mat-icon class="alert-warning-icon">info</mat-icon> No metrics data exist for this version.
+  </mat-card>
+</div>
+
+<div *ngIf="metricsExist" fxFlex>
+  <div class="metrics-container w-100" fxLayout="row">
```
kathy-t, line 8: `metrics-container` doesn't exist anymore

src/app/workflow/executions/executions-tab.component.html
```diff
@@ -0,0 +1,117 @@
...
+    <div class="w-25">
+      <div class="partner-dropdown">
```
kathy-t, line 10: `partner-dropdown` doesn't exist anymore

src/app/workflow/executions/executions-tab.component.html
```diff
@@ -0,0 +1,117 @@
...
+            <mat-select-trigger>
+              {{ currentPartner }}
+            </mat-select-trigger>
+            <mat-option data-cy="metrics-partner-options" [value]="partner" *ngFor="let partner of partners">
+              {{ partner }}
+            </mat-option>
+          </mat-select>
+        </mat-form-field>
+      </div>
+      <div class="execution-counts-container">
```
kathy-t, line 23: Class doesn't exist

src/app/workflow/executions/executions-tab.component.html
```diff
@@ -0,0 +1,117 @@
...
+          <span>{{ successfulExecutions }}</span>
+        </div>
+        <div>
+          <span class="weight-bold">Failed executions: </span>
+          <span>{{ failedExecutions }}</span>
+        </div>
+      </div>
+    </div>
+    <div class="metrics-table-container h-100 w-75" fxFlex>
+      <table mat-table class="metrics-table w-100" aria-label="validations-table" [dataSource]="metricsTable" data-cy="metrics-table">
```
kathy-t, line 39: `metrics-table` class doesn't exist. `aria-label` should be metrics-table

kathy-t (approved):

PR #1761, opened 2023-04-25
Minor Entry Type Changes - SEAB-5365

KimberleyChong:
**Description**
Fixes some minor entry type inconsistencies, some workflow styles were spilling into AppTool and Notebooks, and tools were missing some styles that workflows had. Now from the UI you really can't tell the difference between AppTools and a regular tool.

**Ink Bars:**

Ink bar colours for AppTools notebooks etc. would use the workflow green colour, altered for AppTools, Notebooks and Services to use their corresponding tab group colours (i.e. blue, purple, and orange):
| Before | After |
|:---:|:---:|
| image | image |

**Mat-Dividers:**
- `Mat-cards` in the info tab for Tools were missing `mat-dividers`:
- Before:
image
- After:
image

**Authors:**
- Workflows and AppTools have an author table like this, but tools do not  have this table when it should:
image

- Tools Before:
image

- Tools After:
image

- The `mat-card-title` was also changed from "Tag Information" to "Tool Version Information" to match AppTools, Services, Notebooks and Workflows.
- **NOTE**: For Workflow type entries, the ORCID ID for an author is retrieved using the [workflows/{workflowId}/workflowVersions/{workflowVersionId}/orcidAuthors](https://qa.dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowVersionOrcidAuthors) endpoint. There doesn't seem to be an equivilant endpoint for tools, so the ORCID ID column in the table displays empty always. This isn't a loss of functionality because we didn't display ORCID IDs previously, but if we didn't want to display an empty column we could remove `orcid_id` from the `displayedColumns` array until we implement this endpoint or just leave as so, let me know your opinions!
- Example with `orcid_id` column hidden:
image

**Review Instructions**
Ink bar colours should be the correct colour when viewing an entry, ie. green - workflows, blue - tools, purple - notebooks, orange - services.  Tools should have the new author table and mat-dividers under the titles when viewing the info tab.

**Issue**
SEAB-5365

**Security**


denis-yuen:
> * **NOTE**: For Workflow type entries, the ORCID ID for an author is retrieved using the [workflows/{workflowId}/workflowVersions/{workflowVersionId}/orcidAuthors](https://qa.dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowVersionOrcidAuthors) endpoint. There doesn't seem to be an equivilant endpoint for tools, so the ORCID ID column in the table displays empty always. This isn't a loss of functionality because we didn't display ORCID IDs previously, but if we didn't want to display an empty column we could remove `orcid_id` from the `displayedColumns` array until we implement this endpoint or just leave as so, let me know your opinions!

Thanks for catching this oversight, we should definitely implement a new endpoint. Do you happen  to know if this affects apptools as well?

kathy-t:
> > * **NOTE**: For Workflow type entries, the ORCID ID for an author is retrieved using the [workflows/{workflowId}/workflowVersions/{workflowVersionId}/orcidAuthors](https://qa.dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowVersionOrcidAuthors) endpoint. There doesn't seem to be an equivilant endpoint for tools, so the ORCID ID column in the table displays empty always. This isn't a loss of functionality because we didn't display ORCID IDs previously, but if we didn't want to display an empty column we could remove `orcid_id` from the `displayedColumns` array until we implement this endpoint or just leave as so, let me know your opinions!
> 
> Thanks for catching this oversight, we should definitely implement a new endpoint. Do you happen to know if this affects apptools as well?

This doesn't affect apptools because it uses the workflow component. Example app tool with ORCID author:
image

I think Kimberley's referring to the legacy, non-GitHub app tools, in which case, legacy tools can't have ORCID authors because the ORCID authors are provided through the .dockstore.yml which is why legacy tools don't have an equivalent `orcidAuthors` endpoint.

For the legacy tools, I would recommend removing the `ORCID iD` column unless we plan on adding ORCID author functionality to legacy, non-GitHub app tools.

KimberleyChong:
> I think Kimberley's referring to the legacy, non-GitHub app tools, in which case, legacy tools can't have ORCID authors because the ORCID authors are provided through the .dockstore.yml which is why legacy tools don't have an equivalent `orcidAuthors` endpoint.

Yep, exactly right, I probably should have been more specific when I said "workflow type" -> meaning entries that use workflow components (AppTools, Notebooks, Services)

> For the legacy tools, I would recommend removing the `ORCID iD` column unless we plan on adding ORCID author functionality to legacy, non-GitHub app tools.

I agree in removing the column, let me know if any oppositions!

src/app/workflows/workflows.component.ts
```diff
@@ -49,4 +52,8 @@ export class WorkflowsComponent {
     this.searchPage = this.searchPageUrls.includes(this.urlResolverService.getEntryPathFromUrl());
     this.entryName = this.urlResolverService.getEntryPathFromUrl();
   }
+
+  getEntryTerm(type: newEntryType): string {
```
coverbeck, line 56: WebStorm is giving me compiler errors in workflow.component.html because you're passing a (Swagger) EntryType to this method. I suggest changing the parameter type to (Swagger) EntryType. Then the next line would make more sense to me -- I didn't get why you are casting a `newEntryType` to a `newEntryType`.

src/app/workflows/workflows.component.ts
```diff
@@ -21,6 +21,8 @@ import { SessionService } from 'app/shared/session/session.service';
 import { Observable } from 'rxjs';
 import { EntryType } from '../shared/enum/entry-type';
 import { UrlResolverService } from '../shared/url-resolver.service';
+import { EntryTypeMetadataService } from '../entry/type-metadata/entry-type-metadata.service';
+import { EntryType as newEntryType } from 'app/shared/openapi';
```
coverbeck, line 25: Suggest something like `...as OpenApiEntryType` -- camel case had me confused when I was looking at the code on line 56, since our types are generally title case, plus `new` doesn't convey a lot of info.

src/app/workflow/workflow.component.html
```diff
@@ -46,17 +46,8 @@ <h2 class="text-center">The workflow linked to could not be found!</h2>
...
-        </span>
-        <span *ngIf="entryType === EntryType.BioWorkflow" data-cy="workflow-icon">
-          <img src="../../assets/svg/workflow-circle.svg" alt="workflow icon" />
-        </span>
-        <span *ngIf="entryType === EntryType.Service" data-cy="service-icon">
-          <img src="../../assets/svg/service-circle.svg" alt="service icon" />
-        </span>
-        <span *ngIf="entryType === EntryType.Notebook" data-cy="notebook-icon">
-          <img src="../../assets/svg/notebook-circle.svg" alt="notebook icon" />
+        <span [attr.data-cy]="entryType + '-icon'">
```
coverbeck, line 49: This is much better!

coverbeck (approved):

src/app/workflows/workflows.component.ts
KimberleyChong: Makes sense, I think I got mixed up in the names 😫 I'll also update following you other comment

svonworl (approved):

kathy-t (approved):

denis-yuen (approved):

PR #1762, opened 2023-04-26
SEAB-5118: Adapt search for notebooks

svonworl:
**Description**
This PR adds notebooks support to the Search page.

A notebooks results table and corresponding facets now appear when the (previously placeholder) "Notebooks" tab is selected.  The notebooks results table is similar to the workflow results table, except that there's a new "Language" column, and the "Verified" column has been removed.

This PR supports the display of different facets for each entry type.  For notebooks, we change the label to "Format" on the "Language" facet, add a "Language" facet that displays the ES `descriptorTypeSubclass` field, and remove the "Checker Workflow" facet.

The existing search code saves the "expanded facet" state to local storage and attempts to restore it on occasion.  I was a little concerned that maybe, now that a facet can appear/disappear between entry types, it could cause a problem, but the stored state is keyed on the corresponding ES field, which we generally display consistently between entry types, so it's fine.  I enhanced the code to use the "canned" expanded state as a fallback for the restored values, in case they don't match up one-to-one.  Restored values that don't correspond to a currently-displayed facet are simply ignored. 

Some code that was already duplicated (between tools and workflows) has been duplicated again.  At some future point, we could de-duplicate, but it's probably a low priority, because, relatively speaking, we don't touch the search code very often.

**Review Instructions**
Register a few notebooks on qa, then navigate to the search page, confirm they show up, and that the facets and other search features work as intended.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5118

**Screenshot**
Screen Shot 2023-05-01 at 1 47 39 PM

**Security**


src/app/search/state/search.service.ts
```diff
@@ -83,80 +83,87 @@ export class SearchService {
...
+        esName: 'descriptorType',
+        initiallyExpanded: true,
+      },
+      {
+        friendlyName: 'Language Versions',
+        esName: 'descriptor_type_versions.keyword',
+        tooltip: 'Indicates that the tool or workflow contains at least one version that is written with the workflow language version',
+        initiallyExpanded: false,
+      },
+      ...(tabIndex === SearchService.NOTEBOOKS_TAB_INDEX
```
coverbeck, line 101: Is there a reason to use the spread operator sometimes but not other times?

coverbeck (commented):
Overall, the approach seems reasonable to me.

I do wonder about interaction with the save/restore state, but you already raised it in the PR description.

denis-yuen (commented):
(no objections in principle)

kathy-t (commented):
Approach makes sense to me, no objections

src/app/search/search-notebook-table/search-notebook-table.component.html
```diff
@@ -0,0 +1,80 @@
...
+    matSortActive="starredUsers"
+    matSortDirection="desc"
+    matSortDisableClear
+    data-cy="search-notebook-results-table"
+  >
+    <ng-container matColumnDef="name">
+      <mat-header-cell *matHeaderCellDef mat-sort-header>Name</mat-header-cell>
+      <mat-cell data-cy="notebookColumn" *matCellDef="let notebook">
+        <div>
+          <a [matTooltip]="notebook?.full_workflow_path" [routerLink]="'/notebooks/' + notebook?.full_workflow_path">{{
```
coverbeck, line 18: There's a pipe for the router link

src/app/search/search-notebook-table/search-notebook-table.component.ts
```diff
@@ -0,0 +1,45 @@
...
+})
+export class SearchNotebookTableComponent extends SearchEntryTable implements OnInit {
+  public readonly displayedColumns = ['name', 'all_authors', 'descriptorType', 'descriptorTypeSubclass', 'projectLinks', 'starredUsers'];
+  readonly entryType = 'notebook';
+  public dataSource: MatTableDataSource<Notebook>;
+  constructor(dateService: DateService, searchQuery: SearchQuery, searchService: SearchService) {
+    super(dateService, searchQuery, searchService);
+  }
+
+  privateNgOnInit(): Observable<Array<Notebook>> {
```
coverbeck, line 42: Interesting, I first thought this was a typo, missing a space after `private`, but it's not, we have a `privateNgOnInit` declared in the base class.

src/app/search/search.component.ts
```diff
@@ -161,16 +162,25 @@ export class SearchComponent implements OnInit, OnDestroy {
...
+    this.initializeMappings(0);
     this.clearFacetSearches();
   }
 
+  private tabIndex: number | null = null;
+
+  initializeMappings(newTabIndex: number) {
+    if (newTabIndex !== this.tabIndex) {
+      this.tabIndex = newTabIndex;
+      console.log('generate new mappings');
```
coverbeck, line 174: Debug message

src/app/search/state/search.service.ts
```diff
@@ -83,80 +83,87 @@ export class SearchService {
...
-   * Example: Results returned can be private or public but never both
-   * @memberof SearchService
-   */
-  public exclusiveFilters = this.orderedFacetInfos.filter((facetInfo) => facetInfo.exclusive).map((facetInfo) => facetInfo.esName);
+  private getOrderedFacetInfos(tabIndex: number): Array<FacetInfo> {
+    return [
+      { friendlyName: 'Category', esName: 'categories.name.keyword', initiallyExpanded: true },
+      {
+        friendlyName: tabIndex === SearchService.NOTEBOOKS_TAB_INDEX ? 'Format' : 'Language',
+        esName: 'descriptorType',
```
coverbeck, line 92: Hypothetical problem: I could see wanting to have language expanded for tools/workflows, but not expanded for Notebooks, since we currently only support one format for notebooks. But the way the expansion saving logic looks like it works, keyed off `esName`, it looks like they would clobber each other. No need to change, just noting -- nobody will probably even notice this.

coverbeck (approved):

src/app/search/search.component.ts
svonworl: ack, I thought I took care of all those :)

src/app/search/state/search.service.ts
svonworl: The initial "expanded" setting will be for whatever entry type is searched first, and that will persist across different entry types, until changed by the user.  Not 100% ideal, but probably works good enough.  If we ever decide it doesn't, we can add another key (the entry type) to the save/restore logic.

src/app/search/search-notebook-table/search-notebook-table.component.html
svonworl: There's merge conflict looming, between this and Kimberley's search UI mods on the release branch.  So, for now, I'd suggest we let this ride, as it may soon be adjusted/rewritten then.

src/app/search/search-notebook-table/search-notebook-table.component.scss
```diff
@@ -0,0 +1,11 @@
+.mat-column-descriptorType {
+  max-width: 90px;
+}
+
+.mat-column-projectLinks {
+  max-width: 80px;
+}
+
+td.mat-cell {
+  border-bottom-style: none;
```
kathy-t, line 10: Curious why this is needed for the notebook table, but not the workflows/tools table

src/app/search/search-notebook-table/search-notebook-table.component.scss
```diff
@@ -0,0 +1,11 @@
+.mat-column-descriptorType {
+  max-width: 90px;
+}
+
+.mat-column-projectLinks {
+  max-width: 80px;
+}
```
kathy-t, line 7: Should be in `rem`. I'm wondering if these changes should be in https://github.com/dockstore/dockstore-ui2/blob/release/1.14/src/app/shared/styles/entry-table.scss (showing the file from the release branch because it's in rem) unless there's a reason why the notebooks table needs this override and not the other tables

kathy-t (approved):

src/app/search/search-notebook-table/search-notebook-table.component.html
denis-yuen: Thanks for the heads-up, will do

denis-yuen (approved):

src/app/search/search-notebook-table/search-notebook-table.component.scss
svonworl: This is a verbatim copy of the workflows table styles.  Kimberley made some changes, possibly this related to this, to the release branch.  In this PR, I purposefully tried _not_ to do a lot of refactoring when adding the new notebook functionality, because it would cloud the core changes and make it harder to determine the source of any problems.  Recommend that after we merge the release branch back into develop, we take a look at how it's turned out, and schedule some more work if necessary...

src/app/search/search-notebook-table/search-notebook-table.component.scss
svonworl: See above.

PR #1771, opened 2023-05-11
SEAB-5438: Display old Jupyter notebooks

svonworl:
**Description**
This PR modifies the UI to do a better job of displaying Jupyter notebooks with major nbformat (version) less than 4.

The transition from format format 3 to format 4 appears to have happened approximately 8 years ago.  It's rare, but occasionally, I do see a format 3 notebook, enough so that it seemed important enough to support them.   

The important differences between a format 3 and format 4 Jupyter notebook include:

* In format 3, the a root-level `worksheets` field contains a list of worksheets, each of which contains a `cells` field.  The first contains the notebook cells.  In format 4, `cells` is a root-level field, and `worksheets` has been removed.
* In format 3, rich outputs are embedded directly as fields in the output, denoted by shorthand like `text` and `png`.    In format 4, these outputs are stored in a mime bundle, to which the `data` field is set.
* Various field names and values have changed.

After parsing a pre-format-4 notebook, this code converts the resulting object to a form that vaguely resembles the equivalent format-4 notebook object, which is then processed as before.   The conversion is not perfect or complete, but since the preview code is written to fail gracefully, it works pretty well.  This approach has the advantage of concentrating the pre-format-4 processing in one spot, instead of scattered it piecemeal all over the place. 
 
**Review Instructions**
Register a notebook with nbformat < 4, view its Preview, and confirm that it looks fine.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5438

**Security**

### DIFF
--- a/src/app/notebook/notebook.component.ts
+++ b/src/app/notebook/notebook.component.ts
@@ -120,7 +120,7 @@ export class NotebookComponent implements OnChanges {
       if (this.isArray(cell.outputs)) {
         cell.outputs.forEach(
           (output) =>
-            // Map the previous representation of rich outputs to a "mime bundle'.
+            // Map the old representation of rich outputs to a "mime bundle'.
             (output.data ??= {
               'text/plain': output['text'],
               'text/html': output['html'],
--- a/src/app/notebook/notebook.component.ts
+++ b/src/app/notebook/notebook.component.ts
@@ -63,7 +63,7 @@ export class NotebookComponent implements OnChanges {
                 // Parse the JSON content of the notebook file.
                 const json = JSON.parse(sourceFile.content);
                 // Find the cells and, if necessary, convert them to a representation that approximates nbformat 4.
-                const cells = json?.nbformat >= 4 ? json?.cells : this.cellsToNbformat4(json?.worksheets[0]?.cells);
+                const cells = json?.nbformat < 4 ? this.cellsToNbformat4(json?.worksheets[0]?.cells) : json?.cells;
                 // If the `cells` property is an array, filter spam and "pass" the cells to the template.
                 if (this.isArray(cells)) {
                   this.cells = this.filterSpam(cells);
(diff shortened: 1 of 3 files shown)


src/app/notebook/notebook.component.ts
```diff
@@ -101,4 +103,31 @@ export class NotebookComponent implements OnChanges {
...
+      // Handle some field name changes: 'input' -> 'source', 'prompt_number' -> 'execution_count'
+      cell.source ??= cell['input'];
+      cell.execution_count ??= cell['prompt_number'];
+      // Convert each output.
+      if (this.isArray(cell.outputs)) {
+        cell.outputs.forEach(
+          (output) =>
+            // Map the old representation of rich outputs to a "mime bundle'.
+            (output.data ??= {
+              'text/plain': output['text'],
```
coverbeck, line 125: Is it ok for `output['text']` (or the other ones) to be null?

If it's OK, you should add null to the MimeBundle definition:

```
export interface MimeBundle {
  [mimeType: string]: string | string[] | null;
}
```

coverbeck (approved):

src/app/notebook/notebook.component.ts
```diff
@@ -101,4 +103,31 @@ export class NotebookComponent implements OnChanges {
...
+    if (this.isArray(cells)) {
+      cells.forEach((cell) => this.cellToNbFormat4(cell));
+    }
+    return cells;
+  }
+
+  cellToNbFormat4(cell: Cell): void {
+    if (cell.cell_type === 'code') {
+      // Handle some field name changes: 'input' -> 'source', 'prompt_number' -> 'execution_count'
+      cell.source ??= cell['input'];
```
kathy-t, line 117: TIL `??=` :slightly_smiling_face:

kathy-t (approved):

denis-yuen (approved):

PR #1776, opened 2023-05-15
SEAB-5387: added notebooks count bubble in collections

hyunnaye:
**Description**

**Review Instructions**

**Issue**

**Security**

### DIFF
--- a/src/app/organizations/collection/collection.component.ts
+++ b/src/app/organizations/collection/collection.component.ts
@@ -18,6 +18,7 @@ import { CollectionsQuery } from '../state/collections.query';
 import { CollectionsService } from '../state/collections.service';
 import { OrganizationQuery } from '../state/organization.query';
 import { EntryType } from 'app/shared/enum/entry-type';
+import { CollectionEntry } from '../../shared/openapi';
 
 @Component({
   selector: 'app-collection-entry-confirm-remove',
@@ -65,6 +66,7 @@ export class CollectionComponent implements OnInit {
 
   isAdmin$: Observable<boolean>;
   isCurator$: Observable<boolean>;
+
   constructor(
     private collectionsQuery: CollectionsQuery,
     private organizationQuery: OrganizationQuery,
@@ -87,6 +89,10 @@ export class CollectionComponent implements OnInit {
     this.collectionsService.updateCollectionFromName(organizationName, collectionName);
     this.isAdmin$ = this.userQuery.isAdmin$;
     this.isCurator$ = this.userQuery.isCurator$;
+  }
+
+  checkHasOnlyNotebooks(collection: Collection): boolean {
+    return collection.entries.length === collection.notebooksLength;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/app/organizations/collection/collection.component.ts
```diff
@@ -87,6 +91,13 @@ export class CollectionComponent implements OnInit {
     this.collectionsService.updateCollectionFromName(organizationName, collectionName);
     this.isAdmin$ = this.userQuery.isAdmin$;
     this.isCurator$ = this.userQuery.isCurator$;
+    this.organizationsService.getCollectionByName(organizationName, collectionName).subscribe((openAPICollection) => {
+      this.openAPICollection = openAPICollection;
+    });
+  }
+
+  checkHasOnlyNotebooks(collection: Collection): boolean {
+    return collection.entries.length === this.openAPICollection.notebooksLength;
```
hyunnaye, line 100: This use of both swagger and openapi versions of `Collection` is needed due to this [issue](https://ucsc-cgl.atlassian.net/browse/DOCK-1890). In short, `entries` is only available in the swagger `Collection` and `notebooksLength` is only available in the openapi version of `Collection`.

src/app/organizations/collection/collection.component.ts
```diff
@@ -87,6 +91,13 @@ export class CollectionComponent implements OnInit {
     this.collectionsService.updateCollectionFromName(organizationName, collectionName);
     this.isAdmin$ = this.userQuery.isAdmin$;
     this.isCurator$ = this.userQuery.isCurator$;
+    this.organizationsService.getCollectionByName(organizationName, collectionName).subscribe((openAPICollection) => {
```
coverbeck, line 94: This request can fail, you should handle it.

coverbeck (approved):

kathy-t (approved):

denis-yuen (approved):
With comments above

src/app/organizations/collections/collections.component.html
```diff
@@ -64,6 +64,20 @@ <h4>{{ collection.value.displayName }}</h4>
                   <span *ngIf="collection.value.toolsLength !== 1; else oneTool"> Tools</span>
                   <ng-template #oneTool> Tool</ng-template>
                 </span>
+                <span class="bubble" data-cy="collections-services-count-bubble" *ngIf="collection.value.servicesLength > 0">
+                  {{ collection.value.servicesLength }}
+                  <span *ngIf="collection.value.servicesLength !== 1; else oneService"> Services</span>
+                  <ng-template #oneService> Service</ng-template>
+                </span>
```
svonworl, line 71: Please don't change, but someday, we might consider implementing a `numbered` pipe, that takes a number and some arguments: arg1 is the string to be numbered, and optional arg2 is the plural form of arg1 (default value is arg1 + "s").   Then, we can replace some of the above code with something like:
```
{{ collection.value.servicesLength | numbered:"Service" }}
```

svonworl (approved):

PR #1777, opened 2023-05-16
Minor wording changes for GH App registration

kathy-t:
**Description**
Corresponding documentation PR: https://github.com/dockstore/dockstore-documentation/pull/245

This PR addresses feedback about the GH App registration process. In particular, it changes the wording of "Install our GitHub App on your repository/organization" to "Install our GitHub App on your repository in your personal account or organization" since there was confusion in the ticket because their personal account was never mentioned.

From ticket description:
> If I go back to the Dockstore tools page (I was not sure what to do next), and read the message I see (screenshot attached), it tells me to “do one of the following”, and the two options are to (1) link a [Quay.io](http://quay.io/) account, or (2) press the add tool button to manually add a tool. Neither of those sound like automatically syncing with the GitHub App, so this is also confusing.

This PR removes the word "manually" on the entry page when there are no entries. For the My Tools page, I also removed the first Quay option because it's not the recommended way of registering tools. The user can still read more about the Quay method of registering in the Register a Tool dialog box. I added a Discover Existing Dockstore Tools button in the same mat-card to match the mat-card on the My Workflows page.

Before:
image

After:
Screenshot 2023-05-16 at 14-46-15 Dockstore My Tools

**Review Instructions**
- On staging, log into an account that has no tools and verify that the page looks like the "After" screenshot above.
- Click on Register a Tool and verify that the GitHub App option mentions the personal account.

**Issue**
[SEAB-4410](https://ucsc-cgl.atlassian.net/browse/SEAB-4410)

**Security**

### DIFF
--- a/src/app/container/register-tool/register-tool.component.ts
+++ b/src/app/container/register-tool/register-tool.component.ts
@@ -72,7 +72,7 @@ export class RegisterToolComponent implements OnInit, AfterViewChecked, OnDestro
     {
       label: 'Register using GitHub Apps (Recommended)',
       extendedLabel:
-        'Install our GitHub App on your repository/organization to automatically sync tools with GitHub. Allows you to register a tool descriptor without linking to a Docker image you own.',
+        'Install our GitHub App on your repository in your personal account or organization to automatically sync tools with GitHub. Allows you to register a tool descriptor without linking to a Docker image you own.',
       value: OptionChoice.GitHubApps,
     },
     {
--- a/src/app/mytools/my-tool/my-tool.component.html
+++ b/src/app/mytools/my-tool/my-tool.component.html
@@ -113,14 +113,10 @@
               <mat-card class="alert alert-info mat-elevation-z" role="alert">
                 <p>
                   <mat-icon>info</mat-icon>
-                  There are currently no tools registered under this account, to add your first tool, do one of the following:
+                  There are currently no tools registered under this account. To add your first tool, do one of the following:
                 </p>
                 <ul>
-                  <li>
-                    Go to 'Accounts' and link a Quay.io account that is set up to build Docker images from GitHub, Gitlab, or Bitbucket.
-                    Tools will be synchronized in the linking process.
-                  </li>
-                  <li>Press the 'Add Tool' button to manually add a tool using the registration wizard.</li>
+                  <li>Press the 'Add Tool' button to add a tool using the registration wizard.</li>
                   <button
                     mat-raised-button
                     class="accent-1-dark-btn"
@@ -129,6 +125,17 @@
                     [disabled]="isRefreshing$ | async"
                   >
                     Add Tool
+                  </button>
+                  <li>If there are already tools on Dockstore added by other users that you have access to, press the button below.</li>
+                  <button
+                    mat-raised-button
+                    class="accent-1-dark-btn"
+                    type="button"
+                    (click)="addToExistingTools()"
+                    [disabled]="(isRefreshing$ | async) || (user$ | async) === false"
+                    matTooltip="Discover tools added by others using your linked source control accounts"
+                  >
+                    Discover Existing Dockstore Tools
                   </button>
                 </ul>
                 <div *ngIf="!hasGitHubToken" class="mt-3">
--- a/src/app/myworkflows/my-workflow/my-workflow.component.html
+++ b/src/app/myworkflows/my-workflow/my-workflow.component.html
@@ -157,7 +157,7 @@
                     repositories.
                   </li>
                   <li>
-                    Press the 'Register {{ entryType$ | async | titlecase }}' button to manually add a {{ entryType$ | async }} using the
+                    Press the 'Register {{ entryType$ | async | titlecase }}' button to add a {{ entryType$ | async }} using the
                     registration wizard.
                   </li>
                   <button
--- a/src/app/workflow/register-workflow-modal/register-workflow-modal.component.ts
+++ b/src/app/workflow/register-workflow-modal/register-workflow-modal.component.ts
@@ -87,7 +87,8 @@ export class RegisterWorkflowModalComponent implements OnInit, AfterViewChecked,
   ];
   private githubAppOption = {
     label: 'Register using GitHub Apps (Recommended)',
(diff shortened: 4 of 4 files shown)


kathy-t:
Not sure why the CircleCI jobs are failing the Install Java step :thinking: Looks like CircleCI is currently under maintenance, so maybe that. Will re-run tests later

src/app/container/register-tool/register-tool.component.ts
```diff
@@ -72,7 +72,7 @@ export class RegisterToolComponent implements OnInit, AfterViewChecked, OnDestro
     {
       label: 'Register using GitHub Apps (Recommended)',
       extendedLabel:
-        'Install our GitHub App on your repository/organization to automatically sync tools with GitHub. Allows you to register a tool descriptor without linking to a Docker image you own.',
+        'Install our GitHub App on your repository in your personal account or organization to automatically sync tools with GitHub. Allows you to register a tool descriptor without linking to a Docker image you own.',
```
denis-yuen, line 75: Seems like there are two options:
1) On an personal organization 
2) On a regular organization
3) On a personal repo 
4) On a regular repo

Before I think it listed all but assumed the user knew that the personal org/repo was just like any other.
 
The new one is a little ambiguous, is it "your repository in your personal account or (personal) organization" (which skips 4 and 2) or is it "your repository in your personal account or (non-personal) organization" (which skips 1 and 3)

May be simpler as 
"Install our GitHub App on your repository/organization (whether personal or not)" or something better

kathy-t: I may be missing some background or mixing up terminology, but what's the difference between a personal organization vs a regular one and a personal repo vs a regular repo? 

I worded it like so in the PR because after selecting where to install the app, the configuration page on GitHub is more focused on configuring the repository access

denis-yuen: I don't know if these are official terminology, but I'm interpreting `in your personal account or organization ` in the PR with:
My personal org is https://github.com/denis-yuen
A personal repo is https://github.com/denis-yuen/galaxy-workflow-dockstore-example-2
A regular org is https://github.com/dockstore
A regular repo is https://github.com/dockstore/compose_setup
Otherwise, I don't think there are any real differences  aside from a few subtleties when it comes to permissions , which is my I'm guessing the original text didn't emphasize "personal" at all.

coverbeck: https://docs.github.com/en/get-started/learning-about-github/types-of-github-accounts

So in general I would suggest "personal account" and "organization", even though they say "organization account". In this case, how about dodging the issue a little bit and say "Install our GitHub App in your personal account or organization", and let the finer details of whether to install on individual repos or the whole account/org be left to the GitHub UI?

svonworl: Second Charles' idea.  In github's terminology, I believe that https://github.com/denis-yuen is a "personal account", and https://github.com/dockstore is an "organization".

svonworl (approved):

src/app/container/register-tool/register-tool.component.ts
kathy-t: > In github's terminology, I believe that https://github.com/denis-yuen is a "personal account", and https://github.com/dockstore is an "organization account"

That's how I understood it as well. GitHub documents how to install the GH App on a [personal account](https://docs.github.com/en/apps/using-github-apps/installing-a-github-app-in-your-personal-account) or an [organization](https://docs.github.com/en/apps/using-github-apps/installing-a-github-app-in-your-organization) so I'll implement Charles' idea

src/app/container/register-tool/register-tool.component.html
```diff
@@ -57,7 +57,7 @@ <h4>Register Tool</h4>
       </div>
       <div *ngIf="selectedOption?.value === OptionChoice.GitHubApps">
         <p>
-          Navigate to GitHub to install our GitHub app on your repositories/organizations. See our
+          Navigate to GitHub to install our GitHub app in your personal account or organization. See our
```
denis-yuen, line 60: Sorry to be sticky on this one, but I think this is still ambiguous. 
Consider https://theeditorsblog.net/2015/08/08/one-adjective-paired-with-multiple-nouns-a-readers-question/

> A reader recently had a question about adjectives modifying multiple (coordinate) nouns. An example of coordinate nouns in a sentence would be—
> “Remember to wear your red hat and mittens.”
> In a sentence such as this, we recognize that red modifies both hat and mittens. For emphasis, you could repeat the adjective, but you don’t have to.
> “Remember to wear your red hat and red mittens.”

In this case, it sounds like we're saying that both the account and organization are personal when we only mean the account actually. 

I think the advice works

>  If the adjective is used only for the second noun, it clearly doesn’t modify the first noun. 

So it becomes

> Navigate to GitHub to install our GitHub app in a organization or your personal account. See our ...

src/app/container/register-tool/register-tool.component.ts
```diff
@@ -72,7 +72,7 @@ export class RegisterToolComponent implements OnInit, AfterViewChecked, OnDestro
     {
       label: 'Register using GitHub Apps (Recommended)',
       extendedLabel:
-        'Install our GitHub App on your repository/organization to automatically sync tools with GitHub. Allows you to register a tool descriptor without linking to a Docker image you own.',
+        'Install our GitHub App in your personal account or organization to automatically sync tools with GitHub. Allows you to register a tool descriptor without linking to a Docker image you own.',
```
denis-yuen, line 75: Ditto and the two below

src/app/container/register-tool/register-tool.component.html
kathy-t: Makes sense, will update

.circleci/config.yml
```diff
@@ -578,7 +578,7 @@ commands:
           name: Install java
           command: |
             sudo apt update -y
-            sudo apt install openjdk-17-jdk -yq
+            sudo apt install openjdk-17-jre -yq
```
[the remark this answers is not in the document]
denis-yuen: FYI @y-ng , it would probably be more efficient to just use an image with java built-in, but failing that this could also be an alternative

denis-yuen (approved):
With passing builds

coverbeck (approved):

PR #1787, opened 2023-05-24
SEAB-5470: Use Java image to fix failing installation step

y-ng:
**Description**
Replaced base image with cimg/openjdk:17.0 for some tests to fix failed step involving Java installation.

**Review Instructions**
Check that tests pass using new image.

**Issue**
[SEAB-5470](https://ucsc-cgl.atlassian.net/browse/SEAB-5470), [SEAB-5488](https://ucsc-cgl.atlassian.net/browse/SEAB-5488)

**Security**

[SEAB-5470]: https://ucsc-cgl.atlassian.net/browse/SEAB-5470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SEAB-5488]: https://ucsc-cgl.atlassian.net/browse/SEAB-5488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


y-ng:
Note: The browser variant of the java image additionally has node, selenium, and browser dependencies installed. For example, comparing the run time for `lint_license_unit_test_coverage` (which includes steps to install chrome and container dependencies) shows the job finishes in a shorter amount of time using the browser variant.
- with base image: 6m 44s
- with java image: 6m 3s
- browser variant: 4m 39s

hyunnaye (approved):

.circleci/config.yml
```diff
@@ -102,15 +102,15 @@ jobs:
   lint_license_unit_test_coverage:
     working_directory: ~/repo
     docker:
-      - image: cimg/base:stable-20.04
+      - image: cimg/openjdk:17.0-browsers
```
denis-yuen, line 105: Confirmed that this leads to an improvement for the lint job,

denis-yuen (changes requested):
Looks like the `local_smoke_test` and four integration test jobs are still using `cimg/base:stable-20.04` and installing Chrome and Java. 

Can you try extending this to those jobs and see if there is a similar improvement?

.circleci/config.yml
y-ng: I tried using the java browser image for the integration and smoke tests previously but the cypress tests started to fail related to cy.visit(). Any idea why this would happen?
https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/10130/workflows/9c796a92-81bb-4904-892c-7a4e465344de/jobs/44816
https://cloud.cypress.io/projects/1ya4vj/runs/29321/errors

.circleci/config.yml
```diff
@@ -200,7 +204,7 @@ jobs:
                 role-arn: $AWS_ROLE_ARN
                 role-session-name: "CircleCI-${CIRCLE_WORKFLOW_ID}-${CIRCLE_JOB}"
             - get_workspace
-            - install_container_dependencies
+            - install_npm
```
coverbeck, line 207: npm is already installed in the image. You shouldn't need to do this, unless you need a specific version of npm.

svonworl (commented):
Was here, please re-request review after the other feedback is addressed.  Thanks!

y-ng:
Update: Ended up using a different java browser image that had ubuntu 20.04 installed instead of 22.04, which seemed to fix the failing tests and also resolved the different versions of nginx. Looking at the last run as an example, integration_test_1 took 6m 0s (compared to 7m 27s with the base image) and local_smoke_tests took 4m 12s (compared to 5m 11s).

I'm re-requesting review from some people since there are changes to the PR. Thanks!

coverbeck (approved):
Looks good! FYI, it's good to resolve conversations so we know our feedback has been seen.

denis-yuen (approved):
Can confirm that Java is no longer installed, oddly it still installs Chrome but since the major complaint initially about the Java install and the chrome install is very fast, this seems like a straight improvement
Thanks for powering through this!

.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ orbs:
 executors:
   integration_test_exec: # declares a reusable executor
     docker:
-      - image: cimg/base:stable-20.04
+      - image: cimg/openjdk:17.0.4-browsers
```
kathy-t, line 10: Since the java version is used in 6 places, I suggest making a parameter for it so it's easy to update the version the next time we do. An example is the `java-tag` parameter in the CircleCI config in the webservice [here](https://github.com/dockstore/dockstore/blob/afa15d20dbcc4b8479ac30d9bc8e9a219c4f37a1/.circleci/config.yml#L3-L5) and [here](https://github.com/dockstore/dockstore/blob/afa15d20dbcc4b8479ac30d9bc8e9a219c4f37a1/.circleci/config.yml#L11) is an example of how the parameter is used.

kathy-t (approved):

svonworl (approved):

PR #1788, opened 2023-05-24
SEAB-5530: Smoke tests for notebooks

svonworl:
**Description**
This PR implements some rudimentary smoke tests for notebooks.

The code that monitors workflows was adapted to monitor notebooks.  However, there's no notebooks currently on prod, so the list of notebooks to monitor is empty.  Once we add some notebooks that will persist and propagate back to staging/qa, we can add them to the list.  I did test the `testNotebook` function (by adding a notebook from my local install to my own copy of the code), which will be invoked to check each notebook, and it works.  

The tests are written to gracefully handle the difference between the spartan notebook support in 1.14 (on staging and soon to be on prod) and what's in develop:

1. To find a notebook on staging/prod, the tests use the `/notebooks` url, because "Search" does not display notebooks in 1.14.
2. "Launch with" is not tested on staging/prod, because the launch with buttons are not present in 1.14.

For the new tests to pass, there needs to be at least one published notebook.

I've included `TODO` comments that will help us to remember where to update the smoke tests for 1.15.

A new function `goToRandomEntry` is called to summon a random entry.  Previously, there was a `beforeEach` invocation that, upon casual inspection, appeared to do the same thing, but alas, repeated invocations were causing the wrong code to be called...

I ran this successfully against my local install, and against staging via the following:
```
npm run test-staging-no-auth
```

When we get the circleci problem worked out and can do a clean run, I will update the reference to the correct develop artifacts.

**Review Instructions**
Run the smoke tests against your local install and confirm they succeed.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5530

**Security**

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "config": {
     "webservice_version": "1.15.0",
     "use_circle": true,
-    "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/9915/workflows/774416af-8cf4-452d-8908-e3f30d4ca81e/jobs/33655/artifacts",
-    "circle_build_id": "33655",
+    "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/9918/workflows/b8714bc1-49a2-424d-8b14-527bc3d4a4e3/jobs/33676/artifacts",
+    "circle_build_id": "33676",
     "base_branch": "develop"
   },
   "scripts": {


cypress/e2e/smokeTests/sharedTests/basic-enduser.ts
```diff
@@ -2,17 +2,30 @@ import { ga4ghPath } from '../../../../src/app/shared/constants';
...
   testEntry('Tools');
   testEntry('Workflows');
+  testEntry('Notebooks');
 });
 function testEntry(tab: string) {
-  beforeEach('get random entry on first page', () => {
-    cy.visit('/search');
+  function goToRandomEntry() {
+    // Notebooks search is not functional in staging or prod in 1.14.
+    // TODO after 1.15 release: remove the following code path
```
coverbeck, line 14: Best to create a ticket for it now with a 1.16 milestone.

coverbeck (approved):

kathy-t (approved):

PR #1791, opened 2023-05-29
Determine base branch dynamically

kathy-t:
**Description**
This PR removes `base_branch` from the package.json config <s>and adds a script that can determine the base branch dynamically for PRs. The CircleCI jobs that require the base branch, `audit` and `accessibility_test_base` call this script to get the value. 

Unfortunately there's no built-in way to get the base branch of a PR on CircleCI so the script uses CircleCI's `CIRCLE_PULL_REQUEST` [built-in environment variable](https://circleci.com/docs/variables/#built-in-environment-variables) that contains the PR's URL to extract the PR number then it uses the GitHub API to get the base branch of the PR. Since the script depends on the existence of a PR, the jobs that need a base branch have a step that check if a PR exists before proceeding. If no PR exists, it exits gracefully.</s>

While I was in there, <s>I removed duplicated code for the accessibility test job and</s> I created reusable commands for the repeated commands that I saw.

I also created reusable jobs for the integration tests because the only difference between each definition was the integration test group name.

**EDIT:**
This PR moves the npm audit and accessibility tests from CircleCI to GitHub Actions. GitHub Actions support the `pr_request` event trigger and it provides a built-in environment variable for the base branch of a PR. The workflows will only run when a [pull_request](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) event's activity type is opened, synchronize, or reopened. Thus this will not run when the PR is merged to the target branch, which I think is okay because there's not really a "current"/"base" branch when running on the target branch.

Noting that in moving the audit test to GitHub Actions, the CircleCI `upload_to_s3` job will happen even if the audit fails. It previously required that the unit tests and audit pass.

The CircleCI `audit` job previously was not a required check in GitHub. Should the GitHub Action `audit` job be a required job? Tagging @denis-yuen because I think it has to be configured via the repo settings

Here are the audit and accessibility GitHub Action runs for this PR:
- https://github.com/dockstore/dockstore-ui2/actions/runs/5135102657?pr=1791
- https://github.com/dockstore/dockstore-ui2/actions/runs/5135102648

**Review Instructions**
Create a PR (or find an existing one) and verify that the audit and accessibility jobs are running via GitHub Actions. Can view the jobs accessibility runs [here](https://github.com/dockstore/dockstore-ui2/actions/workflows/accessibility_test.yml) and the audit runs [here](https://github.com/dockstore/dockstore-ui2/actions/workflows/npm_audit_test.yml).

All CircleCI jobs should also pass.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5531

**Security**


.circleci/config.yml
```diff
@@ -218,112 +167,64 @@ jobs:
...
+  integration_test:
+    parameters:
+      integration_test_name:
+        description: The name of the folder containing the integration tests to run
+        type: string
     executor: integration_test_exec
     working_directory: ~/repo
     steps:
       - setup_integration_test
-      - restore_cache:
```
kathy-t, line 271: This was moved into `setup_integration_test`

.circleci/config.yml
```diff
@@ -30,29 +30,6 @@ executors:
           - network.host: 127.0.0.1
           - http.port: 9200
           - discovery.type: single-node
-  accessibility_test_exec:
```
kathy-t, line 33: This executor was a duplicate of `integration_test_exec`, so deleted this and used `integration_test_exec` instead

.circleci/config.yml
```diff
@@ -69,7 +46,6 @@ jobs:
           password: $DOCKERHUB_PASSWORD
     steps:
       - setup_nightly_tests
-      - install_cypress_dependencies
```
kathy-t, line 72: `setup_nightly_tests` calls `setup_integration_test` which calls `install_cypress_dependencies`, so removed that here and below

.circleci/config.yml
```diff
@@ -218,112 +167,64 @@ jobs:
...
+    executor: integration_test_exec
     working_directory: ~/repo
-    parameters:
-      db_dump:
-        type: string
-        default: "db_dump.sql"
     steps:
+      - check_if_pull_request_exists
       - checkout
-      - install_container_dependencies
```
kathy-t, line 291: `install_container_dependencies` was moved into `prepare_accessibility_test`

.circleci/config.yml
```diff
@@ -350,8 +252,6 @@ jobs:
     working_directory: ~/repo
     steps:
       - setup_smoke_tests
-      - restore_cache:
```
kathy-t, line 353: Removed because `setup_smoke_tests` calls `setup_integration_test` which restores the cypress cache

.circleci/config.yml
```diff
@@ -680,47 +626,24 @@ commands:
...
             then
-              bash -i -c 'npm ci'
+              git fetch origin +refs/pull/${CIRCLE_PULL_REQUEST##*/}/merge:
+              git checkout -qf FETCH_HEAD
             fi
-            mkdir -p nightly-test-results/junit
-      - run:
-          name: Prebuild
-          command: bash -i -c 'npm run prebuild.prod'
-  setup_smoke_tests:
```
kathy-t, line 707: FYI, GitHub is saying this job was modified, but this job didn't change

.circleci/config.yml
```diff
@@ -680,47 +626,24 @@ commands:
...
-    steps:
-      - setup_integration_test:
-          db_dump: "smoke_test_db.sql"
-      - run:
-          name: Index ES
-          command: |
-            curl -X POST "localhost:4200/api/api/ga4gh/v2/extended/tools/index" -H "accept: text/plain" -H "Authorization: Bearer fake-admin-token"
-          requires:
-            - setup_integration_test
-  upload_nightly_artifacts:
```
kathy-t, line 717: FYI, GitHub is saying this job was modified, but this job didn't change

scripts/run-pa11y-ci.sh
```diff
@@ -39,11 +38,6 @@ while getopts 'HBCRA' OPTION; do
     B)
       RESULT_FILE="$BASE_BRANCH_RESULT_FILE_NAME"
       ;;
-    C)
```
kathy-t, line 42: Removed this command because the checkout occurs in the CircleCI config now

.circleci/config.yml
```diff
@@ -585,48 +494,12 @@ commands:
...
-        type: string
-        default: "db_dump.sql"
     steps:
-      - browser-tools/install-chrome
-      - browser-tools/install-chromedriver
-      - restore_cache:
-          key: dep-cache-{{ .Environment.CACHE_VERSION }}-{{ checksum "./package-lock.json" }}
-      - run:
-          name: Install dependencies
-          command: bash -i -c 'npm ci'
```
kathy-t, line 599: We were installing the dependencies without checking the cache first. Fixed it so dependencies are only installed if the cache doesn't have them, which saves us about 54 seconds

scripts/get-base-branch.sh
```diff
@@ -0,0 +1,20 @@
...
+set -o xtrace
+
+# This script gets the base branch of the PR and is meant to be used by CircleCI jobs
+
+# CIRCLE_PULL_REQUEST is an environment variable set by CircleCI if a pull request exists
+# https://circleci.com/docs/variables/#built-in-environment-variables
+PR_NUMBER="${CIRCLE_PULL_REQUEST##*/}"
+
+if [[ -z "${PR_NUMBER}" ]]; then
+    echo "A PR number is required to get the base branch"
```
coverbeck, line 14: I should know this, but don't we also build branches without PRs on CircleCI? What will happen when this script fails?

coverbeck (approved):
Always enjoy seeing PRs with duplication removed. :)

scripts/get-base-branch.sh
```diff
@@ -0,0 +1,20 @@
...
+    exit 1
+fi
+
+BASE_BRANCH=$(curl "https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/pulls/${PR_NUMBER}" | jq -r '.base.ref' )
```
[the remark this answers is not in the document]
denis-yuen: > I did keep an eye on the rate-limiting and so far it seems unlikely for us to run into it. The limit is 60 requests per hour, and this request uses up 1. Triggering a new workflow run (for ex: by pushing a new commit) seems to reset the rate limit back to 60.

I think we had an unauthed call in dockstore-deploy that I kept on running into. The problem may be that we're sharing and re-using CircleCI VMs with others

.github/workflows/accessibility_test.yml
```diff
@@ -0,0 +1,131 @@
...
+        branch: [ '', '${{ github.base_ref }}' ]
+
+    # Uses if/else expression evaluation workaround from https://github.com/actions/runner/issues/409#issuecomment-752775072
+    # 'matrix.branch == github.base_ref' is the condition, 'base' is the true value and 'current' is the false value
+    name: Accessibility test (${{ matrix.branch == github.base_ref && 'base' || 'current' }} branch)
+  
+    runs-on: ubuntu-20.04
+    services:
+      postgres:
+        image: postgres:13.3
```
kathy-t, line 19: I didn't provide DockerHub credentials for the postgres and elasticsearch image, but the CircleCI config does:
https://github.com/dockstore/dockstore-ui2/blob/5fc1ed6a35e8be101189fca3cd0f127689355099/.circleci/config.yml#L18-L20

I assume it was added because of rate-limiting? Do we need it here, and if so, do these credentials already exist in our GitHub secrets (I can't see them)?

kathy-t:
Re-requesting reviews because I moved the audit and accessibility tests from CircleCI to GitHub Actions. Read the updated PR description for more details

denis-yuen:
> The CircleCI `audit` job previously was not a required check in GitHub. Should the GitHub Action `audit` job be a required job? Tagging @denis-yuen because I think it has to be configured via the repo settings

Let's see over the next couple of months how it performs

.github/workflows/accessibility_test.yml
denis-yuen: I'd hope that github has some form of Docker image caching and/or a similar deal.

denis-yuen (approved):
Thanks, this ended up being a lot more work than expected, appreciate powering through it

denis-yuen:
> Noting that in moving the audit test to GitHub Actions, the CircleCI `upload_to_s3` job will happen even if the audit fails. It previously required that the unit tests and audit pass.

Think this is generally ok

svonworl (approved):

PR #1793, opened 2023-06-05
SEAB-5570: Fix "non-sticky" search facets

svonworl:
**Description**
In some cases, on the Search page, when you clicked on a facet radiobox, its state would not toggle, and the URL would not update properly.  This behavior could be reproduced by visiting the "Notebooks" tab in search, clicking on a notebook link, using the browser's "Back" button to return to the Search page, and then clicking on the "Python" radiobox in the "Language" facet.  For a split second, the UI appears to be doing the right thing, but then fails.

This PR fixes the problem, by moving the code that initializes the mappings for a given tab index, so that it runs before the facet settings are extracted from the search URL and processed. 

**Review Instructions**
Try the procedure mentioned in the above description, and confirm that the "Language" facet radiobox toggles when clicked, and that the search URL is updated properly.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5570

**Security**

### DIFF
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -265,6 +265,12 @@ export class SearchComponent implements OnInit, OnDestroy {
       this.searchService.setSearchText('');
       this.unsubmittedSearchText = '';
     }
+    if (paramMap.has('entryType')) {
+      const value = paramMap.getAll('entryType');
+      const tabIndex = SearchService.convertEntryTypeToTabIndex(value[0]);
+      this.searchService.saveCurrentTab(tabIndex);
+      this.initializeMappings(tabIndex);
+    }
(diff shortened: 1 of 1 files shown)


coverbeck (approved):

kathy-t (approved):

kathy-t:
> Not sure if related, but do we have an explanation for the failing accessibility results (since this is not a release branch and the related PR has been merged already)

Re-ran it and it passed. It appeared to be some flakiness with the base branch run because the number of errors for the base branch during the failing run was 48, but the number of base branch errors from other PRs was 49. The re-run number of errors was 49. The current branch errors was always 49 for this PR and other ones.

denis-yuen (approved):

PR #1796, opened 2023-06-09
Use OpenAPI instead of Swagger in the UI

coverbeck:
**Description**
Switch the UI to use OpenAPI instead of Swagger (almost) everywhere.

I initially removed Swagger entirely, I wasn't even generating the client code, but I had to restore it until dockstore/dockstore#4422 is resolved.

Mainly, I was able to just replace `swagger` with `openapi` in the imports, but I did run into the following issues:

* The above-mentioned issue with dockstore/dockstore#4422, where I had to revert back to Swagger.
* Parameter order changed in some invocations
* The OpenAPI generated method names were sometimes slightly different, e.g., `refresh1` instead of `refresh`.
* Ran into an odd compilation error during tests, which I fixed based on this [comment](https://github.com/cypress-io/cypress/issues/19066#issuecomment-1144152738).
* Some changes with `descriptorLanguageEnum` and `DescriptorTypeSubclassEnum`

**Review Instructions**
Hard to review, almost every page in the UI could be affected. Going to have to mainly rely on integration and smoke tests. Try visiting commonly used pages in the UI and make sure nothing is broken

**Issue**
dockstore/dockstore#5538

**Security**


svonworl (approved):

kathy-t (approved):
Wrong ticket linked?

THIRD-PARTY-LICENSES.csv
```diff
@@ -50,18 +50,18 @@
 "@babel/helper-function-name@7.19.0","MIT","https://github.com/babel/babel"
 "@babel/helper-hoist-variables@7.18.6","MIT","https://github.com/babel/babel"
 "@babel/helper-member-expression-to-functions@7.18.9","MIT","https://github.com/babel/babel"
-"@babel/helper-module-imports@7.18.6","MIT","https://github.com/babel/babel"
```
denis-yuen, line 53: Is it correct that the 226 lines/dependencies are added due to that odd compilation error?

denis-yuen (commented):
If you were able to eliminate all apis/models but collection, you may be able to cut down on generated code / prevent regressions by doing selective generation for only that api. 
https://openapi-generator.tech/docs/customization#selective-generation

hyunnaye (approved):

THIRD-PARTY-LICENSES.csv
coverbeck: Ran `npm install` again to be sure; I'm afraid so (actually went up two).

image

coverbeck:
> If you were able to eliminate all apis/models but collection, you may be able to cut down on generated code / prevent regressions by doing selective generation for only that api. https://openapi-generator.tech/docs/customization#selective-generation

For OpenAPI, we are using classes in apis/models, it's just that there's a generated index.ts that lets you import them without explicitly specifying the sub-path:

```
export * from './api/api';
export * from './model/models';
export * from './variables';
export * from './configuration';
export * from './api.module';
```

Looks like we could have done that with Swagger too; I hadn't even noticed the IDE had removed those sub-paths until you pointed it out. But I think we need to keep generating them.

My understanding is the Angular build will not include code that's not accessed in an any case.

denis-yuen:
> My understanding is the Angular build will not include code that's not accessed in an any case.

Think I'm more wondering whether a typo would reintroduce code using swagger as opposed whether the code would be included in the build.

THIRD-PARTY-LICENSES.csv
denis-yuen: Idle thought, possibly spin-off ticket. Shouldn't the third party licenses file just be a list of dependencies that end up being part of the build as opposed to dev dependencies? 
i.e. in the old days, I would have thought that if I had like a paid-for licensed Intel C++ compiler that compiled my program, I wouldn't necessarily include that in a list of third party licenses. Is this similar?

denis-yuen (commented):
More interested in the lock-down question

THIRD-PARTY-LICENSES.csv
coverbeck: > Idle thought, possibly spin-off ticket. Shouldn't the third party licenses file just be a list of dependencies that end up being part of the build as opposed to dev dependencies? i.e. in the old days, I would have thought that if I had like a paid-for licensed Intel C++ compiler that compiled my program, I wouldn't necessarily include that in a list of third party licenses. Is this similar?

Hmm, not sure, I poked around, but it's not clear to me; we might need to ask a lawyer. The [license checker](https://www.npmjs.com/package/license-checker) we use has a production only option, so we could exclude the dev dependencies if we wanted to. I'll create a follow-up ticket.

coverbeck: dockstore/dockstore#5549

coverbeck:
> > My understanding is the Angular build will not include code that's not accessed in an any case.
> 
> Think I'm more wondering whether a typo would reintroduce code using swagger as opposed whether the code would be included in the build. By not generating the unused APIs/classes in the first place, a typo couldn't accidentally reintroduce them

The plan is to not even generate the Swagger code once dockstore/dockstore#4422 is fixed, so it won't be possible to accidentally add it. I created a followup ticket, dockstore/dockstore#5550

PR #1807, opened 2023-06-29
Don't include dev dependency licenses

david4096:
**Description**

Changes the operation of license checker to only include `--production` dependencies in the license file, not development dependencies. Goes down to ~250 from where it is now 1463.

**Review Instructions**

`npm run license` should generate the CSV of only production dependencies. The CI jobs should be implicitly testing this.

**Issue**

https://github.com/dockstore/dockstore/issues/5549

**Security**


svonworl (approved):

denis-yuen:
> Changes the operation of license checker to only include `--production` dependencies in the license file, not development dependencies. Goes down to ~250 from where it is now 1463.

For the ticket, I think there is a larger question. Is this removal basically confirming that we only need a list of third party dependencies that we actually include in our distributed artiffacts?  (and that the basic package.json or maven pom files are enough to capture our developer-level dependencies)

denis-yuen (commented):
question

david4096:
> > Changes the operation of license checker to only include `--production` dependencies in the license file, not development dependencies. Goes down to ~250 from where it is now 1463.
> 
> For the ticket, I think there is a larger question. Is this removal basically confirming that we only need a list of third party dependencies that we actually include in our distributed artiffacts? (and that the basic package.json or maven pom files are enough to capture our developer-level dependencies)

Responding in the original ticket

package.json
```diff
@@ -17,7 +17,7 @@
     "build.prod": "npx ng build",
     "build": "npx ng build",
     "watch": "ng build --watch --configuration development",
-    "license": "license-checker --excludePackages dockstore-ui2@$npm_package_version --csv > THIRD-PARTY-LICENSES.csv",
+    "license": "license-checker --production --excludePackages dockstore-ui2@$npm_package_version --csv > THIRD-PARTY-LICENSES.csv",
     "circle-ci-license-test-file": "license-checker --excludePackages dockstore-ui2@$npm_package_version --csv > CIRCLE-THIRD-PARTY-LICENSES.csv",
```
coverbeck, line 21: I think you need to add `--production` here as well.

coverbeck (commented):
You need to run `npm run license` and commit the updated license file. And make the suggested change in the package.json, or the license check will fail (it checks whether the committed THIRD-PARTY-LICENSES.csv matches what's generated from the package.json).

Also, there should probably be PR(s) for our Java repos (can be a followup ticket(s)); assuming there's a way, we should be able to remove `test` scope dependencies from those third party license files.

coverbeck (approved):

denis-yuen (approved):

kathy-t (approved):

PR #1808, opened 2023-07-05
DOCK-2392: Indicate when recent activities are from unknown users

y-ng:
**Description**
This PR addresses the issue where the username is empty when displaying recent activities from unknown users. Now the message shows "Unknown user" before the event description. 

Before:
Screenshot from 2023-07-05 12-12-16

Now: 
Screenshot from 2023-07-05 12-09-37

**Review Instructions**
Check that when you look at the recent activities under starred workflows in the dashboard, the message for events from unknown users is displayed correctly. There should also not be a link to the user when it's not available.

**Issue**
[DOCK-2392](https://ucsc-cgl.atlassian.net/browse/DOCK-2392)

**Security**

[DOCK-2392]: https://ucsc-cgl.atlassian.net/browse/DOCK-2392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/home-page/recent-events/recent-events.component.html
```diff
@@ -18,43 +18,37 @@
...
         </ng-template>
+        <ng-template #withUsername>
+          <strong>
+            <a class="no-underline" [routerLink]="'/users/' + event.initiatorUser?.username">
+              {{ event.initiatorUser?.username }}
+            </a>
+          </strong>
+        </ng-template>
+        <ng-template #withoutUsername>
+          <a class="no-underline"> Unknown user </a>
```
kathy-t, line 29: I think you can remove the `<a>` element because there's no link

kathy-t: You could also add `<strong>` so "Unknown user" is bolded to match the bolded usernames

src/app/home-page/recent-events/recent-events.component.html
```diff
@@ -18,43 +18,37 @@
...
+        </ng-template>
         <div [ngSwitch]="event.type" class="my-3">
           <!-- Entries -->
           <div *ngSwitchCase="EventType.PUBLISHENTRY" class="truncate-text-2">
-            <strong
-              ><a class="no-underline" [routerLink]="'/users/' + event.initiatorUser?.username">{{
-                event.initiatorUser?.username
-              }}</a></strong
-            >
+            <ng-container *ngIf="event.initiatorUser?.username; then withUsername; else withoutUsername"></ng-container>
```
kathy-t, line 34: No need to change because it was already like this, but this `ng-container` duplication can probably be removed by moving it outside of the `ngSwitchCase`s. Looks like there's only three cases where there's no initiator user mentioned. This might require some re-styling and re-arranging, so again, no need to change

coverbeck (approved):
Pending addressing Kathy's comment about anchor.

src/app/home-page/recent-events/recent-events.component.html
```diff
@@ -9,7 +9,7 @@
       <div fxLayout>
         <div *ngIf="event?.initiatorUser?.username; else defaultAvatar">
           <a
-            *ngIf="event.type !== EventType.APPROVEORG && event.type !== EventType.REJECTORG && event.type"
+            *ngIf="event.type !== EventType.APPROVEORG && event.type !== EventType.REJECTORG && event.type !== EventType.REREQUESTORG"
```
y-ng, line 12: FYI added this line because I think it's missing from the `ngIf`

kathy-t (approved):
Nice! Thanks for removing the duplication

PR #1809, opened 2023-07-06
DOCK-2406: Remove failing request on tool page

y-ng:
**Description**
When navigating to a public legacy tool page with browser tools open, there would be a workflow request that returns a 404 (as well as a similar request that succeeds so the page still shows properly). This PR prevents the failing request from being made on the tool page.

**Review Instructions**
With browser tools open, navigate to a public tool and check that there is no 404 in the network tab.

**Issue**
[DOCK-2406](https://ucsc-cgl.atlassian.net/browse/DOCK-2406)

**Security**

[DOCK-2406]: https://ucsc-cgl.atlassian.net/browse/DOCK-2406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/workflow/workflow.component.ts
```diff
@@ -390,39 +390,41 @@ export class WorkflowComponent extends Entry<WorkflowVersion> implements AfterVi
...
-            // Retrieve the workflow path from the URL
-            const splitPath = this.resourcePath.split('/');
-            const workflowPath = splitPath.slice(2, 5);
-            const pathSuffix = workflowPath.join('/');
-
-            // Create suggested paths
-            this.gitlabPath += pathSuffix;
-            this.githubPath += pathSuffix;
-            this.bitbucketPath += pathSuffix;
+    if (!url.includes('containers') && !url.includes('tools')) {
```
y-ng, line 393: I added this line so that the request isn't made on a legacy tool page. The integration and smoke tests pass but I wonder if this would prevent the request from being made on any pages it's actually needed.

coverbeck (commented):
I think the fix may need to be in container.component.html and/or container.component.ts.

In [container.component.html](https://github.com/dockstore/dockstore-ui2/blob/541d618c19aaf7063f5740071952dc0d61e637fd/src/app/container/container.component.html#L446), the intent looks to be that the workflow.component.ts is only instantiated if it's an app tool, but it's getting invoked anyway.

I think the problem is that `tool` is initially undefined, so the template/component is invoked, which causes the the workflow component to get instantiated. Then tool gets assigned a legacy tool, the workflow component goes away, but it already made the Ajax request.

TLDR; this component shouldn't even exist for a legacy tool.

src/app/container/container.component.html
```diff
@@ -13,7 +13,7 @@
   ~    See the License for the specific language governing permissions and
   ~    limitations under the License.
   -->
-<div *ngIf="tool; else workflow">
+<div *ngIf="tool || !displayAppTool; else workflow">
```
coverbeck, line 16: The template and property names were like this already, but this code makes it seem like there are 3 entity types to handle: Tool, AppTool, and Workflow. But this component handles just two types, Tool and AppTool, even though AppTool extends Workflow.

Suggest: change the template id to `appTool`, i.e., `<div *ngIf="tool || !displayAppTool; else appTool">`

coverbeck (approved):

denis-yuen (approved):
Think I accidentally dismissed a passing review when re-requesting

coverbeck (approved):

hyunnaye (approved):

svonworl (approved):

kathy-t (approved):

PR #1811, opened 2023-07-07
Update integration tests with 1.15.0 migration and fix resulting test failures

svonworl:
**Description**
Whilst working on a UI issue, I noticed that the migrations used by the Cypress tests had not been updated for 1.15.0.  So, I made the updates, which caused some notebooks and apptools-related tests to fail.

The solution was to modify the db dumps that insert notebooks/apptools so that they include the `waseverpublic` column.  We didn't notice the problem until now because the inserts worked prior to the 1.15.0 migration.  The 1.15.0 migration adds a constraint that made the previous inserts fail because a value for `waseverpublic` was missing, so the additional notebooks/apptools were not being added to the database as expected.

This PR includes an extra bit of whitespace in `package-lock.json` to trigger the generation/use of a new cache, rather than an older existing cache with a different (unrelated to this PR) problem.  The idea here is that the next time someone makes an update that changes `package-lock.json`, the extra whitespace will go away...

**Review Instructions**
Confirm CircleCI tests are passing.

**Issue**
None

**Security**


denis-yuen:
> The 1.15.0 migration adds a constraint that made the previous inserts fail because a value for `waseverpublic` was missing, so the additional notebooks/apptools were not being added to the database as expected.

For understanding, is there a reason the defaults didn't kick in? 

https://github.com/dockstore/dockstore/pull/5545/files#diff-ec6abf7194a588b9c5424eab0ac536ecf8b139e4a97d0ba81475f2727d5c28ecR110-R158

denis-yuen (approved):
question

kathy-t (approved):

svonworl:
> > The 1.15.0 migration adds a constraint that made the previous inserts fail because a value for `waseverpublic` was missing, so the additional notebooks/apptools were not being added to the database as expected.
> 
> For understanding, is there a reason the defaults didn't kick in?
> 
> https://github.com/dockstore/dockstore/pull/5545/files#diff-ec6abf7194a588b9c5424eab0ac536ecf8b139e4a97d0ba81475f2727d5c28ecR110-R158

Turns out that the new column `waseverpublic` doesn't have a default.  The `value` property in the migration sets the value on existing rows, but doesn't establish a default value for new rows.  I think if we used the liquibase `defaultValue` property instead, that would do the trick, and we probably should modify the webservice accordingly, I can't think of a reason why we wouldn't want to (in this case).

coverbeck (approved):

PR #1816, opened 2023-07-13
Set the Angular strictTemplates compiler option

coverbeck:
**Description**
Sets the `strictTemplates` compiler option to true. This checks that objects and object properties referenced from templates exist and have the correct type.

Turning this on identified several errors, and code fixes were made. Comments are inline with explanations.

**Review Instructions**
1. Not much, but can verify that the [option is true](https://github.com/dockstore/dockstore-ui2/blob/feature/5471/strictness/tsconfig.json#L30)
2. Do a quick sanity check in the UI to make sure nothing's broken. This is extremely vague and so much was touched that not much direction can be provided.

**Issue**
dockstore/dockstore#5471

**Security**


src/app/container/add-tag/add-tag.component.html
```diff
@@ -101,7 +101,7 @@ <h4 mat-dialog-title>Add Version Tag</h4>
     </mat-form-field>
     <div class="form-group">
       <mat-label> CWL Test Parameter File(s): </mat-label>
-      <div *ngFor="let item of unsavedCWLTestParameterFilePaths; let i = index; trackBy: trackByIndex">
```
coverbeck, line 104: `trackByIndex` method did not exist

src/app/container/container.component.html
```diff
@@ -14,16 +14,6 @@
   ~    limitations under the License.
   -->
 <div *ngIf="tool; else workflow">
-  <div class="row m-1" *ngIf="error || missingWarning">
```
coverbeck, line 17: The `error` and `missingWarning` properties do not exist in the controller, so this `*ngIf` never evaluated to true. Just removed the block.

src/app/container/container.component.html
```diff
@@ -342,7 +332,7 @@ <h3>
                 [ngClass]="{ 'btn-copy': toolCopyBtn === 'docker_pull_command' }"
                 type="button"
                 [cdkCopyToClipboard]="dockerPullCmd"
-                (cbOnSuccess)="toolCopyBtnClick('docker_pull_command')"
+                (cdkCopyToClipboardCopied)="toolCopyBtnClick('docker_pull_command')"
```
coverbeck, line 335: This is the correct attribute to use.

src/app/container/info-tab/info-tab.component.html
```diff
@@ -323,7 +323,7 @@
                   <mat-radio-group
                     [(ngModel)]="tool.topicSelection"
                     aria-label="Select an option"
-                    (change)="topicSelectionChange($event)"
+                    (change)="topicSelectionChange()"
```
coverbeck, line 326: The controller method takes no parameter

src/app/container/launch/launch.component.html
```diff
@@ -30,7 +30,7 @@
       <div class="row">
         <div class="col-sm-4">
           <mat-form-field>
-            <mat-select name="toolVersions" [(value)]="currentDescriptor" (selectionChange)="changeDescriptor($event)">
```
coverbeck, line 33: 1. There is already an `ngOnChanges` that handles the change.
2. The `changeDescriptor()` method being called takes 3 parameters, not the one that was being passed to it.

src/app/container/version-modal/version-modal.component.html
```diff
@@ -355,31 +355,31 @@ <h4 mat-dialog-title>{{ TagEditorMode[mode] }} Version Tag</h4>
               </div>
             </div>
           </div>
-          <div class="form-group form-group-sm" *ngIf="!unsavedVersion.create">
```
coverbeck, line 358: There is no `create` property on `unsavedVersion`. If it doesn't exist, then the `*ngIf` always evaluates to true, so just removed the `*ngIf` entirely.

I'm a little wary of this change; I wonder if it was from the previous way Angular used to handle forms? But I tried invoking the UI manually, verified in the debugger that the property doesn't exist, and it seemed to work fine.

src/app/container/version-modal/version-modal.component.html
```diff
@@ -355,31 +355,31 @@ <h4 mat-dialog-title>{{ TagEditorMode[mode] }} Version Tag</h4>
...
+          <div class="form-group form-group-sm">
             <label class="col-sm-3 col-md-3 col-lg-3 control-label"> Last Modified: </label>
             <div class="col-sm-9 col-md-9 col-lg-9">
               <div class="label-value">
-                {{ getDateTimeMessage(unsavedVersion.last_modified) }}
+                {{ getDateTimeMessage(unsavedVersion.last_built) }}
```
coverbeck, line 370: Tags don't have last_modified, they have last_built. Thank you compiler. But, still, a little concerned.

src/app/container/view/view.component.ts
```diff
@@ -39,7 +40,7 @@ import { VersionModalService } from '../version-modal/version-modal.service';
   styleUrls: ['./view.component.css'],
 })
 // This is actually the actions dropdown for tags
-export class ViewContainerComponent extends View implements OnInit {
+export class ViewContainerComponent extends View<Tag> implements OnInit {
```
coverbeck, line 43: Used generics to address some errors

src/app/file-tree/file-tree.component.html
```diff
@@ -1,7 +1,7 @@
 <div mat-dialog-content>
   <mat-form-field appearance="outline" class="w-100">
     <mat-label>Search file...</mat-label>
-    <input type="text" matInput (input)="filter($event.target.value)" />
+    <input type="text" matInput (input)="filter($any($event.target).value)" />
```
coverbeck, line 4: I hate this, but no way around it as far as I can tell.

src/app/home-page/recent-events/recent-events.component.ts
```diff
@@ -86,7 +86,7 @@ export class RecentEventsComponent extends Base implements OnInit {
     private usersService: UsersService,
     private alertService: AlertService,
     private eventsService: EventsService,
-    private orgLogoService: OrgLogoService,
+    public orgLogoService: OrgLogoService,
```
coverbeck, line 89: Template was accessing a private property, which the compiler flagged (worked at runtime because JavaScript doesn't have private)

src/app/loginComponents/accounts/account-sidebar/account-sidebar.component.html
```diff
@@ -38,10 +38,6 @@ <h5 class="weight-bold">GitHub Profile</h5>
         <th>Name</th>
         <td>{{ gitHubProfile.name }}</td>
       </tr>
-      <tr *ngIf="gitHubProfile.email">
```
coverbeck, line 41: Email properties are no longer there.

src/app/loginComponents/accounts/account-sidebar/account-sidebar.component.ts
```diff
@@ -50,21 +52,25 @@ export class AccountSidebarComponent implements OnInit {
...
-    this.usersService.updateLoggedInUserMetadata(service).subscribe(
-      (user: User) => {
-        this.userService.updateUser(user);
-        this.alertService.simpleSuccess();
-        if (service === TokenSource.GITHUB) {
-          this.syncBadgeGit = true;
-        }
-        if (service === TokenSource.GOOGLE) {
-          this.syncBadgeGoogle = true;
+    this.syncing = true;
```
coverbeck, line 55: `syncing` wasn't being set; rest of block is reformatting

src/app/loginComponents/accounts/external/accounts.component.html
```diff
@@ -50,7 +50,6 @@
               matSuffix
               class="accent-1-dark"
               [cdkCopyToClipboard]="dockstoreToken"
-              (cbOnSuccess)="isCopied1 = true"
```
coverbeck, line 53: cbOnSucess does not exist

src/app/mytools/my-tool/my-tool.component.html
```diff
@@ -74,7 +74,7 @@
                     color="primary"
                     type="button"
                     (click)="addToExistingTools()"
-                    [disabled]="(isRefreshing$ | async) || (user$ | async) === false"
```
coverbeck, line 77: `user$` does not exist

src/app/notebook/notebook.component.html
```diff
@@ -13,7 +13,7 @@
         <ng-container *ngIf="cell.cell_type === 'code'">
           <ng-container *ngIf="!cell.metadata?.source_hidden">
             <div class="count">[{{ cell.execution_count ?? ' ' }}]:</div>
-            <app-notebook-source class="source" [cell]="cell" [language]="workflow?.descriptorTypeSubclass"></app-notebook-source>
+            <app-notebook-source class="source" [cell]="cell" [language]="notebook?.descriptorTypeSubclass"></app-notebook-source>
```
coverbeck, line 16: Wrong property name

src/app/organizations/collections/collections.component.ts
```diff
@@ -57,7 +59,7 @@ export class CollectionsComponent implements OnInit, OnChanges {
     this.collectionsService.updateCollections(this.organizationID);
   }
 
-  editCollection(collection: Collection) {
+  editCollection(collection: KeyValue<string, Collection>) {
```
coverbeck, line 62: Update to what's actually being passed from the template

src/app/organizations/collections/create-collection/create-collection.component.ts
```diff
@@ -26,7 +30,7 @@ export class CreateCollectionComponent implements OnInit, OnDestroy {
   constructor(
     private createCollectionQuery: CreateCollectionQuery,
     private createCollectionService: CreateCollectionService,
-    @Inject(MAT_DIALOG_DATA) public data: any,
+    @Inject(MAT_DIALOG_DATA) public data: { collection: KeyValue<string, Collection> | null; mode: TagEditorMode },
```
coverbeck, line 33: any is bad

src/app/organizations/organization/update-organization-description/update-organization-description.component.ts
```diff
@@ -46,7 +46,7 @@ export class UpdateOrganizationOrCollectionDescriptionComponent implements OnIni
     }
   }
 
-  get descriptionValue(): AbstractControl {
+  get descriptionValue(): string {
```
coverbeck, line 49: Returns a `string` in my tests.

src/app/search/search.component.ts
```diff
@@ -403,7 +403,7 @@ export class SearchComponent implements OnInit, OnDestroy {
       value.forEach((innerValue: boolean, innerKey: string) => {
         if (innerValue && this.orderedBuckets.get(key)) {
           if (!this.orderedBuckets.get(key).SelectedItems.get(innerKey)) {
-            this.orderedBuckets.get(key).SelectedItems.set(innerKey, '0');
+            this.orderedBuckets.get(key).SelectedItems.set(innerKey, 0);
```
coverbeck, line 406: Changed type to be number instead of string

src/app/starredentries/starredentries.component.html
```diff
@@ -144,7 +139,7 @@ <h4>
                 <span>
                   <mat-chip-list>
                     <mat-chip class="bubble workflow-background" *ngFor="let label of workflow?.labels | slice: 0:3">{{
-                      label.value
+                      $any(label).value
```
coverbeck, line 142: This code is inside an `ng-template`, and when you do that, you lose type information. An ugly `any`, alas. Only way to workaround it is to have a component instead of a template, I believe.

tsconfig.json
```diff
@@ -23,9 +23,10 @@
         "../node_modules/jszip/dist/jszip.min.js"
       ]
     }
-  }, 
+  },
   "angularCompilerOptions": {
     "skipTemplateCodegen": false,
-    "strictInjectionParameters": true
+    "strictInjectionParameters": true,
+    "strictTemplates": true
```
coverbeck, line 30: This is the change that turns it on.

src/app/starredentries/starredentries.component.html
```diff
@@ -73,12 +73,7 @@
             </mat-card>
             <div fxLayout="column" fxLayoutGap="1rem" class="mb-2 mx-1">
               <mat-card fxFlex *ngFor="let tool of starredTools; let last = last" class="m-1 pb-3">
-                <ng-container
-                  *ngTemplateOutlet="
-                    tool.mode === 'DOCKSTORE_YML' ? workflowSummary : toolSummary;
-                    context: tool.mode === 'DOCKSTORE_YML' ? { workflow: tool, entryType: entryType.AppTool } : { tool: tool }
-                  "
```
kathy-t, line 80: This seems like app tool specific code, what was the reason for removing this?

kathy-t (approved):

denis-yuen (approved):

src/app/starredentries/starredentries.component.html
coverbeck: > This seems like app tool specific code, what was the reason for removing this?

Before the change, the compiler was returning this error with strictTemplates turned on:

```
Error: src/app/starredentries/starredentries.component.html:79:30 - error TS2367: This condition will always return 'false' since the types 'ModeEnum' and '"DOCKSTORE_YML"' have no overlap.

79                     context: tool.mode === 'DOCKSTORE_YML' ? { workflow: tool, entryType: entryType.AppTool } : { tool: tool }
                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

IOW, it always evaluated to false, because DockstoreTool.ModeEnum is defined as `export type ModeEnum = 'AUTO_DETECT_QUAY_TAGS_AUTOMATED_BUILDS' | 'AUTO_DETECT_QUAY_TAGS_WITH_MIXED' | 'MANUAL_IMAGE_PATH' | 'HOSTED';`; there is no `DOCKSTORE_YML`.

So I just  narrowed the code down to the `false` condition is was always evaluating to. But my approach may have just perpetuated a bug; will investigate.

coverbeck: That was a good catch, updated the code.

The issue was that we were casting `Entry` to `DockstoreExtendedTool` in the controller, and `DockstoreExtendedTool` doesn't have a mode that can be `DOCKSTORE_YML`.

The cast was incorrect, as the endpoint can return AppTools, so I changed the type of starredEntries to `starredTools: Array<ExtendedDockstoreTool | ExtendedWorkflow> | null;`.

I then changed the HTML template to use the `entryType` property, which is more obvious than the mode.

kathy-t (approved):

svonworl (approved):

src/app/entry/current-collections/current-collections.component.ts
```diff
@@ -17,7 +17,7 @@ import { OrgLogoService } from '../../shared/org-logo.service';
 export class CurrentCollectionsComponent implements OnInit, OnChanges {
   @Input() id: number;
   // This is bad, should not have an input that's only passed down the chain
-  @Input() versions: Version[];
+  @Input() versions: WorkflowVersion[] | Tag[];
```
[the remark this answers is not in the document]
coverbeck: It's a good idea, but I'm not sure if it's (easily) possible with the code generators we are using. The `@ApiModel` is a Swagger annotation, which we don't use any more; but the `@Schema` annotation looks like it has some attributes that could apply, `subType`, `oneOf`, ... I tried briefly with `oneOf`, and it generated a different openapi.yaml, but didn't seem to change the TS generated code.

@denis-yuen any thoughts or memories about this? I vaguely recall some discussion of hierarchy with Swagger-generated classes, perhaps with Gary, but the memories are so vague, I'm not sure they're real.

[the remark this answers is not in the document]
denis-yuen: Hmmm, I'm not entirely sure how this works myself across languages. 
But taking a look at `Workflow.java` for example. `@ApiModel` is definitely swagger, `@JsonTypeInfo` and `@JsonSubTypes` are jackson. However, The latter two do survive the code generation process and make it into the code generated `Workflow.java`. However, that just lets Jackson know which child class you're de-serializing, it doesn't let you address the object as `Entry` (I think)

I do think it could be worth tinkering with the `oneOf` or similar https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/ (open api doc) 

Is it possible that what you did with `oneOf` was correct, but just that the TS generated code was not up-to-par? Maybe we can look at the Java generated code from the same openapi.yaml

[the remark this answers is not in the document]
coverbeck: dockstore/dockstore#5605

PR #1817, opened 2023-07-14
SEAB-5676: allow admins and curators to see a users GitHub app logs

hyunnaye:
**Description**
This PR allows admins and curators to see users' GitHub app logs through their userpage. This new button is only visible for admins and curators.

Viewing user page as an admin (App logs button) 
Screenshot from 2023-07-14 13-30-43

When button clicked
Screenshot from 2023-07-19 12-03-53

Viewing user page as nonadmin nor curator (App Logs button should be hidden)
Screenshot from 2023-07-14 12-57-57

**Review Instructions**
As an admin or curator, go to  https://qa.dockstore.org/users/hyunnaye and confirm that the button works properly.
Then log out and verify that the button is hidden.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5676

**Security**

### DIFF
--- a/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
+++ b/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
@@ -35,8 +35,8 @@ import { DescriptorLanguageService } from '../../../shared/entry/descriptor-lang
 export class GithubAppsLogsComponent implements OnInit {
   datePipe: DatePipe;
   mapPipe: MapFriendlyValuesPipe;
-  columnsToDisplay: string[] = ['repository', 'reference', 'success', 'type'];
-  displayedColumns: string[] = ['eventDate', 'githubUsername', ...this.columnsToDisplay];
+  columnsToDisplay: string[];
+  displayedColumns: string[];
   lambdaEvents: LambdaEvent[] | null;
   @ViewChild(MatPaginator, { static: true }) paginator: MatPaginator;
   @ViewChild(MatSort, { static: true }) sort: MatSort;
@@ -48,7 +48,7 @@ export class GithubAppsLogsComponent implements OnInit {
   isExpansionDetailRow = (i: number, row: Object) => row.hasOwnProperty('detailRow');
 
   constructor(
-    @Inject(MAT_DIALOG_DATA) public matDialogData: any,
+    @Inject(MAT_DIALOG_DATA) public matDialogData: { userId: number; organization: string },
     private lambdaEventsService: LambdaEventsService,
     private matSnackBar: MatSnackBar,
     private descriptorLanguageService: DescriptorLanguageService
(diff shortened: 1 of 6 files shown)


src/app/user-page/user-page.component.ts
```diff
@@ -17,18 +21,25 @@ import { AlertService } from '../shared/alert/state/alert.service';
 export class UserPageComponent implements OnInit {
   public user: any;
   public username: string;
+  protected lambdaEvents: string[];
```
kathy-t, line 24: This looks unused

src/app/user-page/user-page.component.html
```diff
@@ -43,6 +43,16 @@ <h4 matLine class="accent-1-dark mt-2 mb-0">{{ user?.username }}</h4>
                   <span>{{ gitHubProfile.username }}</span>
                 </a>
               </span>
+              <span *ngIf="loggedInUserIsAdminOrCurator">
+                <button
```
kathy-t, line 47: I know this is only used by our own admins/curators, but it might be helpful to add a tooltip explaining what this button does like, `View the user's GitHub App Logs`, since `Apps Logs` is pretty non-descriptive when it's on the users page

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -48,7 +48,7 @@ export class GithubAppsLogsComponent implements OnInit {
   isExpansionDetailRow = (i: number, row: Object) => row.hasOwnProperty('detailRow');
 
   constructor(
-    @Inject(MAT_DIALOG_DATA) public matDialogData: string,
+    @Inject(MAT_DIALOG_DATA) public matDialogData: any,
```
coverbeck, line 51: Try not to use `any`. I know we use this pattern for dialog data in several places, but our ultimate goal one day is to get rid of all `any`'s, so we should avoid adding new ones.

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -67,22 +67,41 @@ export class GithubAppsLogsComponent implements OnInit {
...
-      )
-      .subscribe(
-        (lambdaEvents) => (this.lambdaEvents = lambdaEvents),
-        (error) => {
-          this.lambdaEvents = null;
-          const detailedErrorMessage = AlertService.getDetailedErrorMessage(error);
-          this.matSnackBar.open(detailedErrorMessage);
-        }
-      );
+    if (this.matDialogData.getUserEvents) {
```
coverbeck, line 70: Nit: we generally don't use `get` to start the name of boolean properties

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -67,22 +67,41 @@ export class GithubAppsLogsComponent implements OnInit {
...
+      this.lambdaEventsService
+        .getUserLambdaEvents(this.matDialogData.value)
+        .pipe(
```
coverbeck, line 73: The blocks 73-86 and 90-103 are identical, I think. You should be able to remove the duplication, e.g., 

```
const lambdaEvents = this.matDialogData.getUserEvents ? this.lambdaEventsService
        .getUserLambdaEvents(this.matDialogData.value) :  this.lambdaEventsService
        .getLambdaEventsByOrganization(this.matDialogData.value); // Or have this logic in a method
lambdaEvents.pipe(
...
```

src/app/user-page/user-page.component.ts
```diff
@@ -17,18 +21,25 @@ import { AlertService } from '../shared/alert/state/alert.service';
...
   public TokenSource = TokenSource;
   public googleProfile: Profile;
   public gitHubProfile: Profile;
   protected ngUnsubscribe: Subject<{}> = new Subject();
```
coverbeck, line 28: It was like this, and  so it was an existing bug, but this `ngUnsubscribe` never has its `done()` method invoked. So your `takeUntil` on line 40 will never trigger.

The easiest fix is to delete this line, and have this class extend `Base`.

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
svonworl, line 51: Should probably replace the `any` with a type, which could be anonymous, maybe something like:
```
@Inject(MAT_DIALOG_DATA) public matDialogData: { value: string | User, getUserEvents: boolean }
```
https://www.typescriptlang.org/docs/handbook/2/objects.html
I could imagine a different structure, where the injected value has two properties `organization` and `user`, and `ngOnInit` takes action based on which is defined.  Might be easier to understand, might not, your call.

svonworl, line 51: Should probably replace the `any` with a type, which could be anonymous, maybe something like:
```
@Inject(MAT_DIALOG_DATA) public matDialogData: { value: string | User, getUserEvents: boolean }
```
https://www.typescriptlang.org/docs/handbook/2/objects.html
I could imagine a different structure, where the injected value has two properties `organization` and `user`, and `ngOnInit` takes action based on which is defined.  Might be easier to understand, might not, your call.

coverbeck (commented):
Screenshot looks good! I was wondering how you were going to show it.

There is a small issue with the UI, I think. When a user views their logs, they know what GitHub organization they are viewing, because they clicked the button next to the org.

In this case, the admin is going to see events for all of the user's GitHub orgs, and won't be able to tell them apart, because the UI doesn't show the GitHub org.

On the other hand, this is more for emergency/debugging situations, and the admin/curator will probably know what repositories they're looking for, so it may not be worth the effort to address it. And the typical user may not belong to many orgs with repos firing events at the same time.

denis-yuen:
> There is a small issue with the UI, I think. When a user views their logs, they know what GitHub organization they are viewing, because they clicked the button next to the org.
> 
> In this case, the admin is going to see events for all of the user's GitHub orgs, and won't be able to tell them apart, because the UI doesn't show the GitHub org.

Agreed on this, is it possible to display the organization or is it not available?

hyunnaye:
> > There is a small issue with the UI, I think. When a user views their logs, they know what GitHub organization they are viewing, because they clicked the button next to the org.
> > In this case, the admin is going to see events for all of the user's GitHub orgs, and won't be able to tell them apart, because the UI doesn't show the GitHub org.
> 
> Agreed on this, is it possible to display the organization or is it not available? Otherwise, the interface looks good

I think I can try displaying the organization!

hyunnaye:
> Screenshot looks good! I was wondering how you were going to show it.
> 
> There is a small issue with the UI, I think. When a user views their logs, they know what GitHub organization they are viewing, because they clicked the button next to the org.
> 
> In this case, the admin is going to see events for all of the user's GitHub orgs, and won't be able to tell them apart, because the UI doesn't show the GitHub org.
> 
> On the other hand, this is more for emergency/debugging situations, and the admin/curator will probably know what repositories they're looking for, so it may not be worth the effort to address it. And the typical user may not belong to many orgs with repos firing events at the same time.

Organization column added. Thank you!

denis-yuen:
Probably should have mentioned this previously, but it seems the "GitHub Username" column could be omitted since it would presumably always be the same in this view (but not committed in the normal view)

svonworl (approved):

denis-yuen:
> I don't think it will necessarily be the same; I believe you'll see all events for the GitHub org, not just the events the user was directly involved in.

Good point, nvm that then

denis-yuen (approved):

coverbeck (approved):

kathy-t (approved):

PR #1819, opened 2023-07-18
SEAB-5655: allow admins and curators to see user's other linked accounts

hyunnaye:
**Description**
This PR adds a new tab "Other linked accounts" on a user's profile page that only admins and curators can see. This allows community leads to view what other types of accounts a user has linked.

Screenshot from 2023-07-21 14-47-09

When there are no other linked accounts
Screenshot from 2023-07-21 13-56-01

Tab is not visible when logged out or not an admin/curator
Screenshot from 2023-07-21 15-13-40

In requests page, the username is linked to their page
Screenshot from 2023-07-21 13-48-35

**Review Instructions**
1. Log into an admin or curator account. 
2. Go to https://qa.dockstore.org/users/denis-yuen and confirm that a zenodo account is linked.
3. Log out and verify that the tab is not visible.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5655

**Security**

### DIFF
--- a/cypress/e2e/immutableDatabaseTests/admin.ts
+++ b/cypress/e2e/immutableDatabaseTests/admin.ts
@@ -13,7 +13,7 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-import { setTokenUserViewPort } from '../../support/commands';
+import { goToTab, setTokenUserViewPort } from '../../support/commands';
 describe('Admin UI', () => {
   setTokenUserViewPort();
   beforeEach(() => {
@@ -27,6 +27,13 @@ describe('Admin UI', () => {
     it('Admin status indicated on profile page', () => {
       cy.get('#dropdown-accounts').click();
       cy.get('[data-cy=account-is-admin]').should('exist');
+    });
+  });
+  describe('Userpage', () => {
+    it('Admin can view other linked accounts of a user', () => {
+      cy.visit('/users/user_A');
+      goToTab('Other Linked Accounts');
+      cy.get('[data-cy=other-linked-accounts-quay]').should('be.visible');
     });
   });
 });
--- a/cypress/e2e/immutableDatabaseTests/admin.ts
+++ b/cypress/e2e/immutableDatabaseTests/admin.ts
@@ -34,6 +34,7 @@ describe('Admin UI', () => {
       cy.visit('/users/user_A');
       goToTab('Other Linked Accounts');
       cy.get('[data-cy=other-linked-accounts-Quay]').should('be.visible');
+      cy.get('[data-cy=Quay-username]').contains('user_A').should('be.visible');
 
       //log out and confirm the tab does not exist
       cy.get('[data-cy=dropdown-main]:visible').click();
--- a/cypress/e2e/immutableDatabaseTests/admin.ts
+++ b/cypress/e2e/immutableDatabaseTests/admin.ts
@@ -33,7 +33,13 @@ describe('Admin UI', () => {
     it('Admin can view other linked accounts of a user', () => {
       cy.visit('/users/user_A');
       goToTab('Other Linked Accounts');
-      cy.get('[data-cy=other-linked-accounts-quay]').should('be.visible');
+      cy.get('[data-cy=other-linked-accounts-Quay]').should('be.visible');
+
+      //log out and confirm the tab does not exist
+      cy.get('[data-cy=dropdown-main]:visible').click();
+      cy.get('[data-cy=dropdown-logout-button]').click();
+      cy.visit('/users/user_A');
+      cy.get('.mat-tab-label').contains('Other Linked Accounts').should('not.exist');
     });
   });
 });
--- a/src/app/user-page/user-page.component.html
+++ b/src/app/user-page/user-page.component.html
@@ -127,38 +127,19 @@
           <span>Other Linked Accounts</span>
         </div>
       </ng-template>
-      <div class="p-2">
-        <div *ngFor="let token of tokens">
-          <mat-card *ngIf="token.tokenSource === 'quay.io'" data-cy="other-linked-accounts-quay">
-            <mat-card-header>
-              <div mat-card-avatar matBadge="✓" class="mr-4">
-                <img src="../../assets/images/account-logos/quay.svg" alt="Quay.io logo" class="w-100" />
-              </div>
-              <mat-card-title class="normal-line-height">Quay.io Account</mat-card-title>
-            </mat-card-header>
+      <div class="w-100">
+        <div *ngIf="otherLinkedAccountsInfo.length === 0">
+          <mat-card class="alert alert-info mat-elevation-z starred-header" role="alert" data-cy="no-other-linked-accounts-banner">
+            This user has no other linked accounts.
           </mat-card>
-          <mat-card *ngIf="token.tokenSource === 'zenodo.org'" data-cy="other-linked-accounts-zenodo">
+        </div>
+        <div *ngFor="let row of otherLinkedAccountsInfo" class="p-3" fxFlex.lt-md="100">
+          <mat-card [attr.data-cy]="'other-linked-accounts-' + row.name">
             <mat-card-header>
               <div mat-card-avatar matBadge="✓" class="mr-4">
-                <img src="../../assets/images/account-logos/zenodo.svg" alt="Zenodo logo" class="w-100" />
+                <img src="../../assets/images/account-logos/{{ row.logo }}" alt="{{ row.name }} logo" class="w-100" />
               </div>
-              <mat-card-title class="normal-line-height">Zenodo Account</mat-card-title>
-            </mat-card-header>
-          </mat-card>
-          <mat-card *ngIf="token.tokenSource === 'bitbucket.org'" data-cy="other-linked-accounts-bitbucket">
-            <mat-card-header>
-              <div mat-card-avatar matBadge="✓" class="mr-4">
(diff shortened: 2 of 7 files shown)


src/app/user-page/user-page.component.html
```diff
@@ -121,6 +121,49 @@ <h4 matLine class="accent-1-dark mt-2 mb-0">{{ user?.username }}</h4>
...
+        <div *ngFor="let token of tokens">
+          <mat-card *ngIf="token.tokenSource === 'quay.io'" data-cy="other-linked-accounts-quay">
+            <mat-card-header>
+              <div mat-card-avatar matBadge="✓" class="mr-4">
+                <img src="../../assets/images/account-logos/quay.svg" alt="Quay.io logo" class="w-100" />
+              </div>
+              <mat-card-title class="normal-line-height">Quay.io Account</mat-card-title>
+            </mat-card-header>
+          </mat-card>
+          <mat-card *ngIf="token.tokenSource === 'zenodo.org'" data-cy="other-linked-accounts-zenodo">
```
coverbeck, line 140: Nit, and optional: the compiler will now catch this if you had made a typo in the string, but I still think it's clearer to have `token.tokenSource === TokenSource.ZENODO`, especially since the type is TokenUser.TokenSourceEnum.

src/app/user-page/user-page.component.ts
```diff
@@ -76,5 +79,8 @@ export class UserPageComponent extends Base implements OnInit {
 
   ngOnInit(): void {
     this.getUserInfo(this.username);
+    this.tokenQuery.tokens$.subscribe((tokens: TokenUser[]) => {
```
coverbeck, line 82: I believe you need a `takeUntil`, or this will cause a leak. Or, don't subscribe at all, and use `async` in the HTML.

src/app/user-page/user-page.component.html
```diff
@@ -121,6 +121,49 @@ <h4 matLine class="accent-1-dark mt-2 mb-0">{{ user?.username }}</h4>
...
     </mat-tab>
+    <mat-tab *ngIf="loggedInUserIsAdminOrCurator">
+      <ng-template mat-tab-label>
+        <div fxLayoutGap="0.5rem">
+          <span>Other Linked Accounts</span>
+        </div>
+      </ng-template>
+      <div class="p-2">
+        <div *ngFor="let token of tokens">
+          <mat-card *ngIf="token.tokenSource === 'quay.io'" data-cy="other-linked-accounts-quay">
```
kathy-t, line 132: Looked at the way the logged-in user's account page lists accounts and the component defines all the possible external accounts:
https://github.com/dockstore/dockstore-ui2/blob/77b17980f764735d9826d1a58ef75b96d2d18105/src/app/loginComponents/accounts/external/accounts.component.ts#L62-L141

I think it would be better to use that list of `AccountInfo` and filter it down to the accounts that the user has tokens. You'll also have to filter out the accounts that are already displayed on the page. The advantage is that you can do something like the following and not have to define a specific case for each account type:
https://github.com/dockstore/dockstore-ui2/blob/77b17980f764735d9826d1a58ef75b96d2d18105/src/app/loginComponents/accounts/external/accounts.component.html#L89-L106

We also don't have to worry about missing an account in this tab if we add another account source.

src/app/user-page/user-page.component.html
```diff
@@ -121,6 +121,49 @@ <h4 matLine class="accent-1-dark mt-2 mb-0">{{ user?.username }}</h4>
...
+            <mat-card-header>
+              <div mat-card-avatar matBadge="✓" class="mr-4">
+                <img src="../../assets/images/account-logos/quay.svg" alt="Quay.io logo" class="w-100" />
+              </div>
+              <mat-card-title class="normal-line-height">Quay.io Account</mat-card-title>
```
kathy-t, line 137: Not sure if it's a privacy issue, but I think it would also be nice to see the username associated with the account

src/app/user-page/user-page.component.html
```diff
@@ -121,6 +121,49 @@ <h4 matLine class="accent-1-dark mt-2 mb-0">{{ user?.username }}</h4>
...
       </div>
     </mat-tab>
+    <mat-tab *ngIf="loggedInUserIsAdminOrCurator">
+      <ng-template mat-tab-label>
+        <div fxLayoutGap="0.5rem">
+          <span>Other Linked Accounts</span>
+        </div>
+      </ng-template>
+      <div class="p-2">
+        <div *ngFor="let token of tokens">
```
kathy-t, line 131: Should handle the case for when there's no additional linked accounts otherwise it's just a blank tab

kathy-t (commented):
Nice to see these features that help make community lead tasks easier. 

The ticket description says that the use case is for organization requests, so I'm going to add a suggestion that would save community leads some clicks, but feel free to create a follow-up ticket if you don't want to address it here :slightly_smiling_face: 

It would be nice if the user mentioned in the organization request (highlighted in red) was linked to the user's page so community leads can click on it and view the linked accounts:
image

Can be a follow-up ticket

src/app/user-page/user-page.component.html
hyunnaye: @denis-yuen any thoughts?

cypress/e2e/immutableDatabaseTests/admin.ts
```diff
@@ -29,4 +29,11 @@ describe('Admin UI', () => {
       cy.get('[data-cy=account-is-admin]').should('exist');
     });
   });
+  describe('Userpage', () => {
+    it('Admin can view other linked accounts of a user', () => {
+      cy.visit('/users/user_A');
+      goToTab('Other Linked Accounts');
+      cy.get('[data-cy=other-linked-accounts-quay]').should('be.visible');
+    });
+  });
```
svonworl, line 38: Probably would be good to have a test that confirms that non-admin third parties _can't_ see the other linked accounts (assuming that's the desired behavior), so we don't accidentally leak the info.

src/app/user-page/user-page.component.html
denis-yuen: Since this is admin and curator only, I think it would be fine. 
The actual endpoint would be protected anyway

hyunnaye: added

src/app/user-page/user-page.component.ts
```diff
@@ -76,5 +82,14 @@ export class UserPageComponent extends Base implements OnInit {
 
   ngOnInit(): void {
     this.getUserInfo(this.username);
+    this.tokenQuery.tokens$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((tokens: TokenUser[]) => {
+      for (const account of this.accountsInfo) {
+        const found = tokens.find((token) => token.tokenSource === account.source);
+        if (found && !['GitHub', 'Google', 'ORCID'].includes(account.name)) {
```
coverbeck, line 88: Instead of using strings where there is the possibility of a typo mismatch, you could use/rename the unused `show` property, after updating its declarations accordingly. It also makes is easy to add new accounts that we might want to show without needing to update this code. Something like `const found = tokens.find((token) => token.tokenSource === account.source && account.showForAdmins)`.

I know I keep harping on using types instead of strings, and while you don't have a typo mismatch here, it's better to have the compiler do the check than manually verifying it. Plus, if they ever change the spelling/casing, it's easier to update in one place rather than figure out this code is related to AccountInfo.

src/app/user-page/user-page.component.ts
```diff
@@ -76,5 +82,14 @@ export class UserPageComponent extends Base implements OnInit {
...
+          account.username = found?.username;
+          this.otherLinkedAccountsInfo.push(account);
```
coverbeck, line 90: I am a little wary of modifying the "global" AccountInfo instances with the username. When you leave this page, `account.username` is going to be set to the user. If another component uses the `AccountInfo` instances, the instances will have the left-over username. While it may not matter today in terms of what's displayed in the UI, it could cause some confusion.

I suggest cloning the values, using Object.assign or the spread operator.

src/app/user-page/user-page.component.ts
```diff
@@ -25,14 +27,20 @@ export class UserPageComponent extends Base implements OnInit {
   public googleProfile: Profile;
   public gitHubProfile: Profile;
   public loggedInUserIsAdminOrCurator: boolean;
+  protected otherLinkedAccountsInfo: AccountInfo[] = [];
+  accountsInfo: Array<AccountInfo> = Object.assign([], accountInfo);
```
coverbeck, line 31: This creates a new array, but it doesn't create copies of the array elements, so the the code on line 91 is still modifying the global `AccountInfo` objects.

src/app/user-page/user-page.component.ts
```diff
@@ -76,5 +84,14 @@ export class UserPageComponent extends Base implements OnInit {
 
   ngOnInit(): void {
     this.getUserInfo(this.username);
+    this.tokenQuery.tokens$.pipe(takeUntil(this.ngUnsubscribe)).subscribe((tokens: TokenUser[]) => {
+      for (const account of this.accountsInfo) {
+        const found = tokens.find((token) => token.tokenSource === account.source);
+        if (found && !this.publicAccountsSource.includes(account.source)) {
+          account.username = found?.username;
+          this.otherLinkedAccountsInfo.push(account);
```
coverbeck, line 92: Should be something like `this.otherLinkedAccountsInfo.push(Object.assign({username: found?.username}), account)`, which will create a copy of the account object.

denis-yuen (approved):

kathy-t (approved):

svonworl (approved):

PR #1823, opened 2023-07-20
DOCK-2411: Fix console errors when viewing dag tab

y-ng:
**Description**
When looking at the workflow DAG tab, sometimes there would be an error related to 'edges' being undefined and the tab would be blank. This is because sometimes `dagResults` would be an empty object so this PR adds a check to see if the results contain `edges` and `nodes`, and then display the missing message in the DAG tab.

**Review Instructions**
Go to the DAG tab inside a workflow (such as the workflow mentioned in the linked issue) and check that there are no errors in the browser console.

**Issue**
[DOCK-2411](https://ucsc-cgl.atlassian.net/browse/DOCK-2411)

**Security**

[DOCK-2411]: https://ucsc-cgl.atlassian.net/browse/DOCK-2411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/workflow/dag/dag.component.html
+++ b/src/app/workflow/dag/dag.component.html
@@ -21,9 +21,12 @@
         <mat-icon class="alert-warning-icon">warning</mat-icon>&nbsp;DAG display is still experimental for Nextflow.
       </mat-card>
     </div>
-    <div fxFlex *ngIf="dagType === 'classic' && (isNFL$ | async) === false && (dagResult$ | async) === false">
+    <div
+      fxFlex
+      *ngIf="dagType === 'classic' && (isNFL$ | async) === false && (!(dagResult$ | async) || (dagResult$ | async | json) === '{}')"
+    >
       <!-- This covers CWL, WDL, Service, and other plugin languages (everything except NFL) -->
-      <mat-card class="alert alert-warning mat-elevation-z" role="alert" *ngIf="(dagResult$ | async) === false && !selectedVersion?.valid">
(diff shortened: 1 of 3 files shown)


david4096 (approved):

src/app/workflow/dag/dag.component.html
```diff
@@ -21,9 +21,12 @@
         <mat-icon class="alert-warning-icon">warning</mat-icon>&nbsp;DAG display is still experimental for Nextflow.
       </mat-card>
     </div>
-    <div fxFlex *ngIf="dagType === 'classic' && (isNFL$ | async) === false && (dagResult$ | async) === false">
+    <div
+      fxFlex
+      *ngIf="dagType === 'classic' && (isNFL$ | async) === false && ((dagResult$ | async) === null || (dagResult$ | async | json) === '{}')"
```
coverbeck, line 26: Shouldn't `((dagResult$ | async) === null || (dagResult$ | async | json) === '{}')` instead be `(missingTool$ | async)`?

Otherwise, there is the hypothetical case where the API response is something like `{nodes: [], edges: []}`, which wouldn't meet this `*ngIf` condition, and the code on line 40 would never get evaluated and that HTML never displayed.

y-ng: Good point! Then I suppose the type of warning displayed (missing the descriptor or missing required tools) would mainly depend on the value of `selectedVersion?.valid`.

kathy-t (approved):

PR #1826, opened 2023-07-25
SEAB-5576: Allow entries to be deleted

svonworl:
**Description**
This PR implements the UI elements that allow entries to be deleted in certain cases.

The primary changes are:
1. The addition of a "Delete" button on the non-public page for an entry, which is displayed only when the webservice reports that the entry is deletable, which currently, for new entries, is when the entry has never been published and is not checked or a checker workflow.
2. A new dialog that pops up when someone presses the "Delete" button, which details what will happen when the entry is deleted and asks the user to confirm that they want to delete the entry.  It includes some instructions about modifying/deleting the source .dockstore.yml so that the entry is no longer described, to prevent it from reappearing upon a future push.

When I began this PR, it had been so long since I had interacted with an old-style tool that I forgot that they already could be deleted via the UI.  So, I left the existing old-style tool interface as-is.  The UI changes in this PR appear for entries that are a derivative of `Workflow`: `BioWorkflows`, `AppTools`, `Services`, and `Notebooks`.  There's some differences between the two deletion mechanisms: the old-style tool deletion interface allows tools to be deleted if they're not published (but have previously been published), the "Delete" button is always displayed but greyed-out appropriately, and the dialog is different.  If desired, we can make the two consistent, the new deletion scheme would work for old-style tools.
 
**Screenshots**
Screen Shot 2023-07-25 at 11 41 53 AM

Screen Shot 2023-07-25 at 11 43 28 AM

**Review Instructions**
Create an unpublished entry via the .dockstore.yml registration path.  Visit the non-public entry page, and "Delete" it via the button.  Confirm the entry is gone.
Create an unpublished entry via the .dockstore.yml registration path.  Publish the entry.  Visit the non-public entry page, and confirm that the "Delete" button is not displayed.
Create an unpublished entry via the .dockstore.yml registration path.  Publish the entry, then, unpublish it.  Visit the non-public entry page, and confirm that the "Delete" button is not displayed.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5576

**Security**


david4096 (approved):

david4096:
Looks good to me. Non-blocking quibbles/thoughts I had.

* For entries using the old Refresh code three buttons will appear which may lead to crowding. I don't have an immediate suggestion of where else it should go, perhaps on the bottom of the Info tab?
* Considering UX, a trash icon or red for the delete button itself might be good
* Nice-to-have would be redirecting back to my-workflows or my-notebooks depending on the type being deleted rather than the dashboard
* Presuming the reason we need them to disconnect the GitHub app is we didn't request that permission? Another nice-to-have would be to disconnect this for them somehow

svonworl:
Some of my rationale/thoughts on these:

> * For entries using the old Refresh code three buttons will appear which may lead to crowding. I don't have an immediate suggestion of where else it should go, perhaps on the bottom of the Info tab?
> * Considering UX, a trash icon or red for the delete button itself might be good

We could reduce the amount of space that the "Delete" button consumes by changing the text label to a trash icon.  The blue button color was intentional, to avoid distracting/drawing too much attention, but I'm not opposed to making it red either.

> * Nice-to-have would be redirecting back to my-workflows or my-notebooks depending on the type being deleted rather than the dashboard

I went with `/dashboard` because it's more likely [in the future] to do the right thing if the entry deleted was the last of its type (ex: the last notebook was deleted and there's none left), but `/my-<entry-type>s` is better otherwise.

> * Presuming the reason we need them to disconnect the GitHub app is we didn't request that permission? Another nice-to-have would be to disconnect this for them somehow

Yeah, I think the user has to deinstall the GitHub app themselves.  In some [most?] cases, they won't, though, because they'll simply be fixing a naming problem or removing an experiment or something like that, and continuing to register/update entries from the repo...

src/app/entry/delete/dialog/delete-entry-dialog.component.html
```diff
@@ -0,0 +1,22 @@
+<h1 mat-dialog-title>Are you sure you want to delete this Dockstore {{ term }}?</h1>
```
coverbeck, line 1: I suggest specifying the name of the entry. If not here, because it may cause wrapping issues in the title, then at least on line 4.

I sometimes run into with the AWS Console UI, where it says "Are you sure you want to delete this <entity type>. I may not be 100% what they're referring to (the dialog may be covering, and I get paranoid it's referring to some crucial resource, so I dismiss the dialog to make sure I'm deleting the right thing. And then I bring up the dialog again, but even then, did I mis-click?

In your screenshot, you can see the selected/targeted notebook, but there's not guarantee it won't be covered up on a smaller screen or if we reorg things some day.

src/app/entry/delete/dialog/delete-entry-dialog.component.html
```diff
@@ -0,0 +1,22 @@
...
+  <p>You are about to permanently delete this {{ term }}. After you delete it, it will no longer appear on Dockstore.</p>
+  <p>A {{ term }} can be deleted if it:</p>
+  <ul>
+    <li>Has never been published,</li>
+    <li>Was created on/after April 1, 2022, and</li>
+    <li>Is not a checked or checker workflow.</li>
+  </ul>
+
+  <p>
+    After you delete the {{ term }}, you must uninstall the Dockstore GitHub App from the source GitHub repo, or edit/remove the
```
coverbeck, line 13: ~~I forget the details of your webservice PR, are we only allowing deletion of GitHub App entries? If yes, never mind; if no, this text may not always be applicable.~~

OK, read your PR description after I entered the above... Do we bring up a different dialog for non-GitHub app entries? Does it look radically different?

src/app/entry/delete/dialog/delete-entry-dialog.component.ts
```diff
@@ -0,0 +1,89 @@
...
+          this.close();
+        })
+      )
+      .subscribe(
+        () => {
+          this.inform(`The ${this.term} was deleted.`);
+          this.redirect();
+        },
+        (error: HttpErrorResponse) => {
+          this.inform(`Could not delete this ${this.term}.`);
```
coverbeck, line 85: 1. Would be nice to pass along error details to `inform()`.
2. If there's an error, the dialog is going to disappear right away because of line 76. Our snackbar messages are pretty subtle (that's another issue unrelated to this PR), so it may not be immediately apparent to the user what just happened. But I'm OK saying that's beyond the scope of this.

src/app/shared/entry-actions/workflow-actions.component.ts
```diff
@@ -65,4 +68,8 @@ export class WorkflowActionsComponent extends EntryActionsComponent implements O
     const prefix = this.sessionQuery.getValue().entryType === EntryType.Service ? ga4ghServiceIdPrefix : ga4ghWorkflowIdPrefix;
     this.refreshService.refreshWorkflow(prefix + this.workflow.full_workflow_path, versionName);
   }
+
+  delete() {
+    this.dialog.open(DeleteEntryDialogComponent, { width: '600px', data: this.workflow });
```
coverbeck, line 73: Good thing Angular doesn't type check the `data` here with what's injected into the dialog, otherwise this would fail, as a `Workflow` is not an `Entry`, at least not until we solve dockstore/dockstore#5605. :)

src/app/entry/delete/dialog/delete-entry-dialog.component.html
```diff
@@ -0,0 +1,22 @@
...
+    .dockstore.yml file so that it no longer describes the {{ term }}. If you do not, the {{ term }} may reappear on Dockstore the next time
```
svonworl, line 14: The dialog should probably only display this paragraph if the entry was created via the .dockstore.yml scheme, otherwise it could be a little confusing.

src/app/entry/delete/dialog/delete-entry-dialog.component.ts
svonworl: Now that you mention it, isn't our general policy to make the dialog persist on failure?  If so, it's easy to make it conform.

denis-yuen:
Reading from the top (some of these may be covered already)


> * The addition of a "Delete" button on the non-public page for an entry, which is displayed only when the webservice reports that the entry is deletable, which currently, for new entries, is when the entry has never been published and is not checked or a checker workflow.

Wonder if it would make sense to show it greyed out for non-deletable entries but with explanation. Or maybe just a "i" info type building

> When I began this PR, it had been so long since I had interacted with an old-style tool that I forgot that they already could be deleted via the UI. So, I left the existing old-style tool interface as-is.  

Makes sense

denis-yuen (approved):
pending suggestions above

src/app/entry/delete/dialog/delete-entry-dialog.component.html
```diff
@@ -0,0 +1,22 @@
...
+    you push to the repo.
+  </p>
+  <p>Are you sure you want to delete this {{ term }}?</p>
+  <div class="btn-group" fxLayout="row" fxLayoutAlign="end center" fxLayoutGap="1rem">
+    <button mat-flat-button (click)="no()" data-cy="delete-no" [disabled]="clicked" mat-dialog-close class="secondary-1">No</button>
+    <button mat-flat-button (click)="yes()" data-cy="delete-yes" [disabled]="clicked" color="warn">Yes, delete this {{ term }}</button>
```
kathy-t, line 20: I think we should avoid using `No/Yes` although I think we have a few dialogs that do this on the accounts page. Most of our other dialogs use `Cancel/Verb` and the examples in the [Google Material design ](https://m3.material.io/components/dialogs/guidelines) guide do that too. I suggest `Cancel/Delete this {{term}}` or just `Cancel/Delete`

src/app/shared/entry-actions/workflow-actions.component.ts
kathy-t, line 73: Should use one of these sizes:
https://github.com/dockstore/dockstore-ui2/blob/d77d52aff6a67d9f739b91959701abae09c697e7/src/app/shared/constants.ts#L28-L30

src/app/shared/entry-actions/workflow-actions.component.html
```diff
@@ -32,4 +32,15 @@
   >
     {{ 'Refresh' }}
   </button>
+  <button
+    *ngIf="workflow?.deletable"
```
kathy-t, line 36: Similar to Denis' comment:
> Wonder if it would make sense to show it greyed out for non-deletable entries but with explanation. Or maybe just a "i" info type building

I personally prefer showing the greyed out button and providing information to the user about why they can't delete. This way, they know that deleting is an option under the correct circumstances

svonworl: > Similar to Denis' comment:
> 
> > Wonder if it would make sense to show it greyed out for non-deletable entries but with explanation. Or maybe just a "i" info type building
> 
> I personally prefer showing the greyed out button and providing information to the user about why they can't delete. This way, they know that deleting is an option under the correct circumstances

Understood, but here's an argument for the button just disappearing:
An entry can only be deleted if it's never been published.  In other words, once an entry is published, it will never again be deletable, there's nothing the user can do to make it so.  So, the corresponding greyed-out button will sit in the UI next to the never-again-deletable entry, consuming space and attention, taunting the user about what might have been, for eternity.

denis-yuen: > taunting the user about what might have been, for eternity

:laughing: 
This was kinda my thoughts behind the "info" 
Screenshot from 2023-07-26 15-04-37
Because it's less of an actionable thing but more of a BTW

kathy-t: > never-again-deletable entry, consuming space and attention, taunting the user about what might have been, for eternity.

Good point :grin: I like Denis' "info" idea too. 

Even if we don't keep the button/info, it would be worth documenting this new functionality since there are conditions that need to be met in order to delete and a screenshot of the delete dialog box might be of interest to a user who can't view it (because they published their workflow, for example)

src/app/entry/delete/dialog/delete-entry-dialog.component.html
svonworl: For classic tools, the "Delete" button is red[-ish], and we display a much more spartan dialog:
Screen Shot 2023-07-26 at 2 17 14 PM

svonworl: (The new delete functionality in this PR will apply to all derivates of `Workflow`, not matter how they were registered.)

svonworl:
Ok, I made the following changes:
1. The dialog text regarding adjusting `.dockstore.yml` is only displayed if the entry was created via the `.dockstore.yml` registration scheme.
2. The path of the entry is now included in the dialog.
3. Buttons -> "Cancel", "Delete this <term>"
4. Dialog now persists on failure.
5. Dialog is now bootstrap4largeModalSize wide
6. Dialog now uses the `AlertService`, which will display some persistent failure information in the dialog and on the entry page upon deletion failure.

svonworl:
Revised dialog:
Screen Shot 2023-07-27 at 12 25 42 PM

kathy-t (approved):

src/app/entry/delete/dialog/delete-entry-dialog.component.html
```diff
@@ -0,0 +1,24 @@
+<h1 mat-dialog-title>Are you sure you want to delete this Dockstore {{ term }}?</h1>
+<div mat-dialog-content>
+  <app-alert></app-alert>
+  <p>
+    You are about to permanently delete the {{ term }} <b>{{ path }}</b
+    >. After you delete the {{ term }}, it will no longer appear on Dockstore.
+  </p>
+  <p>A {{ term }} can be deleted if it:</p>
+  <ul>
+    <li>Has never been published,</li>
```
coverbeck, line 10: Do these 3 restrictions apply to non-GitHub app entries? I don't know. But if they don't apply, then I think you'd need the `*ngIf`.

coverbeck (approved):
Assuming the answer to my one question is no.

src/app/entry/delete/dialog/delete-entry-dialog.component.html
svonworl: Yup, the restrictions apply to all entries.

PR #1827, opened 2023-08-01
DOCK-2428: Display markdown tables correctly

y-ng:
**Description**
Some markdown tables do not render properly due to the presence of tab characters in the table. This PR removes tabs from table headers before parsing so that the html is generated correctly for the table.

Before:
Screenshot from 2023-08-01 16-48-48

After:
Screenshot from 2023-08-01 16-47-06

**Review Instructions**
Go to the info tab for a workflow with a table in the description (e.g., the workflow mentioned in the related ticket /workflows/github.com/aofarrel/tree_nine/tree_nine:main?tab=info) and make sure the table is displayed properly.

**Issue**
[DOCK-2428](https://ucsc-cgl.atlassian.net/browse/DOCK-2428)

**Security**

[DOCK-2428]: https://ucsc-cgl.atlassian.net/browse/DOCK-2428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
+++ b/src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
@@ -107,6 +107,6 @@ export class MarkdownWrapperService {
    * @returns {string} The modified string where all tabs in lines beginning with '|' are replaced with four spaces
    */
   removeTabsFromTableHeaders(data: string): string {
-    return data.replace(/(^\|.*)/gm, (match) => match.replace(/\t/g, '    '));
+    return data.replace(/(^ {0,3}\|.*)/gm, (match) => match.replace(/\t/g, '    '));
   }
 }
--- a/src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
+++ b/src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
@@ -87,7 +87,9 @@ export class MarkdownWrapperService {
    */
   customCompile(data, baseUrl): string {
     const parseOptions = { markedOptions: { baseUrl: baseUrl } };
-    return this.markdownService.parse(data, parseOptions);
+    // Remove tab characters from markdown table headers or they won't display properly
+    const markdownData = data.replace(/(?<=\|[-:\t ]*)\t(?=[-:\t ]*\|)/g, '');
+    return this.markdownService.parse(markdownData, parseOptions);
   }
 
   customSanitize(html): string {
--- a/src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
+++ b/src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
@@ -87,8 +87,7 @@ export class MarkdownWrapperService {
    */
   customCompile(data, baseUrl): string {
     const parseOptions = { markedOptions: { baseUrl: baseUrl } };
-    // Remove tab characters from markdown table headers or they won't display properly
-    const markdownData = data.replace(/(?<=\|[-:\t ]*)\t(?=[-:\t ]*\|)/g, '');
+    const markdownData = this.removeTabsFromTableHeaders(data);
     return this.markdownService.parse(markdownData, parseOptions);
   }
 
@@ -98,5 +97,18 @@ export class MarkdownWrapperService {
       FORBID_TAGS: this.forbidTags,
       FORBID_ATTR: this.forbidAttr,
     });
+  }
+
+  /**
+   * Removes tab characters from markdown table headers or they won't display properly
+   * Markdown uses pipes with three or more hyphens in between to create columns and headers, and colons to align text
+   * E.g., |---|---|---| or | :--- | ---: | where there can be spaces between pipes and hypens but not tabs
+   * @param {string} data A string containing the markdown data
+   * @returns {string}
+   */
+  removeTabsFromTableHeaders(data: string): string {
+    return data
+      .replace(/(?<=\|[\t ]*)\t(?=[\t ]*:?-+:?[\t ]*\|)/g, '    ') // remove tabs after the starting pipe before the hyphens
+      .replace(/(?<=\|[\t ]*:?-+:?[\t ]*)\t(?=[\t ]*\|)/g, '    '); // remove tabs after the hyphens before the ending pipe
   }
 }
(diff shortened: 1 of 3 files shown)


src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
```diff
@@ -87,7 +87,9 @@ export class MarkdownWrapperService {
    */
   customCompile(data, baseUrl): string {
     const parseOptions = { markedOptions: { baseUrl: baseUrl } };
-    return this.markdownService.parse(data, parseOptions);
```
denis-yuen, line 90: While it is good that you can reproduce (and fix)!

I'm always a little wary of regular expressions. 
e.g. could some other content accidentally get modified or could tab characters get left in accidentally

Wondering if you tried the `gfm` option to see if it matches github's behaviour 

I also see some changes  in how tabs/tables are handled in v4.0.14 through v4.0.11 so you could try updating past them to see if that changes anything 
https://github.com/markedjs/marked/releases/tag/v4.0.14
https://github.com/markedjs/marked/releases/tag/v4.0.12
https://github.com/markedjs/marked/releases/tag/v4.0.11

y-ng: Setting `gfm: true` didn't fix the table formatting, and I also tried updating the version (we were already using `marked@4.2.3` so I tried v4.3.0) but this unfortunately doesn't resolve the issue either.

src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
```diff
@@ -87,7 +87,9 @@ export class MarkdownWrapperService {
...
+    // Remove tab characters from markdown table headers or they won't display properly
+    const markdownData = data.replace(/(?<=\|[-:\t ]*)\t(?=[-:\t ]*\|)/g, '');
```
coverbeck, line 91: I'm wary of the regex as well, but if it's the only way, then I guess there's no other option.

* I suggest making a method out of this and adding a bunch more unit tests, creating a markdown-wrapper.service.spec.ts. That way it's easier to have more tests focused on just this regex. It's minor, but it's easier to test more inputs than the more verbose tests you need in the component spec.
```
expect(service.removeTabsFromHeaders(input1).toEqual(output1);
expect(service.removeTabsFromHeaders(input2).toEqual(output2);
expect(service.removeTabsFromHeaders(input3).toEqual(output3);
....
```
* Do you need a `^\s*` at the beginning of the regex? I think, but please confirm, the `|---` table syntax has to be the first non-blank characters of the line to be a table, doesn't it? IOW, if you have the line `I am not a table | col 1 | col 2 |`, wouldn't that match your regex, but it shouldn't?
* Should you replace the tab with a space instead of an empty string? I'm not sure, just asking. If you limit it to inside table headers, maybe it doesn't matter.
* Some more detailed comments on what exactly the regex is doing would be nice; I always forget the more detailed regex syntax, so it's hard for me to understand it. If you move it to a method, you could make it a top-level comment.

denis-yuen:
> warning The version of Java (11.0.14.1) you have used to run this analysis is deprecated and we will stop accepting it soon. Please update to at least Java 17.
> Read more [here](https://docs.sonarcloud.io/appendices/scanner-environment/)

WIll create ticket

src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
```diff
@@ -97,4 +98,17 @@ export class MarkdownWrapperService {
...
+   * Markdown uses pipes with three or more hyphens in between to create columns and headers, and colons to align text
+   * E.g., |---|---|---| or | :--- | ---: | where there can be spaces between pipes and hypens but not tabs
+   * @param {string} data A string containing the markdown data
+   * @returns {string}
+   */
+  removeTabsFromTableHeaders(data: string): string {
+    return data
+      .replace(/(?<=\|[\t ]*)\t(?=[\t ]*:?-+:?[\t ]*\|)/g, '    ') // remove tabs after the starting pipe before the hyphens
+      .replace(/(?<=\|[\t ]*:?-+:?[\t ]*)\t(?=[\t ]*\|)/g, '    '); // remove tabs after the hyphens before the ending pipe
+  }
```
y-ng, line 113: I changed the regex a bit so it's more specific (and is potentially less likely to modify other content) but I wonder if it's become too complicated. The comments generally describe what's happening but on line 111 (and similarly for line 112), for example, tabs are removed when:
1. There is a preceding pipe `?<=\|` followed by zero or more spaces/tabs `[\t ]*`
2. And the tab comes before zero or more spaces/tabs `?=[\t ]*`, followed by zero or one colon`:?`, followed by more than one hyphen`-+`, followed by zero or one colon `:?`, followed by zero or more spaces/tabs `[\t ]*`, followed by a pipe `\|`

Also, to be safe the tab is being replaced by four spaces instead of an empty string.

svonworl:
Would it work to replace all tabs with four spaces, but just in the lines that start with `|`?

y-ng:
> Would it work to replace each tab with four spaces, but just in the lines that start with `|`? I'm a little bit concerned about the SonarCloud "Security Hotspots", it's tough to tailor complicated regexps so that they're not DOS-able.

Yeah, the tables would still render correctly since the tabs are removed (and this gets rid of the security hotspots which is good). The only thing is that tabs in other lines beginning with '|" would also be replaced with spaces, but that might not be too big of a concern if it's not very common.

coverbeck (approved):

src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
```diff
@@ -97,4 +98,15 @@ export class MarkdownWrapperService {
...
+
+  /**
+   * Removes tab characters from markdown table headers or they won't display properly
+   * Markdown uses pipes with three or more hyphens in between to create columns and headers, and colons to align text
+   * E.g., |---|---|---| or | :--- | ---: | where there can be spaces between pipes and hyphens but not tabs
+   * @param {string} data A string containing the markdown data
+   * @returns {string} The modified string where all tabs in lines beginning with '|' are replaced with four spaces
+   */
+  removeTabsFromTableHeaders(data: string): string {
+    return data.replace(/(^\|.*)/gm, (match) => match.replace(/\t/g, '    '));
```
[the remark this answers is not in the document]
y-ng: Good point, I'll account for that. It seems you can only put a maximum of three leading spaces for it to be a table though.

With three leading spaces:
   |col1|col2|
   |---|---|
   |1|2|

With four leading spaces:
    |col1|col2|
    |---|---|
    |1|2|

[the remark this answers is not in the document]
y-ng: FYI I'm using ' ' instead of '\s' to account for spaces only rather than any whitespace character (tabs, linebreaks, etc.)

svonworl (approved):

kathy-t (approved):

PR #1844, opened 2023-09-20
Update expired artifacts

coverbeck:
**Description**
Fix smoke test failures

**Review Instructions**
Nightly smoke tests should pass.

**Issue**
None

**Security**

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "config": {
     "webservice_version": "1.15.0",
     "use_circle": true,
-    "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/10298/workflows/6d3ce13c-045a-4c13-be33-d2b406130c24/jobs/36411/artifacts",
-    "circle_build_id": "36411"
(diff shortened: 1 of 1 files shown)


coverbeck:
Sheesh, why did build fail?

denis-yuen:
Think the accessibility test compares the branch with the base branch to see if things are getting worse. So I think it is expected to fail in this case

denis-yuen (approved):

coverbeck:
I think dockstore/dockstore#5643 broke things. ☹️. Workflow no longer has an id, among other things.

coverbeck:
Updated OpenAPI generator to 5.1.0, fixes missing properties, but now enum casing is different. :(

coverbeck:
> Updated OpenAPI generator to 5.1.0, fixes missing properties, but now enum casing is different. :(

Fixed that by setting a property in swagger-config.json. New problem is that workflowVersions and other collections are now a Set instead of an array.

svonworl (approved):

hyunnaye (approved):

denis-yuen (commented):
Will probably need a new one

PR #1847, opened 2023-09-29
SEAB-5913: Improve search relevance

svonworl:
**Description**
This PR improves the relevance of Dockstore search results.

To run a search, the UI synthesizes a search query using the bodybuilder library and submits it to ES.  Before continuing, please read the following, which explains the structure of typical ES search queries:

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html

In the previous implementation:
* Keyword subqueries were run in a `filter` context, preventing ES from scoring the results.
* The UI always sorted the results, per the header control settings, so even if ES returned results sorted by relevance, the UI would reorder them (in descending star order by default).
* Essentially, when we ran a search, ES was determining all of the hits that matched the search criteria, sorting them by stars, and returning an arbitrary 200 of them.  There was no ordering by relevance, so for searches with many hits, it was entirely possible that the most relevant entry could be omitted from the results.

In this implementation:
* Keyword subqueries run in a `should` context, which allows ES to calculate a relevance score for each hit by combining the scores of the subqueries and order the returned hits by score.  Each hit's score is returned in the query response.
* We change the UI so sorting is initially disabled, and the results are displayed in the order that ES returns them.  Each header sort control now has a "none" setting, in addition to "ascending" and "descending".
* When we run a keyword search,  ES now orders the hits by score.  Otherwise, it orders the results by stars as it did before.
* The query synthesizer now splits the search string into terms, and creates individual subqueries for each term.   The net effect is that entries which contain both terms are marked as more relevant than before.
* To improve relevance, weights have been introduced, via the `boost` setting in each subquery.
* The topic and categories are now included in keyword subqueries.

When reviewing the code, keep in mind that bodybuilder's `orQuery` adds a subquery to the `should` context, and that we force a minimum number of "shoulds" to match via the `minimum_should_match` field.

There is a companion PR that tweaks the webservice to further improve search relevance:
https://github.com/dockstore/dockstore/pull/5686

In combination, these two PRs greatly improve the quality of Dockstore keyword search results.  On prod, Ash's "samtools index" keyword search test ranks his and Milan's workflows, which are probably the most relevant hits, somewhere between the middle and bottom of all results.

After the improvements, his and Milan's workflows appear as results 1 and 3, with Julian's, which is also quite relevant, as result 2:

image

Yay!

**Review Instructions**
Go to Search, and run some keyword, facet, and advanced searches, looking for inconsistencies in the behavior or results, or regressions in search quality.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5913

**Security**

### DIFF
--- a/src/app/search/query-builder.service.ts
+++ b/src/app/search/query-builder.service.ts
@@ -115,8 +115,11 @@ export class QueryBuilderService {
     tableBody = tableBody.query('match', '_index', index);
     tableBody = this.appendQuery(tableBody, values, advancedSearchObject, searchTerm);
     tableBody = this.appendFilter(tableBody, null, filters, exclusiveFilters);
-    // most popular results should be returned first
-    tableBody = tableBody.sort('stars_count', 'desc');
+    // the default sort order is by ES score
+    // only sort by stars if there's no search term
+    if (this.isEmpty(values)) {
+      tableBody = tableBody.sort('stars_count', 'desc');
+    }
     const builtTableBody = tableBody.build();
     const tableQuery = JSON.stringify(builtTableBody);
     return tableQuery;
@@ -191,7 +194,7 @@ export class QueryBuilderService {
    */
   appendQuery(body: any, values: string, oldAdvancedSearchObject: AdvancedSearchObject, searchTerm: boolean): any {
     const advancedSearchObject = { ...oldAdvancedSearchObject };
-    if (values.toString().length > 0) {
+    if (!this.isEmpty(values)) {
       body = this.searchEverything(body, values);
     } else {
       body = body.query('match_all', {});
@@ -208,6 +211,10 @@ export class QueryBuilderService {
     return body;
   }
 
+  isEmpty(values: string): boolean {
+    return values.toString().trim().length <= 0;
+  }
+
   /**
    * Appends search-everything filter to the query
    * Currently searches sourcefiles, description, labels, author, and path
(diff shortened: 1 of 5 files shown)


src/app/search/query-builder.service.ts
```diff
@@ -221,16 +247,23 @@ export class QueryBuilderService {
...
-        .orFilter('match_phrase', 'tags.sourceFiles.content', searchString)
-        .orFilter('match_phrase', 'description', searchString)
-        .orFilter('match_phrase', 'labels', searchString)
-        .orFilter('match_phrase', 'author', searchString)
-    );
+    // Extract each search term from the search string, limiting to a maximum of 20 terms to prevent a DOS attack
+    const terms = searchString.trim().split(' ').slice(0, 20);
+    terms.forEach((term) => {
+      body
+        .orQuery('wildcard', 'full_workflow_path', { value: '*' + term + '*', case_insensitive: true, boost: 7 })
```
denis-yuen, line 254: Neat! The boost categories are pretty readable/tweakable in the future

denis-yuen (approved):
:partying_face:

kathy-t (approved):

src/app/search/query-builder.service.ts
```diff
@@ -191,23 +195,45 @@ export class QueryBuilderService {
...
+        body = this.searchEverything(body, values);
+      } else {
+        body = body.query('match_all', {});
+      }
     }
     return body;
   }
 
+  private isEmpty(values: string): boolean {
+    return values == undefined || values.toString().trim().length <= 0;
```
coverbeck, line 217: Nit: `values === undefined`, unless you're trying to catch the null case as well, in which case `!values` is more readable to me.

src/app/search/query-builder.service.ts
```diff
@@ -261,90 +294,38 @@ export class QueryBuilderService {
     return body;
   }
 
-  // Advanced search query related functions
+  advancedSearchDescription(body: any, advancedSearchObject: AdvancedSearchObject) {
+    this.advancedSearch(body, advancedSearchObject, 'description');
+  }
 
-  /* TODO: Make this better */
```
coverbeck, line 266: You made it better! Less code is good.

coverbeck (approved):
> Each header sort control now has a "none" setting, in addition to "ascending" and "descending"

Just curious, where is this in the code? Is it because you don't explicitly set the `start` attribute?

This is great; it seems like we've only been using of a small subset of the ES feature set until now.

david4096 (approved):

svonworl:
> > Each header sort control now has a "none" setting, in addition to "ascending" and "descending"
> 
> Just curious, where is this in the code? Is it because you don't explicitly set the `start` attribute?

Previously, we disabled the "none" setting with `matSortDisableClear`:
https://material.angular.io/components/sort/overview

PR #1852, opened 2023-10-19
DOCK-1195: Add archived entry UI

svonworl:
**Description**
This PR modifies the UI to accommodate archived entries, which are intended to be read-only and deemphasized in the UI.  We imagine that users will archive old/retired/obsolete entries that they aren't allowed to delete.  To that end, this PR:

* Adds "Archive"/"Unarchive" buttons to the "actions" buttons that appear to the upper right on the entry screen.
* Gates the "Archive" button with a dialog that explains what happens when the entry is archived.
* Marks archived entries as "Archived".
* Disables/hides UI elements that could be used to modify an archived entry.
* Adds a new "Archived" column to the "my entries" sidebar accordion, which appears as necessary to list archived entries.
* Sorts the Search results so that archived entries always appear last, and adds a collapsed "Archived" facet.

See below for a screenshot.

In retrospect, this should have been split into at least two PRs, but by the time I realized so, it was too late.

Editorial: this PR reminded me of how very difficult it can be to correctly make substantial changes to the current UI.  There's many contributing factors: the relative heaviness of Angular, the tool/workflow split, extensive Observable manipulation, state scattered everywhere, race conditions, lack of documentation, duplicate+ low-level code/HTML, etc.  In many systems, relative to the backend, it's much easier to experiment with and iterate the UI, but not here.  In the longer term, it'd probably benefit us to consider how to improve the situation.  Retire the old-style tool concept?  Simplify the UI to simplify the code?  Refactor (to varying degrees)?  UI3?  IMHO, all of those options should be on the table...

In that vein, there are numerous slight inconsistencies between how the new archive UI appears/works for workflows/notebooks/apptools versus old-style tools, much like the existing UI. :)

**Screenshots**
The gate dialog:
image

The "my" page an archived workflow:
image

**Review Instructions**
Register and publish an entry on qa, then navigate to the entry page and archive it.  Confirm that when the entry is archived, the entry is marked "Archived", the UI becomes read-only, the entry appears in the "Archived" section of your entry sidebar, and that the entry appears last in the appropriate Search results.  Unarchive the entry and confirm that all of these changes have been reversed.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1195
dockstore/dockstore#3199

**Security**


coverbeck:
In your PR description, the link to the GitHub issue is missing dockstore/dockstore (I've made the same mistake many times :) ).

src/app/shared/entry-actions/entry-actions.service.ts
```diff
@@ -184,4 +189,41 @@ export class EntryActionsService {
       );
     }
   }
+
+  archiveEntry(entry: Entry) {
+    this.dialog.open(ArchiveEntryDialogComponent, {
+      width: bootstrap4largeModalSize,
+      data: {
+        entry: entry,
+        callback: (archived: Entry) => {
```
coverbeck, line 198: I don't think we do callbacks with any other dialogs. I'd lean towards doing it the same way for consistency's sake, even though it's arguably more of a hassle to do the subscription.

src/app/mytools/sidebar-accordion/sidebar-accordion.component.html
```diff
@@ -17,40 +17,32 @@
...
-              </a>
-            </mat-list-item>
-            <mat-list-item *ngIf="!orgObj || orgObj?.published.length === 0">
-              <p class="text-center size-small">No published tools.</p>
-            </mat-list-item>
-          </mat-list>
+          <ng-template *ngTemplateOutlet="items; context: { items: orgObj.published, itemState: 'published' }"></ng-template>
         </mat-tab>
         <mat-tab label="Unpublished">
-          <mat-list class="org-accordion-list">
```
coverbeck, line 37: Bravo for de-duplication.

src/app/entry/archive/dialog/archive-entry-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
...
+})
+export class ArchiveEntryDialogComponent {
+  clicked: boolean;
+  term: string;
+  path: string;
+  fromDockstoreYml: boolean;
+
+  constructor(
+    public dialogRef: MatDialogRef<ArchiveEntryDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public data: { entry: Entry; callback: (archived: Entry) => void },
```
coverbeck, line 37: Despite our attempts with the OpenAPI hierarchy, DockstoreTool and Workflow unfortunately do not extend Entry. If they did extend `Entry`, this would be the right way.

Since they don't, I'd suggest making this `entry: Workflow | DockstoreTool`. 

Rationale: if we ever introduce a new Entry that doesn't extend DockstoreTool or Workflow, unlikely though that may be, we would get a compilation error, which we would want, because line 46 would fail at runtime.

That said, you have to go through hoops to enforce type checking with MAT_DIALOG_DATA, which we don't do, so I guess optional.

coverbeck (approved):

src/app/entry/archive/dialog/archive-entry-dialog.component.html
```diff
@@ -0,0 +1,22 @@
+<h1 mat-dialog-title>Are you sure you want to archive this Dockstore {{ term }}?</h1>
+<div mat-dialog-content>
+  <app-alert></app-alert>
+  <p>
```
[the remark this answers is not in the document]
kathy-t: Hmm, I feel like manual topic would be good here. We also have a doc about migrating legacy tools to GitHub App tools and it suggests using the topic to indicate that the legacy tool is "archived" https://docs.dockstore.org/en/stable/getting-started/github-apps/migrating-tools-to-github-apps.html#archiving-or-deleting-your-existing-tool

That doc should probably be updated with this new feature :grin:

src/app/mytools/sidebar-accordion/sidebar-accordion.component.html
```diff
@@ -17,40 +17,32 @@
...
+      <ng-template #items let-items="items" let-itemState="itemState">
+        <mat-list class="org-accordion-list">
+          <mat-list-item *ngFor="let toolObj of items" [ngClass]="{ selectedTool: (toolId$ | async) === toolObj?.id }">
+            <mat-icon class="arrow mat-icon-sm">arrow_forward</mat-icon>
+            <a
+              class="no-wrap"
+              title="{{ toolObj?.name + (toolObj?.toolname ? '/' + toolObj?.toolname : '') }}"
+              [routerLink]="'/my-tools/' + toolObj.tool_path"
+            >
+              {{ toolObj?.name + (toolObj?.toolname ? '/' + toolObj?.toolname : '') }}
```
kathy-t, line 38: The de-duplication is great :grinning: if you want to take it one step further, I found this pipe today that does what this line and a couple line above does https://github.com/dockstore/dockstore-ui2/blob/365875d0ddd27c04f988185e84fe9e284ed30c3f/src/app/shared/entry-to-display-name.pipe.ts#L11

src/app/myworkflows/sidebar-accordion/sidebar-accordion.component.html
```diff
@@ -42,56 +42,40 @@
...
+              selectedTool: (isTool$ | async) && (workflowId$ | async) === workflowObj?.id
+            }"
+          >
+            <mat-icon class="arrow mat-icon-sm">arrow_forward</mat-icon>
+            <a
+              class="no-wrap"
+              title="{{ workflowObj?.repository + (workflowObj?.workflowName ? '/' + workflowObj?.workflowName : '') }}"
+              [routerLink]="'/my-' + (entryType$ | async) + 's/' + workflowObj.full_workflow_path"
+            >
+              {{ workflowObj?.repository + (workflowObj?.workflowName ? '/' + workflowObj?.workflowName : '') }}
```
kathy-t, line 69: Ditto about the pipe https://github.com/dockstore/dockstore-ui2/blob/365875d0ddd27c04f988185e84fe9e284ed30c3f/src/app/shared/entry-to-display-name.pipe.ts#L11

src/app/entry/archive/dialog/archive-entry-dialog.component.html
```diff
@@ -0,0 +1,22 @@
...
+    {{ term }}, it will be:
+  </p>
+
+  <ul>
+    <li>Marked as "Archived".</li>
+    <li>Read-only, except to unarchive.</li>
+    <li>Ranked lower in Dockstore Search.</li>
+    <li>Displayed in the "Archived" section of your Dockstore dashboard.</li>
+  </ul>
+  <p>Later, you can unarchive the {{ term }} to restore it to normal.</p>
```
kathy-t, line 16: Wording suggestion: You can unarchive the <term> to restore it to normal at any time

src/app/workflow/workflow.component.html
```diff
@@ -45,6 +45,7 @@ <h2 class="text-center">The workflow linked to could not be found!</h2>
   <div class="row m-1" *ngIf="!showRedirect" fxLayout.lt-sm="column" fxLayout.gt-sm="row">
     <app-json-ld [json]="schema"></app-json-ld>
     <div fxFlex="80">
+      <div *ngIf="workflow.archived"><span class="bubble mt-3">Archived</span></div>
```
kathy-t, line 48: I feel like we should go with a yellow banner similar to GitHub so it's easier to notice, especially for users who may be browsing public workflows

image

kathy-t (approved):
This feature could use a documentation ticket

src/app/entry/archive/dialog/archive-entry-dialog.component.ts
svonworl: I get your drift, the two approaches have different pros and cons, for sure.  Yes, `Workflow` and `DockstoreTool` don't formally derive to `Entry` right now, but we know they are derivatives.  The choice to use `Entry` was intentional on my part, so it's best IMHO, but you're the lead, so feel free to override... :)

As a compromise, I added some code to line 46 that uses the `Entry.gitUrl` as a fallback, so for a radically-different type of new entry, it's more likely to display something moderately useful, at least...

svonworl:
I addressed most of the feedback, the most visible changes were edits to the dialog text, and a yellow banner above an archived entry (the message text is pretty much the same as github's, it says the two important things.). The text in the banner is left-aligned because given the location of the prominent items on the entry page, there's probably more of a chance that the user will read it.  The date format is one of the several we use...

I've re-requested review.

Screenshots below:
image

image

coverbeck (approved):

kathy-t (approved):
Looks great!

PR #1855, opened 2023-10-21
Use the descriptorLanguage pipe to handle Galaxy on the versions tab

coverbeck:
**Description**
For Galaxy workflows, display `Galaxy` instead of `gxformat2` in the Versions tab.

TLDR; Use the pipe built just for this.

Longer explanation -- we weren't actually displaying the language version twice; we were displaying the language followed by the language version. For example, for WDL versions it displays something like `WDL draft-2`, where `WDL` is the language and  `draft-2` is the language version.

For Galaxy we made a mistake in having the language be `gxformat2` (not to be confused with the language version); the language is Galaxy. But it got baked in, so we use the pipe to display it as Galaxy, but missed this place.

**Review Instructions**
1. Go to a Galaxy workflow
2. Select the Versions table
3. Ensure that it the language versions say `Galaxy gxformat1` instead of `gxformat2 gxformat1`. Unless you pick the one Galaxy workflow currently in our system that is gxformat2, in which case it should say `Galaxy gxformat2`.

**Issue**
dockstore/dockstore#5701

**Security**

### DIFF
--- a/cypress/e2e/group3/deletion.ts
+++ b/cypress/e2e/group3/deletion.ts
@@ -23,12 +23,6 @@ type Entry = {
   myPrefix: string;
 };
 
-const entries: Entry[] = [
-  { table: 'workflow', id: 11, versionId: 14, myPrefix: 'my-workflows', path: 'github.com/A/l' },
-  { table: 'apptool', id: 50, versionId: 1000, myPrefix: 'my-tools', path: 'github.com/C/test-github-app-tools/testing' },
-  { table: 'notebook', id: 1000, versionId: 1000, myPrefix: 'my-notebooks', path: 'github.com/dockstore-testing/simple-notebook' },
-];
-
 function unpublicize(entry: Entry): void {
   invokeSql(`update ${entry.table} set ispublished = false, waseverpublic = false where id = ${entry.id}`);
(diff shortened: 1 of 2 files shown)


kathy-t (approved):

coverbeck:
Restructured deletion tests à la Steve's #1857 to hopefully fix integration test that was failing. Worked once. Re-requesting reviews.

kathy-t (approved):

cypress/e2e/group3/deletion.ts
```diff
@@ -13,75 +13,99 @@
...
+});
+
+describe('Entry Deletion 1', () => {
+  init();
+  deletionTests(entries[1]);
+});
+
+describe('Entry Deletion 2', () => {
+  init();
+  deletionTests(entries[2]);
```
svonworl, line 110: Since we're not iterating through `entries` anymore, we could break out each `Entry` into its own variable (`workflow`, `notebook`, etc) and describe the tests likewise.

```
const notebook = {... }
// ...
describe('Notebook Deletion'), () => {
  init();
  deletionTests(notebook);
}
```

svonworl (approved):

PR #1857, opened 2023-10-23
DOCK-2471: Fix flaky UI integration tests

svonworl:
**Description**
This PR fixes some of the flakiness in the CircleCI UI2 tests, specifically related to failures due to problems in the "my tools" interface.   Since it was related, the fix to https://ucsc-cgl.atlassian.net/browse/SEAB-5970 is included in this PR.

After I fixed the "my tools" issues, the CircleCI runs started consistently failing because of a browser crash in `myworkflows.ts`.  This was weird, because I didn't change anything that should have made it more frequent.  However, it needed to be fixed, and typically, per the internets, the crashes happen because the browser runs out of memory.  One of the `describes` in `myworkflows.ts` was huge, so I split it into pieces, in the hope that maybe it would reduce the memory load somewhere in the CircleCI/Chromium internals (maybe in the recording logic?).   Now, the tests seem to pass, although it's very possible I didn't address the root issue and it is not actually fixed.  Potentially Heisenbuggy.  Let's run it this way for a bit, and if the crashes recur, we can dig deeper...

So, in total, there are four significant changes:
1. In  `MyToolComponent`, we now wait for both the "my" tools and "my" apptools responses to arrive before processing them.  `combineLatest` produces a series of values like so `[(null, null), (A, null), (A, B)]`, and thus, previously, to display the entry denoted by current URL, the code was selecting/requesting an entry from the "my" entries based on incomplete information (from the list of old-style tools or apptools, whichever was updated in the combined observable, but not both), then another entry based on full information (the old-style tools + apptools).  My best theory is that, occasionally, the response for the wrong entry arrived after the response for the correct entry, causing the wrong entry to be displayed, which triggered the test failures.
2. We now observe `workflow$` and set the template flag that causes delegation to the apptool template.
3. We now delegate correctly to the workflow template from the tools template.
4. Split large `describe` block in  `myworkflows.ts`.

**Review Instructions**
Check the CircleCI runs and confirm that the UI2 tests are not failing due to problems in the "my tools" interface  Primarily, this will be failures in `mytools.ts`, `deletion.ts`, and `archival.ts` (forthcoming).  Also, confirm that `myworkflows.ts` is consistently running fine.

**Issues**
https://ucsc-cgl.atlassian.net/browse/DOCK-2471
#5696
https://ucsc-cgl.atlassian.net/browse/SEAB-5970

**Security**

### DIFF
--- a/cypress/e2e/group3/deletion.ts
+++ b/cypress/e2e/group3/deletion.ts
@@ -43,7 +43,7 @@ describe('Entry Deletion', () => {
   function goToPrivatePage(entry: Entry): void {
     cy.visit('/');
     cy.visit(`${entry.myPrefix}/${entry.path}`);
-    cy.wait(2000);
+    cy.contains(entry.path);
   }
 
   it('Should not be able to delete an entry that is published', () => {
@@ -81,8 +81,6 @@ describe('Entry Deletion', () => {
       cy.contains('button', 'Delete').should('be.visible').should('not.be.disabled').click();
       cy.get('[data-cy=delete-yes]').should('be.visible').should('not.be.disabled').click();
       cy.wait(1000);
-      cy.contains(entry.path).should('not.exist');
-      goToPrivatePage(entry);
       cy.contains(entry.path).should('not.exist');
     });
   });
--- a/src/app/container/container.component.html
+++ b/src/app/container/container.component.html
@@ -13,7 +13,8 @@
   ~    See the License for the specific language governing permissions and
   ~    limitations under the License.
   -->
-<div *ngIf="tool || !displayAppTool; else appTool">
+<app-workflow *ngIf="displayAppTool" [isWorkflowPublic]="isToolPublic"></app-workflow>
+<div *ngIf="tool && !displayAppTool">
   <div
     class="row m-1"
     *ngIf="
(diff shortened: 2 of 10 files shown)


src/app/container/container.component.html
```diff
@@ -13,7 +13,8 @@
   ~    See the License for the specific language governing permissions and
   ~    limitations under the License.
   -->
-<div *ngIf="tool || !displayAppTool; else appTool">
+<app-workflow *ngIf="displayAppTool" [isWorkflowPublic]="isToolPublic"></app-workflow>
+<div *ngIf="tool && !displayAppTool">
```
svonworl, line 17: The previous logic that controlled whether to use the tool template (for old-style tools) or the workflow template) for apptools) was incorrect.  It's fixed here, and I moved the code that conditionally delelegates to the workflow template to the top, to make sure it's clear what's happening without having to scan the entire template.

kathy-t (approved):

cypress/e2e/group3/deletion.ts
```diff
@@ -43,7 +43,7 @@ describe('Entry Deletion', () => {
   function goToPrivatePage(entry: Entry): void {
     cy.visit('/');
     cy.visit(`${entry.myPrefix}/${entry.path}`);
-    cy.wait(2000);
+    cy.contains(entry.path);
```
coverbeck, line 46: 👍 This is indeed better than a hard-coded wait.

coverbeck (approved):

hyunnaye (approved):

PR #1859, opened 2023-10-24
Disable manual topic editing in the UI for GitHub App entries

kathy-t:
**Description**
This PR goes with https://github.com/dockstore/dockstore/pull/5713 and disables manual topic editing in the UI for GitHub app entries since they should be updating it in the .dockstore.yml instead. Topic selection is left as is, so the user can still change the selection for a GitHub app entry.

Also updated some tooltips for the topics.

**Review Instructions**
In addition to the instructions in https://github.com/dockstore/dockstore/pull/5713, verify that the `Edit` button for editing manual topics is disabled for GitHub App entries.

**Issue**
dockstore/dockstore#5625
[DOCK-2444](https://ucsc-cgl.atlassian.net/browse/DOCK-2444)

**Security**

[DOCK-2444]: https://ucsc-cgl.atlassian.net/browse/DOCK-2444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -261,8 +261,12 @@
                     [disabled]="workflow?.mode === WorkflowType.ModeEnum.DOCKSTOREYML"
                     [matTooltip]="
                       workflow?.mode === WorkflowType.ModeEnum.DOCKSTOREYML
-                        ? 'Edit the manual topic for this ' + (entryType$ | async) + ' in the .dockstore.yml file on GitHub'
-                        : ''
+                        ? 'Edit the manual topic for this ' +
+                          (entryType$ | async) +
+                          ' in the .dockstore.yml file on GitHub. Editing the topic in the UI is deprecated for GitHub App ' +
+                          (entryType$ | async) +
+                          's.'
+                        : 'Edit the manual topic for this ' + (entryType$ | async)
                     "
                   >
                     <mat-icon *ngIf="topicEditing" class="mat-icon-sm">save</mat-icon>
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -198,11 +198,25 @@
                 }}</span>
               </div>
               <div *ngIf="!isPublic">
-                <span><strong matTooltip="Short description of the workflow">Topic Automatic</strong>:&nbsp;</span>
+                <span
+                  ><strong
+                    matTooltip="Short description of the {{
+                      entryType$ | async
+                    }}, retrieved automatically from the GitHub repository description"
+                    >Topic Automatic</strong
+                  >:&nbsp;</span
+                >
                 <span>{{ workflow?.topicAutomatic }}</span>
               </div>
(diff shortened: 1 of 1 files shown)


src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -198,11 +198,25 @@
...
+                  >:&nbsp;</span
+                >
                 <span>{{ workflow?.topicAutomatic }}</span>
               </div>
               <form *ngIf="!isPublic" #editTopic="ngForm" fxLayout="row" fxLayoutAlign="start center" class="w-100">
-                <span><strong matTooltip="Short description of the workflow">Topic Manual</strong>:&nbsp;</span>
+                <span
+                  ><strong
+                    matTooltip="Short description of the {{ entryType$ | async }}, entered manually in the {{
+                      workflow?.mode === WorkflowType.ModeEnum.DOCKSTOREYML ? '.dockstore.yml' : 'UI'
```
svonworl, line 215: This message might be a little confusing for `.dockstore.yml` users that have previous entered a topic in the UI, because it'll tell them that it was entered in the `.dockstore.yml`.

kathy-t: Interesting you bring that up, I suppose this feature should come with some type of notification to .dockstore.yml users that they should modify their .dockstore.yml to include `topic`s that they might have added in the UI. The webservice PR https://github.com/dockstore/dockstore/pull/5713 won't overwrite existing `topicManual` if they don't do this so we won't lose it, but they should know about it

coverbeck: FWIW, I find tooltips are good to have, but half the users don't know to look for them (I often forget myself on other websites), and they shouldn't be relied as the sole way of conveying key info. Could you convey the info some other way as well? Ideas:

* Next to/below the Topic Manual label, have some text saying the text must be edited in the .dockstore.yml.
* When the user clicks the Edit button, a dialog pops saying edit the .dockstore.yml. I'm not 100% wild about this, but this is a brainstorm.

These are suggestions, not blocking, and it could stay the way it is as I suspect most users won't run into this.

coverbeck (approved):

src/app/workflow/info-tab/info-tab.component.html
kathy-t: > FWIW, I find tooltips are good to have, but half the users don't know to look for them (I often forget myself on other websites), and they shouldn't be relied as the sole way of conveying key info

I lean towards keeping it as is because this pattern exists for the other fields on the info tab. If we decide to re-design how to display tooltip info, we can migrate them all at once so the UI is consistent

kathy-t: > This message might be a little confusing for .dockstore.yml users that have previous entered a topic in the UI, because it'll tell them that it was entered in the .dockstore.yml.

I specified in the disabled Edit button for GitHub app entries that editing in the UI is deprecated

PR #1869, opened 2023-11-07
GitHub Apps Logs Pagination - UI

hyunnaye:
**Description**
This PR updates the UI to reflect the webservice changes to the pagination and filtering of github apps logs. Now, the total count of the events will show the total amount of events existing for that user/organization. However, flipping through the pagination or sorting will call to the endpoint with its respective offset, limit, and filter values.

View of apps logs (with results)
Screenshot from 2023-11-07 16-07-41

View of apps logs (no results after filtering)
Screenshot from 2023-11-07 16-07-54

When there are no apps logs for that repo
Screenshot from 2023-11-15 15-05-39

**Review Instructions**
Go to your workflows in qa and test out the github apps logs component. Change the sorting, test some filters and verify it works.

Also, go to http://qa.dockstore.org/users/hyunnaye and do the same to the user's apps logs. Note you will need to be logged in as an admin or curator.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2440

**Security**


src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -66,20 +85,129 @@ export class GithubAppsLogsComponent implements OnInit {
...
+      this.loadEvent(this.matDialogData, this.paginator.pageIndex, this.paginator.pageSize, null, null, null);
+      this.paginatorService.setPaginator(this.type, this.paginator.pageSize, this.paginator.pageIndex);
+
+      // Handle paginator changes
+      merge(this.paginator.page)
+        .pipe(distinctUntilChanged())
+        .subscribe(() => {
+          this.loadEvent(
+            this.matDialogData,
+            this.paginator.pageIndex * this.paginator.pageSize,
```
svonworl, line 110: This code appears to use `pageIndex` in two different ways.  At line 101, we call `loadEvent` with the raw `pageIndex` value, and here, we multiply `pageIndex` by the `pageSize`.  It should probably be consistently one or the other, and if we used the db code as a guide, an "index" would be an offset from 0, measured in the number of results.

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -88,23 +216,20 @@ export class GithubAppsLogsComponent implements OnInit {
...
+      this.showContent = 'table';
     }
   }
-
-  applyFilter(event: string) {
-    this.dataSource.filter = event.trim().toLowerCase();
-    if (this.dataSource.paginator) {
-      this.dataSource.paginator.firstPage();
-    }
+  ngOnDestroy() {
```
coverbeck, line 231: Have your class `extends Base`, then you don't need this method.

coverbeck (commented):
Works well when I tested it locally!

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
svonworl: Yeah, sorry, I think I misinterpreted the code, the `pageIndex` is coming from `MatPaginator`, which [I assume] defines the page index as the number of pages of items.  Would still recommend that in the first `loadEvents` call, if you use `pageIndex` rather than zero, you multiply it by `pageSize` as you do here, just in case it's not zero for some reason.

svonworl (approved):

cypress/e2e/group2/myworkflows.ts
```diff
@@ -113,6 +116,32 @@ describe('Dockstore my workflows', () => {
       // cy.contains('2020-02-19T18:20');
       // cy.contains('2020-06-05T07:40');
       cy.contains('1 – 2 of 2');
+
+      //Filtering
+      const filteredResponse = [
```
coverbeck, line 121: Missed this in my first pass, but you should explicitly declare the type here: `const filteredResponse: LambdaEvent[] = [`.

That way, if we ever change the Lambda event schema in the web service, this code will fail to compile, which will be good, because it will force us to keep the mock in sync with the web service schema.

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -66,20 +86,109 @@ export class GithubAppsLogsComponent implements OnInit {
...
-      ? this.lambdaEventsService.getUserLambdaEvents(this.matDialogData.userId)
-      : this.lambdaEventsService.getLambdaEventsByOrganization(this.matDialogData.organization);
+  }
+
+  ngAfterViewInit() {
+    this.loadAppsLogs();
+  }
+
+  loadAppsLogs() {
+    setTimeout(() => {
```
coverbeck, line 96: Do you need this, and if so why? If you do need it, a code comment would be good.

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -148,34 +141,22 @@ export class GithubAppsLogsComponent implements OnInit, AfterViewInit {
...
-    sortDirection: string,
-    sortCol: string
-  ) {
-    const filtered = filter !== null && filter !== undefined && filter.length > 0;
+  loadEvents(pageIndex: number, pageSize: number, filter: string, sortDirection: string, sortCol: string) {
+    const filtered = filter?.length > 0;
     let lambdaEvents: Observable<HttpResponse<LambdaEvent[]>>;
-    if (matDialogData.userId) {
+    this.loading = true;
+    this.dataSource.data = [];
```
coverbeck, line 158: Lines 157 and 158 now cause a big flashing when paging through the dialog.

Sorry, I know I suggested you might need to move `loading`, but it's causing the bulk of the component to disappear every time you page. So maybe you should do it only once after all, although you won't have an indicator as you're paging.

Line 158 is emptying the table before it's fetched the new content, so you see the table contents shrink then grow. You might only want to set to an empty array if the request fails.

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
hyunnaye: I thought I needed it because I was using https://github.com/dockstore/dockstore-ui2/blob/62fa718933b3eb852602cb10defab94c503faeb7/src/app/shared/tool-lister.ts#L68 to help create this ticket but when I tested it without it, it looks to work the same!

denis-yuen (approved):
With notes/changes from others, glad to see this coming together!

kathy-t (commented):
Looks great! I did notice one thing when testing locally though. I'm using a database with prod data, so none of the lambda events have entry name set (they are all `null`).

Sorting by entry name results in no events, even though there are events.

image

image

coverbeck (approved):

hyunnaye:
> Looks great! I did notice one thing when testing locally though. I'm using a database with prod data, so none of the lambda events have entry name set (they are all `null`).
> 
> Sorting by entry name results in no events, even though there are events.
> 
> image
> 
> image

this is fixed with the webservice PR: https://github.com/dockstore/dockstore/pull/5749

PR #1871, opened 2023-11-09
DOCK-2373: Fix inline markdown images

svonworl:
**Description**
This PR modifies the UI's markdown service to generate correct HTML for relative inline images when the `baseUrl` references github, thus displaying them properly in an entry's description, and in a notebook's "Preview". 

Previously, Dockstore displayed relative inline markdown images as broken image icons, for example, see the broken flowchart images midway down the description here:
https://dockstore.org/workflows/github.com/aofarrel/myco/myco_sra:main?tab=info

The UI calculated a base URL, which it passed to the markdown converter, which prepended it to the relative generated `img` tag urls.  Alas, our base URL (and the URL constructed from it) referenced the github resource embedded _within a github page_.  So, when browser retrieved the image from the URL, it got back a blob of github HTML, rather than the expected image, causing it to barf.  You can't fix the problem simply by changing the base URL we pass to the markdown converter to reference the raw resource, because then the relative outbound links don't reference the github pages anymore.

As it so happens, if you add the query `raw=true` to the url for a github page that references a resource, it redirects to the raw resource.  For example:
https://github.com/aofarrel/myco/blob/main/doc/doc_making_resources/myco_sra_flowchart_1.png (github page)
https://github.com/aofarrel/myco/blob/main/doc/doc_making_resources/myco_sra_flowchart_1.png?raw=true (raw resource)

So, this PR simply adds `raw=true` to all `img` urls that reference github.

The conversion step could happen either before (on the markdown) or after (on the html) the markdown conversion.  I picked after, so that if we ever rewrite relative `img` tags, which are legal in markdown, those would work, too.

A more sophisticated solution would probably generate the appropriate URL for both the outbound link (`a`) and embedded image (`img`) cases, sans redirects, for all of our supported source code repositories, but...

This is a quick hack that probably works well enough for now.

**Review Instructions**
Create a version of an entry that has a README that contains an inline markdown image (the syntax is `description`) with a relative link that points to an image in the repo.  Register the version on Dockstore, view the entry's info page, and confirm that the image is displayed in the description. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2373

**Security**
We are manipulating HTML which is then displayed on our site.  Possibly, there could be security implications, but we run the results through multiple sanitizers, so it's probably no more of an issue than HTML that users embed in their markdown.

### DIFF
--- a/src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
+++ b/src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
@@ -115,7 +115,7 @@ export class MarkdownWrapperService {
    * Add the query 'raw=true' to all img tag src urls that refer to github, causing github to redirect to the raw image.
    */
   makeGitHubImagesRaw(html: string): string {
-    return html.replace(/(<img.*? src=")(.*?)(".*?>)/gm, (all, before, url, after) => {
+    return html.replace(/(<img[^>]*? src=")([^"]*?)("[^>]*?>)/gm, (all, before, url, after) => {
       if (url.startsWith('https://github.com/')) {
         const separator = url.includes('?') ? '&' : '?';
         return `${before}${url}${separator}raw=true${after}`;
--- a/src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
+++ b/src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
@@ -88,7 +88,8 @@ export class MarkdownWrapperService {
   customCompile(data, baseUrl): string {
     const parseOptions = { markedOptions: { baseUrl: baseUrl } };
     const markdownData = this.removeTabsFromTableHeaders(data);
-    return this.markdownService.parse(markdownData, parseOptions);
+    const html = this.markdownService.parse(markdownData, parseOptions);
+    return this.makeGitHubImagesRaw(html);
   }
 
   customSanitize(html): string {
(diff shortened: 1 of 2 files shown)


src/app/shared/markdown-wrapper/markdown-wrapper.service.ts
```diff
@@ -109,4 +110,18 @@ export class MarkdownWrapperService {
   removeTabsFromTableHeaders(data: string): string {
     return data.replace(/(^ {0,3}\|.*)/gm, (match) => match.replace(/\t/g, '    '));
   }
+
+  /**
+   * Add the query 'raw=true' to all img href urls that refer to github, causing github to redirect to the raw image.
+   */
+  makeGitHubImagesRaw(html: string): string {
```
coverbeck, line 117: This would be a good candidate for some unit tests, like there are for `removeTabsFromTableHeaders`, especially since regular expressions are hard, for me at least, to figure out the details of what they're doing without going back to a reference manual.

coverbeck (commented):
The approach makes sense; as mentioned in an inline comment, it would be really good to have some unit tests.

kathy-t (commented):
Approach makes sense. 

Just throwing it out there that another option is to override the existing `image` renderer, which would remove the need for you to look for the `img` tags yourself, but it's up to you if you want to go down that route https://marked.js.org/using_pro#renderer.

svonworl:
> Approach makes sense.
> 
> Just throwing it out there that another option is to override the existing `image` renderer, which would remove the need for you to look for the `img` tags yourself, but it's up to you if you want to go down that route https://marked.js.org/using_pro#renderer.

Overriding the renderer is a good idea, but alas, marked only seems to allow _completely_ overriding the renderer for a given type of element (or sometimes, to either override or completely fall back).  What I'd like to do is take the _output_ from the default `image` renderer and filter _that_, so that marked could do the heaving lifting (attach the base url, escape the fields, etc) and we'd just swoop in at the end and tweak the urls as needed.  Alas, I haven't been able to figure out a way to do that using the marked configuration system, and so, this other approach...

david4096 (approved):

kathy-t (approved):

coverbeck (approved):

hyunnaye (approved):

PR #1874, opened 2023-11-21
SEAB-6021: Fix stuck profile page

svonworl:
**Description**
This PR fixes the UI so that when the profile page switches from user X's to user Y's profile page, user Y's information is displayed.  Previously, the UI was broken, and would get "stuck" on user X's page.  The URL would change, but the information on the profile page did not.

The fix was to the `UserPageComponent` and `RecentEventsComponent` components, which comprise most of the profile page, and previously both read the username from the route upon construction.  Now, they observe the route parameters and update the user when they change.  

Both components were lightly refactored to make their internals more similar to each other.  Also, I deleted the unused property `username`, please inspect and confirm that it is unused. 

I kinda feel like both components should be embedded in a "profile page" component, which passes the username to them as an Input.  But that's a substantial [philosophical] difference and a big change for "right before the release" so I left it the way it is.

**Review Instructions**
Log into qa, navigate to `qa.dockstore.org/users/svonworl`, and use the dropdown menu at upper right to see your profile.  Confirm that all information on the profile page changes from svonworl's to yours.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6021

**Security**
No concerns.

### DIFF
--- a/cypress/e2e/immutableDatabaseTests/dropdown.ts
+++ b/cypress/e2e/immutableDatabaseTests/dropdown.ts
@@ -58,6 +58,22 @@ describe('Dropdown test', () => {
     });
   });
 
+  describe('Go to profile page from another profile page', () => {
+    it('Should show user correct profile', () => {
+      const fromUser = 'potato';
+      const currentUser = 'user_curator';
+      cy.visit(`/users/${fromUser}`);
+      cy.contains('Activity');
+      cy.get('app-user-page').contains(fromUser);
+      cy.get('app-user-page').contains(currentUser).should('not.exist');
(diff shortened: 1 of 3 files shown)


kathy-t (approved):

coverbeck (approved):

denis-yuen:
> I kinda feel like both components should be embedded in a "profile page" component, which passes the username to them as an Input. But that's a substantial [philosophical] difference and a big change for "right before the release" so I left it the way it is.

:+1:

denis-yuen (approved):

PR #1875, opened 2023-11-21
DOCK-2487: Generalize recentEvents pipe

svonworl:
**Description**
This PR modifies the `recentEvents` pipe to use the referenced entry's type metadata so that it works for all entry types.  Previously, it malfunctioned when passed an event that referenced a notebook, service, or apptool. 

The PR adds some opinionated (in that they prescribe the underlying implementation in some cases) unit tests that exercise the new code.  Lots of duck typing, quaaack.

**Review Instructions**
On qa, register and publish a notebook, service, or apptool.  Navigate to your profile, and confirm that the entries in the "Recent Events" section are formatted and linked correctly. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2487
dockstore/dockstore#5735

**Security**
No concerns.

### DIFF
--- a/src/app/shared/entry/recent-events.pipe.ts
+++ b/src/app/shared/entry/recent-events.pipe.ts
@@ -1,12 +1,11 @@
 import { Inject, Pipe, PipeTransform } from '@angular/core';
-import { Event } from 'app/shared/openapi';
+import { Event, DockstoreTool, Workflow } from 'app/shared/openapi';
 import { EntryToDisplayNamePipe } from '../entry-to-display-name.pipe';
 import { EntryType } from '../../shared/enum/entry-type';
 
 @Pipe({
   name: 'recentEvents',
 })
-// TODO: Accommodate for notebooks and services when we can retrieve those events
 export class RecentEventsPipe implements PipeTransform {
   private EntryType = EntryType;
   constructor(@Inject(EntryToDisplayNamePipe) private entryToDisplayNamePipe: EntryToDisplayNamePipe) {}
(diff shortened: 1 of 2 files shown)


src/app/shared/entry/recent-events.pipe.ts
```diff
@@ -25,27 +24,15 @@ export class RecentEventsPipe implements PipeTransform {
...
-        if (event.workflow) {
-          return '/workflows/' + event.workflow.full_workflow_path;
-        } else if (event.tool) {
-          return '/containers/' + event.tool.tool_path;
-        } else if (event.apptool) {
-          return '/containers/' + event.apptool.full_workflow_path;
-        }
-        break;
+        const entry = this.getEntry(event);
+        return entry && '/' + entry.entryTypeMetadata.sitePath + '/' + this.getPath(entry);
```
svonworl, line 32: Our autoformatter removed the parentheses around the subexpression on the right side of the `&&`, it was better with them.

src/app/shared/entry/recent-events.pipe.ts
```diff
@@ -25,27 +24,15 @@ export class RecentEventsPipe implements PipeTransform {
...
       }
       case 'entryType': {
-        return event.tool || event.apptool ? this.EntryType.Tool : this.EntryType.BioWorkflow;
+        return this.getEntry(event)?.entryTypeMetadata.term ?? 'entry';
```
svonworl, line 35: This code will return "entry" in the error case, when the passed event didn't reference an entry.  I know we don't like exposing the term "entry", but the user shouldn't see it if everything is working right, and, IMHO, it's better than a blank space.

src/app/shared/entry/recent-events.pipe.ts
```diff
@@ -25,27 +24,15 @@ export class RecentEventsPipe implements PipeTransform {
...
-        if (event.workflow) {
-          return this.entryToDisplayNamePipe.transform(event.workflow);
-        } else if (event.tool) {
-          return this.entryToDisplayNamePipe.transform(event.tool);
-        } else if (event.apptool) {
-          return this.entryToDisplayNamePipe.transform(event.apptool);
-        }
-        break;
+        const entry = this.getEntry(event);
+        return entry && this.entryToDisplayNamePipe.transform(entry);
```
svonworl, line 28: Admittedly, the `&&` syntax is terse, but more compact than the equivalent `if`.

src/app/shared/entry/recent-events.pipe.ts
svonworl: Maybe I should change the right subexpression to a template literal.

kathy-t (approved):

coverbeck (approved):

denis-yuen (approved):

PR #1878, opened 2023-11-22
SEAB-6025: Fix alias resolution for AppTools, Notebooks, and Services

svonworl:
**Description**
Previously, the UI's alias resolution code would fail when attempting to redirect to an aliased AppTool, Notebook, or Service.  This PR includes fixes for all current Entry types, engineered to also automagically work when we add the next Entry type, as well as a medium-intensity refactor of the surrounding code.  

Key changes:
1. The previously-separate code paths that handle tools and workflows have been unified, and now use a new webservice endpoint that returns the `Entry` corresponding to an specified alias.  The new endpoint is implemented in this PR: https://github.com/dockstore/dockstore/pull/5763
2. The code that constructs the target URL, both for entries and versions, now uses the entry type metadata from the endpoint responses to determine where to redirect to.  To support this scheme, we also added entry type/metadata information to the `WorkflowVersionPathInfo` class in the webservice PR.
3. The aliases component has been refactored to remove duplication and improve clarity.  The "heavy lifting" now happens in  the `nagivateTo` method, which subscribes to the specified observable and handles values that emerge by optionally filtering, and then computing the target path and changing the route to it. 
4. The "Redirecting" messages have been simplified to a single generic "Redirecting..." message.   The motivation was that this message only appears to the user for a split-second, after the UI changes the route and before the UI elements change in response.  So, the tradeoff is fleeting message detail versus code complexity, simplifying allowed us to delete a substantial amount of code.

**Review Instructions**
Alias a notebook via the API, then attempt to access the notebook via the alias interface, by entering a URL like `qa.dockstore.org/aliases/notebooks/<notebook-alias>` into the address bar of your browser, and confirm that the ui "redirects" to it.  Then, alias a version of the notebook, and access it similarly, via URL like `qa.dockstore.org/aliases/notebook-versions/<version-alias>`.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6025

**Security**
No concerns.


coverbeck (approved):

denis-yuen:
> 1. The previously-separate code paths that handle tools and workflows have been unified, and now use a new webservice endpoint that returns the `Entry` corresponding to an specified alias.  The new endpoint code is on this branch, I'll convert to a PR if we decide to keep it: https://github.com/dockstore/dockstore/tree/feature/seab-6025/fix-aliases-for-apptools-notebooks-services

Nice!

> 2\. Does anyone know why `WorkflowVersionPathInfo` wasn't designed to return a full  `Workflow` and `WorkflowVersion`?

Looks like introduced in https://github.com/dockstore/dockstore/commit/afbbc2447e3b22bd704c7d883e018413a3bd9a8c 
Thinking probably was  to return the minimum necessary at the time

src/app/shared/state/entry-wizard.service.ts
```diff
@@ -117,7 +117,7 @@ export class EntryWizardService {
     this.entryWizardStore.setLoading(true);
     const registryEnum = this.convertSourceControlStringToEnum(repository.gitRegistry);
     this.workflowsService
-      .deleteWorkflow(registryEnum, repository.organization, repository.repositoryName)
+      .deleteWorkflow(registryEnum as 'github.com' | 'bitbucket.org' | 'gitlab.com', repository.organization, repository.repositoryName)
```
kathy-t, line 120: Curious, why did this change?

src/app/aliases/aliases.component.html
```diff
@@ -8,23 +8,8 @@
...
-    <div *ngIf="workflow$ | async as wf" fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="10px">
-      <mat-progress-bar mode="indeterminate"></mat-progress-bar>
-      Redirecting to the workflow <strong>{{ wf.full_workflow_path }}</strong>
-    </div>
-    <div *ngIf="tool$ | async as tool" fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="10px">
-      <mat-progress-bar mode="indeterminate"></mat-progress-bar>
-      Redirecting to the tool <strong>{{ tool.tool_path }}</strong>
-    </div>
-    <mat-card class="alert alert-warning mat-elevation-z" role="alert" *ngIf="types.includes(type) && (aliasNotFound$ | async)">
+    <div *ngIf="validType && found">Redirecting ...</div>
```
kathy-t, line 11: The above "Redirecting..." have a `mat-progress-bar`. Does this new one need it too?

kathy-t (approved):

src/app/aliases/aliases.component.html
svonworl: This might be an illusion caused by the diff, the `mat-card` actually belongs to the next block, the original "Redirecting" `divs` aren't in `mat-cards`.  (I could be wrong.)  The original file https://github.com/dockstore/dockstore-ui2/blob/develop/src/app/aliases/aliases.component.html

src/app/shared/state/entry-wizard.service.ts
svonworl: This was a fix that's since been merged to develop, it now matches.

package.json
```diff
@@ -5,8 +5,8 @@
   "config": {
     "webservice_version": "1.15.0",
     "use_circle": true,
-    "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/10673/workflows/e44fc131-9382-4f1f-b8b3-666468eb7384/jobs/39289/artifacts",
-    "circle_build_id": "39289"
+    "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/10696/workflows/4406a2db-e982-45c0-84be-ef925a4ac7b0/jobs/39419/artifacts",
+    "circle_build_id": "39419"
```
svonworl, line 9: I will point this at the develop artifacts after the companion webservice PR merges.

kathy-t (approved):

denis-yuen (approved):

PR #1880, opened 2023-11-27
SEAB-6024: Modify "Files" tabs, mostly for notebooks

svonworl:
**Description**
The original intention for this PR's backing ticket (https://ucsc-cgl.atlassian.net/browse/SEAB-6024) was to add a "Configuration Files" tab to the notebook entry page, which seemed to be conspicuously missing.   However, it turns out that the code already defined a "Configuration Files" tab (how quickly I forget lol) and it simply wasn't appearing, because the code didn't categorize `.dockstore.yml` or devcontainer files as notebook Configuration files, and the existing tab code was written to always display the first two files tabs and hide the rest if they were empty.  Previously, the notebook "Configuration File" tab was third, and thus did not appear if it had no files.

So, this PR changes the UI to:
* Include `.dockstore.yml` and devcontainer files in the set of notebook configuration files.
* Display all the files tabs, whether they're empty or not.
* Split out the notebooks "Other Files" tab.  These files might be code, but they could also be data, documentation, etc.

These changes are effectively a no-op for GitHub-app-registered workflows, since the `Configuration Files` tab always appears due to the presence of a `.dockstore.yml`.

**Review Instructions**
View a notebook on qa, and confirm that four file tabs appear, and that the `Configuration Files` tab appears and includes `.dockstore.yml`. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6024

**Security**
No concerns.

### DIFF
--- a/src/app/source-file-tabs/source-file-tabs.service.ts
+++ b/src/app/source-file-tabs/source-file-tabs.service.ts
@@ -43,6 +43,7 @@ export class SourceFileTabsService {
     const fileTabsSchematic =
       this.descriptorLanguageService.toolDescriptorTypeEnumToExtendedDescriptorLanguageBean(descriptorLanguage).fileTabs;
 
+    // Display all of the tabs, even if they are empty.
     fileTabsSchematic.forEach((fileTab) => {
       fileTabs.set(fileTab.tabName, []);
     });
--- a/src/app/source-file-tabs/source-file-tabs.service.ts
+++ b/src/app/source-file-tabs/source-file-tabs.service.ts
@@ -43,9 +43,9 @@ export class SourceFileTabsService {
     const fileTabsSchematic =
       this.descriptorLanguageService.toolDescriptorTypeEnumToExtendedDescriptorLanguageBean(descriptorLanguage).fileTabs;
 
-    // Always have the Descriptor Files Tab and Test Parameter Files tab
-    fileTabs.set(fileTabsSchematic[0].tabName, []);
-    fileTabs.set(fileTabsSchematic[1].tabName, []);
+    fileTabsSchematic.forEach((fileTab) => {
+      fileTabs.set(fileTab.tabName, []);
+    });
     if (!sourcefiles || sourcefiles.length === 0) {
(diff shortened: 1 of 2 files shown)


kathy-t (approved):

src/app/source-file-tabs/source-file-tabs.service.ts
```diff
@@ -43,9 +43,10 @@ export class SourceFileTabsService {
     const fileTabsSchematic =
       this.descriptorLanguageService.toolDescriptorTypeEnumToExtendedDescriptorLanguageBean(descriptorLanguage).fileTabs;
 
-    // Always have the Descriptor Files Tab and Test Parameter Files tab
-    fileTabs.set(fileTabsSchematic[0].tabName, []);
-    fileTabs.set(fileTabsSchematic[1].tabName, []);
+    // Display all of the tabs, even if they are empty.
```
coverbeck, line 46: Doesn't this mean there will always be displayed, for example, a configuration tab for legacy workflows and tools, and the tab will never have any files?

If so, what's the thinking? Not necessarily disagreeing at this point, just trying to understand.

svonworl: > Doesn't this mean there will always be displayed, for example, a configuration tab for legacy workflows and tools, and the tab will never have any files?
> 
> If so, what's the thinking? Not necessarily disagreeing at this point, just trying to understand.

Yes.  The rationale is that we got confused when the "Configuration File" tab was hidden, so might as well always display it.  A little more consistency, at the expense of a small increase in UI busyness.  I don't have strong feelings either way, but "display all of them always" also gets points from me because it has the simplest rules.

denis-yuen: Might be good to have a configuration tab for legacy anyway, maybe a follow-up ticket to advertise github apps integration there to further push folks from legacy to github apps

denis-yuen (approved):

coverbeck (approved):

PR #1888, opened 2023-12-05
SEAB-2126: Change registration/login services to handle non-http error

svonworl:
**Description**
This PR changes the UI so that it displays a reasonable/useful message when a login/registration attempt fails with a non-HTTP error, such as when the user closes the GitHub login window.  Previously, in such cases, the UI would display the following error message:

image

Yucky!

As it turns out, the observable which is handled by `LoginService` and `RegistrationService` can produce an error object which is not an `HttpErrorResponse`.  For example, when the login window is closed, an empty object is produced.  The existing code interpreted this empty object as an `HttpErrorResponse`, tried to use values of fields that didn't exist, and produced the  `undefined`s. 

The new code displays a generic failure message if the error object doesn't have a `status` field (and thus, is not an `HttpErrorResponse` object):

image

Otherwise, the new code works as before, displaying a formatted representation of the `HttpErrorResponse`.     The latter case happens primarily when someone tries to register a GitHub user with the same login as an existing Dockstore user, causing the webservice to respond with a 403.
 
**Review Instructions**
On qa, log out, then begin to log in again.  Close the Github login window and confirm that the UI displays a reasonable error message.  Then, begin to register a new user.  Close the Github login window and confirm that the UI displays a reasonable error message.  Then, register a new user with a GitHub login that matches an existing Dockstore login.  Confirm that the error message contains the webservice response message.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-2126

**Security**
This code tweaks the login/registration error handling, but shouldn't cause any issues.


src/app/login/login.service.ts
```diff
@@ -32,8 +32,12 @@ export class LoginService {
           observable.next(user);
           observable.complete();
         },
-        (error: HttpErrorResponse) => {
-          this.alertService.detailedError(error);
+        (error) => {
+          if (error.status !== undefined) {
```
coverbeck, line 36: Why not just `if (error.status)`? You're confident it will never be null?

src/app/login/login.service.ts
```diff
@@ -32,8 +32,12 @@ export class LoginService {
           observable.next(user);
           observable.complete();
         },
-        (error: HttpErrorResponse) => {
-          this.alertService.detailedError(error);
+        (error) => {
```
coverbeck, line 35: It would be nice to preserve the type, now it's being treated as `any` by the compiler: `(error: HttpErrorResponse | undefined)`.

src/app/register/register.service.ts
```diff
@@ -33,8 +33,12 @@ export class RegisterService {
           observable.next(user);
           observable.complete();
         },
-        (error: HttpErrorResponse) => {
-          this.alertService.detailedError(error);
+        (error) => {
```
coverbeck, line 36: Same comments as the ones in login.service.ts

src/app/login/login.service.ts
svonworl: We have code that checks if `error.status` is 0, `if (error.status)` would not trigger on it.
https://github.com/dockstore/dockstore-ui2/blob/9322ad39f1c37046b019a40c9933618b63b6e9ff/src/app/shared/alert/state/alert.service.ts#L91
The goal of the code is to figure out if the error object is an `HttpErrorResponse`, in which case `status` will very likely not be `null`.

src/app/login/login.service.ts
svonworl: The cause of the bug is that the `error` object is not necessarily an `HttpErrorResponse`, it can be some other type of object, for example the empty object `{}` that is produced when the user closes the login window, essentially whatever the underlying `ng2-ui-auth` decides to return.  So, IMHO, it's appropriate to not type the argument.

src/app/login/login.service.ts
svonworl: Postscript: what this code is trying to do is take one action if `error` is an `HttpErrorResponse`, and a different action otherwise.  Right now, it uses the presence of `status` as a proxy for `HttpErrorResponse`-ness, but maybe there's a better way to do it in Typescript?  Here's a stackoverflow thread on how to determine if an object has a certain property, the `in` solution of the first answer might be better... https://stackoverflow.com/questions/1894792/how-to-determine-whether-an-object-has-a-given-property-in-javascript

src/app/login/login.service.ts
```diff
@@ -32,8 +32,12 @@ export class LoginService {
           observable.next(user);
           observable.complete();
         },
-        (error: HttpErrorResponse) => {
-          this.alertService.detailedError(error);
+        (error: HttpErrorResponse | {}) => {
+          if ('status' in error) {
```
coverbeck, line 36: Suggest a comment here and in register.service.ts as to why you're doing this. We all know now why from the PR description, but it might not be obvious a year from now to somebody new looking at the code.

coverbeck (approved):

kathy-t (approved):

src/app/login/login.service.ts
svonworl: Added some comments.

denis-yuen (approved):

PR #1889, opened 2023-12-05
SEAB-6062: Fix topic editing irregularities

svonworl:
**Description**
This PR changes the manual topic editing code to make it work more consistently.

During a planning meeting demo, I noticed that I couldn't edit the manual topic of a notebook.  I could press the "Edit" button, which would bring up the topic editing text box, but the "Save" button was disabled.  So, we wrote this ticket.

We'd forgotten in ???, we disabled manual topic editing for `.dockstore.yml`-based entries.  Now, users are supposed to set the manual topic in `.dockstore.yml`.

However, the UI also wasn't working quite right, and Kathy confirmed that the correct behavior is for the manual topic "Edit" button should be disabled for all `.dockstore.yml`-based entries.  So, to accomplish the above, this PR fixes the relevant button code by consolidating two code paths. 

After this PR, the manual topic "Edit" button will:
* Always be disabled for `.dockstore.yml`-based entries.
* Always display the tooltip advising users of how to change the manual topic for `.dockstore.yml`-based entries. 
* Work as it did before, otherwise, allowing a user to change the manual topic for other types of entries via the UI.

I did some user testing on various types of workflows and notebooks, and it seems to work.  However, please build and test, there's lots of shifting state and special cases in this component, which makes it super easy to mess up...

**Review Instructions**
On qa:
1. Navigate to the info tab of a non-`.dockstore.yml`-based workflow that you own, and confirm you can successfully edit the manual topic.
2. Navigate to the info tab of a `.dockstore.yml`-based workflow that you own, and confirm that the manual topic "Edit" button is disabled and displays a tooltip that discusses how to change the manual topic.
3. Repeat step 2 for a notebook that you own.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6062

**Security**
No concerns.

### DIFF
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -255,7 +255,7 @@
                     mat-raised-button
                     class="small-mat-btn-skin accent-1-dark small-btn-structure mat-elevation-z"
                     (click)="toggleEditTopic()"
-                    data-cy="topicSaveButton"
+                    [attr.data-cy]="topicEditing ? 'topicSaveButton' : 'topicEditButton'"
                     [matTooltip]="
                       workflow?.mode === WorkflowType.ModeEnum.DOCKSTOREYML
                         ? 'Edit the manual topic for this ' +
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -243,21 +243,14 @@
                     <img src="../assets/svg/icons-reject-small.svg" alt="Cancel" class="pr-2 pb-1" />Cancel
                   </button>
                   <button
-                    *ngIf="!workflow?.topic && !topicEditing"
                     [disabled]="
-                      forumUrlEditing || workflowPathEditing || defaultTestFilePathEditing || (isRefreshing$ | async) || !canWrite
+                      forumUrlEditing ||
+                      workflowPathEditing ||
+                      defaultTestFilePathEditing ||
+                      (isRefreshing$ | async) ||
+                      !canWrite ||
+                      workflow?.mode === WorkflowType.ModeEnum.DOCKSTOREYML
                     "
-                    type="button"
-                    mat-raised-button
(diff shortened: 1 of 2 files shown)


denis-yuen:
> I did some user testing on various types of workflows and notebooks, and it seems to do the right thing. However, please build and test, there's lots of shifting state and special cases in this component, which makes it super easy to mess up...

:+1:

denis-yuen:
> However, the UI also wasn't working quite right, and Kathy confirmed that the correct behavior is for the manual topic "Edit" button to be disabled for all `.dockstore.yml`-based entries. To make it so, this PR adjusts the relevant button code by consolidating two code paths.

:+1:

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -243,26 +243,19 @@
...
-                    [disabled]="!canWrite || workflow?.mode === WorkflowType.ModeEnum.DOCKSTOREYML"
-                    type="button"
-                    mat-raised-button
-                    class="small-mat-btn-skin accent-1-dark small-btn-structure mat-elevation-z"
-                    (click)="toggleEditTopic()"
-                    data-cy="topicSaveButton"
+                    [attr.data-cy]="topicEditing ? 'topicSaveButton' : 'topicEditButton'"
                     [matTooltip]="
                       workflow?.mode === WorkflowType.ModeEnum.DOCKSTOREYML
                         ? 'Edit the manual topic for this ' +
```
coverbeck, line 261: I only noticed this because of your PR description, and it's arguably beyond the scope of the PR, but I will suggest anyway to optionally reverse the order of the two sentences. Instead of:

"Edit the manual topic for this entry in the .dockstore.yml file on GitHub. Editing the topic in the UI is deprecated for GitHub App entries."

I would have, stating the reason for the disabling first, then the remedy:

"Editing the topic in the UI is deprecated for GitHub App entries. Edit the manual topic for this entry in the .dockstore.yml file on GitHub."

coverbeck (approved):

denis-yuen (approved):

PR #1891, opened 2023-12-06
finish deprecation of author in favour of authors array

denis-yuen:
**Description**
Supports https://github.com/dockstore/dockstore/pull/5706
Adapts the ui to accept the new authors list

**Review Instructions**
 Build should pass mostly, bonus points for trying out UI pages with authors in QA.

**Issue**
https://github.com/dockstore/dockstore/issues/5467

**Security**


denis-yuen:
Supports https://github.com/dockstore/dockstore/pull/5706

coverbeck (commented):
I think you need to address `ExtendedDockstoreTool.email`. It's confusing and I don't get why it the way it is, but it looks like:

* The ExtendedDockstoreTool interface extends DockstoreTool, where DockstoreTool is the OpenApi generated client class
* For some reason, the interface defines anew `email`, even though it already exists in the base class. But ExtendedDockstoreTool gets its email assigned from DockstoreTool (from the web service).
* So [this code](https://github.com/dockstore/dockstore-ui2/blob/5f81733313bf59701302c9ada03469a03b82e27e/src/app/container/container.component.html#L90), with your websevice PR, will always evaluate to false, I believe.

denis-yuen:
> I think you need to address `ExtendedDockstoreTool.email`. It's confusing and I don't get why it the way it is, but it looks like:
> 
> * The ExtendedDockstoreTool interface extends DockstoreTool, where DockstoreTool is the OpenApi generated client class
> * For some reason, the interface defines anew `email`, even though it already exists in the base class. But ExtendedDockstoreTool gets its email assigned from DockstoreTool (from the web service).
> * So [this code](https://github.com/dockstore/dockstore-ui2/blob/5f81733313bf59701302c9ada03469a03b82e27e/src/app/container/container.component.html#L90), with your websevice PR, will always evaluate to false, I believe.

Good call, had to put in some extra checks as well.

kathy-t (commented):
Will the list tools/workflows page still work? I noticed that those pages still use `<entry>.author` which you removed in your webservice PR

- https://github.com/dockstore/dockstore-ui2/blob/d58322bf080649b630e63820f43c46a6bb00d9df/src/app/containers/list/list.component.html#L57
- https://github.com/dockstore/dockstore-ui2/blob/d58322bf080649b630e63820f43c46a6bb00d9df/src/app/workflows/list/list.component.html#L54

denis-yuen:
> Will the list tools/workflows page still work? I noticed that those pages still use `<entry>.author` which you removed in your webservice PR

Thanks!
A little annoying that the compiler doesn't catch those ones,.

package.json
```diff
@@ -3,8 +3,8 @@
   "version": "2.10.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "1.15.0",
-    "use_circle": true,
+    "webservice_version": "1.15.0-alpha.14-deprecateAuthor",
+    "use_circle": false,
     "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/10707/workflows/cbcf24c8-061a-4003-9d46-9de3b953b25f/jobs/39503/artifacts",
     "circle_build_id": "39503"
```
svonworl, line 9: Will need to be updated to the appropriate artifacts.

svonworl (approved):

denis-yuen:
Re-requesting due to https://github.com/dockstore/dockstore/pull/5706#issuecomment-1854677325 
i.e. this just relies on it being backwards compatible anyway

src/app/container/email.service.ts
```diff
@@ -76,7 +76,10 @@ export class EmailService {
    * @param tool The tool to contact author for
    */
   public composeContactAuthorEmail(tool: DockstoreTool): string {
-    const email = EmailService.getInquiryEmailMailTo(tool.email);
+    let email: string = null;
+    if (tool.authors && tool.authors.length) {
```
coverbeck, line 80: Getting into edge case territory, especially because I think we only have one tool with a private Docker image, but it doesn't look like all authors have emails, at least not in the DB. So ideally you'd check for the first author with an email, and if there aren't any not return anything.

But it's rare enough and was probably already like this, so it could be a followup ticket.

coverbeck (approved):

kathy-t (approved):

src/app/container/email.service.ts
denis-yuen: https://github.com/dockstore/dockstore/issues/5771

PR #1895, opened 2023-12-13
clear sort column on third click and preserve sort order when paging

denis-yuen:
**Description**
Supports https://github.com/dockstore/dockstore/pull/5769
Two complicating sources of confusion,

1) on a third click the UI would clear the direction but not the sort column. 
When the no order is specified but a sort column is specified, the backend will assume the order, so the events would not return to their original order (which is sorted by date, most recent first). 

2) When sorting a column like entryname, paging would lose the sort order

**Review Instructions**
1) In the github events pages, click three times to go between the two orders and the unsorted order to return to the original state (sorted by date)
2) Sort on entryname, click between pages, may need to open network tools to see if order is preserved and matches the UI header

**Issue**
https://github.com/dockstore/dockstore/pull/5769
https://ucsc-cgl.atlassian.net/browse/SEAB-6060

**Security**

### DIFF
--- a/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
+++ b/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
@@ -125,6 +125,10 @@ export class GithubAppsLogsComponent extends Base implements OnInit, AfterViewIn
           } else {
             this.sortCol = this.sort.active;
           }
+          if (this.sort.direction == '') {
+            this.sortCol = null;
+            this.sortDirection = null;
+          }
         }),
         takeUntil(this.ngUnsubscribe)
       )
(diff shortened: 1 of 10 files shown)


denis-yuen:
To be clear, this doesn't solve the problem but makes it more understandable what is going on

denis-yuen:
(noticed second confusing issue in the UI)

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -206,4 +222,22 @@ export class GithubAppsLogsComponent extends Base implements OnInit, AfterViewIn
       this.showContent = 'table';
     }
   }
+
+  mapSortOrder(rawSortOrder: string): 'asc' | 'desc' {
```
denis-yuen, line 226: I'm not entirely sure why this is needed 
Lifted from tool paging which must have had the same issue 
https://github.com/dockstore/dockstore-ui2/blob/develop/src/app/shared/tool-lister.ts#L112-L125

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -125,11 +125,21 @@ export class GithubAppsLogsComponent extends Base implements OnInit, AfterViewIn
           } else {
             this.sortCol = this.sort.active;
           }
+          if (this.sort.direction == '') {
```
kathy-t, line 128: `==` -> `===`

kathy-t (approved):
I think you linked the other PR as the issue instead of the ticket

coverbeck (approved):

PR #1898, opened 2023-12-14
Comment out failing WAF test due to rule change

david4096:
**Description**

A failing smoke test due to change in the WAF rules is being silenced. More investigation to see if we can turn it on in an upcoming sprint.

**Review Instructions**

Smoke tests pass

**Issue**

https://ucsc-cgl.atlassian.net/browse/SEAB-6084

**Security**

### DIFF
--- a/cypress/e2e/smokeTests/sharedTests/waf.ts
+++ b/cypress/e2e/smokeTests/sharedTests/waf.ts
@@ -54,13 +54,6 @@ describe('Test AWS Managed Common Rule Set', () => {
       expect(resp.status).to.eq(403);
     });
   });
-  // EC2MetaDataSSRF_URIPATH 	Inspects for attempts to exfiltrate Amazon EC2 metadata from the request URI path.
-  // it('Try a URL that SHOULD be blocked if EC2MetaDataSSRF_URIPATH were turned on', () => {
-  //   const uglyUrl = '/PARAM=127.0.0.1+-c+0%3B+cat+%2Fetc%2Fpasswd&DIAGNOSIS=PING';
-  //   cy.request({ url: uglyUrl, failOnStatusCode: false }).then((resp) => {
-  //     expect(resp.status).to.eq(403);
-  //   });
-  // });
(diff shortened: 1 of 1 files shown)


cypress/e2e/smokeTests/sharedTests/waf.ts
```diff
@@ -55,12 +55,12 @@ describe('Test AWS Managed Common Rule Set', () => {
     });
   });
   // EC2MetaDataSSRF_URIPATH 	Inspects for attempts to exfiltrate Amazon EC2 metadata from the request URI path.
-  it('Try a URL that SHOULD be blocked if EC2MetaDataSSRF_URIPATH were turned on', () => {
-    const uglyUrl = '/PARAM=127.0.0.1+-c+0%3B+cat+%2Fetc%2Fpasswd&DIAGNOSIS=PING';
-    cy.request({ url: uglyUrl, failOnStatusCode: false }).then((resp) => {
-      expect(resp.status).to.eq(403);
-    });
-  });
+  // it('Try a URL that SHOULD be blocked if EC2MetaDataSSRF_URIPATH were turned on', () => {
```
coverbeck, line 58: If we don't expect to restore the test, we should just delete it. If we think we might restore it, then a explanation/comment would be useful.

coverbeck (approved):

svonworl (approved):

kathy-t (approved):

cypress/e2e/smokeTests/sharedTests/waf.ts
david4096: I tried going through the logs to find any violations of this rule since it change and wasn't able to find any. That was how I constructed the test in the first place. For now, I'll just remove the test since it's not immediately clear how one would construct a violating request.

coverbeck (approved):
If Steve approves, please go ahead and merge (don't wait for Denis), so the nightly tests are passing again.

svonworl (approved):

PR #1916, opened 2024-01-25
SEAB-6188: Enable notebook versions to be hidden

svonworl:
**Description**
Prior to this PR, in the "Edit Notebook Version" dialog, the Git Reference and Notebook Path were displayed as form inputs, and the form validation was failing, because it was applying the regexp that it uses to check workflow paths, which didn't work on notebook paths ending with `.ipynb`.  Because the validation failed, the "Save Changes" button was disabled, thus preventing the only editable setting ("Hidden") from being submitted.

This PR displays a notebook's Git Reference and Notebook Path in `div`s, rather than `form` inputs, thus bypassing the validation, which is unnecessary because they're read only.  Thus, it enables the "Save Changes" button, allowing the user to hide a notebook. 

**Review Instructions**
On qa, select a version of one of your notebooks, and use the Actions menu to edit it and hide it.  Confirm it's hidden.  Then, unhide it.  Next, repeat the process on one of your workflow versions, to confirm that we didn't break anything.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6188

**Security**
No concerns.

### DIFF
--- a/cypress/e2e/group2/notebooks.ts
+++ b/cypress/e2e/group2/notebooks.ts
@@ -82,8 +82,8 @@ describe('Dockstore notebooks', () => {
     cy.get('[data-cy=versionRow]').contains(/jupyter/i);
     // Click on Info button and check content.
     cy.get('[data-cy=versionRow] button').click();
-    cy.get('input[name=reference]').should('have.value', 'simple-published-v1');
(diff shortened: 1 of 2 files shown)


coverbeck (approved):
Does it have the correct target branch? Fix version is 1.15.

kathy-t (approved):

denis-yuen (approved):

PR #1920, opened 2024-01-25
SEAB-6187: Adjust messaging for notebooks

svonworl:
**Description**
This PR adjusts the text of various UI components to mention and/or make sense for notebooks.  Specifically, it addresses the following issues:

* Notebook version “Info” dialog displays “Test Parameter File(s)”, should be “Test Files”.
* Advanced Search dialog box should include “notebooks” alongside “tools and workflows”.
* “Snapshot version” dialog refers to the notebook as a “workflow”.
* “Request DOI” dialog refers to the notebook as a “workflow”.
* “Export to ORCID” dialog refers to the notebook as a “workflow”.
* “Delete notebook” dialog refers to a “checked or checker workflow”, which doesn’t make sense and should be removed for notebooks.

For more details, see this doc: https://docs.google.com/document/d/1s3OU1kDcVwjPIOdgsmmaJTKGD2ZrojgZS5VYUt8njf8

**Review Instructions**
For each of the issues in the above list, navigate to the affected component and confirm that the fix was made correctly.  If this PR introduced a change that allowed the component's text to vary by entry type, also navigate to the same component in the context of a workflow, and confirm that nothing was broken. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6187

**Security**
No concerns.

### DIFF
--- a/src/app/entry/delete/dialog/delete-entry-dialog.component.html
+++ b/src/app/entry/delete/dialog/delete-entry-dialog.component.html
@@ -9,7 +9,7 @@
   <ul>
     <li>Has never been published.</li>
     <li>Was created on/after April 1, 2022.</li>
-    <li *ngIf="isWorkflow">Is not a checked or checker workflow.</li>
+    <li *ngIf="!(isNotebook || isService)">Is not checked or a checker workflow.</li>
   </ul>
   <p *ngIf="fromDockstoreYml">
     After you delete the {{ term }}, you must uninstall the Dockstore GitHub App from the source GitHub repo, or edit/remove the
--- a/src/app/entry/delete/dialog/delete-entry-dialog.component.html
+++ b/src/app/entry/delete/dialog/delete-entry-dialog.component.html
@@ -7,9 +7,9 @@
   </p>
   <p>A {{ term }} can be deleted if it:</p>
   <ul>
-    <li>Has never been published,</li>
-    <li>Was created on/after April 1, 2022, and</li>
-    <li>Is not a checked or checker workflow.</li>
+    <li>Has never been published.</li>
+    <li>Was created on/after April 1, 2022.</li>
+    <li *ngIf="isWorkflow">Is not a checked or checker workflow.</li>
   </ul>
   <p *ngIf="fromDockstoreYml">
     After you delete the {{ term }}, you must uninstall the Dockstore GitHub App from the source GitHub repo, or edit/remove the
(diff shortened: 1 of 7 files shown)


coverbeck (approved):

kathy-t (approved):

src/app/entry/delete/dialog/delete-entry-dialog.component.html
```diff
@@ -7,9 +7,9 @@ <h1 mat-dialog-title>Are you sure you want to delete this Dockstore {{ term }}?<
   </p>
   <p>A {{ term }} can be deleted if it:</p>
   <ul>
-    <li>Has never been published,</li>
```
denis-yuen, line 10: I mean, if we're fixing punctuation, these don't need periods or commas
See the "Research office" example of https://www.imperial.ac.uk/brand-style-guide/writing/grammar/punctuation/bullet-points/

src/app/entry/delete/dialog/delete-entry-dialog.component.html
```diff
@@ -7,9 +7,9 @@ <h1 mat-dialog-title>Are you sure you want to delete this Dockstore {{ term }}?<
...
-    <li>Was created on/after April 1, 2022, and</li>
-    <li>Is not a checked or checker workflow.</li>
```
denis-yuen, line 12: Is the condition right here? 
Should this be displayed for a tool which itself is not a workflow?

denis-yuen (commented):
(more for the condition question)

src/app/entry/delete/dialog/delete-entry-dialog.component.html
svonworl: Yeah, it's wrong, I didn't realize tools could be checked.  Maybe I'll change it to not display for services or notebooks.  I probably should change the last item to say "Is not a checked workflow/tool or checker workflow."  That's kind of a mouthful, maybe "Is not checked or a checker workflow."?

svonworl: Done.

src/app/entry/delete/dialog/delete-entry-dialog.component.html
svonworl: The bullet items "complete the introductory stem", thus each forming full sentences, so according to this, at least, they should be punctuated:
https://www.microsoft.com/en-us/microsoft-365-life-hacks/writing/punctuating-bullet-point-lists

denis-yuen: If you're going with that guide, the first word of each shouldn't be capitalized then

denis-yuen (approved):

PR #1923, opened 2024-01-31
Fix sorting in GitHub App Logs Dialog

coverbeck:
**Description**
Fixes issues with sorting by column in GitHub App Logs.

Although the ticket is written up for being a problem only with entry name, it's not just limited to that column. See [this comment](https://ucsc-cgl.atlassian.net/browse/SEAB-6060?focusedCommentId=46770).

The fundamental problem I believe is that we are using a MatTableDataSource, which is intended for use with client-side pagination, which is being combined with server-side pagination. I couldn't figure how exactly how they were interacting, but I ended up going off this [article](https://blog.angular-university.io/angular-material-data-table/), and implement a new `DataSource`.

It's a big change for this close to the release. Denis pointed out that most users probably won't sort anyway, they'll just be looking at the most recent events, which does work.

~~So I'm creating this PR against the release branch, but leaning towards not merging it. Please voice your opinion.~~

After discussion, targetted the develop branch.

**Review Instructions**
Follow Kathy's instructions [here](https://ucsc-cgl.atlassian.net/browse/SEAB-6060?focusedCommentId=46729).

**Issue**
SEAB-6060

**Security**


src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.ts
```diff
@@ -61,57 +51,44 @@ export class GithubAppsLogsComponent extends Base implements OnInit, AfterViewIn
...
     private lambdaEventsService: LambdaEventsService,
-    private matSnackBar: MatSnackBar,
-    private mapPipe: MapFriendlyValuesPipe,
-    private paginatorQuery: PaginatorQuery
+    private alertService: AlertService
   ) {
     super();
-    this.datePipe = new DatePipe('en');
-    const defaultPredicate = this.dataSource.filterPredicate;
-    this.dataSource.filterPredicate = (data, filter) => {
```
coverbeck, line 76: This code wasn't actually effectively doing anything as far as I could tell. It's for client-side filtering, but before it's triggered, the server-side pagination triggers first, so anything this could filter would get removed, e.g., type "Success", and no records come back from the API.

coverbeck:
> > It's a big change for this close to the release. Denis pointed out that most users probably won't sort anyway, they'll just be looking at the most recent events, which does work.
> > So I'm creating this PR against the release branch, but leaning towards not merging it. Please voice your opinion.
> 
> Out of curiosity, how does the result end up comparing with the `/tools` or `/workflows` pages? But reviewing the change, I do lean towards 1.16

Researching that has led to some interesting discoveries:

* The notebooks table doesn't work at all
* Trying to sort on the author column doesn't work -- generates an API request that fails.

Sorting appears to work so far.

denis-yuen:
> * Trying to sort on the author column doesn't work -- generates an API request that fails.

Probably a missed issue from https://github.com/dockstore/dockstore/issues/5467

src/app/myworkflows/sidebar-accordion/github-apps-logs/lambda-events-datasource.ts
```diff
@@ -0,0 +1,96 @@
...
+export class LambdaEventDataSource implements DataSource<LambdaEvent> {
+  private eventsSubject$ = new BehaviorSubject<LambdaEvent[]>([]);
+  private loadingSubject$ = new BehaviorSubject<boolean>(true);
+  private eventsLengthSubject$ = new BehaviorSubject<number>(0);
+  private showContentSubject$ = new BehaviorSubject<ShowContent>(null);
+
+  public events$ = this.eventsSubject$.asObservable();
+  public loading$ = this.loadingSubject$.asObservable();
+  public eventsLength$ = this.eventsLengthSubject$.asObservable();
+  public showContent$ = this.showContentSubject$.asObservable();
```
svonworl, line 37: Is the reason for this Subject->Observable mapping to prevent the Subjects from being publically exposed as "writable" objects?  A Subject is an Observable: https://rxjs.dev/guide/subject

svonworl (commented):
Does this PR use server-side pagination?  Looks like it, but confirming...

coverbeck:
> Does this PR use server-side pagination? Looks like it, but confirming...

Yes, but the code already did use server-side pagination. The PR description touches on it, but the issue was that the server-side pagination was being combined with `MatTableDataSource`, an Angular Material class for client-side pagination, which is what led to some of the odd results.

This PR implements `DataSource` and doesn't use `MatTableDataSource`, so that it's "purely" server-side.

src/app/myworkflows/sidebar-accordion/github-apps-logs/lambda-events-datasource.ts
coverbeck: > Is the reason for this Subject->Observable mapping to prevent the Subjects from being publically exposed as "writable" objects? A Subject is an Observable: https://rxjs.dev/guide/subject

That was my thinking, yes. I didn't want to give outside consumers of the class the ability to push to feed the Observable.

coverbeck:
>     * The notebooks table doesn't work at all

Logged this as SEAB-6224, where it turned out to be more nuanced; the page offset settings carry over between the different tables, so if you're on a high enough page on workflows, then go to notebooks where such a page doesn't exist, no notebooks are shown.

denis-yuen (approved):
Since it's to develop now

cypress/e2e/group2/myworkflows.ts
```diff
@@ -140,8 +165,8 @@ describe('Dockstore my workflows', () => {
           'X-total-count': '1',
         },
       });
-      cy.get('[data-cy=apps-logs-filter]').type('entry1');
-      cy.contains('2020-02-20T02:20');
+      cy.get('[data-cy=apps-logs-filter]').type(entry1);
+      // cy.contains('2020-02-20T02:20');
```
kathy-t, line 169: Should this comment be uncommented?

kathy-t (approved):

hyunnaye (approved):

svonworl (approved):

PR #1926, opened 2024-01-31
Swap position of Metrics and Tool tabs

coverbeck:
**Description**
Gives a little more prominence to metrics, moving the tools tab out of the initial scrolling view, until we come up with a better solution.

Cypress test changes are annoying and a little hard-coded, but I'm going to limit the scope of this change for the release.

**Review Instructions**
Navigate to a workflow and version with metrics and ensure the metrics tab visible. Something like this:

Screen Shot 2024-01-31 at 11 21 12 AM

**Issue**
SEAB-6206

**Security**

### DIFF
--- a/cypress/e2e/immutableDatabaseTests/workflowDetails.ts
+++ b/cypress/e2e/immutableDatabaseTests/workflowDetails.ts
@@ -107,7 +107,8 @@ describe('Dockstore Workflow Details', () => {
     });
   });
 
-  it('Change tab to tools', () => {
+  it.only('Change tab to tools', () => {
+    cy.get('.mat-tab-header-pagination-after').click();
     goToTab('Tools');
     cy.url().should('eq', Cypress.config().baseUrl + '/workflows/github.com/A/l:master?tab=tools');
   });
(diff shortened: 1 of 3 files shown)


denis-yuen (approved):

cypress/e2e/immutableDatabaseTests/workflowDetails.ts
```diff
@@ -107,7 +107,8 @@ describe('Dockstore Workflow Details', () => {
     });
   });
 
-  it('Change tab to tools', () => {
+  it.only('Change tab to tools', () => {
```
kathy-t, line 110: Remove `.only`

kathy-t (approved):

cypress/e2e/immutableDatabaseTests/workflowDetails.ts
coverbeck: Shoot, I thought I'd caught those. Good catch.

svonworl (approved):
Given the proximity to the release, I support swapping the button positions, as it is relatively safe.

However, in a hotfix, it would be easy to make the buttons less wide, so they all fit and are visible simultaneously, see this styling that I added to the accordion tab bar to make the "Archived" button fit:

https://github.com/dockstore/dockstore-ui2/blob/f0766ab19bfbf2197c1882f015055d7805726475/src/app/shared/styles/my-entry-sidebar.scss#L70

This styling makes the button width adapt to the available space, so they fill what's available, but also can shrink if necessary.

PR #1928, opened 2024-01-31
SEAB-6292: Fix refresh issue when reloading the Metrics tab

hyunnaye:
**Description**
Before, when viewing the metrics tab, if the user refreshes the page, the tab would reset back to the Info tab. 

Thus this PR fixes that issue such that when a user reloads the page, the page will be reloaded back onto the Metrics tab.

**Review Instructions**
1. Go to  https://staging.dockstore.org/workflows/github.com/gatk-workflows/seq-format-conversion/BAM-to-Unmapped-BAM:3.0.0?tab=metrics 
2. Refresh the page and verify that the page loads back onto the Metrics tab.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6202

**Security**

### DIFF
--- a/src/app/workflow/workflow.component.ts
+++ b/src/app/workflow/workflow.component.ts
@@ -161,16 +161,16 @@ export class WorkflowComponent extends Entry<WorkflowVersion> implements AfterVi
     // Entry type is set in the container, workflow, and service routing files.
     this.entryType = this.sessionQuery.getValue().entryType;
     if (this.entryType === EntryType.BioWorkflow) {
-      this.validTabs = ['info', 'launch', 'versions', 'files', 'tools', 'dag'];
+      this.validTabs = ['info', 'launch', 'versions', 'files', 'tools', 'dag', 'metrics'];
       this.redirectToCanonicalURL('/' + myBioWorkflowsURLSegment);
     } else if (this.entryType === EntryType.Tool) {
       this.validTabs = ['info', 'launch', 'versions', 'files'];
       this.redirectToCanonicalURL('/' + myToolsURLSegment);
     } else if (this.entryType === EntryType.Notebook) {
-      this.validTabs = ['info', 'code', 'versions', 'files'];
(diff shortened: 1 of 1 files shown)


src/app/workflow/workflow.component.ts
```diff
@@ -161,16 +161,16 @@ export class WorkflowComponent extends Entry<WorkflowVersion> implements AfterVi
     // Entry type is set in the container, workflow, and service routing files.
     this.entryType = this.sessionQuery.getValue().entryType;
     if (this.entryType === EntryType.BioWorkflow) {
-      this.validTabs = ['info', 'launch', 'versions', 'files', 'tools', 'dag'];
+      this.validTabs = ['info', 'launch', 'versions', 'files', 'tools', 'dag', 'metrics'];
       this.redirectToCanonicalURL('/' + myBioWorkflowsURLSegment);
     } else if (this.entryType === EntryType.Tool) {
       this.validTabs = ['info', 'launch', 'versions', 'files'];
```
coverbeck, line 167: @kathy-t After my questions to you about metrics and tools in Slack regarding my ticket, should `metrics` be here as well?

coverbeck (approved):

src/app/workflow/workflow.component.ts
kathy-t: Hmm, thought this only affected legacy tools, but I guess app tools get these `EntryType.Tool` tabs as well :thinking: I suppose so, so that app tool metrics tab stays when refreshed (@hyunnaye does the metrics tab stay on refresh for app tools with this PR?)

Legacy tools don't have the metrics tab because it doesn't use the workflow component and legacy tools don't have the launch with feature.

hyunnaye: Just checked and looks like app tools had this problem without adding `metrics` on this line. Just added!

denis-yuen (approved):

PR #1932, opened 2024-02-12
DOCK-2484: Keep original URL on Page Not Found

ll5zh:
**Description**
When accessing non-existent pages, the desired behaviour is to display the `PageNotFound` component while keeping the original URL in the address bar; this makes it easy to check for potential typos in the link. See https://dockstore.org/asdf for an example.

Currently, there are some 404 URLs that directly lead to the `page-not-found` path, like https://dockstore.org/workflows/asdf. 

There are also URLs that probably should display the `PageNotFound` component but don't, and also redirect to a strange URL with a double domain (i.e. incorrect path). As an example, visit https://dockstore.org/notebooks/githyb.com/dockstore/dockstore-trs-notebook; it has a typo in "github" and demonstrates this behaviour.

This [comment](https://github.com/dockstore/dockstore/issues/5728#issuecomment-1930368145) contains the list of cases that this PR attempts to resolve.

**Review Instructions**
Try visiting the paths below and check that: 1) the original URL is kept in the address bar, and 2) it displays the expected content (presumably the `PageNotFound` component, since none of the paths are valid entries).

Paths that used to redirect to `page-not-found`:
- /users/asdf
- /organizations/asdf
- /organizations/iwc/collections/asdf (where IWC is a valid organization)
- /containers/asdf
- /search-containers/asdf
- /tools/asdf
- /workflows/asdf

Paths that used to redirect to the double domain URL:
- /search-workflows/asdf
- /notebooks/asdf
- /services/asdf
- /apptools/asdf

In terms of the code, perhaps pay special attention to changes in entry.ts, like the removal of `getPageIndex` (and by extension, `switchToolsToContainers`), and the edits to `decodeURL`. It also might be good to double check `setupPublicEntry` in workflow.component.ts, after removing the regex logic.

**Issue**
https://github.com/dockstore/dockstore/issues/5728
DOCK-2484

**Security**


src/app/shared/entry.ts
```diff
@@ -412,11 +412,14 @@ export abstract class Entry<V extends WorkflowVersion | Tag> implements OnDestro
...
-    const newPath = url.substring(containersIndex);
-    this.location.replaceState(newPath);
+  decodeURL(): void {
+    let url = decodeURIComponent(window.location.href);
+    // Get decoded path by extracting substring of url, starting from third forward slash '/'
+    // This removes the HTTP/HTTPS domain name (e.g. https://dockstore.org) from the url
+    for (let i = 0; i < 3; i++) {
+      url = i < 2 ? url.substring(url.indexOf('/') + 1) : url.substring(url.indexOf('/'));
+    }
+    this.location.replaceState(url);
```
ll5zh, line 422: I think this change resolves some of the weird double domain URL paths. In the case of https://dockstore.org/services/asdf, the `type` passed to `decodeURL` is "workflows", which isn't present in the original URL and thus messes up the new URL at `replaceState`.

Seems to help keep the original URL for 404 paths that look like: `services/*`, `notebooks/*`, `apptools/*`, `search-workflows/*`.

src/app/workflow/workflow.component.ts
```diff
@@ -398,15 +398,8 @@ export class WorkflowComponent extends Entry<WorkflowVersion> implements AfterVi
...
-          const workflowRegex = /\/workflows\/(github.com)|(gitlab.com)|(bitbucket.org)\/.+/;
-          const gitHubAppToolRegex = /\/containers\/(github.com)\/.+/;
-          if (
-            workflowOrgRegex.test(this.resourcePath) ||
-            workflowRegex.test(this.resourcePath) ||
-            gitHubAppToolRegex.test(this.resourcePath)
-          ) {
-            this.router.navigate(['page-not-found']);
+          if (!this.workflow) {
+            this.router.navigate(['page-not-found'], { skipLocationChange: true });
```
ll5zh, line 402: Thought to replace the regex condition to allow invalid workflow paths like `services/asdf`, `notebooks/asdf`, `apptools/asdf`, `search-workflows/asdf` to also display `PageNotFound` instead of a blank page (even though some of these URLs are unlikely to be encountered).

denis-yuen:
> **Issue**
> [DOCK-2484](https://ucsc-cgl.atlassian.net/browse/DOCK-2484)

FYI, not a big deal but link to the github issue for DOCK tickets, it creates nice links to look up the history of an issue

src/app/workflow/workflow.component.ts
denis-yuen: Without testing it out, my initial thought is I'm usually in favour of avoiding a regex if possible. They can be hard to debug

src/app/shared/entry.ts
```diff
@@ -412,11 +412,14 @@ export abstract class Entry<V extends WorkflowVersion | Tag> implements OnDestro
    * Will decode the URL
    * @return {void}
    */
-  decodeURL(type: string): void {
-    const url = decodeURIComponent(window.location.href);
-    const containersIndex = this.getIndexInURL('/' + type);
-    const newPath = url.substring(containersIndex);
-    this.location.replaceState(newPath);
+  decodeURL(): void {
+    let url = decodeURIComponent(window.location.href);
```
coverbeck, line 416: Can this be `const url = decodeURIComponent(window.location.pathname);`, then you don't need the lines 417-421?

src/app/shared/url-resolver.service.ts
```diff
@@ -46,6 +46,9 @@ export class UrlResolverService {
     title = this.decodedString(title.replace(`/containers/`, ''));
     title = this.decodedString(title.replace(`/my-notebooks/`, ''));
     title = this.decodedString(title.replace(`/notebooks/`, ''));
+    title = this.decodedString(title.replace(`/search-workflows/`, ''));
+    title = this.decodedString(title.replace(`/search-containers/`, ''));
+    title = this.decodedString(title.replace(`/apptools/`, ''));
```
kathy-t, line 51: Just going off the function name, it seems like these three shouldn't be here because the URLs don't have the entry paths. For example, clicking on a workflow from `/search-workflows` takes you to `/workflows/<entry-path>`, not `/search-workflows/<entry-path>`

Why are those three needed?

src/app/shared/entry.ts
```diff
@@ -488,7 +482,6 @@ export abstract class Entry<V extends WorkflowVersion | Tag> implements OnDestro
    * Gets the index of /containers or /workflows from the URL
    * @return {number}
    */
-  abstract getPageIndex(): number;
```
kathy-t, line 491: Should delete the function comment above too

src/app/workflow/workflow.component.ts
```diff
@@ -398,15 +398,8 @@ export class WorkflowComponent extends Entry<WorkflowVersion> implements AfterVi
...
-          const workflowOrgRegex = /\/workflows\/.+/;
-          const workflowRegex = /\/workflows\/(github.com)|(gitlab.com)|(bitbucket.org)\/.+/;
-          const gitHubAppToolRegex = /\/containers\/(github.com)\/.+/;
-          if (
-            workflowOrgRegex.test(this.resourcePath) ||
-            workflowRegex.test(this.resourcePath) ||
-            gitHubAppToolRegex.test(this.resourcePath)
-          ) {
-            this.router.navigate(['page-not-found']);
+          if (!this.workflow) {
```
kathy-t, line 401: I think it's better to check that the error code is 404 (not found) instead like the container component does

https://github.com/dockstore/dockstore-ui2/blob/fa61caefdc988edf89e0b3170fc252658914d81c/src/app/container/container.component.ts#L263

src/app/shared/url-resolver.service.ts
ll5zh: I guess if the path happened to be `/search-workflows/<entry-path>` then it would redirect to the correct workflow at `/workflows/<entry-path>`. But you're right, that's not the purpose of those three routes so I think I'll remove those lines.

src/app/workflow/workflow.component.ts
ll5zh: I think checking for a 404 error code causes the [checkerWorkflowFromTools](https://cloud.cypress.io/projects/1ya4vj/runs/33045/overview/d72c1f2b-13e9-45bc-9371-6ed88572d7bd?roarHideRunsWithDiffGroupsAndTags=1) integration test to fail? At 0:40 of the [failed test run](https://cloud.cypress.io/projects/1ya4vj/runs/33045/overview/d72c1f2b-13e9-45bc-9371-6ed88572d7bd/video?roarHideRunsWithDiffGroupsAndTags=1), viewing the parent tool from the checker workflow displays "Page Not Found" instead of the tool entry. It seems like it incorrectly entered the error function's `if (error.status === 404)` block (instead of the error `else` block, which passing tests typically appear to enter -- at 0:43 of this [successful run](https://cloud.cypress.io/projects/1ya4vj/runs/33056/test-results/580ebfdf-da75-41b2-8710-ef6df9913223/video?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%5D&testingTypesEnum=%5B%5D), the page shows "The workflow linked to could not be found!")

.circleci/config.yml
```diff
@@ -490,8 +490,8 @@ commands:
               fi
   install_browser_tools:
     steps:
-      - browser-tools/install-chrome
-      - browser-tools/install-chromedriver
+      - browser-tools/install-browser-tools:
```
kathy-t, line 493: I think this can be reverted since it was fixed in https://github.com/dockstore/dockstore-ui2/pull/1937

kathy-t (approved):

coverbeck (approved):

PR #1934, opened 2024-02-15
Upgrade to Angular 15

coverbeck:
**Description**
Upgrade to Angular 15 and Angular Material 15. 

The ticket title is to upgrade to Angular 16, and the description suggests perhaps even 17. But Angular has a migration tool, which made most of the changes in this PR, and it only upgrades one major version at a time. So this is the first step, and will give us a fallback if upgrading to 16/17 doesn't work.

Most changes made by the Angular Migration tool, but here are some additional notes:

- The Angular Material components we are using have been deprecated and will be removed in Angular Material 17. In preparation for that, they've been renamed `Legacy`, hence most of the changes. I may attempt to migrate to the new components, still on Angular 15, before migration to Angular Material 16.
- The spec files for the unit tests for some reason got fussier about imports being declared.
- The HTML anchor tag does not have a disabled attribute. We were using it before without complaints. I had to change the anchor to a button.

**Review Instructions**
Navigate throughout the UI and ensure no pages seem broken or are behaving differently.

**Issue**
SEAB-6140

**Security**


src/app/docs/docs.component.spec.ts
```diff
@@ -34,15 +36,15 @@ describe('DocsComponent', () => {
...
+  beforeEach(() => {
+    fixture = TestBed.createComponent(DocsComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
 
-  // it('should create', () => {
-  //   expect(component).toBeTruthy();
-  // });
+  it('should create', () => {
```
coverbeck, line 45: This was commented out; in Angular 15 the tests would throw an error for no tests in the spec file.

src/app/shared/register-github-app/register-github-app.component.html
```diff
@@ -29,16 +29,17 @@
     >.
   </p>
   <div fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="2rem" class="mb-4">
-    <a
+    <button
       mat-raised-button
       type="button"
       class="accent-1-dark-btn mat-elevation-z no-underline"
-      [href]="gitHubAppInstallationLink$ | async"
+      (click)="openGitHubApp()"
```
coverbeck, line 36: The anchor tag does not have a disabled attribute (line 39), Angular 15 started error on this, so changed to a button, which doesn't have an href. This covers the changes in the controller as well.

src/app/shared/register-github-app/register-github-app.component.ts
```diff
@@ -7,11 +9,24 @@ import { UserQuery } from '../user/user.query';
...
   @Input() public entryType: string;
+  private gitHubAppInstallationLink: string;
 
-  constructor(private sessionQuery: SessionQuery, private userQuery: UserQuery) {}
+  constructor(private sessionQuery: SessionQuery, private userQuery: UserQuery) {
+    super();
+  }
+
+  ngOnInit(): void {
+    this.gitHubAppInstallationLink$
```
coverbeck, line 24: Needed this for the change to a button and using onClick in the HTML.

src/material.scss
```diff
@@ -20,9 +20,18 @@
 // Include the common styles for Angular Material. We include this here so that you only
 // have to load a single css file for Angular Material in your app.
 // Be sure that you only ever include this mixin once!
-@include mat.core();
+// As of v15 mat.legacy-core no longer includes default typography styles.
```
coverbeck, line 23: This block created by migration tool. I'm not super familiar with typography, but the pages look the same as before, so leaving it as is.

src/app/organizations/collections/remove-collection/remove-collection.component.ts
```diff
@@ -25,11 +25,20 @@ import { CollectionsQuery } from '../../state/collections.query';
 })
 export class RemoveCollectionDialogComponent {
   public collectionsQueryLoading$: Observable<boolean>;
-  constructor(@Inject(MAT_DIALOG_DATA) public data: CollectionDialogData, private collectionsService: CollectionsService, private collectionsQuery: CollectionsQuery) {
+  constructor(
```
coverbeck, line 28: Just the auto-formatter doing its thing

tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "moduleResolution": "node",
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
-    "target": "es2020",
+    "target": "ES2022",
```
svonworl, line 12: Are we pretty confident that everyone who visits Dockstore will have a browser that supports `ES2022`?
Related: https://compat-table.github.io/compat-table/es2016plus/
(https://github.com/tc39/proposal-regexp-match-indices is an issue for some browsers?)
(I copied this question from the PR that upgraded to Angular 14).

coverbeck: Fair point, not sure, rolled back to es2020.

svonworl (approved):

tsconfig.json
denis-yuen: Not sure if I'm reading the chart right, but it looks like ES2021 is as compatible as ES2020. 
Don't know if its worth it

denis-yuen (approved):
That's a lot of deprecation warnings in SonarCloud

denis-yuen:
> So this is the first step, and will give us a fallback if upgrading to 16/17 doesn't work.

This totally makes sense, so no objection. 

But was curious, so looked it up. It seems that this buys us not active support, but three months of security support (till May). 
https://endoflife.date/angular

The L in LTS is a bit relative https://angular.io/guide/releases#actively-supported-versions

tsconfig.json
coverbeck: > Not sure if I'm reading the chart right, but it looks like ES2021 is as compatible as ES2020. Don't know if its worth it

Agreed, it looks the same, but I think I'll keep it to be safe.

coverbeck:
> That's a lot of deprecation warnings in SonarCloud

All the Angular Material components we have been using have been deprecated, and will be phased out in 17. I wanted to tackle those separately, limiting the scope of this PR to the minimum to upgrade to Angular 15.

kathy-t (approved):

PR #1936, opened 2024-02-21
DOCK-2513: Tweak search relevance

svonworl:
**Description**
This PR increases the search query `boost` setting for path matches, so that a hit with a path match ranks higher than an otherwise similar hit with a description match.  14 was the lowest integer boost that seemed to do the trick.  Q: Why did the boost have to be so much higher on the path subqueries?  A: For the answer, we'd have to dive into the details of exactly how ElasticSearch makes the sausage.  Google "BM25" for an idea of what we're talking about here.

While I was doing some experiments, I noticed that entries in our categories weren't ranking as high as I expected.  So I boosted them just a little bit.

**Review Instructions**
Search workflows for the keyword "TheiaProk" and confirm that the hits containing "TheiaProk" in their paths tend to appear first in the results. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2513
dockstore/dockstore#5817

**Security**
No concerns.

### DIFF
--- a/src/app/search/query-builder.service.ts
+++ b/src/app/search/query-builder.service.ts
@@ -274,16 +274,16 @@ export class QueryBuilderService {
     const terms = searchString.trim().split(' ').slice(0, 20);
     terms.forEach((term) => {
       body
-        .orQuery('wildcard', 'full_workflow_path', { value: '*' + term + '*', case_insensitive: true, boost: 7 })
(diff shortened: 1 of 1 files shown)


denis-yuen (approved):
Seems pretty clean

coverbeck (approved):

kathy-t (approved):

PR #1938, opened 2024-02-26
Display skipped executions warning for metrics

kathy-t:
**Description**
Corresponding PRs:
- https://github.com/dockstore/dockstore/pull/5822
- https://github.com/dockstore/dockstore-support/pull/486
- https://github.com/dockstore/compose_setup/pull/251

This PR displays a warning icon if the aggregated metric had executions that were skipped during aggregation. The tool tip displays the number of executions skipped.

image

image

<details>
<summary>Old screenshot</summary>
Screenshot 2024-02-22 at 11-36-26 Dockstore Workflow github com_kathy-t_ghapps-single-workflow
</details>

**Review Instructions**
Verify that this workflow (to be filled in) has a skipped executions warning in the Metrics tab.

**Issue**
[SEAB-6046](https://ucsc-cgl.atlassian.net/browse/SEAB-6046)

**Security**

[SEAB-6046]: https://ucsc-cgl.atlassian.net/browse/SEAB-6046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -4,9 +4,9 @@
   "license": "Apache License 2.0",
   "config": {
     "webservice_version": "1.15.0-rc.2",
-    "use_circle": false,
-    "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/10707/workflows/cbcf24c8-061a-4003-9d46-9de3b953b25f/jobs/39503/artifacts",
-    "circle_build_id": "39503"
+    "use_circle": true,
+    "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/10904/workflows/f9b3b508-768d-42d2-954e-9a23b78952df/jobs/41061/artifacts",
```
kathy-t, line 8: Note to self: update this to a develop artifact when webservice PR is merged

denis-yuen:
> This PR displays a warning icon if the aggregated metric had executions that were skipped during aggregation. The tool tip displays the number of executions skipped.

If I understand correctly, I feel like the warning is a bit too prominent for the following reasons
1) Once we start skipping executions, even if new executions in the future are fixed, the older skipped executions will remain and be skipped in the future
2) Even one skipped execution in thousands of runs could cause the icon to pop up
3) The icon isn't really actionable for end-users, just the platform owner

Brainstorming:
Not sure how to display though, ideas
* "Skipped executions" with a tooltip next to aborted executions without the icon?
* Info icon, not warning icon?
* Display only to admins, curators, and platform owners?

denis-yuen (commented):
technically, looks ok
think warning is a bit too prominent and/or hard to interpret for most users, so may need to consider alternatives

svonworl (approved):
The term "skipped" makes me think of the run not being executed because the workflow engine decided not to or something like that.  I wonder if there's another term or a different way of phrasing it that makes it clearer that we skipped it during the analysis phase.

coverbeck (commented):
Random thoughts, just brainstorming:

*  I'm wary of too much information being in a tooltip; tooltips should supplement info displayed, not be a substitute, because most users don't know to look for them, IMO.
*  How frequently are executions skipped in our current data set? Is this a once in a blue moon sort of thing, or will it be fairly common? If it's reasonably common, I would be tempted to just have another column with the number.
* Does a user need to know in detail that, for example, the memory metric aggregation was skipped 8 times, but the CPU metric was skipped 11 times? I would argue not.
* What if we captured the total number of skipped aggregations at the top along with Total/Successful/Failed/Aborted? I guess the skipped is at different level though, a single execution could have more than one skipped, right?
* Digression, I think the top could be improved; instead of repeating the word "Executions" 4 times, we could have it once.

denis-yuen:
> Does a user need to know in detail that, for example, the memory metric aggregation was skipped 8 times, but the CPU metric was skipped 11 times? I would argue not.

I'm actually toying with the idea that users with the platform partner role may want to know about skipped executions, I'm not sure regular user types should

kathy-t:
>     * How frequently are executions skipped in our current data set? Is this a once in a blue moon sort of thing, or will it be fairly common? If it's reasonably common, I would be tempted to just have another column with the number.

Not frequently -- I don't think we have any skipped executions right now because we've been ingesting data for the platforms. Ideally, the webservice should not let any invalid data get in, which is improved with https://github.com/dockstore/dockstore/pull/5822. So I think the chances are low that there are skipped executions because the webservice would throw an error first.

>     * Does a user need to know in detail that, for example, the memory metric aggregation was skipped 8 times, but the CPU metric was skipped 11 times? I would argue not.

I included it so the user knows how many executions were used to aggregate the metric. For example, an average over 100 executions is different from an average over 1 execution where the other 99 were skipped.

>     * What if we captured the total number of skipped aggregations at the top along with Total/Successful/Failed/Aborted? I guess the skipped is at different level though, a single execution could have more than one skipped, right?

Yeah, it's scoped to the metric level. So for a single execution, it could be skipped for multiple metrics.

coverbeck (approved):

PR #1940, opened 2024-03-01
Add Galaxy planemo information to launch-with page

denis-yuen:
**Description**
Adds Planemo `run` and `workflow_job_init` (parameter file creation)  to launch-with tab

**Review Instructions**
View Galaxy workflows with and without test parameter files already registered, 
Galaxy workflows with test parameter files will give the user the command to download the first one, and show init, and run info. 
Galaxy workflows without them will skip the panel for downloading a test parameter file

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6273

**Security**

### DIFF
--- a/src/app/shared/launch.service.ts
+++ b/src/app/shared/launch.service.ts
@@ -77,6 +77,13 @@ export abstract class LaunchService {
     return `nextflow run https://${workflowPath} -r ${versionName}`;
   }
 
+  getSharedZipString(workflowPath: string, versionName: string) {
+    return `wget -O temp.zip ${Dockstore.API_URI}${ga4ghPath}/tools/${encodeURIComponent(
+      '#workflow/' + workflowPath
+    )}/versions/${encodeURIComponent(versionName)}/GALAXY/files?format=zip
+unzip temp.zip`;
+  }
+
   /**
    * This creates the planemo local init command
    * @param path The GA4GH Tool's path
@@ -84,11 +91,7 @@ export abstract class LaunchService {
    */
   getPlanemoLocalInitString(workflowPath: string, versionName: string, primaryDescriptorPath: string, testParameterPath: string) {
     return (
-      `wget -O foo.zip ${Dockstore.API_URI}${ga4ghPath}/tools/` +
-      encodeURIComponent('#workflow/' + workflowPath) +
-      `/versions/${versionName}/GALAXY/files?format=zip` +
-      '\nunzip foo.zip' +
-      `\nplanemo workflow_job_init ${primaryDescriptorPath} -o ${testParameterPath}`
+      this.getSharedZipString(workflowPath, versionName) + `\nplanemo workflow_job_init ${primaryDescriptorPath} -o ${testParameterPath}`
     );
   }
 
@@ -99,10 +102,7 @@ export abstract class LaunchService {
    */
   getPlanemoLocalLaunchString(workflowPath: string, versionName: string, primaryDescriptorPath: string, testParameterPath: string) {
     return (
-      `wget -O foo.zip ${Dockstore.API_URI}${ga4ghPath}/tools/` +
-      encodeURIComponent('#workflow/' + workflowPath) +
-      `/versions/${versionName}/GALAXY/files?format=zip` +
-      '\nunzip foo.zip' +
+      this.getSharedZipString(workflowPath, versionName) +
       `\nplanemo run ${primaryDescriptorPath} ${testParameterPath} --download_outputs --output_directory . --output_json output.json --engine docker_galaxy`
     );
   }
--- a/src/app/shared/launch.service.ts
+++ b/src/app/shared/launch.service.ts
@@ -78,6 +78,21 @@ export abstract class LaunchService {
   }
 
   /**
+   * This creates the planemo local launch commands
+   * @param path The GA4GH Tool's path
+   * @param versionName The ToolVersion's name
+   */
+  getPlanemoLocalLaunchString(workflowPath: string, versionName: string) {
+    return (
+      `wget -O foo.zip ${Dockstore.API_URI}${ga4ghPath}/tools/` +
+      encodeURIComponent('#workflow/' + workflowPath) +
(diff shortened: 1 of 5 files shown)


denis-yuen:
Screenshot from 2024-03-05 17-51-05

WIP, unreliable since I probably have the Observable and subscribe wrong, but gets the general idea across

denis-yuen:
FYI @mvdbeek before cleaning up (changing the example name from foo.zip and looking at accessibility) are there any other planemo commands that you would like us to highlight on the Galaxy launch-with page?

mvdbeek:
This is a great start, thank you! There's potentially `planemo test workflow.ga` that could be mentioned for running any included tests, but I guess that wouldn't really fit into `Launch` tab.

src/app/shared/launch.service.ts
```diff
@@ -77,6 +77,36 @@ export abstract class LaunchService {
...
+  /**
+   * This creates the planemo local init command
+   * @param path The GA4GH Tool's path
+   * @param versionName The ToolVersion's name
+   */
+  getPlanemoLocalInitString(workflowPath: string, versionName: string, primaryDescriptorPath: string, testParameterPath: string) {
+    return (
+      `wget -O foo.zip ${Dockstore.API_URI}${ga4ghPath}/tools/` +
+      encodeURIComponent('#workflow/' + workflowPath) +
+      `/versions/${versionName}/GALAXY/files?format=zip` +
```
coverbeck, line 89: I think the versionName needs to be encoded as well; I noticed some versions with `/` in them when testing the metrics stuff.

src/app/shared/launch.service.ts
```diff
@@ -77,6 +77,36 @@ export abstract class LaunchService {
...
+  /**
+   * This creates the planemo local launch commands
+   * @param path The GA4GH Tool's path
+   * @param versionName The ToolVersion's name
+   */
+  getPlanemoLocalLaunchString(workflowPath: string, versionName: string, primaryDescriptorPath: string, testParameterPath: string) {
+    return (
+      `wget -O foo.zip ${Dockstore.API_URI}${ga4ghPath}/tools/` +
+      encodeURIComponent('#workflow/' + workflowPath) +
+      `/versions/${versionName}/GALAXY/files?format=zip` +
```
coverbeck, line 104: Encode versionName. This block looks the same as the block on line 86, except for the last line, could optionally reduce duplication that was there.

src/app/shared/launch.service.ts
```diff
@@ -77,6 +77,36 @@ export abstract class LaunchService {
...
   }
 
+  /**
+   * This creates the planemo local init command
+   * @param path The GA4GH Tool's path
+   * @param versionName The ToolVersion's name
+   */
+  getPlanemoLocalInitString(workflowPath: string, versionName: string, primaryDescriptorPath: string, testParameterPath: string) {
+    return (
+      `wget -O foo.zip ${Dockstore.API_URI}${ga4ghPath}/tools/` +
```
coverbeck, line 87: Nit, I'd give it a name like `temp.zip` to better reflect it's purpose.

src/app/workflow/launch/launch.component.ts
```diff
@@ -126,20 +135,45 @@ export class LaunchWorkflowComponent extends EntryTab implements OnInit, OnChang
...
+            this.testParameterPath = undefined;
+          }
+          this.wgetTestJsonDescription = this.launchService.getTestJsonString(
+            ga4ghWorkflowIdPrefix + workflowPath,
+            versionName,
+            descriptorType,
+            this.testParameterPath
+          );
+        }
+        if (descriptorFiles !== undefined && descriptorFiles.length > 0) {
```
coverbeck, line 157: Could simplify to `descriptorFiles?.length > 0`

src/app/shared/launch.service.ts
denis-yuen: Ah, forgot about that

src/app/shared/launch.service.ts
denis-yuen: Oy :grimacing:  (more about the versions with a slash in them, than having to encode)

src/app/workflow/launch/launch.component.ts
```diff
@@ -126,20 +135,45 @@ export class LaunchWorkflowComponent extends EntryTab implements OnInit, OnChang
...
+          }
+          this.wgetTestJsonDescription = this.launchService.getTestJsonString(
+            ga4ghWorkflowIdPrefix + workflowPath,
+            versionName,
+            descriptorType,
+            this.testParameterPath
+          );
+        }
+        if (descriptorFiles?.length > 0) {
+          // ... but primary descriptor is mandatory
```
coverbeck, line 158: This may not be true of the registered Galaxy workflows currently on Dockstore, but I've seen other (invalid) workflows where there are no files associated with a version. I wonder if to be safe you shouldn't set the properties to null or undefined if this condition isn't true.

src/app/workflow/launch/launch.component.ts
```diff
@@ -126,20 +135,45 @@ export class LaunchWorkflowComponent extends EntryTab implements OnInit, OnChang
...
+          this.planemoLocalLaunchString = this.launchService.getPlanemoLocalLaunchString(
+            workflowPath,
+            versionName,
+            this.primaryDescriptorPath,
+            this.testParameterPath === undefined ? 'example-parameter-file.yml' : this.testParameterPath
+          );
+        }
+      },
+      (err) => {
+        console.log(err);
```
coverbeck, line 175: Also suggest setting properties to null/undefined if failure

coverbeck (approved):

src/app/workflow/launch/launch.component.ts
denis-yuen: Added, but note there's a warning " This section is only available for valid versions. A valid version requires a valid descriptor file. " that blanks out the tab anyway

kathy-t (approved):

PR #1941, opened 2024-03-07
SEAB-6014: UI link checks

ll5zh:
**Description**
The goal of this PR is to experiment with integration testing for broken links, which aims to detect instances of broken URLs and images (currently falls under integration_test_4 as linkCheck.ts).

As of now, the test should iterate through relevant pages/selectors on the site (via `urlPages` and `imagePages`) and, for each, perform link checks. It targets anchor links by sending requests to their `href` property ([example](https://www.linkedin.com/pulse/finding-broken-links-using-cypress-brahma-rao-kothapalli)), and targets images by asserting their `naturalWidth` property is greater than zero ([example](https://stackoverflow.com/questions/51246606/test-loading-of-image-in-cypress/58641462#58641462)).

**Review Instructions**
Look over [linkCheck.ts](https://github.com/dockstore/dockstore-ui2/blob/feature/seab-6014/cypress/e2e/immutableDatabaseTests/linkCheck.ts) and the results/outputs of the [Cypress runs](https://cloud.cypress.io/projects/1ya4vj/runs/33225/overview?roarHideRunsWithDiffGroupsAndTags=1); ponder the effectiveness/relevance of a test like this and/or how it could be modified/improved. Would also appreciate any feedback relating to the thoughts included below!

Thoughts:

- Is it inefficient (edit: is it useful/meaningful) to run this test on every push? For example, it's probably unlikely that anchor links to /workflows, /dashboard, etc will ever break, and the same follows for internal images (other than newly added ones) that stay in the repo. In other words, a lot of links are relatively stable.
- ^ A possible solution is only testing external URLs (particularly for anchor links). On the other hand, are there instances of internal links that might break, and thus should be tested?

- For this test to stay up to date (at least with this version), we have to maintain `urlPages` and `imagePages` whenever new pages/components with links are added to the UI. Are the results of this test worth the effort?
- ^ If we keep this method, are there paths/selectors that could be added to/removed from `urlPages` and/or `imagePages`?
- False positives (finds broken links that aren't broken): in previous commits, some [image checks fail on first run](https://cloud.cypress.io/projects/1ya4vj/runs/33196/overview?roarHideRunsWithDiffGroupsAndTags=1), I think this is temporarily fixed by adding `cy.wait(500)` (and looks to be part of a larger flakiness issue, SEAB-6207).
- False negatives (fails to find links that are broken): the link to "[Advanced Features](https://docs.dockstore.org/en/stable/advanced-topics/advanced-features.html)" at /sitemap fails locally, but succeeds in the web run. Seems like the `href` for documentation links isn't formed correctly and takes on the form null/*. (EDIT: fixed by updating test/web.yml, link fails properly)
- There are some links that "fail" but aren't necessarily erroneous (ex. the Twitter link in the footer redirects to login, the token user avatar image in the navbar isn't "visible"). We could filter out such links while checking, but the more we filter out the less useful this test becomes.
- Should a test like this be required for merging PRs?

**Issue**
SEAB-6014

**Security**

### DIFF
--- a/cypress/e2e/immutableDatabaseTests/linkCheck.ts
+++ b/cypress/e2e/immutableDatabaseTests/linkCheck.ts
@@ -1,38 +1,41 @@
 import { setTokenUserViewPort } from '../../support/commands';
 
-describe('Find broken links', () => {
+function formatPaths(paths: string[]) {
+  let formattedPaths = '';
+  paths.forEach((path) => {
+    formattedPaths += '\n- ' + path;
+  });
+  return formattedPaths;
+}
+
+describe('Find broken anchor links', () => {
   setTokenUserViewPort();
 
   function checkUrls(path: string, selector: string) {
+    let brokenUrls = [];
     cy.visit(path);
-    cy.get(selector).within(() => {
-      cy.get('a').each((url) => {
+    cy.get(selector)
+      .find('a')
+      .each((url) => {
         cy.get(url);
         const href = url.prop('href');
         if (href) {
           cy.request({
-            url: url.prop('href'),
+            url: href,
             failOnStatusCode: false,
           }).then((result) => {
             if (result.status != 200) {
+              brokenUrls.push(href);
               cy.log(result.status + ': ' + href);
             }
           });
         }
-      });
-    });
-  }
-
-  function checkImages(path: string, selector: string) {
-    cy.visit(path).wait(500);
-    cy.get(selector).within(() => {
-      cy.get('img').each((image) => {
-        cy.get(image);
-        if (image.prop('naturalWidth') === 0 && image.prop('naturalHeight') === 0) {
-          cy.log('image not visible: ' + image.prop('src'));
+      })
+      .then(() => {
+        if (brokenUrls.length) {
+          throw new Error(`Broken links at "${path}":` + formatPaths(brokenUrls));
         }
       });
-    });
   }
 
   const urlPages = [
@@ -43,6 +46,35 @@ describe('Find broken links', () => {
     ['/docs', 'app-docs'],
     ['/dashboard', 'app-my-sidebar'],
   ];
+
+  urlPages.forEach((urlPage) => {
+    it(`anchor links at "${urlPage[0]}" should work`, () => {
+      checkUrls(urlPage[0], urlPage[1]);
+    });
+  });
+});
+
+describe('Find broken image links', () => {
+  setTokenUserViewPort();
+
+  function checkImages(path: string, selector: string) {
+    let brokenImages = [];
+    cy.visit(path);
+    cy.get(selector)
+      .find('img')
+      .each((image) => {
+        cy.get(image);
+        if (image.prop('naturalWidth') === 0) {
+          brokenImages.push(image.prop('src'));
+          cy.log('image not visible: ' + image.prop('src'));
+        }
+      })
+      .then(() => {
+        if (brokenImages.length) {
+          throw new Error(`Broken images at "${path}":` + formatPaths(brokenImages));
+        }
+      });
+  }
 
   const imagePages = [
     ['/', 'app-root'],
@@ -62,15 +94,9 @@ describe('Find broken links', () => {
     ['/accounts', 'app-accounts-external'],
   ];
 
-  it('should detect broken anchor links', () => {
-    urlPages.forEach((page) => {
-      checkUrls(page[0], page[1]);
-    });
-  });
-
-  it('should detect broken images', () => {
-    imagePages.forEach((page) => {
-      checkImages(page[0], page[1]);
+  imagePages.forEach((imagePage) => {
+    it(`images at "${imagePage[0]}" should work`, () => {
+      checkImages(imagePage[0], imagePage[1]);
     });
   });
 });
--- a/cypress/e2e/immutableDatabaseTests/linkCheck.ts
+++ b/cypress/e2e/immutableDatabaseTests/linkCheck.ts
@@ -1,7 +1,7 @@
 import { setTokenUserViewPort } from '../../support/commands';
 
 // Formats paths to be displayed in an error log
-function formatPaths(paths: string[]) {
+function formatPaths(paths: string[]): string {
   let formattedPaths = '';
   paths.forEach((path) => {
     formattedPaths += '\n- ' + path;
@@ -9,31 +9,33 @@ function formatPaths(paths: string[]) {
   return formattedPaths;
 }
 
-// Represents a page to visit on the site - organizes urlPages, imagePages
-interface Page {
-  path: string;
-  selector: string;
-}
-
 describe('Find broken anchor links', () => {
   setTokenUserViewPort();
 
-  function checkUrls(path: string, selector: string) {
+  let visitedUrls: string[] = [];
+
+  function isDynamicUrl(url: string): boolean {
+    return url.includes('/my-workflows/') || url.includes('/my-tools/');
+  }
+
+  function checkUrls(path: string) {
     let brokenUrls = [];
-    cy.visit(path);
-    cy.get(selector)
-      .find('a')
+    cy.visit(path).wait(500); // Temporary solution to ensure the page loads entirely
+    cy.get('a')
       .each((anchor) => {
         cy.get(anchor).then((anchor) => {
           const href = anchor.prop('href');
-          if (href) {
+          // Send requests to non-dynamic URLs that haven't yet been visited
+          if (href && !visitedUrls.includes(href) && !isDynamicUrl(href)) {
             cy.request({
               url: href,
               failOnStatusCode: false,
             }).then((result) => {
               if (result.status != 200) {
                 brokenUrls.push(href);
-                cy.log(result.status + ': ' + href);
+                cy.log(`${result.status}: ${href}`);
+              } else {
+                visitedUrls.push(href); // Add successful links to visitedUrls so that they won't be visited again
               }
             });
           }
@@ -46,18 +48,11 @@ describe('Find broken anchor links', () => {
       });
   }
 
-  const urlPages: Page[] = [
-    { path: '/', selector: 'app-root' },
-    { path: '/sitemap', selector: 'app-sitemap' },
-    { path: '/about', selector: 'app-about' },
-    { path: '/funding', selector: 'app-funding' },
-    { path: '/docs', selector: 'app-docs' },
-    { path: '/dashboard', selector: 'app-my-sidebar' },
-  ];
+  const urlPages: string = ['/', '/sitemap', '/about', '/funding', '/docs', '/dashboard'];
 
   urlPages.forEach((urlPage) => {
-    it(`anchor links at "${urlPage.path}" should work`, () => {
-      checkUrls(urlPage.path, urlPage.selector);
+    it(`anchor links at "${urlPage}" should work`, () => {
+      checkUrls(urlPage);
     });
   });
 });
@@ -65,16 +60,22 @@ describe('Find broken anchor links', () => {
 describe('Find broken image links', () => {
   setTokenUserViewPort();
 
+  let visitedImages: string = [];
+
   function checkImages(path: string, selector: string) {
     let brokenImages = [];
     cy.visit(path).wait(500);
-    cy.get(selector)
-      .find('img')
+    cy.get('img')
       .each((image) => {
         cy.get(image).then((image) => {
-          if (image.prop('naturalWidth') === 0) {
-            brokenImages.push(image.prop('src'));
-            cy.log('image not visible: ' + image.prop('src'));
+          const src = image.prop('src');
+          if (!visitedImages.includes(src)) {
+            if (image.prop('naturalWidth') === 0) {
+              brokenImages.push(src);
+              cy.log(`image not visible: ${src}`);
+            } else {
+              visitedImages.push(src);
+            }
           }
         });
       })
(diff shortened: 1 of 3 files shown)


denis-yuen:
> * Is it inefficient to run this test on every push?

If I'm reading https://cloud.cypress.io/projects/1ya4vj/runs/33225/test-results/instance/49cf4972-a362-4678-9f4b-8c67560e9dd7/stdout right, it seems like this runs for 1 minute 45 seconds?

ll5zh:
> 
> If I'm reading https://cloud.cypress.io/projects/1ya4vj/runs/33225/test-results/instance/49cf4972-a362-4678-9f4b-8c67560e9dd7/stdout right, it seems like this runs for 1 minute 45 seconds?

Yeah, looks like it lasted 2 minutes 3 seconds and was estimated to be 1 minute 45 seconds? Maybe it's not about efficiency, more so that it seems a bit pointless to test some of those common paths (but if it's not too costly then that might not be too important).

denis-yuen:
> False negatives (fails to find links that are broken): the link to "[Advanced Features](https://docs.dockstore.org/en/stable/advanced-topics/advanced-features.html)" at /sitemap fails locally, but succeeds in the web run. Seems like the `href` for documentation links isn't formed correctly and takes on the form null/*.

We should probably fix the link in the sitemap

coverbeck:
>     * False negatives (fails to find links that are broken): the link to "[Advanced Features](https://docs.dockstore.org/en/stable/advanced-topics/advanced-features.html)" at /sitemap fails locally, but succeeds in the web run. Seems like the `href` for documentation links isn't formed correctly and takes on the form null/*.

This should be fixable by updating https://github.com/dockstore/dockstore-ui2/blob/feature/seab-6014/test/web.yml with the `documentationUrl` property, like [here](https://github.com/dockstore/dockstore/blob/develop/dockstore-integration-testing/src/test/resources/dockstore.yml#L148).

cypress/e2e/immutableDatabaseTests/linkCheck.ts
```diff
@@ -0,0 +1,104 @@
...
+      })
+      .then(() => {
+        if (brokenImages.length) {
+          throw new Error(`Broken images at "${path}":` + formatPaths(brokenImages));
+        }
+      });
+  }
+
+  const imagePages = [
+    ['/', 'app-root'],
```
coverbeck, line 82: Why do you have the selector, `app-root`? Is it to avoid checking the same image twice? E.g., if you go to `/about`, it's a child of `app-about` is a child of `app-root`, so you don't want to check `app-root` again?

Another idea is you could have a set of visited images, so that you don't visit the same image twice.

It would make it easier to maintain -- if we add a new path, e.g., `/wonderful`, we wouldn't have to know the component name to add it to this array.

coverbeck (commented):
This looks great! Some random thoughts:

- I had been envisioning these as smoke tests, running against qa/staging/prod, but running as integration tests is probably better, because we pay more attention to PR checks than we do to smoke test failures (we shouldn't, but we do).
- I think it's OK to just test external links, with a couple of notes:
    1. docs.dockstore.org and discuss.dockstore.org should be considered external
    2. We should check images, internal and external (one of the two cases that spawned the ticket was a missing internal image).
 - You're skipping `/organizations`, which on the one hand, because there is a lot of user content, might be good -- we can't be responsible for somebody putting a bad link in the org description. On the other hand, and I didn't actually check this, maybe there is content that we're responsible for that is getting skipped?

cypress/e2e/immutableDatabaseTests/linkCheck.ts
```diff
@@ -0,0 +1,104 @@
...
+
+describe('Find broken anchor links', () => {
+  setTokenUserViewPort();
+
+  function checkUrls(path: string, selector: string) {
+    let brokenUrls = [];
+    cy.visit(path);
+    cy.get(selector)
+      .find('a')
+      .each((url) => {
```
svonworl, line 19: Suggest a different variable name than `url`.   Here, we're iterating over the elements that were found, the URL is extracted from each further down.

cypress/e2e/immutableDatabaseTests/linkCheck.ts
```diff
@@ -0,0 +1,104 @@
...
+        });
+      })
+      .then(() => {
+        if (brokenUrls.length) {
+          throw new Error(`Broken links at "${path}":` + formatPaths(brokenUrls));
+        }
+      });
+  }
+
+  const urlPages = [
```
svonworl, line 42: Could make this a typed array, that way you could refer to the properties by name (something like "path" and "selector") rather than by array index.

cypress/e2e/immutableDatabaseTests/linkCheck.ts
```diff
@@ -0,0 +1,104 @@
...
+    ['/workflows', 'app-workflows'],
+    ['/notebooks', 'app-workflows'],
+    ['/services', 'app-workflows'],
+    ['/apptools', 'app-workflows'],
+    ['/search-workflows', 'app-workflows'],
+    ['/tools', 'app-containers'],
+    ['/containers', 'app-containers'],
+    ['/search-containers', 'app-containers'],
+    ['/accounts', 'app-account-sidebar'],
+    ['/accounts', 'app-accounts-external'],
```
svonworl, line 96: What is the reason for only inspecting certain selectors?  Guessing that it's at least partially to avoid repeatedly testing the "same" links (the navigation links, for example).

svonworl (commented):
This is great!  I'm gonna advocate for checking internal links, since the bug that triggered the ticket was due to a malformed internal link.

cypress/e2e/immutableDatabaseTests/linkCheck.ts
ll5zh: I think using `app-root` also conveniently performs link checks for `app-navbar` and `app-footer`, alongside the main home component `app-home`. We could also replace it with `['/', 'app-navbar'], ['/', 'app-footer'], ['/', 'app-home']` to make it clearer.

ll5zh: Not sure if this is what you meant by a set of visited images: the test could dynamically add images to a list after they've been visited, so they can be skipped if they appear a second time? Would that significantly add to the test's runtime?

Another thought: if each path had a corresponding list of images to check, then the test could visit each by their `src` tag? This way, we could write the test to avoid duplicate checks: if an image exists on 2+ pages, it only has to be visited on one of them.

^^ Pros: This might also allow us to better filter by which images should (or shouldn't) be checked. For instance, when visiting /organizations, we might not be responsible for an organization's avatar image, but we do want to verify that the organization icon is visible.

^^ Cons: While it does remove the need for component names, you'd have to add a `src` path for every new image, so I can't tell if that makes it easier or harder to maintain. It also might be a bit of effort upfront to cumulate all the images.

cypress/e2e/immutableDatabaseTests/linkCheck.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+describe('Find broken anchor links', () => {
+  setTokenUserViewPort();
+
+  function checkUrls(path: string, selector: string) {
+    let brokenUrls = [];
+    cy.visit(path);
+    cy.get(selector)
+      .find('a')
```
kathy-t, line 18: Similar question to Steve, why are we only checking certain selectors?

If we check all links on a page, this could be simplified to:

```
cy.visit(path);
cy.get('a').each(...your  URL checking code...);
```

cypress/e2e/immutableDatabaseTests/linkCheck.ts
ll5zh: Exactly, mainly so we could avoid checking components like the navbar and footer on every iteration (might make it a bit annoying to update though).

cypress/e2e/immutableDatabaseTests/linkCheck.ts
ll5zh: One reason was to prevent checking navbar/footer links multiple times.

Another reason was to filter out "dynamic" content/links. For example, at /dashboard, the test could avoid checking the links to /my-workflows/..., /my-tools/..., etc in the entry boxes, or the links in Mastodon posts, which might be less valuable to us?

coverbeck: > One reason was to prevent checking navbar/footer links multiple times.

This was the intent of my comment about keeping track of already visited images and links (I'd only said images, which was confusing).

1. You get rid of the selector
2. You navigate to `path`
3.  You visit all links/images in the document, keeping track in a set of the ones you have visited.
4. You go to a new path
5. You get all links in the document, remove the ones you have visited, then check the status of the remaining ones.

That way you don't need the selector, need to know which component corresponds to each path, etc.

On the dynamic content links, not sure, maybe you could exclude them?

ll5zh:
Tried implementing the tracking of already visited URLs/images: if a link passes, it is added to `visitedUrls` or `visitedImages`. If the link is encountered again in the `.each` block, it can be skipped. (If a link fails, ideally we'd want to know every page that includes it + needs to be changed)

Looks like the test ran for [1 minute 5 seconds](https://cloud.cypress.io/projects/1ya4vj/runs/33436/overview/7a9cfd7b-79c6-4be2-a959-161fdf6793b1/stdout?roarHideRunsWithDiffGroupsAndTags=1), although the estimated duration is much higher than before (5 minutes 32 seconds). I think the results are as expected - the only failing link reported is https://docs.dockstore.org/en/latest/advanced-topics/advanced-features.html at /sitemap (which is actually broken).

cypress/e2e/immutableDatabaseTests/linkCheck.ts
```diff
@@ -0,0 +1,116 @@
...
+    formattedPaths += '\n- ' + path;
+  });
+  return formattedPaths;
+}
+
+describe('Find broken anchor links', () => {
+  setTokenUserViewPort();
+
+  let visitedUrls: string[] = [
+    'https://twitter.com/DockstoreOrg', // Skip, test only fails because it redirects to login page if not logged in
```
ll5zh, line 16: Bit of a band-aid solution, but I added the Twitter link here so that it doesn't cause the entire test to fail. It doesn't return the 200 status code because it redirects to Twitter's login page, but it's not really a "broken" link to be fixed.

Similar reason for adding http://localhost:4200/ to `visitedUrls` - it corresponds to the fake user's (nonexistent) avatar image in the navbar, which fails but doesn't actually affect the site's functionality/users.

kathy-t: You could make it more explicit by creating a `skipUrls` array for URLs that are intentionally skipped (similar idea for images too).

kathy-t:
> the only failing link reported is https://docs.dockstore.org/en/latest/advanced-topics/advanced-features.html at /sitemap (which is actually broken).

Is there a ticket to fix this broken link?

cypress/e2e/immutableDatabaseTests/linkCheck.ts
```diff
@@ -0,0 +1,116 @@
...
+  ];
+
+  function isDynamicUrl(url: string): boolean {
+    return url.includes('/my-workflows/') || url.includes('/my-tools/');
```
kathy-t, line 20: Could add `/my-notebooks/` too

cypress/e2e/immutableDatabaseTests/linkCheck.ts
```diff
@@ -0,0 +1,116 @@
...
+        });
+      })
+      .then(() => {
+        if (brokenUrls.length) {
+          throw new Error(`Broken links at "${path}":` + formatPaths(brokenUrls));
+        }
+      });
+  }
+
+  const urlPages: string = ['/', '/sitemap', '/about', '/funding', '/docs', '/dashboard'];
```
kathy-t, line 53: Could also add `/quick-start`

kathy-t (approved):

denis-yuen (approved):
Can either fix ink or create follow-up ticket, whatever is easier

ll5zh:
> the only failing link reported is https://docs.dockstore.org/en/latest/advanced-topics/advanced-features.html at /sitemap (which is actually broken).

Is this the correct "Advanced Features " link...? https://docs.dockstore.org/en/stable/advanced-topics/advanced-topics.html

cypress/e2e/immutableDatabaseTests/linkCheck.ts
```diff
@@ -0,0 +1,116 @@
...
+            cy.request({
+              url: href,
+              failOnStatusCode: false,
+            }).then((result) => {
+              if (result.status != 200) {
+                brokenUrls.push(href);
+                cy.log(`${result.status}: ${href}`);
+              } else {
+                visitedUrls.push(href); // Add successful links to visitedUrls so that they won't be visited again
+              }
```
svonworl, line 41: This code will only add the checked link to `visitedUrls` after a successful visit, so broken links might be checked repeatedly.  That behavior is fine, given that most links aren't broken, but curious if there was a reason for it.

svonworl (approved):

cypress/e2e/immutableDatabaseTests/linkCheck.ts
ll5zh: I think it would help us report every instance of a broken link, so that we know all the places to fix it. Like if https://docs.dockstore.org/en/stable/advanced-topics/advanced-features.html were coded manually onto multiple pages, it'll remind us to change the link on every page because tests for each page would fail separately.

svonworl: yeah, that makes sense, thanks!

coverbeck (approved):

PR #1958, opened 2024-04-03
make zsh compatible and hardcode parameter file name

denis-yuen:
**Description**
* Single quotes needed for wget on zsh
* Hard-coded parameter file name to match WDL

With provided parameter file
Screenshot from 2024-04-05 10-26-52

Without provided parameter file
Screenshot from 2024-04-05 10-27-06

**Review Instructions**
Same as previous PR https://github.com/dockstore/dockstore-ui2/pull/1940

**Issue**
More https://ucsc-cgl.atlassian.net/browse/SEAB-6273

**Security**

### DIFF
--- a/src/app/shared/launch.service.ts
+++ b/src/app/shared/launch.service.ts
@@ -78,9 +78,9 @@ export abstract class LaunchService {
   }
 
   getSharedZipString(workflowPath: string, versionName: string) {
-    return `wget -O temp.zip ${Dockstore.API_URI}${ga4ghPath}/tools/${encodeURIComponent(
+    return `wget -O temp.zip '${Dockstore.API_URI}${ga4ghPath}/tools/${encodeURIComponent(
       '#workflow/' + workflowPath
-    )}/versions/${encodeURIComponent(versionName)}/GALAXY/files?format=zip
+    )}/versions/${encodeURIComponent(versionName)}/GALAXY/files?format=zip'
 unzip temp.zip`;
   }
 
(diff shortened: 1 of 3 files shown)


coverbeck:
I looked at it some more, and it turns out it's a Z Shell thing -- more recent Mac OS versions have zsh as the default, which is why I ran into it. In zsh, the question mark is [used for globbing](https://www.techrepublic.com/article/globbing-wildcard-characters-with-zsh/). I'm able to reproduce the bug in Linux as well if I change my shell to zsh.

Nothing to change to the contents of the PR as a result of the above, just noting it. Well, maybe change the title of PR. :)

src/app/shared/launch.service.ts
```diff
@@ -91,7 +91,9 @@ unzip temp.zip`;
    */
   getPlanemoLocalInitString(workflowPath: string, versionName: string, primaryDescriptorPath: string, testParameterPath: string) {
     return (
-      this.getSharedZipString(workflowPath, versionName) + `\nplanemo workflow_job_init ${primaryDescriptorPath} -o ${testParameterPath}`
+      this.getSharedZipString(workflowPath, versionName) +
+      '\n# Note that if a parameter file has been provided by the workflow author, this will overwrite it.' +
+      `\nplanemo workflow_job_init --force ${primaryDescriptorPath} -o ${testParameterPath}`
```
coverbeck, line 96: This doesn't feel right. The IWC workflows I looked at all have a test parameter file, so this command will fail for  all of them. I'm wondering if we shouldn't just give it a hard-coded name, like Dockstore.yml.

There's something else I missed in the first review that it might be easier to discuss live as well. Perhaps after/during our deploy?

denis-yuen: The command shouldn't fail with the new `--force` flag. 
Can discuss

coverbeck: > The command shouldn't fail with the new `--force` flag. Can discuss

I did overlook that, but still wondering why we're overwriting the test parameter file by default -- wouldn't the user want to use it? But yeah, let's discuss this later.

denis-yuen: Ok, I updated since comparing with the WDL version, I realized that the WDL is actually using a hardcoded parameter file name so matching that behaviour in Galaxy seems simpler anyway than making it dynamic and match the user provided filename.

coverbeck (approved):
In your screenshots, I still find it a little irksome, as we had discussed in the huddle, that the instructions have you download/unzip twice, but I see your point about a user potentially jumping straight to the second box in the UI and skipping the first.

kathy-t (approved):

denis-yuen:
I actually messed up and didn't upload new screenshots and the changes

coverbeck (approved):
Looks good!

kathy-t (approved):

PR #1967, opened 2024-05-02
ui2 - run against webservice develop

denis-yuen:
**Description**
Moves the check_build_develop step earlier (conditionally) and controls it with a new parameter. 
This allows us to run a build nightly, checking for new issues caused by a changing webservice. 

Example of build as usual: 
https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/11796/workflows/0959c4f8-e5d1-42d4-8d0b-2f428b23a2cc

Example  of build with moved check develop step:
https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/11797/workflows/9f04c924-c765-43aa-9f68-8dcc7d5225f5/jobs/53146

On merge, will create a cron trigger, similar to what we have for the CLI. 

**Review Instructions**
Check for nightly builds, ideally should see a build fail due to an incompatible change to the webservice. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6380

**Security**
None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,7 +140,7 @@ jobs:
               command: bash -i -c 'npm config set dockstore-ui2:webservice_version develop'
           - run:
               name: set use_circle to false (openapi.yaml from github)
-              command: bash -i -c 'npm config set dockstore-ui2:use_circle true'              
+              command: bash -i -c 'npm config set dockstore-ui2:use_circle false'              
       - build_ui
       - run:
           name: Install codecov
@@ -287,7 +287,7 @@ jobs:
                 command: bash -i -c 'npm config set dockstore-ui2:webservice_version develop'
             - run:
                 name: set use_circle to false (openapi.yaml from github)
-                command: bash -i -c 'npm config set dockstore-ui2:use_circle true'                 
+                command: bash -i -c 'npm config set dockstore-ui2:use_circle false'                 
             - build_ui
 
 parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -268,6 +268,11 @@ jobs:
           username: dockstoretestuser
           password: $DOCKERHUB_PASSWORD
     steps:
+      - run: # should not need this, but circle ci does not like it when a condition wipes out a whole job https://discuss.circleci.com/t/pipeline-parameters-in-steps-when-conditions-seems-to-have-no-value-using-dynamic-configuration/50524/2
+          name: Java/Maven/Python versions
+          command: |
+            java -version
+            npm --version
       - when:
           condition:
             not: << pipeline.parameters.sanity-check-against-develop >>      
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,7 @@ jobs:
       # Override the webservice version in package.json to instead build against develop webservice
       - run:
           when:
-            condition: <<parameters.sanity_check_against_develop>>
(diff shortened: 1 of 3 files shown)


denis-yuen:
Well, that one is legit

.circleci/config.yml
```diff
@@ -261,12 +268,21 @@ jobs:
...
-          command: bash -i -c 'npm config set dockstore-ui2:webservice_version develop'
-      - build_ui
+      - run: # should not need this, but circle ci does not like it when a condition wipes out a whole job https://discuss.circleci.com/t/pipeline-parameters-in-steps-when-conditions-seems-to-have-no-value-using-dynamic-configuration/50524/2
+          name: Java/Maven/Python versions
+          command: |
+            java -version
+            npm --version
+      - when:
+          condition:
+            not: << pipeline.parameters.sanity-check-against-develop >>
```
kathy-t, line 278: I'm a little confused by the purpose of these steps if the sanity check against develop is also happening in `lint_license_unit_test_coverage` above. Could you elaborate? 

Also should that `not` be there?

.circleci/config.yml
```diff
@@ -131,6 +131,13 @@ jobs:
           command: |
             bash -i -c 'npm run circle-ci-license-test-file'
             bash scripts/detect-package-json-changes.sh
+      # Override the webservice version in package.json to instead build against develop webservice
+      - when:
+         condition: << pipeline.parameters.sanity-check-against-develop >>
+         steps:
+          - run:
+              name: set webservice to develop
+              command: bash -i -c 'npm config set dockstore-ui2:webservice_version develop'
```
kathy-t, line 140: To be safe, I think you should also set the `use_circle` boolean in the package.json to `false` otherwise it will run against the CircleCI artifact defined in the package.json regardless of the `webservice_version`. https://github.com/dockstore/dockstore-ui2/blob/98d079ab6877602b5fb4c2e6133d95c44b4f430e/scripts/generate-openapi-script.sh#L17

.circleci/config.yml
denis-yuen: The basic idea is that it occurs in one place or the other. 
It  used to occur here on, but only on the develop branch and it would only check that the build passed (`build_ui1`) with no tests being executed. When `sanity-check-against-develop` is false, nothing really changes from before. 

When `sanity-check-against-develop` is true, this step basically disappears and moves to `lint_license_unit_test_coverage` so that, hopefully, the integration tests run against develop since the workspace is being attached (which should carryover https://support.circleci.com/hc/en-us/articles/10816400480411-How-to-pass-environment-variables-between-jobs )

package.json
```diff
@@ -3,10 +3,10 @@
   "version": "2.13.0",
   "license": "Apache License 2.0",
   "config": {
-    "webservice_version": "1.16.0",
-    "use_circle": true,
-    "circle_ci_source": "https://app.circleci.com/pipelines/github/dockstore/dockstore/11135/workflows/fa998d0e-13c1-4ba2-8f6c-c535d4731e7d/jobs/42660/artifacts",
-    "circle_build_id": "42660"
+    "webservice_version": "1.16.0-alpha.10",
+    "use_circle": false,
```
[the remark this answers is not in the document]
denis-yuen: I didn't think so, isn't retrieving artifacts from CircleCI still broken?

svonworl (approved):

coverbeck (approved):

PR #1968, opened 2024-05-07
DOCK-2403: Link to individual source files

svonworl:
**Description**
This PR changes the UI so that when the user views a source file, its absolute path is added to the URL as the `file` query parameter.  The resulting URL directly references the source file and can be copied/pasted/saved for future use.

Please view this PR locally and try to make it malfunction.

Believe it or not, the private entry page does not track the selected tab in the URL or allow linking to a particular tab, like the public entry page does.  I poked around and it looks like harmonizing the behavior would be hazardous and quite a chore, so I left it as is.  Thus, the new functionality is only available on the public entry page.

When a version changes, source files can potentially disappear.  When a URL references a now non-existent file, the UI displays a message to that effect, and displays a different file.

As the team informally decided, this PR doesn't implement the "line number" part of DOCK-2403.  However, whilst coding this PR, I realized that it might be relatively easy to do.  If we add a hash of the source file's content to the URL, we can detect that the source file has changed and take appropriate action (for example, no longer highlight the specified line, because the source file content could no longer match up).  If we decide to include the "line number" functionality, I suggest we add it in a follow-up PR.

**Review Instructions**
View an entry's public page.  Click on the "Files" tab.  Confirm that the absolute path of the displayed source file has been added to the URL as the `file` query parameter.  Copy the URL.  Select a different file.  Visit the URL.  Confirm that the original file is displayed.  Paste the URL, corrupt the `file` query parameter, visit the resulting URL, and confirm that an appropriate message is displayed.  Play around with the interface in similar fashion and try to find bugs.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2403
https://github.com/dockstore/dockstore/issues/5523

**Security**
This PR changes the browser URL to include the absolute path of a source file as a query parameter.  The code appears to encode characters correctly, and source files are limited to a small subset of "safe" characters, so I don't believe there is an issue.  However, please scrutinize.

### DIFF
--- a/src/app/source-file-tabs/source-file-tabs.component.ts
+++ b/src/app/source-file-tabs/source-file-tabs.component.ts
@@ -112,7 +112,8 @@ export class SourceFileTabsComponent implements OnChanges {
             if (sourceFile.absolutePath === queryFileName) {
               this.changeTab(tabIndex);
               console.log('INDEXOF ' + sourceFiles.indexOf(sourceFile));
-              this.changeFileType(sourceFiles.indexOf(sourceFile), sourceFiles);
+              this.selectFile(sourceFile);
+              this.changeTab(tabIndex);
               return;
             }
           }
@@ -121,8 +122,9 @@ export class SourceFileTabsComponent implements OnChanges {
 
       // Otherwise, select the first file in the first tab.
       console.log('DEFAULT');
+      const files = this.fileTabs.values().next().value;
+      this.selectFile(files[0]);
       this.changeTab(0);
-      this.changeFileType(0, this.fileTabs.values().next().value);
     }
   }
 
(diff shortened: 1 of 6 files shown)


src/app/source-file-tabs/source-file-tabs.component.ts
```diff
@@ -72,28 +83,57 @@ export class SourceFileTabsComponent implements OnChanges {
...
+              this.selectFile(sourceFile);
+              this.changeTab(tabIndex);
+              return;
+            }
+          }
+        }
+      }
+      // Otherwise, select the first file in the first tab.
+      const files = this.fileTabs.values().next().value;
+      this.selectFile(files[0]);
```
coverbeck, line 119: I don't think you can assume there will be at least one file. I believe several workflow versions without any source files.

svonworl: > I don't think you can assume there will be at least one file. I believe several workflow versions without any source files.

Believe it or not, in Typescript, that'll return `undefined` if there are zero elements.

denis-yuen:
> Thus, the new functionality is only available on the public entry page.

This is fine, most users shouldn't be able to share private urls anyway, unless they belong to the same organization anyway.

hyunnaye (approved):
Tested locally and looks good! Thanks Steve!

src/app/source-file-tabs/source-file-tabs.component.ts
coverbeck: Ah, it does in JavaScript as well... I got confused with Java.

coverbeck (approved):

denis-yuen (approved):

kathy-t (approved):

david4096 (approved):

PR #1969, opened 2024-05-10
Update tar dependency

coverbeck:
**Description**
This is an alternative implementation to #1961, which updates tar by updating Akita. This PR only updates tar, since it's a major version upgrade of Akita, which is risky.

I'm not sure we could get hit by this in the browser, since we don't fetch tar files, but seems safe to do.

**Review Instructions**

```
npm ci
cat node_modules/tar/package.json| jq '.version'
"6.2.1"
# Previous line should be 6.2.1 or greater
```
**Issue**
SEAB-6438

**Security**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dockstore-ui2",
-  "version": "2.10.0",
+  "version": "2.13.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "dockstore-ui2",
-      "version": "2.10.0",
+      "version": "2.13.0",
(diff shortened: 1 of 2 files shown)


denis-yuen (approved):

kathy-t (approved):

coverbeck:
As stated in previous comment, fix didn't work. Looked at vulnerability, and my reading is that it will only affect server-side node applications, so I think we're good.

@david4096 Agree?

coverbeck:
Closing this since the PR didn't actually fix anything, and the risk is deemed accessible, as this issue would only affect server-side Node code.

PR #1970, opened 2024-05-14
redirect artifacts from circle ci to artifactory

denis-yuen:
**Description**
Instead of circle-ci (which seems to be having intermittent auth issues with artifact storage) redirect to artifactory instead.

This was a bit of an opportunity to work with both OSS Artifactory and GitHub packages

On the topic of Artifactory OSS, lacks a search endpoint (which seems petty) https://jfrog.com/help/r/jfrog-rest-apis/artifact-latest-version-search-based-on-layout that we depended on. so I needed to switch to using a Maven plugin which presumably does something else. 

On the topic of GittHub packages, the documentation is somewhat sparse and it requires authorization (uses a classic token as documented https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry ) which is inconvenient. Because we're a public project, I think that means any token from any user will work, but still slightly inconvenient when working locally.  This will make it a little harder to run builds locally since you may need to add your own token to a `~/.m2/settings` file and then fiddle with the scripts a little.

I did get GitHub packages working, so I left some of it in-place as comments just in case we are forced to switch later. 

However, for now I settled on redirecting to artifactory. 

Also that, it's a little non-obvious, but I'd guess that one of the previous use cases of pulling artifacts from circle ci was to build against a webservice branch. However, you can run a build against a specific github branch using something like 

```
  "config": {
    "webservice_version_prefix": "1.16.0",
    "webservice_version": "feature/seab-6420/modify-endpoint-description-for-new-health-checks",
    "use_artifactory": false
  },
```

**Review Instructions**
Build UI either from both github and from artifactory. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5713

**Security**
Adds context to ui2 build with read-only access to github packages for `dockstore/dockstore`


scripts/generate-openapi-script.sh
```diff
@@ -14,9 +14,12 @@ BASE_PATH="https://raw.githubusercontent.com/dockstore/dockstore/$npm_package_co
 wget --no-verbose https://repo.maven.apache.org/maven2/org/openapitools/openapi-generator-cli/${GENERATOR_VERSION}/openapi-generator-cli-${GENERATOR_VERSION}.jar -O openapi-generator-cli.jar
 rm -Rf src/app/shared/openapi
 
-if [ "$npm_package_config_use_circle" = true ]
+if [ "$npm_package_config_use_artifactory" = true ]
 then
-        OPENAPI_PATH=$(./scripts/get-circleci-artifact-url.sh "$npm_package_config_circle_build_id" openapi.yaml)
+        # note that -DremoteRepositories=github-packages:https://maven.pkg.github.com/dockstore/dockstore also works with the right credentials, loaded as -s .github/snapshot-mvn-settings.xml
+        mvn dependency:get -DremoteRepositories=https://artifacts.oicr.on.ca/artifactory/collab-snapshot -Dartifact=io.dockstore:dockstore-webservice:${npm_package_config_webservice_version_prefix}-SNAPSHOT:openapi.yaml:dist -Dtransitive=false --batch-mode -ntp
```
coverbeck, line 20: Presumably this fetches the latest snapshot? Looking at my local Maven repo, there are several SNAPSHOTS, each followed with a date/timestamp.

coverbeck (approved):

scripts/generate-openapi-script.sh
denis-yuen: This is my belief and intent, it seems a little poorly documented, but I think that is what is supposed to happen. 

For example, the second answer in https://stackoverflow.com/questions/5901378/what-exactly-is-a-maven-snapshot-and-why-do-we-need-it

.circleci/config.yml
```diff
@@ -278,16 +278,16 @@ jobs:
...
           condition:
-            not: << pipeline.parameters.sanity-check-against-develop >>      
-          steps:  
+            not: << pipeline.parameters.sanity-check-against-develop >>
+          steps:
             - get_workspace
             # Override the webservice version in package.json to instead build against develop webservice
             - run:
                 name: set webservice to develop
                 command: bash -i -c 'npm config set dockstore-ui2:webservice_version develop'
```
kathy-t, line 287: Do you still need to set the version to develop if you're using the snapshot version?

svonworl:
In the previous system, we could build the ui2 against a particular build of the webservice, which might not be at the head of a branch.  In this new version, we can only do the latter?  Unless we do some additional tagging/branching, in which case, we can probably simulate both?  The ability to do this is occasionally useful.

denis-yuen:
> In the previous system, we could build the ui2 against a particular build of the webservice, which might not be at the head of a branch. In this new version, we can only do the latter? Unless we do some additional tagging/branching, in which case, we can probably simulate both? The ability to do this is occasionally useful.

You can still do this with a hash instead of a branch-name or tag 

```
  "config": {
    "webservice_version_prefix": "1.16.0",
    "webservice_version": "1550b549c45b2f4ba7f205e2f64502d523417dcd",
    "use_artifactory": false
  },
```

Maybe I should document this somewhere, not sure where since JSON doesn't allow comments. Suggestions welcome

scripts/run-webservice-script.sh
```diff
@@ -3,14 +3,18 @@ set -o errexit
 set -o pipefail
 set -o nounset
 set -o xtrace
-if [ "$npm_package_config_use_circle" = true ]
+if [ "$npm_package_config_use_artifactory" = true ]
```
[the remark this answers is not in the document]
denis-yuen: That works too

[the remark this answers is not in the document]
denis-yuen: I guess the other question is, why don't we also grab non-snapshot from artifactory. 
But question for another day/ticket

svonworl (approved):

kathy-t (approved):

PR #1971, opened 2024-05-14
Replace @angular/flex-layout

coverbeck:
**Description**
Replaces `@angular/flex-layout` with `ngbracket/flex-layout`.

***Background***
Angular stopped updating `@angular/flex-layout` as of Angular 15. SEAB-6278 has several links with the background. This will cause us difficulties in upgrading to newer versions of Angular.

Using the module `ngbracket/flex-layout` is the path of least resistance. It is a fork of the `@angular/flex-layout` that is updated for newer versions of Angular. It seems to have decent traction in terms of npm downloads. The other main option is to go to a pure CSS implementation, using CSS Flexbox or tailwind, but those would be a bit more work. We should reconsider those for the future, but for now, to unblock us moving to an Angular 16 upgrade, I think this is best.

**Review Instructions**
Sanity check the main pages in the UI and make sure the layout looks unchanged.

**Issue**
SEAB-6278

**Security**


package.json
```diff
@@ -4,7 +4,7 @@
   "license": "Apache License 2.0",
   "config": {
     "webservice_version": "1.16.0-alpha.10",
-    "use_circle": true,
+    "use_circle": false,
```
coverbeck, line 7: Workaround for CircleCI artifact issue.

kathy-t (approved):

kathy-t:
More general question, but going forward, should we avoid using flex layout for new pages in the UI?

denis-yuen:
> More general question, but going forward, should we avoid using flex layout for new pages in the UI?

I'd guess so https://github.com/ngbracket/ngx-layout has a few linked alternatives. 

https://github.com/angular/flex-layout/issues/1426 has some discussion, @coverbeck how did you pick this particular one for the record?

coverbeck:
> More general question, but going forward, should we avoid using flex layout for new pages in the UI?

I think we should keep using it for now. We don't know what we would do as an alternative, so it's probably best to replace it all at once.

coverbeck:
> > More general question, but going forward, should we avoid using flex layout for new pages in the UI?
> 
> I'd guess so https://github.com/ngbracket/ngx-layout has a few linked alternatives.
> 
> [angular/flex-layout#1426](https://github.com/angular/flex-layout/issues/1426) has some discussion, @coverbeck how did you pick this particular one for the record?

* Option 1 was dropping flex-layout entirely, and replacing it with plain CSS. There were several posts, including variants using the tailwind.css library, on how you could do that.
* Option 2 was using a different library that was compatible with the existing. I saw 2 in the thread, the one I'm using here, and [this one](https://github.com/alessiobianchini/ng-flex-layout).

My CSS skills have faded, and the ones I had were flex layouts, so I didn't feel comfortable resolving option 1 in a short period of time. I feel it's more important to unblock the Angular upgrade; option 2 is the past of least resistance. In the future we might consider taking out flex-layout, but this is the less disruptive for now.

On why this library vs the other one, this one has a lot more activity.

svonworl (approved):

denis-yuen (approved):

PR #1975, opened 2024-05-30
Get QA Auth Smoke Tests passing

coverbeck:
**Description**
Get rid of the `--record` option when running auth tests against QA. The big con is that if the tests fail, we'll have to manually run them locally to see what the problem is, as the `--record` option lets us save videos where we can see the failures.

I spent about 8 hours trying to get the tests to pass on CircleCI and this is all I could figure out to get them passing. Here's what I tried, none of which worked; if anybody else has any other ideas, I'm happy to give them a shot first.

- Upped the CircleCI resource_class from medium (the default) to both medium+ and large.
- Tried a newer Docker image, that went from Node 16 to Node 18 (as well as updated the Java version). Not only did it not work, it also broke the [regular integration tests](https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/11935/workflows/f52d39e1-56e0-4fb1-a048-fce1b53d1f1d). I didn't bother looking into the regular failures since it didn't fix the smoke tests anyway.
- Tried a different testing reporter, `teamcity` instead of `junit`.
- Tried a new cache
- Tried restructuring the auth-tests.ts file.

Some additional observations:

- Auth tests run locally against QA with the record option, although I'm on a Mac/ARM, so it's not an exact comparison.
- The non-auth tests run on CircleCI against QA with record with no problems.

**Review Instructions**
Nightly auth tests should be passing on CircleCI.

**Issue**
SEAB-6459

**Security**


denis-yuen (commented):
I don't remember the backstory on this. 

Why do only the `test-local-no-auth`, `test-qa-no-auth` and `test-qa-auth` jobs have `--record` in the first place. i.e. we don't record in staging or prod 

Overall, not objecting, but how did you come to suspect this as a potential fix and is there a theory for why?

coverbeck:
> I don't remember the backstory on this.
> 
> Why do only the `test-local-no-auth`, `test-qa-no-auth` and `test-qa-auth` jobs have `--record` in the first place. i.e. we don't record in staging or prod

I don't know the background, but my guess is because QA is changing daily, the possibility for breakage is higher, and it's more convenient to have the artifacts so you can debug. Whereas once the tests are passing in staging/prod, it'll be months before the possibility of them breaking again happens (unless they tests themselves change).

For `test-local-no-auth`, my guess is so that we could test the Cypress Cloud/record integration.
 
> Overall, not objecting, but how did you come to suspect this as a potential fix and is there a theory for why? Or why it continues to work for those other jobs?

I looked at what's different between the passing and failing jobs and the `--record` was a thing. I was surprised it made a difference, especially given that test-qa-no-auth works with the switch. 

After that worked, I looked around, and there were some bug reports/comments of `--record` being an issue. [Here](https://github.com/cypress-io/cypress/issues/6695#issuecomment-852811773) (old ticket but still) and [here](https://github.com/cypress-io/cypress/issues/27722#issuecomment-1704504453).

Also, FWIW, I noted this in the JIRA issue but not in the PR, these failures started when Cypress 13 was merged to develop.

In any case, my theory is there is some flakiness in Cypress. :)

denis-yuen (approved):

kathy-t:
Does `staging-auth` work with recording? 

Also, this probably won't make a difference but I noticed that there's a new browser-tools version so could try updating that to 1.4.8 https://github.com/dockstore/dockstore-ui2/blob/6a4ee281ce5a18e7daf1ff7ca8b40130ae4a52c6/.circleci/config.yml#L6

coverbeck:
> Does `staging-auth` work with recording?

Good idea, that would see if the problem stems from something new in Dockstore 1.16 or if it's purely Cypress.

> Also, this probably won't make a difference but I noticed that there's a new browser-tools version so could try updating that to 1.4.8
> 
> https://github.com/dockstore/dockstore-ui2/blob/6a4ee281ce5a18e7daf1ff7ca8b40130ae4a52c6/.circleci/config.yml#L6

I think I tried that and it didn't make a difference. But I'll try again to be sure, and we might as well update anyway.

david4096 (approved):

svonworl (approved):
> "Get rid of the --record option when running auth tests against QA"

Recording the tests was causing them to fail?  If yes, is there clues/evidence as to why that's happening?

svonworl:
> > "Get rid of the --record option when running auth tests against QA"
> 
> Recording the tests was causing them to fail? If yes, is there clues/evidence as to why that's happening?

Ooops, nevermind. missed some of the comments, they discuss it.

kathy-t (approved):

coverbeck:
> Does `staging-auth` work with recording?

It also [failed](https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/11959/workflows/3f4085d7-6d1e-4c04-a944-df236d3fd9b2/jobs/53859) against staging with recording.

kathy-t:
Thanks for testing it!

PR #1976, opened 2024-06-04
DOCK-2527: Improve DOI labels

hyunnaye:
**Description**
This PR changes the labels from just "DOI" to "Concept DOI" and "Version DOI"

Before fix:
image

After fix (Note the DOI labels have changed):
image

Tooltip:
image

**Review Instructions**
Go to https://qa.dockstore.org/workflows/github.com/kathy-t/ghapps-single-workflow:master?tab=info and verify that the labels have changed

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2527

**Security**

### DIFF
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -416,7 +416,14 @@
             </div>
           </li>
           <div>
-            <strong matTooltip="Digital Object Identifier for all workflow versions">Concept DOI</strong>:
+            <strong matTooltip="Digital Object Identifier for all workflow versions">Concept DOI</strong>
+            <img
+              matTooltip="This DOI represents all versions, and will always resolve to the latest one"
+              src="../../../../assets/svg/help.svg"
+              alt="help"
+              class="doi-help-icon"
+            />
+            :
             <a
               [href]="'https://doi.org/' + workflow.conceptDoi"
               *ngIf="workflow.conceptDoi"
--- a/src/app/workflow/info-tab/info-tab.component.html
+++ b/src/app/workflow/info-tab/info-tab.component.html
@@ -416,7 +416,7 @@
             </div>
           </li>
           <div>
-            <strong matTooltip="Digital Object Identifier for all workflow versions">DOI</strong>:
+            <strong matTooltip="Digital Object Identifier for all workflow versions">Concept DOI</strong>:
             <a
               [href]="'https://doi.org/' + workflow.conceptDoi"
               *ngIf="workflow.conceptDoi"
@@ -513,7 +513,7 @@
         {{ selectedVersion.versionMetadata.engineVersions }}
       </div>
       <div *ngIf="(entryType$ | async) === EntryType.BioWorkflow">
-        <strong matTooltip="Digital Object Identifier">DOI</strong>:
+        <strong matTooltip="Digital Object Identifier">Version DOI</strong>:
         <a
           [href]="'https://doi.org/' + selectedVersion.doiURL"
           *ngIf="selectedVersion.doiURL"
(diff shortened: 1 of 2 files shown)


denis-yuen (commented):
On second though, "Concept DOI" may be a bit obscure, can we also put a question mark like 

Screenshot from 2024-06-04 17-44-56

And for the mouseover text like "You can cite all versions by using the DOI [10.5281/zenodo.592056](https://zenodo.org/doi/10.5281/zenodo.592056). This DOI represents all versions, and will always resolve to the latest one" from https://zenodo.org/records/10815677

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -513,7 +513,7 @@
         {{ selectedVersion.versionMetadata.engineVersions }}
       </div>
       <div *ngIf="(entryType$ | async) === EntryType.BioWorkflow">
-        <strong matTooltip="Digital Object Identifier">DOI</strong>:
+        <strong matTooltip="Digital Object Identifier">Version DOI</strong>:
```
coverbeck, line 516: I would just leave this as DOI, although others may disagree (please opine if you do). You're already in a box named "Workflow Version Information", and we don't repeat "version" in front of other labels.

In my mind, it just introduces confusion.. why does it say "Version DOI" but not "Version Description"?

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -416,7 +416,7 @@
             </div>
           </li>
           <div>
-            <strong matTooltip="Digital Object Identifier for all workflow versions">DOI</strong>:
+            <strong matTooltip="Digital Object Identifier for all workflow versions">Concept DOI</strong>:
```
coverbeck, line 419: Agreed with Denis' other comment about adding the circled "i".

src/app/workflow/info-tab/info-tab.component.html
svonworl: >  (please opine if you do)

I don't mind the more descriptive labels.  There's two different kinds of DOIs, and if we include "Concept" and "Version" in the labels, the user (who probably doesn't have our unique point of view and background knowledge) doesn't need to read "DOI", parse the surrounding context, and then reason to determine which type it is).

kathy-t (commented):
Was here, agree with Denis' feedback

hyunnaye:
Updated screenshots

denis-yuen (approved):

src/app/workflow/info-tab/info-tab.component.html
```diff
@@ -416,7 +416,14 @@
             </div>
           </li>
           <div>
-            <strong matTooltip="Digital Object Identifier for all workflow versions">DOI</strong>:
+            <strong matTooltip="Digital Object Identifier for all workflow versions">Concept DOI</strong>
+            <img
+              matTooltip="This DOI represents all versions, and will always resolve to the latest one"
+              src="../../../../assets/svg/help.svg"
+              alt="help"
+              class="doi-help-icon"
```
coverbeck, line 424: I believe the convention should be to use `.pl-2` instead. Instead of introducing a new class that duplicates an existing one. @kathy-t can confirm.

coverbeck (approved):

svonworl (approved):

src/app/workflow/info-tab/info-tab.component.html
kathy-t: Yes, I recommend using `pl-2` to avoid duplication https://github.com/dockstore/dockstore-ui2/blob/b4a52e7864b579354c79a12f28d732d922082384/src/bootstrap4.scss#L280-L283

PR #1977, opened 2024-06-05
Use standalone Angular

coverbeck:
**Description**
Converts to a mainly [standalone](https://v15.angular.io/guide/standalone-components) Angular app. Essentially, it allows to get rid of modules, which reduces the need for `NgModule`. The `NgModule` was confusing -- I never quite understood it, and we weren't the only ones, which is why they moved towards this simplication.

This was mainly done by running their migration tool, although manual fixup was required afterwards.

I'm doing this as a prerequisite to the Angular 16 migration. When I first tried the migration, I ran into compile errors using MatIcon not being recognized, and some links I found suggested it would more likely work with standalone components, so here we are.

Also hopefully reduced the flakiness of some tests due to unsafe usage of `scrollIntoView()`, see discussion [here](https://ucsc-gi.slack.com/archives/C05EZH3RVNY/p1717782826180929?thread_ts=1717781995.867669&cid=C05EZH3RVNY).

Finally, while working on this, smoke tests broke due to auto DOI generation. I commented out the test and logged SEAB-6508.

**Review Instructions**
Try as many pages and dialogs of the UI and make sure nothing is broken.

**Issue**
SEAB-6150

**Security**


svonworl (approved):
The previous module system confused me, so I'm glad to see it go.  That said, I'm a little concerned about the code bloat and ease of modification.  One idea for later is to come up with lists of imports that cover particular functional areas, and use them to simplify the `import` lists.  Questions is, are we recreating modules in the process?  Don't think so, but not sure, never really completely understood Angular modules, lol...

kathy-t (approved):

coverbeck:
Created a followup ticket for smoke tests: https://ucsc-cgl.atlassian.net/browse/SEAB-6508

coverbeck:
> The previous module system confused me, so I'm glad to see it go. That said, I'm a little concerned about the code bloat and ease of modification. One idea for later is to come up with lists of imports that cover particular functional areas, and use them to simplify the `import` lists. If we do that, question is, are we effectively recreating the modules functionality in the process? Don't think so, but not sure, never really completely understood Angular modules, lol...

I'm hoping this is less confusing; whereas before you had to go look for the dependency of a component in a module, or a module referenced by another module, now it's just directly declared. With services being an annoying exception.

PR #1984, opened 2024-06-27
Bring ng2-ui-auth into dockstore-ui2

coverbeck:
**Description**
Commit files from ronzeidman/ng2-ui-auth to our repo, and tweak. We've been using ng2-ui-auth as an npm dependency.

- ng2-ui-auth does not build/load with newer versions of Angular
- It has not been updated for 5 years.
-  I investigated and tried using other libraries instead,  including [angular-oauth2-oidc](https://www.npmjs.com/package/angular-oauth2-oidc) and [angular-auth-oidc-client](https://www.npmjs.com/package/angular-auth-oidc-client), but they're both OIDC-centric, and GitHub does not even support OIDC. Digression: I was incorrectly thinking OAuth and OIDC are basically the same, but they're not.
- Started to write my own code, but then I realized I was duplicating/rewriting what ng2-ui-auth already does, so,
- Just copied the files into our repo.
- That way we can tweak the files for future Angular upgrades as necessary.

More info in the issue, dockstore/dockstore#3641.

Also ran into, and hopefully fixed, flaky test issues. Latest theory is that animation starts when you click, and we need to let if finish by waiting before doing another selector. Looks promising so far, although the sample size is small.

**Review Instructions**
Ensure that logging in/out, and linking 3rd party accounts, all still work.

**Issue**
dockstore/dockstore#3641
DOCK-1475

**Security**
This is used for authN/Z, but it's the same code as before, just now committed to our repository, so risks should be no worse than before.


src/app/ng2-ui-auth/lib/ng2-ui-auth.module.ts
```diff
@@ -0,0 +1,58 @@
...
+import { LocalService } from './local.service';
+import { AuthService } from './auth.service';
+
+@NgModule({
+  imports: [HttpClientModule],
+  declarations: [],
+  exports: [],
+})
+export class Ng2UiAuthModule {
+  static forRoot(configOptions?: IPartialConfigOptions, defaultJwtInterceptor = true): ModuleWithProviders<Ng2UiAuthModule> {
```
coverbeck, line 40: This was the main change, adding the type to `ModuleWithProviders<Ng2UiAuthModule>`.

cypress/e2e/group2/myworkflows.ts
```diff
@@ -542,6 +542,7 @@ describe('Dockstore my workflows part 3', () => {
       cy.wait(1000);
       cy.get('#2-register-workflow-option').click();
       cy.contains('button', 'Next').click();
+      cy.wait(2000); // Give animation a chance to kick in.
```
coverbeck, line 545: Here and in mytools.ts, this seems to fix flaky tests in registration dialog, so far.

svonworl: yay

src/app/ng2-ui-auth/README.md
```diff
@@ -0,0 +1,4 @@
+Files in this directory are originally from https://github.com/ronzeidman/ng2-ui-auth
```
svonworl, line 1: I can't find a license in https://github.com/ronzeidman/ng2-ui-auth, although at least one linked repo is MIT-licensed.   Should probably confirm that we're allowed to use it.

Will return and do a deeper review later...

david4096 (approved):
Unfortunate situation with a fine approach. Curious if there is a reason why you opted to move the code into this repo instead of forking and modifying it? I'm fine with this as is but it seems cleaner to be able to do use the npm machinery and maintain the dependency in a different repo. Also avoids needing to change the path to the deps in the individual source files (which you've already done so it's fine).

https://docs.npmjs.com/cli/v6/configuring-npm/package-json#github-urls

coverbeck:
> Unfortunate situation with a fine approach. Curious if there is a reason why you opted to move the code into this repo instead of forking and modifying it? I'm fine with this as is but it seems cleaner to be able to do use the npm machinery and maintain the dependency in a different repo. Also avoids needing to change the path to the deps in the individual source files (which you've already done so it's fine).
> 
> https://docs.npmjs.com/cli/v6/configuring-npm/package-json#github-urls

We haven't published anything on `npm`; I'm personally unfamiliar with the process, it just seemed easier to use locally. I did publish a personal thing the central Maven repo once, and it was a bit of a hassle. No idea if the npm publishing is easier/harder. Plus I was viewing this as more of a stopgap measure, so it didn't seem like we should.

If we can get the licensing issue sorted out, it might be the better approach.

denis-yuen (approved):
Pending license situation

david4096:
@coverbeck Sorry my comment was unclear. The link I sent shows how you install directly from a github url. So in our package json we could add 

```
  "dependencies": {
    "ng2-ui-auth": "dockstore/ng-2-ui-auth#main"
  }
```

This gives us the ease of using `npm install` to get the dependencies but the files themselves are all stored on github.

coverbeck:
@david4096 , Oh, my bad. I just assumed, incorrectly, that I knew what you were talking about and didn't click your link. I do like your proposal.

If we get the license issue sorted out, I agree, we should go with your approach.

kathy-t (commented):
Was here, it sounds like the license situation was figured out and you might be going with David's approach? If so I will wait for those changes, but re-request my review if I misunderstood the plan and this PR is good as is :slightly_smiling_face:

coverbeck:
Update: I've spent over a day trying to get dockstore/ng2-ui-auth work as a library. It works when I run `ng serve`, e.g., running locally, but not when I do an `ng build` and run against those assets; I get provider and injection errors and I'm unable to find a fix. I'm sure there's a way, but it's eluding me. There's a whole deep dive into how `npm` packages work that I'm not familiar with. 

At this point, I think it's just better to go with the approach of this PR and unblock the move to Angular 16, which is more important.

The license issue was resolved in the source repo.

cypress/e2e/group2/myworkflows.ts
kathy-t: Ohh just saw this, do you have a sense of how reliable this fix is? Perhaps I should see how this fix plays out before merging https://github.com/dockstore/dockstore-ui2/pull/1991?

kathy-t (approved):

cypress/e2e/group2/myworkflows.ts
coverbeck: `wait`s always seem like a stopgap solution. I'd rather go with yours. Mine has worked 2 out of 2 times, less conclusive than your testing.

david4096:
My apologies for inadvertently leading you down that rabbit hole. I had thought it would be interchangeable

coverbeck:
> My apologies for inadvertently leading you down that rabbit hole. I had thought it would be interchangeable

No need to apologize; I think your suggestion is the right approach; it would let us stay in sync more easily with any changes to the upstream repo or forks. I wish it had worked more easily.

svonworl (approved):

PR #1988, opened 2024-07-08
Add "Configure your authorized organizations" message to My Tools page

kathy-t:
**Description**
This PR adds the "Configure your authorized organizations" message to the My Tools page so it matches the My Workflows page.

**Review Instructions**
Verify that the My Tools page has the same "Configure your authorized organizations" as the My Workflows page in the left side bar.

Click on the "Configure" link and verify that it takes you to configure the app for the correct environment. For example, if you're in QA, then it should display the GitHub OAuth app for QA.

**Issue**
dockstore/dockstore#4852
https://ucsc-cgl.atlassian.net/browse/DOCK-2136

**Security**

### DIFF
--- a/src/app/mytools/my-tool/my-tool.component.html
+++ b/src/app/mytools/my-tool/my-tool.component.html
@@ -102,7 +102,7 @@
               <small
                 >Don't see an organization?
                 <a
-                  href="https://github.com/settings/connections/applications/7ad54aa857c6503013ea"
+                  href="https://github.com/settings/connections/applications/{{ Dockstore.GITHUB_CLIENT_ID }}"
                   target="_blank"
                   rel="noopener noreferrer"
                   >Configure</a
--- a/src/app/mytools/my-tool/my-tool.component.html
+++ b/src/app/mytools/my-tool/my-tool.component.html
@@ -98,6 +98,17 @@
                 [groupEntriesObject]="groupEntriesObject$ | async"
                 [refreshMessage]="isRefreshing$ | async"
               ></app-sidebar-accordion>
+              <br />
+              <small
+                >Don't see an organization?
+                <a
+                  href="https://github.com/settings/connections/applications/7ad54aa857c6503013ea"
+                  target="_blank"
+                  rel="noopener noreferrer"
+                  >Configure</a
+                >
+                your authorized organizations.</small
+              >
             </div>
           </div>
         </mat-sidenav>
--- a/src/app/myworkflows/my-workflow/my-workflow.component.html
+++ b/src/app/myworkflows/my-workflow/my-workflow.component.html
@@ -101,7 +101,7 @@
               <small
                 >Don't see an organization?
                 <a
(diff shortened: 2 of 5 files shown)


denis-yuen (approved):

svonworl (approved):

src/app/mytools/my-tool/my-tool.component.html
```diff
@@ -98,6 +98,17 @@
                 [groupEntriesObject]="groupEntriesObject$ | async"
                 [refreshMessage]="isRefreshing$ | async"
               ></app-sidebar-accordion>
+              <br />
+              <small
+                >Don't see an organization?
+                <a
+                  href="https://github.com/settings/connections/applications/7ad54aa857c6503013ea"
```
coverbeck, line 105: I didn't know we had such a link. :)

No need to change, could be a followup ticket, could be overkill and not worth it since this will rarely come into play, but ideally the installation id would come from the config.json, so that it would work on staging and qa. And also if it ever changes -- I don't know if they're immutable or not, couldn't tell from a quick Google search.

coverbeck (approved):

src/app/mytools/my-tool/my-tool.component.html
kathy-t: This is the oauth app's client ID according to this https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#directing-users-to-review-their-access

We have access to this value here https://github.com/dockstore/dockstore-ui2/blob/e221a20669591eb9b25ebc13267eab7f2a03598c/src/app/shared/dockstore.model.ts#L50

Seems like a simple fix so I'll do it in this PR and re-request review after testing it :slightly_smiling_face:

kathy-t:
See discussion in https://github.com/dockstore/dockstore-ui2/pull/1988#discussion_r1670867271

I changed the "Configure" link so that it uses the GitHub client ID from the configuration file and redirects you to the settings page for the environment's OAuth app. Previously, it always sent you to the settings page of prod's OAuth app because the link was hardcoded.

Re-requesting review

src/app/myworkflows/my-workflow/my-workflow.component.html
```diff
@@ -101,7 +101,7 @@
               <small
                 >Don't see an organization?
                 <a
-                  href="https://github.com/settings/connections/applications/7ad54aa857c6503013ea"
+                  href="https://github.com/settings/connections/applications/{{ Dockstore.GITHUB_CLIENT_ID }}"
```
svonworl, line 104: Cool.  A bit out of scope, but if we're making this parameterizable, might as well also do it in the other places where it appears (grep for `7ad54aa857c6503013ea`)

svonworl (approved):

coverbeck (approved):
I didn't realize it was the GitHub Client id, I thought it was a different value, lol.

PR #1991, opened 2024-07-11
Use Chrome browser for Cypress tests to reduce flakiness and remove scrollIntoView

kathy-t:
**Description**
Was going to put these changes into one of my open PRs but I wanted to isolate the changes, thus this PR exists. I was running into flaky tests in mytools.ts (integration_test_3)  and myworkflows.ts (integration_test_2) so I did some investigating. This PR attempts to reduce the flakiness by running the tests using the Chrome browser instead of the Electron browser (default by Cypress). With these changes, I ran the tests 4 consecutive times and they all passed on [CircleCI](https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/12188).

We initially thought the flakiness was due to `scrollIntoView` ([slack discussion](https://oicr.slack.com/archives/C05EZH3RVNY/p1717781995867669)) but we were still running into flakiness after fixing it.

I noticed that whenever these tests did fail, it had this error message ([example failed build](https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/12175/workflows/fd4e2646-a427-41b8-bd4e-7c3a8d43eae9/jobs/55035)):
```
We detected that the Electron Renderer process just crashed.

We have failed the current spec but will continue running the next spec.

This can happen for a number of different reasons.

If you're running lots of tests on a memory intense application.
  - Try increasing the CPU/memory on the machine you're running on.
  - Try enabling experimentalMemoryManagement in your config file.
  - Try lowering numTestsKeptInMemory in your config file during 'cypress open'.
```

Once I changed it to use the Chrome browser, the tests did not fail due to this error again. Perhaps something weird was going on with the Electron browser causing it to crash. Either way, I don't believe `scrollIntoView` was the real issue.
- Side note: Is there a reason why we used the Electron browser instead of Chrome since we do install the Chrome browser?

Even though I did not encounter this crashing error again, there was some still flakiness in mytools.ts involving typing in the source code repository input field. My fix was to copy whatever myworkflows.ts was doing because it did not fail at all after I changed the browser to Chrome and that job ran at least 5 times.

I also removed the use of `scrollIntoView` because Cypress will automatically scroll elements into view if an action is performed on it ([docs](https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Actionability)). I believe we initially started using it in https://github.com/dockstore/dockstore-ui2/pull/1966 because of the flakiness in the registration modal, but the tests are passing without it now.

**Review Instructions**
Builds should pass. 

**Issue**
n/a

**Security**


cypress/e2e/group3/mytools.ts
```diff
@@ -275,11 +275,15 @@ describe('Dockstore my tools', () => {
       // Make sure page is loaded first
       cy.get('#tool-path').should('be.visible');
       cy.get('#register_tool_button').click();
+      // TODO: Fix this.  When 'Next' is clicked too fast, the next step is empty
+      cy.wait(1000);
```
kathy-t, line 279: Copied from myworkflows.ts. Not ideal to have a wait, but I think this was the main fix that allowed the test to properly type in the source code repository input.

coverbeck (approved):
Great find!

As mentioned in standup, I don't recall the reason why we chose Electron. One guess is because it's hypothetically more "neutral" and isn't tying us too closely to one particular browser. But given the grief the flakiness has caused us, I think this is fine.

svonworl (approved):
Yay

denis-yuen (approved):

denis-yuen:
> As mentioned in standup, I don't recall the reason why we chose Electron. One guess is because it's hypothetically more "neutral" and isn't tying us too closely to one particular browser. But given the grief the flakiness has caused us, I think this is fine.

By default, I would think Neutron is Chromium under the hood, meaning very similar to Chrome anyway. 
So would think this is a no-op essentially, but in practice if it is more stable :shrug:

PR #1998, opened 2024-08-06
DOCK-2137: fix checker workflow field refresh issue

hyunnaye:
**Description**
This PR fixes the issue where the checker workflow field briefly disappears when toggling between the different topic sentences.

**Review Instructions**
Go to one of your workflows and change the topic sentence from Automatic to Manual. The checker workflow field should not disappear.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2137

**Security**

### DIFF
--- a/src/app/shared/entry/info-tab-checker-workflow-path/info-tab-checker-workflow-path.component.html
+++ b/src/app/shared/entry/info-tab-checker-workflow-path/info-tab-checker-workflow-path.component.html
@@ -15,7 +15,7 @@
       >
         <mat-icon class="mat-icon-sm">visibility</mat-icon> View
       </a>
-      <span *ngIf="(isPublic$ | async) === false">
+      <span *ngIf="(isPublic$ | async) === false && (isRefreshing$ | async) === false">
         <button
           id="addCheckerWorkflowButton"
           #addCheckerWorkflowButton
--- a/src/app/shared/entry/info-tab-checker-workflow-path/info-tab-checker-workflow-path.component.html
+++ b/src/app/shared/entry/info-tab-checker-workflow-path/info-tab-checker-workflow-path.component.html
@@ -1,4 +1,4 @@
-<li *ngIf="((checkerId$ | async) || (canAdd$ | async)) && (isRefreshing$ | async) === false">
+<li *ngIf="(checkerId$ | async) || (canAdd$ | async)">
   <form #editCheckerWorkflowPathForm="ngForm" class="form-inline">
     <div class="form-group" fxFlex="noshrink" fxLayout fxLayoutAlign=" center">
       <strong matTooltip="Checker workflow of this tool/workflow that tests it">Checker Workflow</strong>:


coverbeck (commented):
I tried running this locally, and I still see a flash. Does anybody else?

It may not be worth fixing, but just want to be sure I'm looking at things correctly.

kathy-t:
> I tried running this locally, and I still see a flash. Does anybody else?

It doesn't flash for me locally

src/app/shared/entry/info-tab-checker-workflow-path/info-tab-checker-workflow-path.component.html
```diff
@@ -1,4 +1,4 @@
-<li *ngIf="((checkerId$ | async) || (canAdd$ | async)) && (isRefreshing$ | async) === false">
+<li *ngIf="(checkerId$ | async) || (canAdd$ | async)">
```
kathy-t, line 1: The `Add` button on line 19 should also have ` && (isRefreshing$ | async) === false` in its `ngIf` condition so that users can't click it while the page is refreshing.

denis-yuen (approved):
Seems  ok on my end with caveat above

svonworl (approved):
Not sure I'm testing it right, but running locally, when I view one of my workflows and change the topic's source from (Manual to Automatic, etc) in the Topic "Edit" dialog, the "Checker Workflow" line does not appear to blink.

Related thought: some bad behaviors will only be apparent in situations where there high latency between the browser and the dockstore api, long enough that the flickering (or whatever) spans enough clicks that it registers visually.  Someday, we might try to come up with a testing setup wherein we simulate high round trip latency (other side of world, 300ms ping?), and run our UI through its paces to discover areas that need improvement.

(We did this at an old startup of mine to test our rich-media Java banner units, it would help to ensure that the banners were usable for everyone, everywhere, and occasionally turn up some pretty funky race-condition-type stuff).

kathy-t (approved):

hyunnaye:
> Not sure I'm testing it right, but running locally, when I view one of my workflows and change the topic's source from (Manual to Automatic, etc) in the Topic "Edit" dialog, the "Checker Workflow" line does not appear to blink.
> 
> Related thought: some bad behaviors will only be apparent in situations where there high latency between the browser and the dockstore api, long enough that the flickering (or whatever) spans enough clicks that it registers visually. Someday, we might try to come up with a testing setup wherein we simulate high round trip latency (other side of world, 300ms ping?), and run our UI through its paces to discover areas that need improvement.
> 
> (We did this at an old startup of mine to test our rich-media Java banner units, it would help to ensure that the banners were usable for everyone, everywhere, and occasionally turn up some pretty funky race-condition-type stuff).

It should be when you save the topic. So the Checker Workflow line shouldn't blink after you saved the changes.

denis-yuen:
> Someday, we might try to come up with a testing setup wherein we simulate high round trip latency (other side of world, 300ms ping?), and run our UI through its paces to discover areas that need improvement.

If manual, we sometimes use https://developer.chrome.com/docs/devtools/network/reference/#throttling

svonworl:
> > Someday, we might try to come up with a testing setup wherein we simulate high round trip latency (other side of world, 300ms ping?), and run our UI through its paces to discover areas that need improvement.
> 
> If manual, we sometimes use https://developer.chrome.com/docs/devtools/network/reference/#throttling

Cool, looks like you can set the latency in a "Custom" throttling profile, thanks!

coverbeck:
OK, I have wide-ish screen and was only focusing on the right portion of the screen, where you click to edit the topic. The View with an eye icon for the checker workflow does disappear then reappear, both with and without the fix. And that's what I was referring to, and incorrectly got it mixed in with the line as a whole.

The `Checker Workflow` label does remain and that's what the bug was written up against.

coverbeck (approved):

PR #2001, opened 2024-08-07
Display engine version for Nextflow workflows

kathy-t:
**Description**
This PR displays the Engine Versions column instead of Language Versions for Nextflow workflows.

Screenshot 2024-08-07 at 10-53-16 Dockstore Workflow github com_nf-core_rnaseq

**Review Instructions**
On QA, navigate to https://qa.dockstore.org/workflows/github.com/nf-core/rnaseq:dev?tab=versions and verify that there's an Engine Versions column and no Language Versions column for Nextflow workflows. Go to a non-Nextflow workflow, click the Versions tab and verify that it has a Language Versions column.

**Issue**
https://github.com/dockstore/dockstore/issues/5918
https://ucsc-cgl.atlassian.net/browse/DOCK-2536

**Security**

### DIFF
--- a/src/app/workflow/versions/versions.component.html
+++ b/src/app/workflow/versions/versions.component.html
@@ -118,6 +118,24 @@
       </td>
     </ng-container>
 
+    <ng-container matColumnDef="engineVersions">
+      <th
+        scope="col"
+        mat-header-cell
+        *matHeaderCellDef
+        mat-sort-header
+        matTooltip="The engine version used by the version's descriptor file(s)."
+        matTooltipPosition="above"
+      >
+        Engine Versions
+      </th>
+      <td mat-cell *matCellDef="let version">
+        <div *ngFor="let engineVersion of version.versionMetadata?.engineVersions">
+          <div class="no-wrap">{{ version.versionMetadata?.engineVersions }}</div>
+        </div>
+      </td>
+    </ng-container>
+
     <ng-container matColumnDef="valid">
       <th
         scope="col"
--- a/src/app/workflow/versions/versions.component.html
+++ b/src/app/workflow/versions/versions.component.html
@@ -130,9 +130,9 @@
         Engine Versions
       </th>
       <td mat-cell *matCellDef="let version">
-        <div *ngFor="let engineVersion of version.versionMetadata?.engineVersions">
-          <div class="no-wrap">{{ engineVersion }}</div>
-        </div>
+        <span>{{
+          workflow.descriptorType | descriptorLanguage | descriptorLanguageVersions: version.versionMetadata?.engineVersions:false
+        }}</span>
       </td>
     </ng-container>
 
--- a/src/app/workflow/versions/versions.component.html
+++ b/src/app/workflow/versions/versions.component.html
@@ -131,7 +131,7 @@
       </th>
       <td mat-cell *matCellDef="let version">
         <div *ngFor="let engineVersion of version.versionMetadata?.engineVersions">
-          <div class="no-wrap">{{ version.versionMetadata?.engineVersions }}</div>
+          <div class="no-wrap">{{ engineVersion }}</div>
         </div>
       </td>
     </ng-container>
--- a/src/app/workflow/versions/versions.component.html
+++ b/src/app/workflow/versions/versions.component.html
@@ -124,7 +124,7 @@
         mat-header-cell
         *matHeaderCellDef
         mat-sort-header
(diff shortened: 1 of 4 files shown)


src/app/workflow/versions/versions.component.html
```diff
@@ -118,6 +118,24 @@
...
+        *matHeaderCellDef
+        mat-sort-header
+        matTooltip="The engine version used by the version's descriptor file(s)."
+        matTooltipPosition="above"
+      >
+        Engine Versions
+      </th>
+      <td mat-cell *matCellDef="let version">
+        <div *ngFor="let engineVersion of version.versionMetadata?.engineVersions">
+          <div class="no-wrap">{{ engineVersion }}</div>
```
kathy-t, line 134: Made this `no-wrap` otherwise it wraps the nextflow engine version in an unfriendly way like below

Screenshot 2024-08-07 at 11-35-01 Dockstore Workflow github com_nf-core_rnaseq

When there are multiple engine versions, they are displayed one per line

denis-yuen (approved):
Screenshots look good

src/app/workflow/versions/versions.component.ts
```diff
@@ -89,6 +89,7 @@ export class VersionsWorkflowComponent extends Versions implements OnInit, OnCha
   workflow: ExtendedWorkflow;
   entryType = EntryType;
   DoiInitiatorEnum = Doi.InitiatorEnum;
+  engineVersionsList: string[] = ['Nextflow !>=22.10.1', 'Nextflow !>=23.04.0'];
```
denis-yuen, line 92: A little curious about this hard coding, this doesn't look like a test at first glance

kathy-t: Oops, my mistake, used this for local testing but forgot to remove it. Removed now :slightly_smiling_face:

svonworl (approved):

src/app/workflow/versions/versions.component.html
coverbeck: To avoid the repetition and because we're used to seeing the most significant part of a phrase on the left of text (in left-to-right languages), I wonder if we should strip `Nextflow` for the version column values, e.g., just show `!>=22.10.1` instead of `Nextflow !>=22.10.1`. It would also make the table a little less crowded.

I think it still makes sense to have the full string, with the `Nextflow` word on the facets page, as there are all types of workflows on the search page, whereas for this table the column is only visible on a Nextflow workflow.

src/app/workflow/versions/versions.component.html
```diff
@@ -118,6 +118,24 @@
       </td>
     </ng-container>
 
+    <ng-container matColumnDef="engineVersions">
+      <th
+        scope="col"
+        mat-header-cell
+        *matHeaderCellDef
+        mat-sort-header
+        matTooltip="The engine version used by the version's descriptor file(s)."
```
coverbeck, line 127: Nit, I suggest "The minimum required Nextflow version." or "The minimum required Nextflow engine version". See https://www.nextflow.io/docs/latest/config.html#scope-manifest

coverbeck (approved):

src/app/workflow/versions/versions.component.html
kathy-t: Seems intentional that we're storing the engine name based on this comment 

https://github.com/dockstore/dockstore/blob/ee8503a70285afdc8fcfdb70580c59cdf8130889/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextflowHandler.java#L492-L494

Proposing we leave it as is for now and create a new ticket if it becomes an issue since it'll require a webservice change

coverbeck: Lol, citing my own comment against my argument. :) 

I'd still argue that it's redundant on this page for Nextflow. My suggestion, which wasn't clear, would be to have a pipe here, not to change it on the backend -- I think it still makes sense to have `Nextflow !>=22.10.1` on the facets page for differentiation when we add other engines for other languages. But here it's already a Nextflow workflow, and there are no other engines for Nextflow other than Nextflow itself, so it's not really adding anything and slightly getting in the way of viewing the significant info.

coverbeck (approved):
Per my comment, I still feel we should drop `Nextflow`, but I don't feel strongly enough to block, and if you still think we should keep it, I'm good with that.

kathy-t:
Updated so that 'Nextflow' is removed from the engine version value for the Versions tab.

image

coverbeck (approved):

PR #2008, opened 2024-08-20
DOCK-2557: Adds topic sentences in collection page

hyunnaye:
**Description**
This PR adds topic sentences to each entry in the collection page

Before:
image

After:
image

AI Bubble
image

**Review Instructions**
Go to https://qa.dockstore.org/organizations/BroadInstitute and verify that there are topic sentences for each entry.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2557

**Security**

### DIFF
--- a/src/app/organizations/collection/collection.component.html
+++ b/src/app/organizations/collection/collection.component.html
@@ -108,6 +108,7 @@
                         {{ entry.entryPath + (entry.versionName ? ':' + entry.versionName : '') }}
                       </h5>
                     </div>
+                    <span>{{ entry.topic }}</span>
                     <div *ngIf="canEdit$ | async" fxLayout="row" fxLayoutAlign="end center">
                       <span>
                         <button
--- a/src/app/organizations/collection/collection.component.html
+++ b/src/app/organizations/collection/collection.component.html
@@ -133,7 +133,12 @@
                       </span>
                     </div>
                   </div>
-                  <span>{{ entry.topic }}</span>
+                  <div class="truncate-text-2">
+                    <app-ai-bubble *ngIf="entry.topicSelection === TopicSelectionEnum.AI && !entry.isApprovedAITopic"></app-ai-bubble>
+                    <span class="size-small" [matTooltip]="entry.topic" matTooltipPosition="left">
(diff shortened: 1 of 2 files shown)


src/app/organizations/collection/collection.component.html
```diff
@@ -133,6 +133,7 @@ <h5>
                       </span>
                     </div>
                   </div>
+                  <span>{{ entry.topic }}</span>
```
kathy-t, line 136: Suggesting that we make the styling similar to the search page (smaller text, truncate to two lines with the full topic displayed in a tooltip).

We should also make sure we display the AI bubble if it's an unapproved AI topic

image

coverbeck (approved):
Approved, pending addressing of Kathy's comment.

src/app/organizations/collection/collection.component.html
coverbeck, line 136: On further reflection, shouldn't this be causing a build error until dockstore/dockstore#5980 is merged?

If so, the dependency might be worth clarifying in the PR description.

coverbeck (commented):
Left an additional comment.

src/app/organizations/collection/collection.component.html
svonworl: > On further reflection, shouldn't this be causing a build error until [dockstore/dockstore#5980](https://github.com/dockstore/dockstore/pull/5980) is merged?
> 
> If so, the dependency might be worth clarifying in the PR description.

Would be great if there was a lightweight way to point the PR at a version of the webservice that supported the new `topic` field, so that reviewers could easily do things like check out the branch and test it.

svonworl (approved):

src/app/organizations/collection/collection.component.html
denis-yuen: Think we have part of this with https://github.com/dockstore/dockstore-ui2?tab=readme-ov-file#build pointing at branches or commits. I think the next part would be to upload artifacts to artifactory from branches other than develop

denis-yuen (approved):

kathy-t (approved):

PR #2016, opened 2024-09-09
DOCK-2469: Display "Warnings" indicator in lambda logs

svonworl:
**Description**
This PR changes the UI to display the text "Warnings" in the "Success" column of the lambda logs, when the event is successful (`success` == true) and a message is included.  This approach works because currently, lambda log messages are composed solely of warnings and errors.

In the longer term, we might want to implement a more "correct"/robust solution, which might involve collapsing the four existing "outcomes" of a lambda event ("Success", "Failed", "Warnings", and "Ignored") into a single enum/column.

But the latter is a lot more work than what's implemented here, which works fine in the meantime and helps us to hit our 1.16 release date.

**Review Instructions**
View the lambda logs, and confirm that for events that were successful but have a message, "Warnings" is displayed in the "Success" column.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2469

**Security**
No concerns.

### DIFF
--- a/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.html
+++ b/src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.html
@@ -62,7 +62,7 @@
             </td>
           </div>
           <div *ngSwitchCase="'success'">
-            <th mat-header-cell *matHeaderCellDef mat-sort-header>Success</th>
+            <th mat-header-cell *matHeaderCellDef mat-sort-header>Status</th>
             <td mat-cell *matCellDef="let element">
               {{
                 element.ignored ? 'Ignored' : element.success && element.message ? 'Warnings' : (column | mapFriendlyValue: element[column])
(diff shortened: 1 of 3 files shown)


src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.html
```diff
@@ -63,7 +63,11 @@ <h1 mat-dialog-title>GitHub App Logs</h1>
           </div>
           <div *ngSwitchCase="'success'">
             <th mat-header-cell *matHeaderCellDef mat-sort-header>Success</th>
```
kathy-t, line 65: Wondering if we should rename the Success column to Status because there are 4 possible values now

denis-yuen (approved):

coverbeck (approved):

src/app/myworkflows/sidebar-accordion/github-apps-logs/github-apps-logs.component.html
svonworl: Renamed "Success" column to "Status".

PR #2021, opened 2024-09-18
Smoke Tests fixes

coverbeck:
**Description**
Auth smoke tests were failing on QA. Fixing it led to a chain of other things. The original problem was due to the versions table in QA now requiring scrolling at a lower resolution with the addition of the DOI column. Interacting with the Actions dropdown in Cypress near the edge of the viewport seemed to be flaky.  This led to:

* Increasing the resolution that tests are run in. See Slack [discussion](https://ucsc-gi.slack.com/archives/C05EZH3RVNY/p1726595220382829). Although this punts on the scrolling issue, we were also running at an odd resolution, so why fight it.
* Some of the integration tests started failing after increasing the resolution, sigh. I think some of our tests are just flaky, and changing the resolution moved the flakiness around. Hopefully reduced flakiness:
    * Waiting for certain Ajax requests to complete before proceeding
    * Use `trackBy` in some template loops; this causes DOM elements to be reused instead of recreated. This fixed some Cypress errors about DOM elements no longer being present.
* If the tests failed, the resetDockstoreTestUser4 step would not get executed. This would then cause the next run to fail, etc. Always have it execute, and for good measure, execute it before running the tests (in case somebody runs the tests from a local box and doesn't reset).

**Review Instructions**
Ensure that all the smoke tests are passing nightly for a couple of days in a row on [CircleCI](https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2?branch=develop).

**Issue**
SEAB-6674

**Security**


cypress/e2e/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -228,7 +227,7 @@ function testTool(registry: string, repo: string, name: string) {
   });
 }
 
-function testWorkflow(registry: string, repo: string, name: string) {
```
coverbeck, line 231: These parameters were not being used, so I just removed them. Their existence was throwing me off as I was trying to understand the code.

svonworl (approved):

denis-yuen (approved):

PR #2025, opened 2024-09-26
SEAB-6508: Adjust smoke tests for new auto-DOI generation feature

svonworl:
**Description**
This PR attempts to adjust the smoke tests so that they do not create lots of "orphaned" Zenodo DOIs, by tweaking the tests so that a test workflow is not restubbed in the prod tests, preventing new tag-backed versions from being created and each triggering the generation of a new DOI.  The assumption here is that we don't want to create lots of orphaned production Zenodo DOIs.  If we decide that some degree of orphaning is acceptable, we might partially or completely reverse this change.

This PR reflects a pragmatic but imperfect-balancing of the various tradeoffs, and is probably good enough for now.

Originally, I had planned to tweak the smoke tests so that in prod, the "refresh from stub" path used a tagless fork of the currently-used repo.  However, that approach would have required a lot of new code, and would have been vulnerable to us accidentally adding tags to the tagless repo.

The drawback of this approach is that we're doing less smoke testing on the "refresh from stub" path.  I believe that between the integration tests, the remaining smoke tests in prod, and the full smoke tests in staging, that we've still got enough test coverage to flag the vast majority of failure modes.

This PR represent a first swing in what will be an ongoing game of whack-a-mole.  It's easy to imagine how we might mistakenly modify the smoke tests and start generating undesired DOIs, and perhaps be unaware of it until the community started to fill up with them. 

To foreclose that possibility, down the road, we might consider a different solution.  Perhaps, we could modify the webservice to not generate production Zenodo DOIs for a version, if said version corresponds to one of our testing users or repositories?  We don't like putting "environment-specific" code in the webservice, so it'd be hard to sell, but the advantage would be 100% certainty that we weren't creating a bunch of DOIs inadvertently, no matter what we put in the smoke tests. 

**Review Instructions**
After DOI generation is enabled on prod, and the smoke tests have run, check our community in production Zenodo and make sure that it's not filling with DOIs that were generated by the smoke tests.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6508

**Security**

### DIFF
--- a/cypress/e2e/smokeTests/sharedTests/auth-tests.ts
+++ b/cypress/e2e/smokeTests/sharedTests/auth-tests.ts
@@ -40,6 +40,11 @@ function storeToken() {
   }
 }
 
+function hasSandboxZenodo() {
+  const baseUrl = Cypress.config('baseUrl');
+  return baseUrl === 'https://qa.dockstore.org' || baseUrl === 'https://staging.dockstore.org' || baseUrl === 'https://dev.dockstore.net';
+}
+
 function unpublishTool() {
   it('unpublish the tool', () => {
     storeToken();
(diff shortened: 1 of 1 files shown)


denis-yuen (approved):

cypress/e2e/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -40,6 +40,11 @@ function storeToken() {
   }
 }
 
+function hasSandboxZenodo() {
+  const baseUrl = Cypress.config('baseUrl');
+  return baseUrl === 'https://qa.dockstore.org' || baseUrl === 'https://staging.dockstore.org' || baseUrl === 'https://dev.dockstore.net';
```
coverbeck, line 45: The develop branch runs of the smoke tests run against 1.15 prod, so this means we won't be restubbing on prod as soon as this is merged.

Is that the intent? It may be tricky to manage if we want to keep the tests running as is until 1.16 is deployed, so this could be OK, just making sure.

coverbeck (approved):

cypress/e2e/smokeTests/sharedTests/auth-tests.ts
svonworl: > The develop branch runs of the smoke tests run against 1.15 prod, so this means we won't be restubbing on prod as soon as this is merged.
> 
> Is that the intent? It may be tricky to manage if we want to keep the tests running as is until 1.16 is deployed, so this could be OK, just making sure.

Ok with me, it's a side affect of our smoke testing scheme that it's probably not worth working around, in this case.

kathy-t (approved):

PR #2027, opened 2024-10-03
SEAB-6696: Improve search result presentation

svonworl:
**Description**
This PR makes some relatively-minor changes to the presentation of Dockstore's Search page results, with the overarching goal of making each hit's topic sentence more prominent. readable, and completely-displayed.

This Slack thread contains some of the related discussion:
https://ucsc-gi.slack.com/archives/C05EZH3RVNY/p1727899641002089

Specifically, this PR:

* increases topic font size
* removes "Verified" and "Links" columns
* decreases "Author" value font size and column width
* tightens line height
* fixes vertical alignment of authors block and horizontal alignment of star counts
* increases truncation limit of topic and authors values to three lines
* adds a small "entry" icon in front of each hit
* fixes a smattering of inconsistencies

The net effect is to widen the column that contains the topic sentence, thus allowing more of it to be displayed, in a larger font. 

Screenshots of non-keyword results, before PR ([dockstore.org](http://dockstore.org/)) and after (localhost):
dump 2024-10-03 at 9 41 05 AM
dump 2024-10-03 at 9 43 23 AM

Screenshots of keyword results, before PR ([dockstore.org](http://dockstore.org/)) and after (localhost):
dump 2024-10-03 at 9 41 35 AM
dump 2024-10-03 at 9 42 46 AM

There's another set of pages that list all of the workflows/tools/notebooks (for example: https://dockstore.org/workflows), which this PR doesn't currently change.  Although these pages are public, we generally don't point at or advertise them.  Should we update them with the same changes as we made above?  Not sure.  It could be useful to leave them the way they are, to have more info to aid with debugging, for example...

Longer term, IMHO, we should probably format the search results as "cards", in a style that's unified with the current collection entry scheme.  After this PR merges, I will write a ticket which describes that work.
 
**Review Instructions**
Go to the search page on qa, click on each of the "Workflows", "Tools", and "Notebooks" tabs, and confirm that the changes appear and look fine.  Inspect similarly after clicking a facet, and searching for a keyword.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6696

**Security**

### DIFF
--- a/src/app/search/search-notebook-table/search-notebook-table.component.html
+++ b/src/app/search/search-notebook-table/search-notebook-table.component.html
@@ -54,7 +54,7 @@
       <mat-header-cell *matHeaderCellDef mat-sort-header start="desc">Stars</mat-header-cell>
       <mat-cell class="starred-cell weight-bold" *matCellDef="let notebook"
         ><mat-icon class="star-icon" *ngIf="notebook?.source.starredUsers?.length > 0">star_rate</mat-icon>
-        {{ !notebook?.source.starredUsers || notebook?.source.starredUsers.length === 0 ? '' : notebook?.source.starredUsers?.length }}
+        {{ !notebook?.source.starredUsers || notebook?.source.starredUsers.length === 0 ? '' : notebook?.source.starredUsers.length }}
       </mat-cell>
     </ng-container>
     <!-- Search highlighting column - this row is made up of this one column that spans across all columns -->
--- a/src/app/search/search-notebook-table/search-notebook-table.component.html
+++ b/src/app/search/search-notebook-table/search-notebook-table.component.html
@@ -25,14 +25,11 @@
     </ng-container>
     <ng-container matColumnDef="all_authors">
       <mat-header-cell fxShow fxHide.lt-sm *matHeaderCellDef mat-sort-header>Author</mat-header-cell>
-      <mat-cell
-        fxShow
-        fxHide.lt-sm
-        class="duration-cell truncate-text-2"
-        *matCellDef="let notebook"
-        [matTooltip]="notebook?.source.all_authors | getSearchAuthorsHtml: false"
-        [innerHTML]="notebook?.source.all_authors | getSearchAuthorsHtml"
-      ></mat-cell>
+      <mat-cell fxShow fxHide.lt-sm *matCellDef="let notebook" [matTooltip]="notebook?.source.all_authors | getSearchAuthorsHtml: false">
+        <div class="truncate-text-3 size-small">
+          <span [innerHTML]="notebook?.source?.all_authors | getSearchAuthorsHtml"></span>
+        </div>
+      </mat-cell>
     </ng-container>
     <ng-container matColumnDef="descriptorType">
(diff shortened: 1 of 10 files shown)


src/app/search/search-notebook-table/search-notebook-table.component.html
```diff
@@ -54,19 +50,11 @@
...
-      </mat-cell>
-    </ng-container>
     <ng-container matColumnDef="starredUsers">
       <mat-header-cell *matHeaderCellDef mat-sort-header start="desc">Stars</mat-header-cell>
-      <mat-cell class="description-cell" *matCellDef="let notebook"
-        >{{ !notebook?.source.starredUsers || notebook?.source.starredUsers.length === 0 ? '' : notebook?.source.starredUsers?.length }}
-        <mat-icon class="star-icon" *ngIf="notebook?.source.starredUsers?.length > 0">star_rate</mat-icon>
+      <mat-cell class="starred-cell weight-bold" *matCellDef="let notebook"
+        ><mat-icon class="star-icon" *ngIf="notebook?.source.starredUsers?.length > 0">star_rate</mat-icon>
+        {{ !notebook?.source.starredUsers || notebook?.source.starredUsers.length === 0 ? '' : notebook?.source.starredUsers?.length }}
```
coverbeck, line 57: Nit: you don't need the question marks on the far right `notebook?.source.starredUsers?.length` because you already checked they exist earlier in the expression. It doesn't hurt obviously, but it makes it slightly less readable, to me, because I was trying to figure out why the checks were there.

coverbeck (approved):
Definitely an improvement in the UI!

kathy-t (approved):

kathy-t:
> There's another set of pages that list all of the workflows/tools/notebooks (for example: https://dockstore.org/workflows), which this PR doesn't currently change. Although these pages are public, we generally don't point at or advertise them. Should we update them with the same changes as we made above? 

I'm okay leaving them as is since we don't display topics there and it's not advertised to users. We could overhaul them if we want to when we transition to cards, but for now it doesn't seem worth it

src/app/search/search-notebook-table/search-notebook-table.component.html
svonworl: This is code I copied/adapted from the workflows template because the previous code didn't format consistently (when I first wrote it, I probably copied it from the wrong place).   Since it's a tiny change, I will make it in each of the templates, to be consistent.  Could be written a lot better, though.

Makes me pine for a framework which makes it easy to define functions that generate little HTML snippets that you can use wherever.  Part of the reason this code isn't already in a pipe is because creating and deploying an Angular pipe is a pain in the butt, relatively speaking...

david4096 (approved):

denis-yuen:
> Should we update them with the same changes as we made above? Not sure. It could be useful to leave them the way they are, to have more info to aid with debugging, for example...

Personally, I like to have them as
1) a back-up for developer debugging when elastic search is down or potentially bugged
2) help google, but maybe not so important if sitemap is working (I haven't checked whether it is working)

Not sure that answers the question, but may help

PR #2031, opened 2024-10-15
DOCK-2592: Fix faceted search for complex queries involving ORs

svonworl:
**Description**
This PR changes the UI's Elasticsearch query-construction code to add OR subqueries in such a way that they produce correct results when included in complex queries.

In an ES query, `must` clauses corresponds to ANDs, and `should` clauses correspond to ORs.  We assemble a query as a collection of `musts` and `shoulds`, the `musts` corresponding to a facet with a single value selected, and the `shoulds` to a facet with multiple values selected.

Things work more or less as you'd expect, except for a quirk: if a query contains both `musts` and `shoulds`, the `shoulds` are effectively ignored for hit selection purposes (used only for ranking), and the query returns all hits that match the `musts`.  For example, in the case of the bug reported in the ticket, the query returned all Nextflow workflows, because there were multiple organization values selected in the facet, so the organizations were effectively ignored.

So, essentially, prior to this PR, all complex queries involving a combination of singly and multiply-selected facets were broken, and such queries returned all hits matching the singly-selected facets.  Probably affected keyword searches in combination with a multiply-selected facet, too.

The solution was to nest the OR (`should`) subqueries in their own `bool` filter, requiring a rewrite of the query construction logic.

In a nutshell: `musts` at the same level of `shoulds` override the `shoulds`, so we nest the `shoulds` in their own subfilter to avoid that.

I didn't totally grok the "Return some garbage filter" code, so I left it in there, functionally unchanged. 

We might be able to use a lightly-massaged version of the "OR" (multiple-value) query construction code to handle the "AND" (single-value) case, too.  But, that'd be a big enough change that I'd worry about unforeseen performance/functional gotchas, so for safety sake, we'll leave it the olde way for now...

This PR is targetted to the release branch.  It also contains the "AI Bubble FAQ link" change, which didn't get merged to develop in time to make it into the release.

**Review Instructions**
On staging, try the search in the ticket, and confirm that the results look good.  Play with the facets in similar ways and confirm the behavior is correct.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2592
dockstore/dockstore#6020
https://ucsc-cgl.atlassian.net/browse/SEAB-6308

**Security**

### DIFF
--- a/src/app/search/query-builder.service.ts
+++ b/src/app/search/query-builder.service.ts
@@ -180,19 +180,18 @@ export class QueryBuilderService {
         values.forEach((value) => {
           body = body.notFilter('term', 'some garbage term that hopefully never gets matched', value);
         });
-      } else if (values.size > 1) {
-        // Add a filter that matches one or more of multiple values
+      } else if (values.size == 1) {
+        // Add a filter that matches a single value
+        const [value] = values;
+        const convertedValue = isExclusiveFilter ? this.convertIntStringToBoolString(value) : value;
+        body = body.filter('term', key, convertedValue);
+      } else {
+        // Add a filter that matches at least one of multiple values
         body = body.filter('bool', (b) => {
           for (const value of values) {
             b = b.orFilter('term', key, value);
           }
           return b;
-        });
-      } else {
-        // Add a filter that matches a single value
-        values.forEach((value) => {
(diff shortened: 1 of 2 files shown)


coverbeck (approved):
In the PR Issue section, I suggest adding a link to the other issue you're fixing via cherry-pick.

kathy-t (approved):

PR #2035, opened 2024-10-24
DOCK-2594: Fix TRS ID "copy button" bug

svonworl:
**Description**
This PR fixes the "copy" button (on the workflow page, in the "Workflow Information" section, at the right end of the "TRS" line) so that when two workflows are viewed successively, the copied value is correct.  Previously, if you visited the pages for two workflows A and B, if you used the button to copy the TRS ID on workflow B's page, the copied value was the TRS ID for workflow A.

The cause was that we were updating the TRS ID field of the component when an "entryType" observable issued forth a new value.  The assumption was probably that if we visited successive workflows, we'd see successive "workflow" entry type values on the line.  But, for whatever reason, that's not how it was working...

Editorial: this type of bug suggests that our UI code (and more specifically, the architecture and constructs typically employed within) is making it hard for us to do simple, straightforward things (such as, in this case, display the TRS ID corresponding to the current workflow).

The branch has the wrong issue number, woops.

**Review Instructions**
Go to search, reproduce the steps in the first paragraph of the "Description" section above, use the button to copy the TRS ID, and confirm it has the correct value.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2594
dockstore/dockstore#6024

**Security**

### DIFF
--- a/src/app/workflow/info-tab/info-tab.service.ts
+++ b/src/app/workflow/info-tab/info-tab.service.ts
@@ -189,7 +189,7 @@ export class InfoTabService {
 
   getTRSId(workflow: Workflow | undefined): string {
     if (workflow) {
-      return workflow.trsId;
+      return workflow.entryTypeMetadata.trsPrefix + workflow.full_workflow_path;
     } else {
       return '';
     }
--- a/src/app/workflow/info-tab/info-tab.service.ts
+++ b/src/app/workflow/info-tab/info-tab.service.ts
@@ -189,7 +189,7 @@ export class InfoTabService {
 
   getTRSId(workflow: Workflow | undefined): string {
     if (workflow) {
-      return workflow.entryTypeMetadata.trsId;
+      return workflow.trsId;
     } else {
       return '';
     }
--- a/src/app/workflow/info-tab/info-tab.service.ts
+++ b/src/app/workflow/info-tab/info-tab.service.ts
@@ -187,11 +187,12 @@ export class InfoTabService {
     );
   }
 
-  getTRSId(workflow: Workflow | undefined, entryType: EntryType): string {
-    if (!workflow) {
+  getTRSId(workflow: Workflow | undefined): string {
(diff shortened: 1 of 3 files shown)


coverbeck (approved):

src/app/workflow/info-tab/info-tab.service.ts
```diff
@@ -187,11 +187,12 @@ export class InfoTabService {
     );
   }
 
-  getTRSId(workflow: Workflow | undefined, entryType: EntryType): string {
-    if (!workflow) {
+  getTRSId(workflow: Workflow | undefined): string {
+    if (workflow) {
+      return workflow.entryTypeMetadata.trsPrefix + workflow.full_workflow_path;
```
kathy-t, line 192: No need to change, but FYI, there is a `trsId` field now:  https://github.com/dockstore/dockstore/blob/c8459501ac20132294376702c6c8a81338d3c9bf/dockstore-webservice/src/main/resources/openapi3/openapi.yaml#L12529-L12530

kathy-t (approved):

david4096 (approved):

src/app/workflow/info-tab/info-tab.service.ts
kathy-t: the `trsId` should be available for `Workflow` https://github.com/dockstore/dockstore/blob/c8459501ac20132294376702c6c8a81338d3c9bf/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java#L969-L971

svonworl: Forgot about that, changed.  Thanks Kathy!

svonworl: Woops, something is wrong.  Being that I'm leaving soon and we're trying to deploy again, reverting to original solution...

PR #2040, opened 2024-10-29
SEAB-6767: Fix launch-with button "has no content" bug

svonworl:
**Description**
This PR fixes the "has no content" launch-with button bug, which is described in https://ucsc-cgl.atlassian.net/browse/SEAB-6767.

The bug is caused by the complicated, asynchronous information flow through our UI, due to the web of Observables and Subscriptions that we have woven.  We have discussed removing Akita, and if/when we do, we should consider reevaluating this part of our UI's information management model. 

Because our UI code sets the workflow and version name at different clicks, `ngOnChanges` is called twice when the launch-with component is initialized, and the second call has consistent/correct values.  We move the primary/secondary descriptor retrieval code to `ngOnChanges` to more closely couple it to the final `ngOnChanges` call.

**Review Instructions**
Confirm that you cannot reproduce the bug, using the steps detailed in the Slack thread which is linked in the ticket.  Then, view successive entry pages (by navigating our UI, without reloading the page) and confirm that it does not feel laggier.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6767

**Security**

### DIFF
--- a/src/app/workflow/launch-third-party/launch-third-party.component.ts
+++ b/src/app/workflow/launch-third-party/launch-third-party.component.ts
@@ -3,7 +3,7 @@ import { Component, Input, OnChanges, OnInit, SimpleChanges } from '@angular/cor
 import { MatLegacyDialog as MatDialog, MatLegacyDialogModule } from '@angular/material/legacy-dialog';
 import { DescriptorLanguageService } from 'app/shared/entry/descriptor-language.service';
 import { combineLatest, Observable, Subscription } from 'rxjs';
-import { map, shareReplay } from 'rxjs/operators';
+import { map, shareReplay, takeUntil } from 'rxjs/operators';
 import { Base } from '../../shared/base';
 import { Dockstore } from '../../shared/dockstore.model';
 import { CloudInstance, CloudInstancesService, User, Workflow, WorkflowVersion } from '../../shared/openapi';
@@ -297,11 +297,13 @@ export class LaunchThirdPartyComponent extends Base implements OnChanges, OnInit
         this.descriptorLanguageService.workflowDescriptorTypeEnumToExtendedDescriptorLanguageBean(descriptorType).descriptorLanguageEnum;
       this.primaryDescriptorSubscription = this.workflowsService
         .primaryDescriptor1(this.workflow.id, descriptorType, this.selectedVersion.name)
+        .pipe(takeUntil(this.ngUnsubscribe))
         .subscribe((sourceFile) => {
           this.descriptorsService.updatePrimaryDescriptor(sourceFile);
         });
       this.secondaryDescriptorsSubscription = this.workflowsService
         .secondaryDescriptors1(this.workflow.id, descriptorLanguageEnum, this.selectedVersion.name)
+        .pipe(takeUntil(this.ngUnsubscribe))
         .subscribe((sourceFiles: Array<SourceFile>) => {
           this.descriptorsService.updateSecondaryDescriptors(sourceFiles);
         });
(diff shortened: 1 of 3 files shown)


src/app/workflow/launch-third-party/launch-third-party.component.ts
```diff
@@ -319,6 +292,19 @@ export class LaunchThirdPartyComponent extends Base implements OnChanges, OnInit
...
       this.selectedVersionWorkflowPathAsQueryValue = this.encode(this.prependIfNotPrefix('/', this.selectedVersion.workflow_path));
+      const descriptorType = this.workflow.descriptorType;
+      const descriptorLanguageEnum =
+        this.descriptorLanguageService.workflowDescriptorTypeEnumToExtendedDescriptorLanguageBean(descriptorType).descriptorLanguageEnum;
+      this.primaryDescriptorSubscription = this.workflowsService
+        .primaryDescriptor1(this.workflow.id, descriptorType, this.selectedVersion.name)
+        .subscribe((sourceFile) => {
+          this.descriptorsService.updatePrimaryDescriptor(sourceFile);
+        });
+      this.secondaryDescriptorsSubscription = this.workflowsService
```
coverbeck, line 303: Before your PR, this request was being done after the request to fetch the primary descriptor. In your PR it's being done in parallel.

Just checking if we're OK switching that. It is broken now, and this approach should be faster, so I suppose it's OK.

coverbeck (approved):
NIt-picky stuff; I was intentionally going over it extra carefully given that it's coming in so late in the release cycle.

src/app/workflow/launch-third-party/launch-third-party.component.ts
```diff
@@ -279,37 +277,12 @@ export class LaunchThirdPartyComponent extends Base implements OnChanges, OnInit
...
-                });
-            }
-          });
-        }
-      });
   }
 
   ngOnChanges(changes: SimpleChanges): void {
     this.descriptorsQuery.clear();
+    this.primaryDescriptorSubscription?.unsubscribe();
```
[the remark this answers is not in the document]
svonworl: > These `unsubscribe`s are here in case the user switches versions navigating away from the page, correct?

The `unsubscribe`s are so we don't handle any in-process requests that are left over from the previous set of inputs.  Imagine one of those request gets delayed, and then we receive the response, and handle it.  It could "overwrite" the actions of a superceding request, which would be bad.

> Are you sure there won't be a memory leak when the component is destroyed? I don't think so, but double-checking. On the one hand, I think subscriptions are not garbage collected if the observable hasn't been completed, on the other hand, http requests complete as soon as they are done, and once an observable is done, the subscriber will be garbage collectable.

Supposedly, http-based observables complete themselves after the result arrives:
https://stackoverflow.com/questions/35042929/is-it-necessary-to-unsubscribe-from-observables-created-by-http-methods-to-avoid

> In any case, I think you should also add the `takeUntil(this.ngUnsubscribe)`. If the user navigates away from the page while these requests are still happening, they'll get cancelled.

However, this is a good idea.  Say a request gets delayed, and we navigate away from the component.  We don't want to change state later when the response finally arrives.  Added.

src/app/workflow/launch-third-party/launch-third-party.component.ts
svonworl: Yeah, the tradeoff here is we might do an extra request or two, but they are initiated earlier and in parallel.  I think the new way is ok, but we'll need to make sure it doesn't feel any laggier after we've deployed to staging/production.

coverbeck: > Yeah, the tradeoff here is we might do an extra request or two, but they are initiated earlier and in parallel. I think the new way is ok, but we'll need to make sure it doesn't feel any laggier after we've deployed to staging/production.

My edge-case concern was more that there was some weird UI interaction, e.g., the display of the secondary descriptors somehow depended on the primary descriptor being displayed first. Just wanted to make triple-sure... But I think it's OK.

kathy-t (approved):

hyunnaye (approved):

denis-yuen (approved):

PR #2043, opened 2024-11-14
SEAB-6777: Improve formatting of workflow "Versions" table

svonworl:
**Description**
This PR improves the workflow page's "Versions" table so that it fits horizontally into our UI when the browser window is wide.  Previously, the table was too wide for space afforded to it by our UI, causing the rightmost elements to be hidden behind other elements, wide content to be wrapped over multiple lines, and a scrollbar to appear at bottom.

Specifically, this PR:
1. removes the "sort" control from all but the "Git Reference" and "Date Modified" columns.
2. decreases the padding between adjacent columns.

These changes allow the column headers to fit within the available space, and provide extra horizontal space that columns containing wide content can scale into.

These changes make a big difference, improving the spacing and flow of the table for most workflows.  Arguably, the table is less useful because we now can't sort some columns, but IMHO, the readability improvement greatly outweighs any drawbacks.

Before (develop):
dump 2024-11-13 at 5 54 39 PM

After (this PR):
dump 2024-11-13 at 5 53 46 PM

The tool table contains less information, and didn't need these changes.

At narrow browser widths, the table still gets mangled/hidden.   A possible next step might be to remove the "information circles" and instead directly link the table heading text to the documentation. 

**Review Instructions**
Scale your browser to maximum width, view a workflow's "Versions" table, and confirm that it fits horizontally and that there is no scroll bar.  Do the same for some other types of workflows.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6777

**Security**

### DIFF
--- a/src/app/workflow/versions/versions.component.scss
+++ b/src/app/workflow/versions/versions.component.scss
@@ -49,3 +49,14 @@ tr.mat-row.highlight {
 tr.highlight > .mat-cell:first-of-type > span {
   padding-left: 5px;
 }
+
+.mat-cell:not(:first-of-type),
+.mat-header-cell:not(:first-of-type) {
+  padding-left: 1rem;
+}
+
+.mat-cell:not(:last-of-type),
+.mat-header-cell:not(:last-of-type),
+table th {
+  padding-right: 1rem;
+}
--- a/src/app/workflow/versions/versions.component.scss
+++ b/src/app/workflow/versions/versions.component.scss
@@ -50,6 +50,8 @@ tr.highlight > .mat-cell:first-of-type > span {
   padding-left: 5px;
 }
 
+/* Narrow the version table columns by decreasing their padding. */
+/* See https://ucsc-cgl.atlassian.net/browse/SEAB-6777 */
 .mat-cell:not(:first-of-type),
 .mat-header-cell:not(:first-of-type) {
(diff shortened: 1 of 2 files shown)


denis-yuen:
> At narrow browser widths, the table still gets mangled/hidden. A possible next step might be to remove the "information circles" and instead directly link the table heading text to the documentation.

Seems a bit counter-intuitive. One other quick change would be to move verified into the overview page. Metrics basically replaces verified going forward.

coverbeck (approved):

src/app/workflow/versions/versions.component.scss
```diff
@@ -49,3 +49,14 @@ tr.mat-row.highlight {
 tr.highlight > .mat-cell:first-of-type > span {
   padding-left: 5px;
 }
+
+.mat-cell:not(:first-of-type),
+.mat-header-cell:not(:first-of-type) {
+  padding-left: 1rem;
```
kathy-t, line 55: Would recommend adding a comment about why these styles override the default https://github.com/dockstore/dockstore-ui2/blob/f145992b8085007f0c52b1c010dbaa7d1103ef27/src/styles.scss#L1249-L1259

kathy-t (approved):

david4096 (approved):

PR #2048, opened 2024-12-05
SEAB-6341: Add Metrics type icons in versions tab and Rescent Versions sidebar

hyunnaye:
**Description**

**Review Instructions**

**Issue**

**Security**

### DIFF
--- a/src/app/workflow/versions/versions.component.html
+++ b/src/app/workflow/versions/versions.component.html
@@ -217,7 +217,18 @@
         >
       </th>
       <td mat-cell *matCellDef="let version">
-        <mat-icon data-cy="metrics" *ngIf="version.metricsByPlatform && (version.metricsByPlatform | json) !== '{}'">check</mat-icon>
+        <img
+          *ngIf="hasExecutionMetrics"
+          src="assets/svg/manufacturing.svg"
+          alt="Execution Metrics Icon"
+          matTooltip="This version contains execution metrics."
+        />
+        <img
+          *ngIf="hasValidationMetrics"
+          src="assets/svg/data_check.svg"
+          alt="Validation Metrics Icon"
+          matTooltip="This version contains validation metrics."
+        />
       </td>
     </ng-container>
 
--- a/src/app/workflow/versions/versions.component.ts
+++ b/src/app/workflow/versions/versions.component.ts
@@ -75,6 +75,9 @@ export class VersionsWorkflowComponent extends Versions implements OnInit, OnCha
   @Input() versions: Array<WorkflowVersion>;
   @Input() workflowId: number;
   @Input() verifiedVersionPlatforms: Array<VersionVerifiedPlatform>;
+  @Input() hasExecutionMetrics!: boolean;
+  @Input() hasValidationMetrics!: boolean;
+
   _selectedVersion: WorkflowVersion;
   Dockstore = Dockstore;
   @Input() set selectedVersion(value: WorkflowVersion) {
--- a/src/app/workflow/workflow.component.html
+++ b/src/app/workflow/workflow.component.html
@@ -343,6 +343,8 @@
                 [canRead]="canRead"
                 [canWrite]="canWrite"
                 [isOwner]="isOwner"
+                [hasExecutionMetrics]="hasExecutionMetrics"
+                [hasValidationMetrics]="hasValidationMetrics"
               >
               </app-versions-workflow>
             </div>
@@ -456,7 +458,12 @@
             </mat-tab>
           </div>
           <mat-tab id="metricsTab" label="Metrics">
-            <app-executions-tab [entry]="workflow" [version]="selectedVersion"></app-executions-tab>
+            <app-executions-tab
+              [entry]="workflow"
+              [version]="selectedVersion"
+              (showExecutionIcon)="showExecutionMetricsIcon($event)"
+              (showValidationIcon)="showValidationMetricsIcon($event)"
+            ></app-executions-tab>
           </mat-tab>
           <mat-tab *ngIf="!isPublic() && isHosted() && isOwner" id="permissionsTab" label="Permissions">
             <app-permissions [workflow]="workflow"></app-permissions>
@@ -514,6 +521,18 @@
                       </a>
                     </mat-chip>
                   </span>
+                  <img
+                    *ngIf="hasExecutionMetrics"
+                    src="assets/svg/manufacturing.svg"
+                    alt="Execution Metrics Icon"
+                    matTooltip="This version contains execution metrics."
+                  />
+                  <img
+                    *ngIf="hasValidationMetrics"
+                    src="assets/svg/data_check.svg"
+                    alt="Validation Metrics Icon"
+                    matTooltip="This version contains validation metrics."
+                  />
                 </div>
                 <small>{{ sortedVersion.last_modified | date }}</small>
               </div>
--- a/src/app/workflow/executions/executions-tab.component.ts
+++ b/src/app/workflow/executions/executions-tab.component.ts
@@ -13,7 +13,7 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-import { Component, Input, OnChanges, OnInit } from '@angular/core';
+import { Component, EventEmitter, Input, OnChanges, OnInit, Output } from '@angular/core';
 import { EntryTab } from '../../shared/entry/entry-tab';
 import {
   CloudInstance,
@@ -122,6 +122,9 @@ export class ExecutionsTabComponent extends EntryTab implements OnInit, OnChange
   @Input() entry: BioWorkflow | Service | Notebook;
   @Input() version: WorkflowVersion;
 
+  @Output() showExecutionIcon = new EventEmitter<boolean>();
+  @Output() showValidationIcon = new EventEmitter<boolean>();
+
   constructor(
     private extendedGA4GHService: ExtendedGA4GHService,
     private alertService: AlertService,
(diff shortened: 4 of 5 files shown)


denis-yuen (commented):
This is annoying feedback since I see this was the case back in the storytime mock-up, while I get "validation" from the validation icon, I'm not really getting a sense of "statistics" from the execution metrics icon. 

I don't feel strongly about it but am curious how others feel and/or what the alternatives are in terms of icons

src/app/workflow/versions/versions.component.html
```diff
@@ -217,7 +217,18 @@
         >
       </th>
       <td mat-cell *matCellDef="let version">
-        <mat-icon data-cy="metrics" *ngIf="version.metricsByPlatform && (version.metricsByPlatform | json) !== '{}'">check</mat-icon>
+        <img
+          *ngIf="hasExecutionMetrics"
+          src="assets/svg/manufacturing.svg"
```
coverbeck, line 222: I think you need a relative path here, `../../assets`, or however many double dots are necessary, or it won't work in prod. Here's where it matters: https://github.com/dockstore/dockstore-ui2/blob/develop/.circleci/config.yml#L247

src/app/workflow/versions/versions.component.html
```diff
@@ -217,7 +217,18 @@
...
+          alt="Execution Metrics Icon"
+          matTooltip="This version contains execution metrics."
+        />
+        <img
+          *ngIf="hasValidationMetrics"
+          src="assets/svg/data_check.svg"
```
coverbeck, line 228: Same comment about relative path.

src/app/workflow/workflow.component.html
```diff
@@ -514,6 +521,20 @@ <h3>
                       </a>
                     </mat-chip>
                   </span>
+                  <img
+                    *ngIf="hasExecutionMetrics"
+                    src="assets/svg/manufacturing.svg"
```
coverbeck, line 526: Relative path

src/app/workflow/workflow.component.html
```diff
@@ -514,6 +521,20 @@ <h3>
...
+                    alt="Execution Metrics Icon"
+                    data-cy="execution-metrics-icon"
+                    matTooltip="This version contains execution metrics."
+                  />
+                  <img
+                    *ngIf="hasValidationMetrics"
+                    src="assets/svg/data_check.svg"
```
coverbeck, line 533: Relative path

src/app/workflow/versions/versions.component.ts
```diff
@@ -82,6 +82,9 @@ export class VersionsWorkflowComponent extends Versions implements OnInit, OnCha
   @Input() workflowId: number;
   @Input() verifiedVersionPlatforms: Array<VersionVerifiedPlatform>;
   @Input() publicPage: boolean;
+  @Input() hasExecutionMetrics!: boolean;
+  @Input() hasValidationMetrics!: boolean;
```
coverbeck, line 86: What does the exclamation point do?

src/app/workflow/versions/versions.component.html
```diff
@@ -217,7 +217,18 @@
         >
       </th>
       <td mat-cell *matCellDef="let version">
-        <mat-icon data-cy="metrics" *ngIf="version.metricsByPlatform && (version.metricsByPlatform | json) !== '{}'">check</mat-icon>
+        <img
+          *ngIf="hasExecutionMetrics"
```
coverbeck, line 221: Instead of relying on event emitter, can't you just tell by looking a `version.metricsByPlatform` where there are execution and validation metrics or not? I may be missing something, and you may need a pipe to figure it out, but it seems more straightforward than having an event emitter from the executions-tab-component.

src/app/workflow/versions/versions.component.ts
hyunnaye: Weird, I think Intellij added that on its own when I created the inputs through their dialog

svonworl:
> This is annoying feedback since I see this was the case back in the storytime mock-up, while I get "validation" from the validation icon, I'm not really getting a sense of "statistics" from the execution metrics icon.

Some kind of graph [bar graph?] might be good as the "executions metrics" icon, the gears typically indicate "setup" or "configuraton" in most UIs.  Also, both current icons have a strong diagonal visual weight, which looks kinda funny when they're displayed together, so making one less so would be an improvement.

svonworl (approved):
Don't really care for the metrics icons in the "Recent Versions" box, it's added visual clutter and feels tangential to the box's purpose.  But if that's what was in the spec and the team decided was best, no need to change it.

kathy-t:
> This is annoying feedback since I see this was the case back in the storytime mock-up, while I get "validation" from the validation icon, I'm not really getting a sense of "statistics" from the execution metrics icon.
> 
> I don't feel strongly about it but am curious how others feel and/or what the alternatives are in terms of icons

We could re-purpose the verified icon which is just a checkmark? But then it would need to be in a bubble and might look confusing in the versions table which already has a bunch of checkmarks

image

kathy-t (commented):
Was here, following discussion. 

The Recent Versions box does look busy when both icons are there and the box is kind of small. Maybe it's not worth it to have it there, or we just display a generic metrics icon, but then there would be 3 metrics icons and that might be too much

denis-yuen:
> Don't really care for the metrics icons in the "Recent Versions" box, it's added visual clutter and feels tangential to the box's purpose. But if that's what was in the spec and the team decided was best, no need to change it.

On the fence on this one, I agree that the icons in the more detailed versions tab is more intuitive

hyunnaye:
Updated screenshots - gotta wait for https://github.com/dockstore/dockstore/pull/6055 for tests to pass

kathy-t:
> Updated screenshots to show the new execution metrics icon and removed the icons on the recent versions sidebar. Need to wait until [dockstore/dockstore#6055](https://github.com/dockstore/dockstore/pull/6055) merge

FYI, you can update the webservice version in the package.json so it builds against your webservice branch if you want to see if the UI builds pass before the webservice PR is merged.

You would modify it like https://github.com/dockstore/dockstore-ui2/pull/2046/files, except your `webservice_version_prefix` would be `1.17.0-feature-seab-6341`

src/app/workflow/versions/versions.component.html
```diff
@@ -210,7 +210,20 @@
         >
       </th>
       <td mat-cell *matCellDef="let version">
-        <mat-icon data-cy="metrics" *ngIf="version.metricsByPlatform && (version.metricsByPlatform | json) !== '{}'">check</mat-icon>
+        <img
+          *ngIf="showExecutionMetricsIcon(version)"
```
coverbeck, line 214: Best practice is that you should use a pipe. Up to the team to decide if it's necessary or not.

src/app/workflow/versions/versions.component.html
```diff
@@ -210,7 +210,20 @@
...
+          src="../assets/svg/execution-icon.svg"
+          alt="Execution Metrics Icon"
+          data-cy="execution-metrics-icon"
+          matTooltip="This version contains execution metrics."
+        />
+        <img
+          *ngIf="showValidationMetricsIcon(version)"
```
coverbeck, line 221: Same as other comment

src/app/workflow/versions/versions.component.ts
```diff
@@ -245,4 +247,20 @@ export class VersionsWorkflowComponent extends Versions implements OnInit, OnCha
   trackBy(index: number, item: WorkflowVersion) {
     return item.id;
   }
+
+  showExecutionMetricsIcon(version: WorkflowVersion) {
```
coverbeck, line 251: Nit, whether you make this a pipe or not, I suggest passing `version.metricsByPlatform` as the parameter, that way you can get rid of the 2 `version.` strings inside the method.

src/app/workflow/versions/versions.component.ts
```diff
@@ -245,4 +247,20 @@ export class VersionsWorkflowComponent extends Versions implements OnInit, OnCha
...
+    if (version.metricsByPlatform !== null) {
```
coverbeck, line 252: Can it be `undefined`? I suspect the API always returns something, so probably not, but just double-checking. If unsure, it could be safer to do `if (version.metricsByPlatform) {`

coverbeck (approved):

denis-yuen (approved):
Ok with current set of icons, after addressing feedback from others

src/app/workflow/versions/versions.component.ts
```diff
@@ -245,4 +247,20 @@ export class VersionsWorkflowComponent extends Versions implements OnInit, OnCha
...
+      const metrics = version.metricsByPlatform[PartnerEnum.ALL];
+      return metrics?.executionStatusCount != null;
```
kathy-t, line 254: `!=` -> `!==`

src/app/workflow/versions/versions.component.ts
```diff
@@ -245,4 +247,20 @@ export class VersionsWorkflowComponent extends Versions implements OnInit, OnCha
...
+    }
+    return false;
+  }
+
+  showValidationMetricsIcon(version: WorkflowVersion) {
+    if (version.metricsByPlatform !== null) {
+      const metrics = version.metricsByPlatform[PartnerEnum.ALL];
+      return metrics?.validationStatus != null;
```
kathy-t, line 262: Ditto

kathy-t (approved):
Design looks good. Would second Charles' feedback about the pipe

kathy-t (approved):

PR #2050, opened 2024-12-11
DOCK-2591: Use webservice-defined sort ordering in workflow versions table by default

svonworl:
**Description**
As it turns out, the UI was already using the webservice-defined sort order when the versions table was loaded.  I confirmed this by observing that the request to the webservice did not include a `sortCol` parameter, and that initially, the default version always appeared at the top, no matter when it was last updated.  However, the UI was incorrectly indicating that it was sorting by "Last Modified" in descending order (the black arrow appeared next to the "Last Modified" header).
 
This PR changes the UI so that the version table sort controls now have three states ("none", "ascending", and "descending"), they start off set to "none". and they display their state correctly.

This is similar to how the search table works: the controls initially don't specify any sort ordering. In that case, we supply a default sort order in the ES query.

For the workflow versions table, we let the webservice decide the order, because it has access to all of the versions and we can code whatever ordering logic we like, allowing us the freedom to implement whatever default order is appropriate.  Currently, we implement the "default version at top" rule, but later, we could do other things, like bubble release tags towards the top, `main` and `develop`, etc... This could be rule-based: for example, `main`, then `develop`, then release tags, then other versions.  Or, the sort query could calculate a numerical "importance" for each version, based on whether it was the default branch, a release tag, a mainline branch, feature branch, creation/modification date, etc., and sort on it.

**Review Instructions**
Find a workflow that has a default version that's not the last modified version.  Make sure that when the workflow's versions table is first displayed, the default version appears first, and then is followed by other versions in descending "Last Modified" order.  Confirm that the sort controls are in "none" state.  Then, toggle them and confirm that the versions sort correctly.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2591
dockstore/dockstore#6019

**Security**

### DIFF
--- a/cypress/e2e/group1/checkerWorkflowFromWorkflow.ts
+++ b/cypress/e2e/group1/checkerWorkflowFromWorkflow.ts
@@ -123,9 +123,6 @@ describe('Checker workflow test from my-workflows', () => {
       cy.url().should('eq', Cypress.config().baseUrl + '/my-workflows/github.com/A/l');
 
       goToTab('Versions');
-      // Need to press this twice to get the sort into "last modified last" order
-      cy.get('[data-cy=date-modified-header]').should('be.visible').click();
-      cy.get('[data-cy=date-modified-header]').should('be.visible').click();
       cy.wait(1000); //waits for the sorting to finish to ensure Actions button is clicked for the right version (was encountering flakiness)
 
       cy.contains('button', 'Actions').click();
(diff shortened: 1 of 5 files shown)


svonworl:
Unexpected integration test failures, please hold off on reviewing for now.

coverbeck (approved):

coverbeck:
> Unexpected integration test failures, please hold off on reviewing for now.

Whoops, reviewed before noticing this.

david4096 (approved):

denis-yuen (commented):
still holding off, request review when needed, just managing notifications on my end

PR #2052, opened 2024-12-19
Migrate from Material legacy components to MDC components

coverbeck:
**Description**
Upgrades to Angular MDC Material from the legacy components we are using today.

Why? Because Angular Material 17 dropped the legacy components. This means we can't upgrade to Angular 17 and stay current. Angular 16, our current version, is already no longer [actively supported](https://angular.dev/reference/releases#actively-supported-versions).

***The Migration***
* Following [this guide](https://v16.material.angular.io/guide/mdc-migration).
* Main issues
    * Selectors in the cypress tests relying on "internal classes"
    * Styling (hard to find)

**Continued PR Description by Kathy**
I recommend reviewing without whitespace and running this locally to check out the changes.

Recall that this was where Charles left off https://github.com/dockstore/dockstore-ui2/pull/2052#issuecomment-2583596348

When fixing the styles, I went with the default Material style where possible because Angular Material strongly discourages people from overriding internal CSS classes. An example of why this should be avoided is that although the migration guide says that "CSS classes applied to components use the `mat-mdc-` prefix, whereas before it was simply a `mat-` prefix", I found that some classes did not apply the style as we expected because Angular Material changed the internal structure of some components. It's hard to find these without reviewing every single style where we targeted internal Material classes to make sure that the style is applied properly.

I found that using the Angular Material tokens allowed me to target the styling components that we wanted without having to define CSS classes that target deeply nested Material CSS classes. These tokens start with `--mdc` or `--mat`. The newest Angular Material version (19) recommends using tokens to override component styles https://material.angular.io/guide/theming#component-tokens and Angular Material 19 component documentation now have a Styling tab which contains the tokens that can be used like for buttons https://material.angular.io/components/button/styling. When we do get up to Angular 19, we will have to make some modifications to the token names (as far as I can tell, it's just removing the `--mdc/--mat` prefix), but it will be simpler because the tokens are documented.

Changes:
- Used the default Material component styles where possible. In the migration, the tool marked several CSS styles with a `TODO(mdc-migration)` comment where an automatic migration could not happen. These were styles that overrode the internal Material classes. 
  - In many cases, I just removed them because the default style was appropriate and this would make it more maintainable in the future. Where it made sense, I created a CSS class that did not rely on the internal Material class and added the new class to the Material component element in the HTML.
  - For example, in the multi-cloud launch with menu, we override the internal Material styles to make the radio buttons look more dense and the selected radio button color doesn't have an accent color. IMO, we can live without it and it doesn't look that different.

    Before:
    image

    After:
    image

- Consolidated several button styles to make it more consistent.
  - In particular: `private-btn`, `accent-1-dark`, `accent-1-dark-btn`, and `small-btn-structure` classes were used in combination with each other through out the UI, and some of those classes actually overlap in styles.
  - What also prompted this was that the mat-elevation classes were no longer working due to the new internal class structure of the MDC components, which had a higher specificity that overrode the mat-elevation classes we were using. 
  - This was specifically problematic for buttons because we were using the `mat-elevation-z` class to make buttons appear flat. For example, we would do `<mat-raised-button class="mat-elevation-z">` which made a raised button flat. IMO, that type of styling should be avoided because there's already a flat Material button style, `mat-flat-button`. I removed the `mat-elevation-z` styles and used the appropriate Material button style instead.
- Removed `mat-elevation-z` from mat-cards
  - Similar to the buttons above, we were using `mat-elevation-z` to make a mat-card appear flat. However, there is already a flat mat-card `<mat-card appearance="outlined">`, so we don't need to use `mat-elevation-z` (which did not work anyway due to the reason mentioned in the buttons point above. Found an issue for it https://github.com/angular/components/issues/26094#issue-1465231931).
- Converted all form fields to outline form fields
  - Previously, we had a mixture of "fill" and "outline" form fields, likely from our migration to the new UI redesign. I made all form fields have the outline style by default because that is the form field style in all of our UI redesign mockups in Zeplin. Furthermore, fill form fields have a bug where the form field label disappears if the density is less than -1, which leads to weird alignment issues in the form field itself (issue https://github.com/angular/components/issues/26579)
  - One notable change due to this is the Files tab. The dark primary color background made it difficult to read the form field label, so I changed it to match the [zeplin mockups](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/6021ab0fb454f610f8815657).
    
    Before:
    
    image

    After:

    image

- Made "bubble" styles the default style for Material chips so we don't need to add the `bubble` class to chips or spans
- I went through the test changes that Charles did where he had to update the `.mat` class references to `.mat-mdc` so that they no longer rely on Material's internal classes. For examples, instead of looking for `.mat-mdc-tab` when clicking a tab, we now look for the HTML `role=tab` instead. IMO, this is less likely to change than the class names. In more simple cases, I just added a `data-cy` attribute so that we can reference it in tests. This could be an enhancement in the future where we look through the tests and make sure we're using data-cy where possible.

**Things to look out for**
- Padding for cards and dialogs. In the migration, we have to specifically add `mat-card-content` and `mat-dialog-content` so that the content has padding (before, it had padding by default).
- Button styles
- Spacing for form fields, in register entry/organization/collection dialogs in particular (where there are multiple form fields stacked on top of each other)
- Bubbles/chips like labels, descriptor type bubbles, AI bubbles, etc.

Screenshots of things that look a little different that were not already mentioned. Please do let me know if you feel like we should keep certain styles.

<details>
<summary>Cards</summary>

For raised cards, we used to override the border and box-shadow style. I removed these so we are using the default raised style which has no border.

Before:
image

After:
image

</details>

<details>
<summary>Menus</summary>

Menu panels used to have bold text. Now, we are using the default font weight which is not bold. Note that it seems like in Angular Material 19, the menu content is bold again, so if you liked that, you will see it again eventually :grin: 

Before:
image

After:
image

For the Actions menu in the Versions tab, it used to be a size 13px font, but I kept it the same as all of our other menus, which is 14px.

Before:
image

After:
image

</details>

<details>
<summary>Form fields</summary>

Before:
image

After:
image

Before:
image

After:
image

</details>

<details>
<summary>Slide toggles</summary>

The color styling here appears different because I had to override the slide toggle styles using CSS classes. In the new Angular Component, the slide toggle component actually uses colors of the palette that we have not defined. We've defined palettes with keys from 1-6, which does not follow Material's guide https://v16.material.angular.io/guide/theming#create-your-own-palette.

>The map must define hues for 50 and each hundred between 100 and 900

We should eventually look at defining a full palette according to Material's guide.

Before:
image

After:
image

</details>

<details>
<summary>Buttons</summary>

Buttons with the class `private-btn` no longer have a shadow. This class previously added a box-shadow and border to buttons which made it like a `mat-stroked-raised-button` (combo of `mat-stroked-button` and `mat-raised-button`). To make our styling more intentional, these are now all `mat-stroked-button`s with no shadow.

Before:
image

After:
image

</details>

**Enhancements/thoughts for the future**
- Define a full color palette according to https://v16.material.angular.io/guide/theming#create-your-own-palette
- There are several bootstrap styles that conflict with the Material styles so we have to override the Bootstrap ones to make things look aligned. Wondering if we plan to remove Bootstrap in the future?
- Remove unused CSS classes
- We should use mat-icons over uploading our own images where possible. This is more apparent in buttons because when a button is disabled, a mat-icon has an appropriate disabled color while an `img` keeps its original color.
- We use alot of `accent-1-dark` across our UI, which is not what we defined as our `primary` color. A search for `color="primary"` gives me 19 results and a search for `mat.get-color-from-palette($dockstore-app-primary` gives me 15 results for a total of 34 results. A search for `accent-1-dark` gives me 92 results. Maybe worth changing primary colors at some point? 

**Review Instructions**
Go through the UI and verify that things look okay. Keep an eye out for alignment/padding issues.
- Padding for cards and dialogs. In the migration, we have to specifically add `mat-card-content` and `mat-dialog-content` so that the content has padding (before, it had padding by default).
- Button styles
- Spacing for form fields, in register entry/organization/collection dialogs in particular (where there are multiple form fields stacked on top of each other)
- Bubbles/chips like labels, descriptor type bubbles, AI bubbles, etc.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6279

**Security**


coverbeck:
Latest status before transferring this. The good news is all integration and unit tests are passing. The bad news is:

* I haven't looked not tried, but I wouldn't be surprised if some of nightly smoke tests are using selectors that no longer exist.
* Known problems, and it's most likely not everything:

**Dashboard, padding around Register Workflow button not right, plus `Workflows` not vertically centered**
Screen Shot 2025-01-10 at 10 43 50 AM

**Dropdown on files tab has wrong background color, perhaps wrong font**
Screen Shot 2025-01-10 at 10 44 47 AM

**Horizontal scrolling has crept back into versions tab, probably a font size issue, but just guessing**
Screen Shot 2025-01-10 at 10 44 36 AM

**Left-side of dashboard totally off, padding, colors**
Screen Shot 2025-01-10 at 10 43 02 AM

**Organizations page, content between different boxes not aligning. Compare to prod where the divider lines up exactly between boxes on the same row**
Screen Shot 2025-01-10 at 10 42 50 AM

**Home page, starred-entries page. "chips" are larger, with the bonus of throwing off the wrapping on the home page**
Screen Shot 2025-01-10 at 10 42 39 AM

src/app/search/search.component.scss
```diff
@@ -196,41 +96,9 @@ mat-expansion-panel mat-expansion-panel-header[aria-expanded='false'] {
...
-  padding: 0rem;
-}
-
-// Form field text animation re-implementation
-:host ::ng-deep .mat-form-field-appearance-outline.mat-form-field-can-float.mat-form-field-should-float .mat-form-field-label {
-  transform: translateY(-1.1em) scale(0.75);
-  width: 133.33333%;
-  top: 1.84375em;
-}
-
```
kathy-t, line 230: These styles made the outline form field smaller than the default appearance. In this PR, we achieve the same effect by setting the form-field density to -4 in `_material-overrides.scss`. I would recommend not targeting internal Material CSS classes that are deeply nested because it's hard to maintain and it may misalign things, causing a domino effect of adding more internal Material styles to fix the misalignment.

Before: 
image

After:
image

src/app/confirmation-dialog/confirmation-dialog.component.html
```diff
@@ -1,11 +1,11 @@
-<h1 *ngIf="data?.title" mat-dialog-title>{{ data.title }}</h1>
+<h1 *ngIf="data?.title" mat-dialog-title data-cy="confirmation-dialog-title">{{ data.title }}</h1>
 
 <div *ngIf="data?.message" mat-dialog-content [innerHTML]="data.message"></div>
-<div mat-dialog-actions fxLayout="row" fxLayoutAlign="end center" fxLayoutGap="10px">
-  <button data-cy="cancel-dialog-button" mat-flat-button [mat-dialog-close]="false" class="secondary-1" cdkFocusInitial>
+<mat-dialog-actions [align]="'end'">
```
kathy-t, line 4: I changed the uses of fxLayout in dialogs like this line to use the existing `mat-dialog-actions` align property to align it to the end of the row. It achieves the same effect without requiring the flex module.

denis-yuen (approved):
Got to trust in the demo due to the size of the PR. 
Let''s get this out to QA and break out follow-up PRs if needed.

svonworl (approved):
Thanks for grinding through this, Kathy, true grit!

I was able to get this running locally without any fuss, after which I did a quick run through the UI.  Overall, looks good.  There are a few issues that probably could use some improvement before we release.  I will catalog them.

hyunnaye (approved):

PR #2071, opened 2025-03-24
SEAB-6966: add relevance sort in search

hyunnaye:
**Description**
This PR adds the relevance sort back to the search page and adds it as default.

Screenshot example when searched for "test"
image

When set back to Most stars
image

**Review Instructions**
Go to the search page on qa and verify the relevance sort works

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6966

**Security**

### DIFF
--- a/src/app/search/search-entry-table.component.ts
+++ b/src/app/search/search-entry-table.component.ts
@@ -128,11 +128,15 @@ export class SearchEntryTableComponent extends Base implements OnInit {
     ['categories.displayName', 'Category'],
   ]);
   public defaultSortOption: SortOption = {
-    label: 'Most Stars',
-    sort: { active: 'starredUsers', direction: 'desc' },
+    label: 'Relevance',
+    sort: { active: null, direction: 'desc' },
   };
   public sortOptions: SortOption[] = [
     this.defaultSortOption,
+    {
+      label: 'Most Stars',
+      sort: { active: 'starredUsers', direction: 'desc' },
+    },
     {
       label: 'Recently Updated',
       sort: { active: 'last_modified_date', direction: 'desc' },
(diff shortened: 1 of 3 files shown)


src/app/search/search-entry-table.component.ts
```diff
@@ -128,11 +128,15 @@ export class SearchEntryTableComponent extends Base implements OnInit {
     ['categories.displayName', 'Category'],
   ]);
   public defaultSortOption: SortOption = {
```
kathy-t, line 130: When a user navigates to the search page on prod or when the user selects facets, the default sort appears to be by stars https://dockstore.org/search?entryType=workflows&searchMode=files and it's like that because ES returns it sorted by stars if there's no search term. The sort by relevance only shows up when there is a text search (basic or advanced). 

In this PR, the default sort by relevance when there's no search term is really sorted by stars.

Not sure if it's worth differentiating this, but noting it and curious what others think

svonworl: Yeah, interesting dilemma.  On the one hand, currently, we only really sort by "relevance" when there's a search term, so maybe the "Sort by Relevance" menu option doesn't make sense otherwise.  On the other hand, someday, we may have a default non-search-term sort order that's relevance-based, using signals like usage, entries that we've chosen to feature, membership in categories/collections, etc.

When I first pondered this ticket, I figured we'd not display the "Sort by Relevance" menu option when there's no search term.  For now, at least.  However, might be tricky to implement...

denis-yuen: > Not sure if it's worth differentiating this, but noting it and curious what others think


> When I first pondered this ticket, I figured we'd not display the "Sort by Relevance" menu option when there's no search term. For now, at least. However, might be tricky to implement... 

Agree with both, worth a try if easy, could ignore for now if not

hyunnaye: I tried to do this differentiation but is pretty tricky to implement.. Maybe we can leave it since the organisation search also doesn't say it's sorted by stars on default. https://qa.dockstore.org/organizations

denis-yuen (approved):

svonworl (approved):

kathy-t (approved):

PR #2076, opened 2025-04-04
DOCK-2610: Include Toil launch instructions

svonworl:
**Description**
Ticket https://ucsc-cgl.atlassian.net/browse/DOCK-2610 prescribes that we append the version to the TRS ID in our UI, so that when it's copied into a Toil command, there's enough information to download the correct version and run it.  However, a TRS ID containing the version is not spec compliant, and we determined that it would break Galaxy's import page, and perhaps other software.

So, instead, as agreed upon, this PR adds Toil launch instructions to the workflow page "Launch" tab for published CWL and WDL workflows.

Also, this PR:
* Removes AGC instructions from the "Launch" tab.
* Changes "www.dnastack.com" to "dnastack.com"  in links, because the former was hanging and causing the tests to fail.

The new Toil instructions are an abbreviated form of what Adam suggested to me in a Slack exchange.  In the spirit of the other example commands, the Toil command is a basic, non-intimidating incantation that will successfully run the workflow, is short enough to fit entirely within the available space (typically), and is easy to read, understand, and copy.  As a nod to more advanced uses, I included a couple of Toil documentation links below it.

In the Toil command, we reference the workflow via its TRS ID because that's what we tend to prefer.

Please fire this up and click around to make sure the links point at the right pages.

**Screenshot**
Snippet of the bottom of a workflow page, after change:
image

**Review Instructions**
For a published CWL workflow, view the "Launch" tab and confirm that the Toil launch instructions appear, and that the example command and links within are correct.  Do the same for a published WDL workflow.  View the "Launch" tab for a Galaxy workflow, and confirm that Toil launch instructions do not appear.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2610
dockstore/dockstore#6067

**Security**

### DIFF
--- a/src/app/about/about.component.html
+++ b/src/app/about/about.component.html
@@ -64,7 +64,7 @@
             Dockstore never requires a user to log in to search and inspect contents for workflows and tools. Links to source repositories
             are provided. Once found, analyses can be moved from environment-to-environment (e.g.
             <a target="_blank" rel="noopener noreferrer" href="https://terra.bio/">Terra</a>,
-            <a target="_blank" rel="noopener noreferrer" href="https://www.dnastack.com/">DNAstack</a>, and
+            <a target="_blank" rel="noopener noreferrer" href="https://dnastack.com/">DNAstack</a>, and
             <a target="_blank" rel="noopener noreferrer" href="https://www.dnanexus.com/">DNAnexus</a>) and yet be guaranteed to run on
             anything that supports Docker.
           </p>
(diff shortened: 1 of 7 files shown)


denis-yuen:
> Snippet of the bottom of a workflow page, after change:

The Toil launch-with instructions should be ahead of the checker launch instructions as they are more important/probable in terms of being useful

denis-yuen (approved):
Apologies, must have missed this completely

svonworl:
> The Toil launch-with instructions should be ahead of the checker launch instructions as they are more important/probable in terms of being useful

Moved the Toil "launch" instructions to just above the checker workflow instructions.

kathy-t (approved):

PR #2079, opened 2025-04-10
Fix app header

kathy-t:
**Description**
The app headers for non-”My Entry” pages are not padded properly. This PR fixes that so that the app header is properly padded with a consistent height for all pages. Discovered during [usability test](https://docs.google.com/document/d/1vKoDCVQXEWG8AefEiLdr5tw44LLDzRY9LFH6bK3sEvk/edit?usp=sharing). 

Problematic app headers before fix:
image

image

image

image

image

App headers after fix:
image

image

image

image

image

image

**Review Instructions**
Navigate to different pages and verify that the app header is properly padded.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6279

**Security**

### DIFF
--- a/src/app/header/header.component.ts
+++ b/src/app/header/header.component.ts
@@ -17,12 +17,13 @@ import { Component } from '@angular/core';
 import { Router } from '@angular/router';
 import { toExtendSite } from '../shared/helpers';
 import { AlertComponent } from '../shared/alert/alert.component';
+import { FlexModule } from '@ngbracket/ngx-layout';
 
 @Component({
   selector: 'app-header',
   templateUrl: './header.component.html',
   standalone: true,
-  imports: [AlertComponent],
+  imports: [AlertComponent, FlexModule],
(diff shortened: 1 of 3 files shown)


src/app/header/header.component.ts
```diff
@@ -17,12 +17,13 @@ import { Component } from '@angular/core';
...
 import { toExtendSite } from '../shared/helpers';
 import { AlertComponent } from '../shared/alert/alert.component';
+import { FlexModule } from '@ngbracket/ngx-layout';
 
 @Component({
   selector: 'app-header',
   templateUrl: './header.component.html',
   standalone: true,
-  imports: [AlertComponent],
+  imports: [AlertComponent, FlexModule],
```
kathy-t, line 26: The HTML was using flex layout but the module was not included so it had no effect

denis-yuen (approved):

PR #2080, opened 2025-04-15
Address Angular Material component migration usability feedback

kathy-t:
**Description**
This PR addresses a couple usability feedback points, listed below. See the original feedback [here](https://docs.google.com/document/d/1vKoDCVQXEWG8AefEiLdr5tw44LLDzRY9LFH6bK3sEvk/edit?usp=sharing).

Expand each item below to view the screenshots.

<details>
<summary>Smaller font for Featured Content is less eye-catching. This is on the logged in dashboard page. Fixed by making the title an h3 element</summary>
  
  Before
  image

  After
  image

</details>

<details>
<summary>Incorrect header color for hosted files. Turns out this affects all tab groups with nested tab groups. Fixed it so that nested tab groups have the same color as the parent tab groups. Also removed extra padding separating the parent and child tab groups in the Files tab.</summary>

  Before
  image

  After
  image

</details>

<details>
<summary>Upper edge of cards is hard to see. Fixed by adding a border to all cards.</summary>

  Before
  image

  After
  image

</details>

<details>
<summary>Bubble have different padding (more horizontal padding than prod). Fixed by removing the mat chip border radius that we override. The default border radius is rounder, taking up less visual space.</summary>

  Before
  image

  image

  After
  image

  image

</details>
  
**Review Instructions**
Verify that the usability issues above have been fixed in QA.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-7085

**Security**

### DIFF
--- a/src/app/source-file-tabs/source-file-tabs.component.html
+++ b/src/app/source-file-tabs/source-file-tabs.component.html
@@ -1,6 +1,6 @@
 <app-loading class="w-100" [loading]="loading">
-  <div *ngIf="version" class="p-3">
-    <mat-card appearance="outlined" class="alert alert-warning" *ngIf="validationMessage">
+  <div *ngIf="version && validationMessage" class="p-3">
+    <mat-card appearance="outlined" class="alert alert-warning">
       <mat-card-content>
         <mat-icon class="alert-warning-icon">warning</mat-icon>
         &nbsp;
--- a/src/_functions.scss
+++ b/src/_functions.scss
@@ -78,11 +78,11 @@
   @include mat-icon-size(16px);
 }
 
-// Sets the color of a selected tab's underline color
+// Sets the color of a selected tab's underline color.
+// Applies the style to children tab groups as well
 @mixin tab-group-indicator-color($color) {
   &.mat-mdc-tab-group,
   .mat-mdc-tab-group {
-    // Applies the style to children tab groups as well
     --mdc-tab-indicator-active-indicator-color: #{$color};
     --mat-tab-header-active-hover-indicator-color: #{$color};
     --mat-tab-header-active-focus-indicator-color: #{$color};


src/app/home-page/dashboard/dashboard.component.html
```diff
@@ -19,7 +19,7 @@
         <app-mastodon-timeline fxFlex.md="35" fxFlex.gt-md="35" fxFlex.lt-md="100"></app-mastodon-timeline>
       </div>
       <div class="w-100">
-        <div class="mat-card-title text-align-center">Featured Content</div>
```
kathy-t, line 22: This no longer worked because we were applying an internal Material class to a div and this class no longer exists. Instead of using the new class name, I made this an `h3` instead so it's less likely to break in the future

src/app/source-file-tabs/source-file-tabs.component.html
```diff
@@ -1,6 +1,6 @@
 <app-loading class="w-100" [loading]="loading">
-  <div *ngIf="version" class="p-3">
-    <mat-card appearance="outlined" class="alert alert-warning" *ngIf="validationMessage">
+  <div *ngIf="version && validationMessage" class="p-3">
```
kathy-t, line 2: Noticed that this extra padding was always being applied. Changed it so it only applies if there's validation messages, which is similar to the other mat-card alerts below

src/_material-overrides.scss
```diff
@@ -155,3 +154,16 @@ table th {
...
+.mat-mdc-card {
+  // All mat-cards have this border to make the top edge of a raised card more prominent
+  border: solid 1px mat.get-color-from-palette($dockstore-app-gray, 7) !important;
+}
+
+// Checkboxes
+.mat-mdc-checkbox {
+  // Make the checkbox size smaller than the default.
+  // Especially important for the search page facets so that they aren't too large.
+  --mdc-checkbox-state-layer-size: 24px;
```
kathy-t, line 168: Default is 28px, picked 24px because it looked similar enough to prod

denis-yuen (approved):

svonworl (approved):

PR #2090, opened 2025-04-25
Add buttons to Manage DOIs dialog to create/delete edit link for Dockstore DOIs

kathy-t:
**Description**
This PR adds buttons to the Manage DOIs dialog so that a user can create an edit access link to edit the metadata of an automatically generated Dockstore DOI on Zenodo. This single link can be used to edit all versions of the DOI. In order for the changes to be saved on Zenodo, the user must be logged into their Zenodo account, otherwise the attempt to publish the changes will fail with a 403.

Once the user is done editing the DOI, they can delete the access link from the Manage DOIs dialog. 

Take note of the sentence after the Dockstore DOIs. There is a "Show more" button that hides the details of the edit link so that the dialog is not too cluttered.

I'm not completely satisfied with the wording, so please feel free to leave suggestions.

image

image

image

**Review Instructions**
- Navigate to a workflow that you have with an automatically generated DOI
- Click the Manage DOIs button
- In the Dockstore card, click the "Show more" button
- Click the Create Edit Link button. A link should be generated and displayed in the card
- Click on the edit link. You should be directed to the DOI on Zenodo. Make sure you're logged into Zenodo.
- Click the Edit button on the right hand side then edit the metadata of the DOI (ex: modify the description). Then click the Publish button.
- The changes should be saved. Refresh the page and verify that your modification shows up.
- Go back to your Dockstore page. In the Manage DOIs dialog, delete the edit link. You should not be able to edit the DOI anymore on Zenodo.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6462

**Security**

  - Should update the docs in https://docs.dockstore.org/en/stable/end-user-topics/dois.html


src/app/shared/entry/doi/manage-dois/manage-dois-dialog.component.html
```diff
@@ -87,6 +87,41 @@ <h1 mat-dialog-title>Manage DOIs</h1>
...
+              <div class="size-small gray-caption">
+                You can edit the metadata of a published Dockstore DOI at anytime.
+                <button mat-button (click)="toggleShowMoreEditLinkInfo()" class="private-btn">
+                  <mat-icon class="mat-icon-sm">{{ showMoreEditLinkInfo ? 'keyboard_arrow_up' : 'keyboard_arrow_down' }}</mat-icon
+                  >{{ showMoreEditLinkInfo ? 'Collapse' : 'Show more' }}
+                </button>
+              </div>
+              <div *ngIf="showMoreEditLinkInfo">
+                <p class="size-small gray-caption">
+                  Dockstore uses Zenodo's link sharing feature to create a secret link with edit access so you can edit the DOI. You must be
```
denis-yuen, line 101: Ditto here

src/app/shared/entry/doi/manage-dois/manage-dois-dialog.component.html
```diff
@@ -87,6 +87,41 @@ <h1 mat-dialog-title>Manage DOIs</h1>
...
+                  logged into your Zenodo account when you open the link. When you are done editing the DOI, delete the link on Dockstore.
```
denis-yuen, line 102: Is there a reason to delete the link security wise or is this just being tidy?

src/app/shared/entry/doi/manage-dois/manage-dois-dialog.component.html
```diff
@@ -87,6 +87,41 @@ <h1 mat-dialog-title>Manage DOIs</h1>
...
+                </p>
+                <span *ngIf="dockstoreDoiAccessLink">
+                  <a
+                    [href]="
+                      doiInfo.value.versionDoi?.name ?? doiInfo.value.conceptDoi.name | zenodoAccessLink: dockstoreDoiAccessLink.token
+                    "
+                    target="_blank"
+                    rel="noopener noreferrer"
+                  >
+                    Edit DOI on Zenodo<mat-icon class="mat-icon-sm">open_in_new</mat-icon>
```
denis-yuen, line 112: Ok with this as a title, but when explaining more, kinda prefer record or deposit vs DOI

denis-yuen (commented):
For discussion

src/app/shared/entry/doi/manage-dois/manage-dois-dialog.component.html
svonworl: > Is there a reason to delete the link security wise or is this just being tidy?

Is it true that once someone possesses the secret link, they can log into their Zenodo account and edit the record, even if they aren't associated with it?  If yes, we probably want the edit links to be short-lived, to minimize the potential damage if they leak.  And, maybe we should filter the edit links from our censored database dump?

denis-yuen: Worthy questions.

A little confused by "You must be logged into your Zenodo account when you open the link" but I guess it's saying that it can be any zenodo account unlike Google Drive where you can share by link but still limit it to specific accounts. 
 https://help.zenodo.org/docs/share/user-sharing/ vs https://help.zenodo.org/docs/share/link-sharing/

It's probably kind of late in the game, did we consider user sharing and the pros/cons

kathy-t: > Is it true that once someone possesses the secret edit link, they can log into their Zenodo account and edit the record, even if they aren't associated with it? If yes, we probably want the edit links to be short-lived, to minimize the potential damage if they leak.

Yeah that's true, anyone that's logged in with the link can edit the record. There is an option to specify an expiration date in the Zenodo API when requesting the link, but that requires modifying the webservice. Another option is we automatically delete the edit link when the dialog is closed. Let me know what is preferred.

>  And, maybe we should filter the edit links from our censored database dump?

Good news is that we don't store the token needed for the access link :slightly_smiling_face:. More details in the Shared Access Links section of this PR https://github.com/dockstore/dockstore/pull/5879, but we only store the access link ID. Every time we need an edit link, we call the Zenodo API to get the edit link token using the edit link ID.

> did we consider user sharing and the pros/cons

No, I don't think I ever presented the user sharing option. It seems a bit stricter since you need the user's Zenodo name/username/email to share the record with them. I also don't see the user sharing API documented https://inveniordm.docs.cern.ch/reference/rest_api_drafts_records/#access-links.

denis-yuen: > Let me know what is preferred. 

I don't really have a preference here, think it can be left as is, especially for 1.17. 

>  It seems a bit stricter since you need the user's Zenodo name/username/email to share the record with them 

Interesting. Presumably they would need an account anyway even for link sharing, but not having access to the API is definitely a problem. I think this may be a follow-up ticket.

svonworl: I'm ok with the current implementation in this PR, if it's the case that if the access link leaks, the user has a straightforward way to disable it (via the Dockstore or Zenodo UI).  If not, we should probably make it expire.  A bad scenario is if the user generates the link, leaves Dockstore, leaks the link, and there's no easy way to disable it.

kathy-t: > I'm ok with the current implementation in this PR, if it's the case that if the edit link leaks, the user has a straightforward way to disable it (via the Dockstore or Zenodo UI)

The user will be able to delete in the Dockstore UI via the Manage DOIs dialog at anytime. I'll create a follow-up ticket to set an expiration date anyway in the case that the user forgets to delete it

src/app/shared/entry/doi/manage-dois/manage-dois-dialog.component.ts
```diff
@@ -90,6 +101,21 @@ export class ManageDoisDialogComponent {
...
 
+  requestDoiAccessLink() {
+    this.manageDoisDialogService.requestDoiAccessLink(this.entry).subscribe((accessLink) => {
+      this.dockstoreDoiAccessLink = accessLink;
+    });
+  }
+
+  deleteDoiAccessLink() {
+    this.manageDoisDialogService.deleteDoiAccessLink(this.entry);
+    this.dockstoreDoiAccessLink = undefined;
```
svonworl, line 112: Borderline nitpicky, but we might not unset `dockstoreDoiAccessLink` until the observable has successfully completed.  That way, the user can try again in the case of failure (due to network problems or whatever).  With this change, the implementation/flow of `deleteDoiAccessLink` looks more similar to `requestDoiAccessLink`.

svonworl (approved):

src/app/shared/entry/doi/manage-dois/manage-dois-dialog.component.ts
```diff
@@ -90,6 +101,22 @@ export class ManageDoisDialogComponent {
     this.manageDoisDialogService.saveDoiSelection(this.entry, this.selectedOption);
   }
 
+  requestDoiAccessLink() {
+    this.manageDoisDialogService.requestDoiAccessLink(this.entry).subscribe((accessLink) => {
+      this.dockstoreDoiAccessLink = accessLink;
```
svonworl, line 106: Out of curiosity, do `workflowsService.getDOIEditLink` (called in `ngOnInit`) and `manageDoisDialogService.requestDoiAccessLink` do the same thing?

svonworl (approved):

src/app/shared/entry/doi/manage-dois/manage-dois-dialog.component.ts
kathy-t: Not quite: `workflowsService.getDOIEditLink` gets the edit link if it exists, otherwise returns a 404. It does not create edit links. `manageDoisDialogService.requestDoiAccessLink` creates an edit link if it does not exist and throws an error if it does.

src/app/shared/entry/doi/manage-dois/manage-dois-dialog.component.html
```diff
@@ -87,6 +87,42 @@ <h1 mat-dialog-title>Manage DOIs</h1>
...
+                You can edit the metadata of a published Dockstore DOI at anytime.
+                <button mat-button (click)="toggleShowMoreEditLinkInfo()" class="private-btn">
+                  <mat-icon class="mat-icon-sm">{{ showMoreEditLinkInfo ? 'keyboard_arrow_up' : 'keyboard_arrow_down' }}</mat-icon
+                  >{{ showMoreEditLinkInfo ? 'Collapse' : 'Show more' }}
+                </button>
+              </div>
+              <div *ngIf="showMoreEditLinkInfo">
+                <p class="size-small gray-caption">
+                  Dockstore uses Zenodo's link sharing feature to create a secret link with edit access so you can edit the record. You must
+                  be logged into your Zenodo account when you open the link. When you are done editing the record, delete the link on
```
denis-yuen, line 102: Maybe "your Zenodo account" should "a Zenodo account" on the off chance that the account they link in the accounts page is different. Minor point

denis-yuen (approved):

kathy-t:
FYI, I changed the browser used for the integration tests to Chrome. After integration_test_2 failed 5 times, I made the switch and it passed on the first try. Maybe a coincidence but we can monitor to see if the flakiness continues with Chrome.

kathy-t:
Created https://ucsc-cgl.atlassian.net/browse/SEAB-7141 and https://ucsc-cgl.atlassian.net/browse/SEAB-7140

PR #2091, opened 2025-04-25
SEAB-7015: add feature flag and preview buttons for snakemake

hyunnaye:
**Description**
This PR adds the preview bubble and header to registered Snakemake workflows and to Snakemake workflow registration radio buttons.

This PR adds an `enableSnakemake` feature flag which controls whether or not we display the Snakemake descriptor language in the Register Workflow dialog. When the feature is set to `true`, the Snakemake registration option is displayed in the UI. If it is set to `false`, then the Snakemake registration option is hidden. Once a Snakemake workflow is registered, the workflow will be displayed on the site with a preview banner and bubble regardless of the feature flag state.

The preview banner is displayed:
- On the Register Workflow dialog for custom and hosted workflows
- On the My Workflow page of the Snakemake workflow

The preview bubble is displayed:
- On the search page, beside the Snakemake bubble for a snakemake workflow
- On the Workflow page (private and public) beside the Snakemake descriptor type in the Workflow Information box
- On the Register Workflow dialog beside the Snakemake radio button for registering hosted and custom workflows

<details>
<summary>Screenshots</summary>

Private page
image

Public page 
image

Search page
image

Register Workflow dialog if feature is enabled for custom workflows:

image

image

Preview banner not displayed if a non-Snakemake option is selected
image

Register Workflow dialog if feature is enabled for hosted workflows:

image

image

Preview banner not displayed if a non-Snakemake option is selected
image

</details>

**Review Instructions**

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-7015
**Security**

### DIFF
--- a/src/app/preview-warning/preview-warning.component.html
+++ b/src/app/preview-warning/preview-warning.component.html
@@ -1,4 +1,4 @@
-<mat-card appearance="outlined" class="alert alert-warning w-100" data-cy="preview-warning">
+<mat-card appearance="outlined" class="alert alert-warning w-100 mb-0" data-cy="preview-warning">
   <mat-card-content>
     <mat-icon class="alert-warning-icon">warning</mat-icon>
     <strong> Notice:</strong> {{ featureName }} is an early access feature. Consider this a prototype that may be more likely to have
--- a/src/app/preview-warning/preview-warning.component.html
+++ b/src/app/preview-warning/preview-warning.component.html
@@ -1,7 +1,17 @@
-<mat-card appearance="outlined" class="alert alert-warning w-100 mb-0" data-cy="preview-warning">
+<mat-card *ngIf="!isBubble" appearance="outlined" class="alert alert-warning w-100 mb-0" data-cy="preview-warning">
   <mat-card-content>
     <mat-icon class="alert-warning-icon">warning</mat-icon>
     <strong> Notice:</strong> {{ featureName }} is an early access feature. Consider this a prototype that may be more likely to have
     breaking changes going forward than other Dockstore features.
   </mat-card-content>
 </mat-card>
+<mat-chip
+  *ngIf="isBubble"
+  class="preview-bubble"
+  [matTooltip]="
+    featureName +
+    ' is an early access feature. Consider this a prototype that may be more likely to have
+breaking changes going forward than other Dockstore features.'
+  "
+  >PREVIEW</mat-chip
+>
--- a/src/app/preview-warning/preview-warning.component.html
+++ b/src/app/preview-warning/preview-warning.component.html
@@ -1,17 +1,7 @@
 <mat-card *ngIf="!isBubble" appearance="outlined" class="alert alert-warning w-100 mb-0" data-cy="preview-warning">
   <mat-card-content>
     <mat-icon class="alert-warning-icon">warning</mat-icon>
-    <strong> Notice:</strong> {{ featureName }} is an early access feature. Consider this a prototype that may be more likely to have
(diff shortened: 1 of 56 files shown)


denis-yuen:
> Search page with banner and badge:

This one  is a bit overkill unless it only shows up when the SMK facet is selected. Otherwise, I think we can skip. 
The filetree one is also overkill (i.e. we don't have workflows that fix languages ... yet)

Remaining place:

This dialog (and the very similar one for hosted workflows)
Screenshot from 2025-04-25 13-40-22

hyunnaye:
> > Search page with banner and badge:
> 
> This one is a bit overkill unless it only shows up when the SMK facet is selected. Otherwise, I think we can skip. The filetree one is also overkill (i.e. we don't have workflows that fix languages ... yet)
> 
> Remaining place:
> 
> This dialog (and the very similar one for hosted workflows) Screenshot from 2025-04-25 13-40-22

I added the preview button on this dialog but I cannot get the screenshots/see how it looks due to some weird error on my local where I can't see snakemake as one of the options in the dialog. So whoever takes this ticket, please try on your local and see how it looks because I cannot test it.

src/app/preview-warning/preview-warning.component.html
```diff
@@ -1,7 +1,17 @@
-<mat-card appearance="outlined" class="alert alert-warning w-100" data-cy="preview-warning">
+<mat-card *ngIf="!isBubble" appearance="outlined" class="alert alert-warning w-100 mb-0" data-cy="preview-warning">
   <mat-card-content>
     <mat-icon class="alert-warning-icon">warning</mat-icon>
     <strong> Notice:</strong> {{ featureName }} is an early access feature. Consider this a prototype that may be more likely to have
```
denis-yuen, line 4: Worth making/possible to make this String a readonly variable in the service?

src/app/shared/dockstore.model.ts
```diff
@@ -104,5 +104,6 @@ export class Dockstore {
     enableLaunchWithGalaxy: true,
     enableMultiCloudLaunchWithDNAstack: false,
     enableOrcidExport: true,
+    enableSnakemake: false,
```
denis-yuen, line 107: think we can turn this on for a staging release

src/app/preview-warning/preview-warning.component.html
kathy-t: Yeah it's doable and a good idea since it's also the bubble's tooltip, will implement

src/app/shared/dockstore.model.ts
kathy-t: Hmm this value is controlled by the presence of `?snakemake` in the URL. Are you saying we should turn it on for staging regardless of the `?snakemake` parameter in the URL?

denis-yuen: > I don't see the purpose in having the URL "turn off" the feature temporarily if the default is on.

Yeah, you're probably right. I was probably more interested in being able to turn on/off the feature globally. Agreed, the url control probably only be useful if if the default was off. 

> In the future, it might be better to control the features from the webservice through the web.yaml so that we don't need a new UI release to turn a feature on/off 

I can see the argument, config change vs new build is more convenient. Feels a little strange to flip a change in the UI from the webservice, but I don't see a specific problem.

kathy-t: I'll set the snakemake default as true so it's on and remove the URL control

svonworl (approved):

denis-yuen (approved):

PR #2092, opened 2025-04-29
SEAB-7081: Add launch information for snakemake

svonworl:
**Description**
This PR adds some rudimentary instructions to the "Launch" tab for snakemake workflows, and makes some cosmetic changes to the order of some properties/methods in the interest of improving clarity.

Per a discussion with Denis, the instructions recommend a "github clone" file download and `snakemake` invocation.

The good thing about a "github clone" download is that it:
* downloads all of the necessary files (including those which our new plugin might have [inadvertently] not pulled into Dockstore).
* works for unpublished workflows.

The bad thing about the "github clone" approach is that it:
* doesn't support non-GitHub-based workflows (hosted, GitLab, etc).
* might download a bunch of extra/unnecessary junk.

If the workflow isn't GitHub-based, no instructions appear.

I'm guessing the number of non-GitHub-based snakemake workflows, at least in the short term, will be close to zero, so the above is probably ok.

Sidebar:

This PR is representative of the difficulty of making simple modifications to our UI code.  In a typical codebase, this change would have required about a dozen lines of code, tops, and been fairly localized.  The layered services, indirection, webs of observables, and duplicate code [due to the inherent friction of Angular], all contribute to making it tedious and time-consuming.

**Review Instructions**
View a snakemake workflow on qa, and confirm the instructions appear.  Give the instructions a try, and confirm that they do the right thing.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-7081

**Security**

### DIFF
--- a/src/app/workflow/launch/launch.component.html
+++ b/src/app/workflow/launch/launch.component.html
@@ -138,35 +138,6 @@
           </mat-card-content>
         </mat-card>
 
-        <mat-card *ngIf="(isSMK$ | async) && DockstoreToolType.ModeEnum.HOSTED !== mode && WorkflowType.ModeEnum.HOSTED !== mode">
-          <mat-card-content>
-            <div [matTooltip]="snakemakeTooltip">
-              To run a Snakemake workflow locally, download the workflow files:
-              <button
-                mat-icon-button
-                class="copy-button"
-                matTooltip="Copy command"
-                [cdkCopyToClipboard]="snakemakeGetWorkflowCommand"
-                appSnackbar
-              >
-                <mat-icon>file_copy</mat-icon>
-              </button>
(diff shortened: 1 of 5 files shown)


src/app/workflow/launch/launch.component.html
```diff
@@ -168,6 +168,35 @@
           </mat-card-content>
         </mat-card>
 
+        <mat-card *ngIf="(isSMK$ | async) && 'github.com' === workflow.sourceControl">
+          <mat-card-content>
+            <div [matTooltip]="snakemakeTooltip">
+              Download the workflow files
```
denis-yuen, line 174: "Clone the workflow files" , shorter and more descriptive  IMHO

denis-yuen (approved):

kathy-t (approved):

PR #2093, opened 2025-04-30
DOCK-2617: Fix workflow versions dropdown

svonworl:
**Description**
This PR fixes the "versions" dropdown on the private workflows page, so that the selection changes when a version in the list below is clicked.  The problem was that `WorkflowVersion`s were being compared for equality, but at some point (the "paged version" implementation), our code changed so that the compared versions were sometimes different objects, and thus "not equal", even though their contents were the same.

The fix is to add a `compareWith` function to the `mat-select` that compares by version ID.

There are about 26 `mat-select`s in the current UI code, one wonders how many are similarly afflicted.

One proactive defense against future problems would be to create a `compareWith` function that would compare by ID and, if not possible, fall back to normal equality.  Something like:

```
    equivalent(a: any, b: any): boolean {
        return (a.id !== undefined || b.id !== undefined) ? a.id === b.id ? a === b;
    }
```
And add it to all of the mat-select`s.  Or maybe there's a way to specify it as the default, without having to change each one.

Might be too risky/late in the release cycle to do it for 1.17, though.

**Review Instructions**
Follow the steps in https://ucsc-cgl.atlassian.net/browse/DOCK-2617 that detail how to reproduce the bug, and confirm that it no longer happens.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2617
https://github.com/dockstore/dockstore/issues/6102

**Security**

### DIFF
--- a/src/app/workflow/workflow.component.html
+++ b/src/app/workflow/workflow.component.html
@@ -107,6 +107,7 @@
           <mat-select
             (selectionChange)="onSelectedVersionChange(selectedVersion)"
             [(value)]="selectedVersion"
+            [compareWith]="equalIds"
             [formControl]="workflowVersionsCtrl"
           >
             <mat-option>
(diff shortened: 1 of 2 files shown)


denis-yuen (approved):

kathy-t (approved):
Good find, Steve! I was scratching my head over why it only affected workflows :slightly_smiling_face:

PR #2094, opened 2025-05-01
Fix staging/QA auth tests

kathy-t:
**Description**
This PR fixes the failing auth tests in staging and QA. The fixes may not follow Cypress best practices (i.e. using data-cy), but it is a stop-gap fix that works for all 3 environments. This PR adds several data-cy attributes to the elements used in the auth smoke tests, but since prod won't have these attributes immediately, using these data-cy attributes in this PR will cause prod smoke tests to fail. Shortly after this PR is deployed to prod, we can fix up the auth tests so that they use these data-cy attributes.

See that the auth smoke tests pass for all three environments here https://app.circleci.com/pipelines/github/dockstore/dockstore-ui2/13739/workflows/d82cc822-4663-4799-850d-25524b09d024

**Review Instructions**
Auth tests for all environments should pass

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-7129

**Security**

### DIFF
--- a/cypress/e2e/smokeTests/sharedTests/auth-tests.ts
+++ b/cypress/e2e/smokeTests/sharedTests/auth-tests.ts
@@ -96,11 +96,7 @@ function registerQuayTool(repo: string, name: string) {
     });
     cy.contains('mat-option', repo).should('be.visible').click();
     cy.wait('@repos');
-    cy.get('mat-dialog-content').within(() => {
-      cy.contains('div', name).within(() => {
-        cy.contains('mat-icon', 'sync').should('be.visible').click();
-      });
-    });
+    cy.get('[data-cy=refresh-repo-button]').should('be.visible').click();
     cy.wait('@containers');
     cy.contains('button', 'Finish').should('be.visible').click();
     cy.url().should('contain', toolName);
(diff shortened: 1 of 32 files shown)


cypress/e2e/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -40,11 +40,6 @@ function storeToken() {
   }
 }
 
-function hasSandboxZenodo() {
```
kathy-t, line 43: Removed because unused

cypress/e2e/smokeTests/sharedTests/auth-tests.ts
```diff
@@ -215,10 +205,10 @@ function testTool(registry: string, repo: string, name: string) {
       cy.visit('/my-tools');
       goToVersionsTab();
       toggleHiddenToolVersion();
-      cy.get('[data-cy=hiddenCheck]').should('have.length', 1);
+      cy.get('.mat-column-hidden').contains('mat-icon', 'check').should('have.length', 1);
```
kathy-t, line 208: Changed because the data-cy name changed so that the tool and workflow tables have the same data-cy name. This is only temporary and will be changed back to using data-cy when this PR is merged to develop.

svonworl (approved):

denis-yuen (approved):

PR #2132, opened 2025-09-05
SEAB-7194: Add "small multiples" weekly execution count graphs to versions table

svonworl:
**Description**
This PR adds replaces the "Metrics" column in the workflow "Versions" table with a "Weekly Runs" column, which contains a thumbnail-sized bar graph (histogram) of the weekly execution counts (when present) for each version.  The graphs display a single bar per week which represents the execution count for ALL platforms and ALL statuses.  If a weekly execution count time series has not been aggregated for a version, no graph is displayed.

All of the graphs have the same X axis range (same weeks) and same Y axis (a range from zero to the maximum weekly execution count for any version).  Formatted thusly, the graphs can easily be compared with one another, to determine, for example, which version is used more, or which version has the most recent usage.

When the version table component starts, it retrieves the maximum weekly execution count from the webservice using the new endpoint described in https://github.com/dockstore/dockstore/pull/6156.  Thus, we avoid the need to pull a bunch of version information (beyond what's on the page) over to the client-side to compute the value, which could be very slow and consume lots of bandwidth.

Right now, this PR points at the webservice branch that has the corrected `openapi.yaml` description as described in https://github.com/dockstore/dockstore/pull/6157.  Once it merges, I will retarget `package.json` appropriately.

To implement the graphs, we create a new component `ThumbnailTimeSeriesGraphComponent`.  Eventually, we should also similarly componentize the metrics graphs that appear in the "Metrics" tab, because it will make the code clearer and be easier  to reuse them elsewhere (in the workflow-level metrics display, on the search page, etc).

One can imagine multiple layers of graph components, wherein the lowest layer represents the basic types of graphs (bar chart, line graph, pie chart, etc), and higher-level components that adapt our data to these basic types (for example, component that translates a `TimeSeriesMetric` to the values/labels/legends/colors etc which are fed into the basic bar graph component).

For now, mouseover details are disabled because the default display is rendered within the canvas, which is too small to contain all the information.  We can implement a different mechanism later if we feel like it is important.

Helper time-series-related functions have moved to `TimeSeriesService` so they can be used in multiple places.

I didn't link the graphs to the Metrics tab, because, as it turns out, the previous column values weren't linked, either.  Such linking has historically been a source of bugs and is a little bit fiddly to implement, so I left it out for now.

Right now, there's 11 weeks of data (bars) in the graphs because that's what seemed to look best.

In the long term, it might be good to aggregate monthly execution counts, also, and use that data in the thumbnails (so that the graphs in the "Versions" tab and graph in the "Metrics" tab both represent approximately one year). 

**Screenshot**
image

**Review Instructions**
On staging, after a full aggregation has been run, view the workflow page for some workflows that have execution metrics, select the "Versions" tab, and confirm that the graphs look correct, by comparing to the raw time series returned by the webservice, and also verifying that the graphs on the "Metrics" tab are consistent.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-7194

**Security**

### DIFF
--- a/src/app/shared/graphs/thumbnail-time-series-graph.component.ts
+++ b/src/app/shared/graphs/thumbnail-time-series-graph.component.ts
@@ -56,6 +56,7 @@ export class ThumbnailTimeSeriesGraphComponent implements OnChanges {
             },
             grid: {
               tickLength: 0,
+              color: '#ddd',
             },
           },
           y: {
@@ -66,6 +67,7 @@ export class ThumbnailTimeSeriesGraphComponent implements OnChanges {
             },
             grid: {
               tickLength: 0,
+              color: '#ddd',
             },
           },
         },
--- a/src/app/workflow/versions/versions.component.html
+++ b/src/app/workflow/versions/versions.component.html
@@ -216,13 +216,16 @@
         >
       </th>
       <td mat-cell *matCellDef="let version">
+        @let timeSeries = version?.metricsByPlatform?.['ALL']?.executionStatusCount?.count?.['ALL']?.weeklyExecutionCounts;
         <div class="ml-4">
           <app-thumbnail-time-series-graph
-            [timeSeries]="version?.metricsByPlatform?.['ALL']?.executionStatusCount?.count?.['ALL']?.weeklyExecutionCounts"
+            *ngIf="timeSeries"
+            [timeSeries]="timeSeries"
             [now]="now"
             [binCount]="numberOfWeeksToGraph"
             [maxValue]="maxWeeklyExecutionCount"
           ></app-thumbnail-time-series-graph>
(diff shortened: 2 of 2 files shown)


denis-yuen (commented):
question about charts

denis-yuen:
> My personal feeling is that the end user doesn't care whether there's a metrics entry in the db or not, they just want to see how often the version has been run. 

I think there's a grey area when the workflow was run a lot previously but the runs have scrolled off to the left. 
i.e. a user would still prefer a workflow with some metrics even if they're old vs no metrics at all

Maybe making the X-Y axes darker and a tooltip (if not already present)  might be sufficient

denis-yuen (commented):
comment above

PR #2175, opened 2026-02-05
SEAB-7439: Fix flash of stale search results

svonworl:
**Description**
Previously, in certain situations, stale results briefly appear on the search page.  Typically, this was most noticeable when the user runs a search, navigates to another page, and then returns to the search page, in which case the previous results would appear for a split second before being replaced.  For example, if the user ran a "covid" search, then left and returned again via the navbar, the "covid" results would appear briefly.

This PR partially addresses the problem by more completely clearing the state when the search component initializes.

I also tried a different solution, which was to simply not display the results when `isLoading` was true.  However, it felt too jolty for parts of the search UI to completely blink out and then back in again.  I kinda like the way the search results lazily update...

**Review Instructions**
On qa, try the triggering navigation sequence as detailed above, and confirm that stale results do not briefly appear when you return to the search page.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-7439

**Security**

### DIFF
--- a/src/app/search/search.component.ts
+++ b/src/app/search/search.component.ts
@@ -261,6 +261,7 @@ export class SearchComponent implements OnInit, OnDestroy {
     });
 
     this.hits = [];
+    this.searchService.setHits([], [], []);
 
     this.aNDSplitFilterText$ = this.advancedSearchQuery.aNDSplitFilterText$;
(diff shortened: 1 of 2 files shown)


denis-yuen:
re-ran failed build

denis-yuen (commented):
Build failure seems consistent

denis-yuen (approved):

PR #2199, opened 2026-04-08
minimally viable PKCE working for github login and account linking

denis-yuen:
**Description**
Passes through the state and pkce code challenge values to github (so that they can then be sent to the webservice via a 302 redirect), and then the original code verifier value can be sent to github to exchange for an access token (after checking that the state value matches, so we're not just being attacked)
See comments linked in JIRA ticket

**Review Instructions**
Login via GitHub, do account linking on the accounts page, and use the link button on the my tools//my workflows page. Track the calls going to github and redirected back to Dockstore. Should see additional query parameters for state in both directions and a code challenge doing toward GitHub. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-7521
See tickets linked to https://github.com/dockstore/dockstore/pull/6246

**Security**
Should be improvement, but it would be nice to get this looked at with a more dedicated eye


denis-yuen:
Asking for review to check out the general structure and ideas behind the PR, will not merge until dealing with the DB spin-off ticket

Also ideas for cleaning up the code and making it more "angular"

svonworl (commented):
Something like this could work, but I have a few observations:
* IMHO, it would be more simple, and closer to the expected information flow, if the client generated the verifier and state and stored them client-side (in sessionStorage?) for later use.
* This PR might not be performing all of the security checks, like verifying that the `state` specified in the request matches the `state` in the response.  (I may have missed it.)
* The accompanying webservice PR appears to use the non-Dockstore-controlled `state` value (currently extracted from the redirect) to look up the verifier on the webservice.  This makes me feel a little funny.  Not sure what could go wrong here, safety or otherwise, but security is hard to get right.  This bullet point is less relevant if we actually check that `state` has the right value.
  
If I understand correctly, the canonical reference for PKCE-based Github auth is:
https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#web-application-flow

Basically, the flow is as follows:
1. Client generates random code `verifier` and `state`.
2. Client hits the `oauth/authorize` endpoint, including the `state` and hashed `challenge` generated from the code `verifier`.
3. Client receives redirect that contains `code` and `state`, and stops if the received `state` value doesn't match.
4. Client hits the `access_token` endpoint, including `code` and the `verifier`.
5. Client receives bearer token, ...

In our current system, sans PKCE, steps 1-3 happen in the browser, and the browser hands off to the webservice, which performs the rest of the steps.  So, as architected above, to support PKCE, we'd need to pass the verifier to the webservice prior to step 4 (probably by adding it to the parameters of the endpoint that triggers step 4). 

Why not implement it that way?

denis-yuen:
Thanks for the questions, it's good to ponder this. 

> IMHO, it would be more simple, and closer to the expected information flow, if the client generated the verifier and state and stored them client-side (in sessionStorage?) for later use.

I don't think this would work for a couple reasons. 

The first is that before this PR, it's the webservice in our architecture that exchanges the `code` and `client_secret` for an `access token`. After this PR, it adds the `verifier` after checking that the `state` is valid. If the Angular client generated those two values, they would essentially be stranded and unable to be used by the webservice for the exchange. 
The second is that it's the webservice which is secure and the UI/client which is not. If the client was in charge of generating those values, it seems that an attacker could fake those values?  


> * The accompanying webservice PR appears to use the non-Dockstore-controlled `state` value (currently extracted from the redirect) to look up the verifier on the webservice.  This makes me feel a little funny.  Not sure what could go wrong here, safety or otherwise, but security is hard to get right.  This bullet point is less relevant if we actually check that `state` has the right value.

The state is generated by the webservice here https://github.com/dockstore/dockstore/pull/6246/changes#diff-b6618b98b37958d417997af083f649fdc1fd603a59a702027f5cebb6c7d9ceecR606 and then checked here https://github.com/dockstore/dockstore/pull/6246/changes#diff-b6618b98b37958d417997af083f649fdc1fd603a59a702027f5cebb6c7d9ceecR672

In other words, the state value is controlled (randomly generated) by the webservice and returned by GitHub via redirect. Checking that they match is seems to be the extent to what a "right value" is. I'm not sure there are any more security checks that can be done with the state specifically. 
e.g. https://auth0.com/docs/secure/attack-protection/state-parameters#csrf-attacks 

Or from another POV, an attacker pretending to be the UI client would need to guess random state values, send them to GitHub and then we'd get random requests from GitHub with the wrong state value?


> If I understand correctly, the canonical reference for PKCE-based Github auth

I think you misunderstand (or I do, very possible!)
I was under the impression the reason why the UI should not be generating verifier and state values is that the UI is untrusted. Not specifically our Angular UI, but if someone wrote a UI that looks like our UI to steal credentials. 

The attack would look like
1. Fake UI generates state and verifier values
2. The user instructs the fake UI to hit GitHub and the UI gets a `code`.
3. Since the Fake UI has the verifier, it does the exchange of a `code` with GitHub along with the`verifier` value (and the `client_secret` somehow which is what I don't understand)  to do the exchange instead and gets an access token instead of the webservice and does malicious things. 

Instead, because only the webservice has access to the verifier, only the webservice can do the exchange and receive the access token.

denis-yuen:
I guess another way of answering this 

> Why not implement it that way? 

Is that if the UI can generate the verifier, state, and code challenge then the UI is only one secret (the client secret) away from doing the entire exchange itself. At that point, we're basically dealing with "implicit flow" which is what OAuth 2.1 seems to be recommending against

denis-yuen:
> 21:31:49.347 INFO  Check ALM binding of project 'dockstore_dockstore-ui2'
> 21:31:49.508 INFO  Detected project binding: BOUND
> 21:31:49.509 INFO  Check ALM binding of project 'dockstore_dockstore-ui2' (done) | time=161ms
> 21:31:49.513 INFO  Load branch configuration
> 21:31:50.372 ERROR Could not find the pullrequest with key 'null'
> 21:31:50.706 INFO  EXECUTION FAILURE

Seems like something has gone awry, will retry tomorrow or perhaps open a new  PR entirely

Edit to add: now seems resolved

src/app/login/login.service.ts
```diff
@@ -15,34 +15,51 @@
...
-          if ('status' in error) {
-            this.alertService.detailedError(error);
-          } else {
-            this.alertService.customDetailedError('Login failed', 'Could not login to Dockstore.');
-          }
-        }
-      );
+      //  looks like we can insert state and challenge here and thus hook in PKCE
+      this.tokenService
+        .getGitHubCodeChallenge()
```
svonworl, line 39: We should probably modify this block of code so that for all outcomes, we perform the appropriate `alertService` call.  As the code is now, if `getGitHubCodeChallenge` fails, we emerge without taking any action, the `alertService` is left "open", and the failure is silent.  This is also true of the similar code in `RegisterService`.  The minimal fix would be to pass a similar error handling function to the outer `subscribe`.  I believe there's also more elegant ways to chain two observables together, but the details escape me at the moment...

svonworl (approved):

PR #2222, opened 2026-06-16
fix github app notification paging

denis-yuen:
**Description**
Turns out we're just not taking into account pageSize when switching between pages, so there's overlap especially with relatively small numbers of notifications

**Review Instructions**
Get 6 of more notifications. They will overlap in production 1.19.X but should not overlap after this PR

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2635

**Security**
None

### DIFF
--- a/src/app/myworkflows/sidebar-accordion/github-apps-logs/recommended-actions.component.ts
+++ b/src/app/myworkflows/sidebar-accordion/github-apps-logs/recommended-actions.component.ts
@@ -113,6 +113,6 @@ export class RecommendedActionsComponent extends Base implements OnInit, AfterVi
   }
 
   private loadEvents() {
-    this.dataSource.loadEvents(this.paginator.pageIndex, this.paginator.pageSize);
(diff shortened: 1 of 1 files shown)


src/app/myworkflows/sidebar-accordion/github-apps-logs/recommended-actions.component.ts
```diff
@@ -113,6 +113,6 @@ export class RecommendedActionsComponent extends Base implements OnInit, AfterVi
   }
 
   private loadEvents() {
-    this.dataSource.loadEvents(this.paginator.pageIndex, this.paginator.pageSize);
+    this.dataSource.loadEvents(this.paginator.pageIndex * this.paginator.pageSize, this.paginator.pageSize);
```
denis-yuen, line 116: Compare with https://github.com/dockstore/dockstore-ui2/blob/2.16.1/src/app/shared/tool-lister.ts#L131

svonworl (approved):

PR #2235, opened 2026-07-03
SEAB-7551: Add owner category curation dialog

svonworl:
**Description**
This PR adds a dialog which an entry owner can use to approve/remove AI-curated membership in a category.  This new dialog is displayed when the entry owner clicks the new "Manage Categories" button, which appears below the categories on the private entry page.

A screenshot of the dialog is worth a thousand words:
image

Essentially, next to each AI-curated category membership, two buttons are displayed.  The "X" button will remove the entry from the category, and the "check" button will approve the entry's membership in the category.  The owner makes the desired changes, than presses the "Save" button at bottom, which persists the changes in the webservice.

To persists each change, the "save" code hits a webservice endpoint.  Where there are `N` changes, the code executes `N-1` changes in parallel, and after they all finish, it executes the single remaining change.  Thusly, all of the previous database changes are visible to the final endpoint invocation, and the corresponding ES indexing request that it submits always arrives last to the ES server, presumably giving it priority over everything that happened before it. 

**Review Instructions**
After one of your entries has AI-curated category memberships, go to the private entry page, click "Manage Categories", and make some adjustments.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-7551

**Security**

### DIFF
--- a/src/app/categories/manage/manage-categories-dialog.component.html
+++ b/src/app/categories/manage/manage-categories-dialog.component.html
@@ -1,5 +1,11 @@
 <h1 mat-dialog-title>Manage Categories</h1>
 <mat-dialog-content>
+  <p class="gray-caption">
+    Dockstore uses AI to classify {{ entryTypeMetadata.termPlural }} into various types of categories. For any AI-curated classification,
+    click the "check" button to approve of your {{ entryTypeMetadata.term }}'s membership, or the "X" button to remove your
+    {{ entryTypeMetadata.term }} from the category. Press the "Save" button at bottom when you are done.
+  </p>
+
   <app-alert></app-alert>
 
   <div *ngIf="categories.length === 0" class="mt-2 mb-2">
--- a/src/app/categories/manage/manage-categories-dialog.component.html
+++ b/src/app/categories/manage/manage-categories-dialog.component.html
@@ -1,0 +1,84 @@
+<h1 mat-dialog-title>Manage Categories</h1>
+<mat-dialog-content>
+  <app-alert></app-alert>
+
+  <div *ngIf="categories.length === 0" class="mt-2 mb-2">
+    <p class="gray-caption">This entry is not in any categories.</p>
+  </div>
+
+  <ng-container *ngFor="let group of groupOrder">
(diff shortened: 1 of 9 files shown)


src/app/categories/manage/manage-categories-dialog.component.html
```diff
@@ -0,0 +1,90 @@
+<h1 mat-dialog-title>Manage Categories</h1>
+<mat-dialog-content>
+  <p class="gray-caption">
+    Dockstore uses AI to classify {{ entryTypeMetadata.termPlural }} into various types of categories. For any AI-curated classification,
+    click the "check" button to approve of your {{ entryTypeMetadata.term }}'s membership, or the "X" button to remove your
```
denis-yuen, line 5: Wonder if you could use the actual checkmark image or a unicode https://www.compart.com/en/unicode/U+2713 character

src/app/categories/manage/manage-categories-dialog.component.html
```diff
@@ -0,0 +1,90 @@
...
+    {{ entryTypeMetadata.term }} from the category. Press the "Save" button at bottom when you are done.
+  </p>
+
+  <app-alert></app-alert>
+
+  <div *ngIf="categories.length === 0" class="mt-2 mb-2">
```
denis-yuen, line 11: Does it make sense to even allow the user to click in if there are no categories and we have no way (?) to submit new categories short of swagger.

denis-yuen (approved):
A couple comments/questions

src/app/categories/manage/manage-categories-dialog.component.html
svonworl: When there are no categories, the private entry page doesn't display the "Manage Categories" button, so this code is a backup, and very rarely will be executed.

src/app/categories/manage/manage-categories-dialog.component.html
svonworl: Good ideas, but let's stick with what's here, it's probably sufficient.  If we display anything other than run-of-the-mill Unicode, we risk there not being coverage in all browsers.  We could display the buttons themselves, but there's formatting (line height) considerations...

PR #2265, opened 2026-08-04
allow publishing of wdl 1.1+

denis-yuen:
**Description**
Temporary workaround to allow publishing of workflows detected as WDL 1.1 and above. 
Pending a proper update to our cromwell code for parsing 
 On merge, I will also cherry-pick this to a hotfix branch

**Review Instructions**
Can test with https://github.com/DockstoreTestUser2/TestEntryValidation/blob/master/valid_1_1_Workflow.wdl 
Register and publish with only a WDL 1.1 version registered and visible on the versions page

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1700
https://github.com/dockstore/dockstore/issues/4021

**Security**
None

### DIFF
--- a/src/app/shared/entry-actions/entry-actions.service.ts
+++ b/src/app/shared/entry-actions/entry-actions.service.ts
@@ -97,11 +97,11 @@ export class EntryActionsService {
   }
 
   private isWorkflow(entry: Entry): entry is Workflow {
-    return entry.entryType === 'WORKFLOW';
+    return entry.entryType === OpenApiEntryType.WORKFLOW;
   }
 
   private isTool(entry: Entry): entry is DockstoreTool {
-    return entry.entryType === 'TOOL' || entry.entryType === 'APPTOOL';
+    return entry.entryType === OpenApiEntryType.TOOL || entry.entryType === OpenApiEntryType.APPTOOL;
   }
 
   private isEntryValid(entry: Entry | null): boolean {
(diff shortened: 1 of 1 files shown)


denis-yuen:
See also https://github.com/dockstore/dockstore/pull/6327

svonworl (approved):

src/app/shared/entry-actions/entry-actions.service.ts
```diff
@@ -106,6 +106,17 @@ export class EntryActionsService {
     if (!versionTags) {
       return false;
     }
+    // WDL 1.1 special case (version is not blank and doesn't start with draft and is 1.1 or higher)
+    if (
```
svonworl: Do we need to check if the entry is a workflow and WDL, somewhere in here?

src/app/shared/entry-actions/entry-actions.service.ts
```diff
@@ -95,6 +96,14 @@ export class EntryActionsService {
     }
   }
 
+  private isWorkflow(entry: Entry): entry is Workflow {
```
denis-yuen, line 99: This works locally but is dumb, is there a better way to do this in the codebase right now? 
Google-fu failing me @svonworl

svonworl: There's an `entryType` field in the entry objects that you can use.  But watch out, because there's two different `EntryType` classes floating around.  The OpenAPI version is the one you want (I think):
```
import { EntryType } from 'app/shared/openapi';
[...] entry.entryType === EntryType.WORKFLOW [...]
```

svonworl (approved):

PR #2275, opened 2026-08-20
SEAB-7673: Upgrade to Angular 21

svonworl:
**Description**
This PR updates the UI to use Angular 21 (from Angular 20). We performed the steps outlined by the Angular upgrade guide (https://angular.dev/update-guide?v=20.0-21.0&l=3).  As is typical, this involved running the migration tool and bumping packages as required.  Beyond that, relatively, there's a lot of changes, as compared to the preceding Angular 19 and Angular 20 migrations.

The fundamental change is a conversion of the template control flow, from the old structural directives (`*ngIf`, `*ngFor`, `*ngSwitch`) to the optimized block-based control flow syntax (`@if`, `@for`, `@switch`).  At first, I tried to resist this change, but it turned out that it was difficult to suppress, and it's the future.  So, I did it.  The conversion script was not able to convert the recent-events component, so it still uses the old structural directives.  It is tagged to so that `angular-eslint` will not barf on them.  We should convert it at some later time, but in a separate ticket, because it may be tricky...

To bolster my own QA, I tasked Claude with finding mistakes in this step, and it actually found/corrected a bug regarding a change of the indentation of the pre-formatted text of the `token` field of the CLI config file, which was impressive.

Other noteworthy changes:
* Changed in the Advanced Search dialog's "files/description" dropdown.  Previously, the template would try to hide the selected value from the dropdown menu.  This seemed to interact poorly with the newer Angular material mechanics.  The solution was to simply display both values ("files" and "description") in the dropdown, as would normally happen.
* Corrected a problem in the metrics tests, purportedly (as reported by Claude) caused by the ephemeral CDK backdrop overlap element sometimes absorbing a click destined for a menu.  As always, this could be a flake, and it is difficult to tell if our fix actually fixes the problem, or even if the problem is actually what we think it is.
* Tweaked the formatting of the category bubbles to avoid a ghost rectangle appearing on their right side.
* Added a matching "glob" in the test section of the angular config file, so the tests properly handle the ace assets.

**Review Instructions**
Confirm that the tests run successfully, and that the UI is functioning properly.  Given that most of the component templates have undergone major syntactic changes, spend a more effort than you usually would.  Note that this PR modifies the dropdown in the Advanced Search dialog, and changes the mechanics of the styling of the category bubbles (which should have the same visual appearance and spacing as before).  Please scrutinize.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-7673

**Security**


package.json
```diff
@@ -60,7 +60,7 @@
     "@ngneat/forms-manager": "^2.3.0",
     "academicons": "^1.8.6",
     "ace-builds": "^1.4.12",
-    "angular-tag-cloud-module": "^20.0.0",
+    "angular-tag-cloud-module": "^21.0.0",
     "bodybuilder": "^2.4.0",
```
denis-yuen, line 64: snyk warning is about a dependency through bodybuilder, but it isn't changing which is weird. Can chalk it up to a false positive?

THIRD-PARTY-LICENSES.csv
```diff
@@ -1,156 +1,69 @@
...
 "ajv@6.12.3","MIT","https://github.com/ajv-validator/ajv"
-"angular-tag-cloud-module@20.0.0","MIT","https://github.com/d-koppenhagen/angular-tag-cloud-module"
+"angular-tag-cloud-module@21.0.0","MIT","https://github.com/d-koppenhagen/angular-tag-cloud-module"
 "aproba@1.2.0","ISC","https://github.com/iarna/aproba"
 "assign-symbols@1.0.0","MIT","https://github.com/jonschlinkert/assign-symbols"
 "balanced-match@1.0.2","MIT","https://github.com/juliangruber/balanced-match"
-"bluebird@3.7.2","MIT","https://github.com/petkaantonov/bluebird"
-"bodybuilder@2.4.0","MIT","https://github.com/danpaz/bodybuilder"
+"bluebird@3.7.1","MIT","https://github.com/petkaantonov/bluebird"
+"bodybuilder@2.5.1","MIT","https://github.com/danpaz/bodybuilder"
```
denis-yuen, line 42: Perhaps, my theory is correct. In any case, this means that this is a newer version, albeit one with a reported security flaw. Would suggest checking if the flaws synk is reporting with 2.5.1 are already present in 2.4.0 (i.e. we're not making it worse) or not.

THIRD-PARTY-LICENSES.csv
```diff
@@ -1,156 +1,69 @@
...
-"@mermaid-js/parser@1.1.1","MIT","https://github.com/mermaid-js/mermaid"
 "@ngbracket/ngx-layout@15.1.3","MIT","https://github.com/ngbracket/ngx-layout"
-"@ngneat/forms-manager@2.5.0","MIT",""
-"@popperjs/core@2.11.2","MIT","https://github.com/popperjs/popper-core"
+"@ngneat/forms-manager@2.5.1","MIT",""
+"@popperjs/core@2.11.8","MIT","https://github.com/popperjs/popper-core"
 "@schematics/angular@8.3.29","MIT","https://github.com/angular/angular-cli"
 "@schematics/update@0.803.29","MIT","https://github.com/angular/angular-cli"
-"@types/d3-array@3.2.1","MIT","https://github.com/DefinitelyTyped/DefinitelyTyped"
-"@types/d3-axis@3.0.6","MIT","https://github.com/DefinitelyTyped/DefinitelyTyped"
```
denis-yuen, line 34: A lot of d3 dependencies disappearing, intentional?

denis-yuen (commented):
There's a lot here so I could only skim, a couple minor questions

THIRD-PARTY-LICENSES.csv
svonworl: As far as I can tell, this is ok.

THIRD-PARTY-LICENSES.csv
svonworl: Pinned lodash.unset to 4.18.0, got rid of snyk warning.

denis-yuen (approved):

